PR #2, opened 2025-07-28
Added basic support for the tabernacle experience.

soukigabriel:
- Added ext_interactiveBible and ext_tabernacle to the list of imported extensions.
- Brought Sujan's modifications to ThePage to support Locations interactions (required for the tabernacle experience)
- Modified ThePage and Section components to support interactions with the Tabernacle experience. Mostly added memoized variables, functions and shouts.
- Added styles at app.styles.page.css for interactions with the Tabernacle experince.


package-lock.json
KallynGowdy, line 1: @soukigabriel could you try to use [PNPM](https://pnpm.io/installation) for installing dependencies from NPM? It is the package manager that I originally setup with this project and it works better for making sure dependencies don't break over time.

soukigabriel: I've switched to PNPM as requested: removed package-lock.json. Let me know if everything looks good now!

packages/seed-bible/extra.aux
KallynGowdy, line 1: @soukigabriel are these bots supposed to be included in the seed-bible AUX? All of these bots don't have a `system` tag and seem like they might be dynamically created/updated on a per-experience basis.

packages/seed-bible/app/components/thePage.tsx
KallynGowdy, line 1: @mazenemad can you review these changes and merge if they look good?

packages/seed-bible/extra.aux
soukigabriel: It seemed to me like they were dynamically created, intended to be temporal, but set to shared instead, and as they didn't showed up in the system portal just remained there, but I couldn't find where they come from.

Most of them were just duplicates carrying the same webp images as the source. In case we need to do more research about this I could do a checkout to a previous version.

Screenshot 2025-07-30 181229

KallynGowdy (approved):

KallynGowdy:
@soukigabriel Does this still need merging?

soukigabriel:
> @soukigabriel Does this still need merging?

Yes, but @mazenemad's review is very important here, because it makes important changes to the ThePage and Section components.

KallynGowdy:
@soukigabriel Does this still need merging? Or is it in the latest SeedBible?

KallynGowdy:
@soukigabriel I'm closing this. I'm assuming all these changes have either already been integrated into the SeedBible package or are in #9 .

PR #9, opened 2025-08-13
🧩 Added interactiveBibleExtension and tabernacleExtension to the packages folder

soukigabriel:


KallynGowdy:
@soukigabriel Looks good but there's just a couple linting issues that need to be fixed.

KallynGowdy:
@soukigabriel Are these the most recent versions of these extensions?

PR #11, opened 2025-09-06
Playlist v1

Kabi4:


KallynGowdy:
@Kabi4 Is this feature already in the SeedBible bundle (version 32)?

KallynGowdy:
@Kabi4 Actually disregard my last comment. This is for the playlist extension, right?

Can you put it outside the `seed-bible` package folder in `packages/playlist` (instead of `packages/seed-bible/playlist`)? My goal is to have one core package for the `seed-bible` and all other packages/extensions should be in their own folder so that it is easy for us to bundle.

PR #40, opened 2025-09-26
Take Id from old data if present do not create new id or record

Kabi4:


KallynGowdy:
@Kabi4 could you make the function take a separate id parameter that defaults to a uuid instead of trying to grab it from the data?

The data won't have the id on it unless it is added manually.

Look at the Annotation interface, the annotation has id and bookId and chapter number, but the data is just whatever is attached to the annotation.

PR #46, opened 2025-09-26
Playlist improvements

Kabi4:


KallynGowdy (changes requested):
@Kabi4 it looks like the playlist got duplicated in the seed-bible folder. Can you make sure that it is in the `packages/Playlist` folder?

KallynGowdy (changes requested):
@Kabi4 there's a bunch of files that got deleted that the deployment system needs to work like  extra.aux. Also it looks like the playlist folder got renamed to be lowercase instead of starting with an uppercase letter (it needs to match the name of the extension). Have you been using the helper commands listed in the Readme? (`pnpm extension unpack Playlist`) These commands make it easy to download the extension from the records system.

PR #67, opened 2025-09-30
Fixing study note UI

Ahmedki1l:
Added the More Resources button


packages/seed-bible/app/sn_components/currentStudyNote.json
```diff
@@ -1,237 +1,110 @@
 [
```
KallynGowdy: @Ahmedki1l can you remove this file so that we're not accidentally including any copyrighted content?

KallynGowdy (changes requested):

packages/seed-bible/app/sn_components/currentStudyNote.json
Ahmedki1l: deleted it

KallynGowdy (approved):

PR #268, opened 2025-10-24
Add functions for saving reading history permanently

KallynGowdy:
Added the following functions to `db.annotations.library` for saving user reading history permanently:

Saving history:
- `saveUserReadingHistory(bookId, chapter)` - Saves the reading history for the book and chapter current user in their user record. This function should be called periodically as long as a user is active. Does nothing if the user is not logged in.
- `saveReadingHistory(recordName, userId, bookId, chapter)` - Saves the reading history for the given user (userId) in the record with the given name (recordName). The user should have the ability to read and write to the given record. This function should be called periodically as long as the user is active. `saveUserReadingHistory()` should be preferred except in cases when trying to save history for a group of users together.

Reading history:
- `getUserReadingHistorySummary(startTime, endTime)` - Gets a summary of reading history for the current user between the given start and end times (which are in unix time in seconds).
- `getReadingHistorySummary(recordName, startTime, endTime)` - Gets a summary of reading history stored in the given record between the given start and end times (which are in unix time in seconds).

Time helpers:
- `getTodayTimeSpan()` - Gets the start and end times of today.
- `getPastYearTimeSpan()` - Gets the start and end times of the past year. That is, one year from today in the past until the end of today.
- `getCurrentYearTimeSpan()` - Gets the start and end times of the current year. That is, January 1 this year until the end of today.


craigbbradley:
@KallynGowdy @soukigabriel @mazenemad for users that launch Seed Bible sessions that anonymous users join, I do think it would be good to keep a record of how many total anonymous users have joined their sessions and total anonymous user reading time

craigbbradley:
If it makes a difference i'm also good with doing min instead of sec intervals

packages/seed-bible/db/annotations/library.tsx
```diff
@@ -268,3 +268,406 @@ export async function loadTranslationDocument(recordName: string, bookId: string
...
+export async function saveReadingHistory(recordName: string, userId: string, bookId: string, chapter: number, recencyThresholdSeconds: number = 30 * 60): Promise<void> {
+    console.log(`Saving reading history for user ${userId}, book ${bookId}, chapter ${chapter}`);
+    const currentTimeSeconds = Math.floor(Date.now() / 1000);
+    const currentYear = new Date().getUTCFullYear();
+
+    const doc = await getReadingHistoryDocument(recordName, currentYear);
+    const recencyThreshold = currentTimeSeconds - recencyThresholdSeconds;
+    const array = doc.getArray('events');
+    const event = findMostRecentReadingEvent(array, userId, bookId, chapter, recencyThreshold);
+    const now = Math.floor(Date.now() / 1000);
```
soukigabriel: now is redundant with currentTimeSeconds

packages/seed-bible/db/annotations/library.tsx
```diff
@@ -268,3 +268,406 @@ export async function loadTranslationDocument(recordName: string, bookId: string
...
+
+/**
+ * Finds the most recent reading event for the given user, book, and chapter.
+ * @param events The list of reading events.
+ * @param userId The ID of the user.
+ * @param bookId The ID of the book.
+ * @param chapter The chapter number.
+ * @returns The most recent reading event, or null if no event was found.
+ */
+function findMostRecentReadingEvent(events: SharedArray<SharedMap<any>>, userId: string, bookId: string, chapter: number, oldestTime: number): ReadingEvent | null {
```
soukigabriel, line 668: Maybe we might want to put the oldestTime comprobation above the userId/bookId/chapter comprobation. If not, the function will always return an event that matches the criteria, even if it is outside of the specified threshold.

KallynGowdy: Oh yeah that's a bug

PR #465, opened 2025-11-25
for new instances the added line of code will treat them as newly joi…

Bored-Wizard:
…ned inst and install the main packages for them


KallynGowdy:
@Bored-Wizard Is this supposed to fix an issue? What does this change?

Bored-Wizard:
> @Bored-Wizard Is this supposed to fix an issue? What does this change?

@KallynGowdy if a new user joins then he definetly won't have the installedPackages tag, so by checking that I am just initializing the code in the onEggHatch which will install the other packages. it solves the issue with not having any packages installed when new user joins a shared inst.

PR #552, opened 2025-12-10
the global function and variables other devs might need

Bored-Wizard:


Bored-Wizard:
@mazenemad @soukigabriel @Kabi4 can you just briefly checkout the functions and variable mentioned in the file if  they are good to go or not.

PR #604, opened 2025-12-19
added the onActiveTabChanged listener, it is being used to change tra…

Bored-Wizard:
…nslation in bible nav ATM to change bible nav active translation


craigbbradley:
@KallynGowdy we may want to make sure @mazenemad reviews before approving to make sure this doesn't break stuff :)

Bored-Wizard:
@craigbbradley the PR doesn't contain anything which modifies anything in seedBible and just access/shouts the updates from there.

PR #632, opened 2025-12-29
Fixing issue of uninstall

Kabi4:


KallynGowdy:
@Kabi4 How is this supposed to fix the uninstall? Also isn't the extension supposed to be called "Discover" now?

PR #651, opened 2026-01-02
.gitignore Security Refactor

LandonSiler:
The following entries have been added to the .gitignore:
```.gitignore
.claude
bun.lock
.vscode/*
!.vscode/settings.json
```
These entries aim to ignore redundant or (intended to be) local only files / directories.


KallynGowdy:
@LandonSiler Is there a reason why this PR affects so many files? It looks like a bunch of files got formatted for some reason. Also could you make sure that TypeScript validation is enabled by default?

LandonSiler:
@KallynGowdy  when I merged in all the updates from develop, there appears to have (through some means) been a mass auto format... I would suspect the auto format was either from husky, or my IDE.
The main focus were the commits from December 19th:
MainCommits

I'm going to rebase and remove the fluff!

Also typescript validation was disabled previously; I had re-enabled Javascript validation by removing the 
configuration:
```json
{
 "javascript.validate.enable": false
}
```
I believe the reason typescript validation was disabled— was partially due to the lack of a declaration file for some environment specific types. 
[`AuxLibraryDefinitions`](https://github.com/casual-simulation/casualos/blob/f3d6226ae026587648cab4363c89f6d0eb6f0b26/src/aux-runtime/runtime/AuxLibraryDefinitions.def) from CasualOS could probably fix this?
There were also issues from the custom import path model that maps bot systems to their virtual fs equivalent.

KallynGowdy:
@LandonSiler ok that makes sense. I would like typescript validation to be enabled so that we can encourage devs to fix typescript issues over time, and I fixed some issues related to missing types/declaration files. It's not perfect obviously, but we need to be progressively fixing/improving things as we go.

Also can you fix the merge conflicts?

LandonSiler:
@KallynGowdy Conflicts resolved!
There's actually one less commit; develop had a commit which had taken care of the disabled JavaScript validation.

PR #714, opened 2026-01-16
New Annotation tags System

Kabi4:


packages/Playlist/playlist/playlistMode/AddAnotationUI.tsx
```diff
@@ -979,6 +979,7 @@ const AddAnotationUI = ({
       });
     }
     try {
+      await os.sleep(1000);
```
KallynGowdy: @Kabi4 Why are we waiting for 1 second here?

packages/Playlist/playlist/playlistMode/AddAnotationUI.tsx
```diff
@@ -1133,6 +1138,8 @@ const AddAnotationUI = ({
 
     setLoading(true);
 
+    await os.sleep(1000);
```
KallynGowdy: @Kabi4 And again here?

packages/seed-bible/webhook/API/defineGlobal.tsx
```diff
@@ -1,3 +1,70 @@
...
+    });
+  
+    return doc.body.innerHTML;
+  }
+  
+  function randomColor() {
+    return `hsl(${Math.random() * 360}, 70%, 55%)`;
+  }
+  
+globalThis.ColorizeParagraphs = autoTagHTML;
```
KallynGowdy: @Kabi4 Can we refactor this to use module exports instead?

packages/seed-bible/webhook/API/defineGlobal.tsx
```diff
@@ -1,3 +1,70 @@
...
+  const doc = new DOMParser().parseFromString(html, "text/html");
+
+  // Get only visible text (ignores tags like <span>, <p>, <br>, etc.)
+  const text = doc.body.textContent || "";
+
+  // Extract + remove duplicates
+  return [...new Set(text.match(hashtagRegex) || [])];
+}
+
+globalThis.ExtractHashtagsFromHTML = extractHashtagsFromHTML;
```
KallynGowdy: @Kabi4 Can we refactor this to use module exports instead?

KallynGowdy (changes requested):

packages/Playlist/playlist/playlistMode/AddAnotationUI.tsx
Kabi4: Actually  to not make Load on FE indentify tags we wait for 1 second after user has stopped typing so to make that thing consider we have done it.

packages/Playlist/playlist/playlistMode/AddAnotationUI.tsx
Kabi4: Actually  to not make Load on FE indentify tags we wait for 1 second after user has stopped typing so to make that thing consider we have done it.

Both are same one is edit one is save

PR #734, opened 2026-01-21
added a listener so that if the book selector is open and user presse…

Bored-Wizard:
…s back the book selector is closed


KallynGowdy:
@Bored-Wizard Have you tested that this works in Chrome?

Bored-Wizard:
@KallynGowdy yes it works

PR #808, opened 2026-01-31
Lazy load

Kabi4:


packages/seed-bible/db/annotations/library.tsx
```diff
@@ -176,7 +176,7 @@ export function getAnnotationMarker(
 export async function getAnnotationRecord(forceLogin?: boolean) {
   const injectedRecordKey = configBot.tags.annotationRecordKey;
   if (injectedRecordKey) {
-    return injectedRecordKey;
+    return injectedRecordKey.replace(".subjectless", "");
```
KallynGowdy: @Kabi4 You shouldn't be changing record keys like this. It isn't supported and is what is causing the system to not find the record.

PR #849, opened 2026-02-09
added shortName in the tab and book title, also had to add it to a lo…

Bored-Wizard:
…t of places. also updated the tests as the UI also includes translation shortName


KallynGowdy:
@Bored-Wizard can you fix the tests to trim whitespace when checking the book titles?

Bored-Wizard:
@KallynGowdy done

PR #861, opened 2026-02-12
change in settingconfig and themes

AmitPant9999:


AmitPant9999:
@craigbbradley  @mazenemad can you review the pr as i have made some modifications in settingsConfig file

mazenemad:
We shouldn’t switch between minimalTheme and defaultTheme within the core app components.

I’ve added the presets tag for testing and made it dynamic, so you can easily clone it and work on it independently.

PR #881, opened 2026-02-18
Feature/new annotation UI

Kabi4:


packages/Playlist/playlist/playlistMode/fetchAnnotationsData.tsx
```diff
@@ -9,27 +10,27 @@ const chapter = isPrev
...
 
-if (!globalThis.AnnotationsData) {
-  globalThis.AnnotationsData = {};
+if (!G.AnnotationsData) {
+  G.AnnotationsData = {};
 }
 
-const userRecord = await getAnnotationRecord();
-
-if (!userRecord) {
```
KallynGowdy, line 18: @Kabi4 Users don't always have to be logged in in order to fetch annotations. If there is a `annotationsRecordKey`, then they should be able to see annotations even when not logged in. Please restore this code so that it tries to fetch annotations as long as there is a `userRecord`.

KallynGowdy (changes requested):

PR #906, opened 2026-02-23
Task/improve reference

Bored-Wizard:


packages/References/references/manager/GetReferences.tsx
```diff
@@ -29,3 +48,82 @@ export const GetReferences = async (props: {
     verse,
   };
 };
+
+export const GetChapterContent = async (props: {
+  bookId: string;
+  chapter: number;
+  reference: ReferenceInterface;
+}) => {
```
KallynGowdy, line 69: @Bored-Wizard This should accept a `translationId` parameter so that we can fetch the translation that the user is using when looking up cross-references. If the related chapter isn't found for that translation, then we should default back to the `BSB` translation.

PR #1010, opened 2026-03-15
Add/mobile for preset

AmitPant9999:


AmitPant9999:
@KallynGowdy it is for making work our  easy for ligonier and reflection-ministries as we do not have to fix conflits every time when we are merging core with ligonier or r-m

PR #1211, opened 2026-05-14
Dev/gabriel souki

soukigabriel:
Merging with Scripture Map 2D in a functional state. User presence and UI design are a work in progress.


KallynGowdy:
@soukigabriel Can you remove the `coverage` directory and add it to the `.gitignore`?

Blitzy (approved):
Hey @soukigabriel this is a good first pass. There are some UX and polish and a few bugs stuff that we will iterate on in separate tracked issues here in GitHub but going to merge this into `develop` as everything is in working order with the refactor of the refactor.

PR #1231, opened 2026-05-28
updated ui as per latest specification

Bored-Wizard:


KallynGowdy:
@Bored-Wizard Somehow all the translation files got saved with a UTF-8 Byte Order Mark, and that was causing the build to fail because [it isn't allowed in JSON](https://stackoverflow.com/a/38036753/1832856). I fixed it but wanted you to know.

PR #1246, opened 2026-06-04
updated extension title and added a check for pattern param

Bored-Wizard:


packages/twitchPub-extension/ext_twitchPub/host/twitchPubManager.tsx
```diff
@@ -49,10 +49,12 @@ const getUrl = (
 ) => {
   const redirectUri = new URL(configBot.tags.url ?? "https://ao.bot/");
   redirectUri.search = "";
-  redirectUri.searchParams.set(
-    "pattern",
-    configBot.tags.pattern || "SeedBible"
-  );
+  if (redirectUri.origin !== "https://seedbible.org") {
```
KallynGowdy: @Bored-Wizard We don't need to check the origin. If a `pattern` is present in the original, then it should also be present in the new link. If there is no `pattern`, then it shouldn't exist in the generated link either.

Bored-Wizard: got it

packages/twitchPub-extension/ext_twitchPub/host/twitchPubManager.tsx
```diff
@@ -49,10 +49,12 @@ const getUrl = (
 ) => {
   const redirectUri = new URL(configBot.tags.url ?? "https://ao.bot/");
   redirectUri.search = "";
-  redirectUri.searchParams.set(
-    "pattern",
-    configBot.tags.pattern || "SeedBible"
-  );
+  if (redirectUri.searchParams.get("pattern") === null) {
```
KallynGowdy: @Bored-Wizard Looks like this would add the `SeedBible` pattern every time since the search params are cleared on line 51. You should check if a pattern was specified on the configBot instead

Bored-Wizard: checking configbot's pattern now

PR #1258, opened 2026-06-05
Add chat features

### LINKED ISSUE #1232: Build AI Chat integrations
We need support for AI chat that enables the following:

- Extensions can register their own chat providers
- Multiple chat providers can be registered at once
- Users can chat with a provider by @'ing it by name
- Users can @everyone to chat with all providers at once
- Chats can be synchronized to shared sessions
- If the user is not in a shared session, then chat messages go to the default AI if none is @'ed
- If the user is in a shared session but does not @ a provider, then it just goes to the other users.
- Users can see typing indicators for AI and other users.

KallynGowdy:
Closes #1232 

Added support for chat features. Still needs an additional design pass, but all of the functionality is there.

### Features

- Supports Chat Providers
  - Extensions can register their own chat providers to respond to mentions and interact in solo chats.
- Shared chats
  - Every shared session now gets a shared chat
  - Shared chats can have chat providers as participants (users can only message their own providers, but everyone can see the responses)
  - User presence and typing indicators are supported (for chat providers as well)
- Local chats
  - Users can start new local chats with a chat provider of their choice
  - Additional chat providers can be added by mentioning them

### Visuals

By default, there is no option to chat unless there is a chat provider registered or a shared session open.

#### Toolbar

Desktop:
image

Mobile:
image

#### Chat List (Empty)

Desktop (1440px):
image

Laptop (1024px):
image

Mobile (375px):
image

#### Chat List

Desktop:
image

Mobile:
image

#### Chat View

Desktop:
image

Mobile:
image


KallynGowdy:
@Blitzy This should be good to review now

Blitzy:
Very cool, this looks good to go.

PR #1294, opened 2026-06-29
twitch pub now has access to ai transcript ref extraction function fr…

Bored-Wizard:
…om the live ref extract feature. It now automatically allow host to go to ref just by saying the reference to the verse


KallynGowdy:
@Bored-Wizard So this will allow hosts to turn on transcription and then automatically navigate based on what they say? Does that require access to the OpenAI Realtime API?

Also, instead of making the twitch extension dependent on transcription, could we just make the transcription extension support navigating based on detected references? This way, the twitch extension wouldn't need to be dependent on transcription and it would be more flexible

Bored-Wizard:
@KallynGowdy yes, it still uses OpenAI Realtime API to generate transcript which is then given to ai.chat to detect references.

I guess I can separate the core features and the UI from it and make 2 extentions out of it.

PR #1299, opened 2026-06-30
Add support for saving installed extensions and automatically re-installing them on refresh

### LINKED ISSUE #1203: Save installed extensions to the user's profile and auto-install them on load
We should save the list of extensions that the user has chosen to install. When they load a new tab (new inst), we should load the extensions that are in their profile data.

KallynGowdy:
Fixes #1203 

If the user isn't logged in, then installed extensions are saved to local storage. On login, the extensions from local storage are merged (via union) with the ones from the user's profile.

### DIFF
--- a/packages/seed-bible/seed-bible/managers/ExtensionManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ExtensionManager.tsx
@@ -1,8 +1,13 @@
-import { signal } from "@preact/signals";
+import { effect, signal } from "@preact/signals";
 import { orderBy, union } from "es-toolkit";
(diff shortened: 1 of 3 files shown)


Blitzy (approved):
Verified working:

1. Signed in.
2. Installed Audio Reader extension.
3. Opened incognito tab (anonymous user).
5. Verified that Audio Reader was **not** installed.
4. Installed Twitch Host extension.
5. Signed in again.
6. Verified both Audio Reader and Twitch Host were installed.
7. Signed out.
8. Opened new tab.
9. Signed in.
10. Verified both Audio Reader and Twitch Host were re-installed.

PR #1300, opened 2026-06-30
fix: align reader bookmark button with the title row

mazenemad:
Center the 36px bookmark button (and the quick toolbar beside it) on the title's first line instead of pinning them 14px higher, so they no longer look offset against the title on narrow readers.


Blitzy (approved):
Simple alignment of the bookmark and quick toolbar buttons with the desktop mode book title header. Looks good!

PR #1301, opened 2026-06-30
fix: stop the reader title overlapping the bookmark button

mazenemad:
Reserve trailing-edge padding on the chapter title equal to the bookmark button's footprint so long titles wrap/stop before it instead of running underneath on narrow readers.


Blitzy (changes requested):
Some of the quick toolbar buttons like the button for the Audio Reader still overlap the title.

Image

packages/seed-bible/seed-bible/app/main.css
```diff
@@ -237,7 +237,7 @@ body {
  * browsers don't show the default white "overscroll" gap when the user
  * pulls the page up past the bottom toolbar. Scrolling happens inside
  * the inner panes (which have their own overflow scrollers). */
-@media (max-width: 768px) {
+@media (max-width: 480px) {
```
Blitzy, line 240: These seem like accidental changes for this particular fix?
The mobile width should stay at 768px. Please revert all of the mobile width changes in this commit.

Blitzy (changes requested):
There appears to have been an accidental inclusion of mobile width changes in this commit that shouldnt be there. Please revert them so mobile width is still at 768px.

Blitzy:
@mazenemad there is an issue with still being able to click on the toolbar while the sidebar is open in overlay mode.

Recording 2026-07-03 115108

PR #1311, opened 2026-07-02
feat: enhance local config caching for user settings in createConfig

### LINKED ISSUE #1308: Scripture font size is reverting after closing "Display & Theme" settings page while not signed in.
## What is happening:

While not signed in (anonymous user), if i change that scripture font size to anything other than the default, it will revert back to the default once the "Display & Theme" settings page is closed.

_Setting scripture font to largest settings as anonymous user:_
Image

_After closing the "Display & Theme" panel, the scripture font reverts back to its default settings:_
Image

## What I expected:

The scripture font size that I set as an anonymous user would stay what I set it for the duration of the current session.

mazenemad:
Fixes #1308

### DIFF
--- a/packages/seed-bible/seed-bible/managers/ConfigManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ConfigManager.tsx
@@ -89,6 +89,16 @@ export function createConfig(
   login: LoginManager,
   navigation: NavigationManager
 ) {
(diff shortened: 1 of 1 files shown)


KallynGowdy:
@mazenemad can you make this persist the local config to local storage so that their settings will be preserved across browser refreshes?

KallynGowdy (approved):

PR #1313, opened 2026-07-02
feat: enhance onboarding and tutorial flow for session link users

### LINKED ISSUE #1309: Don't show welcome screen/tutorials when joining a shared session
When a user opens the seed bible via a link that auto-joins them to a shared session (contains `sessionId` in the query), we should not show that user the welcome screen or tutorials. If they revisit the seed bible but without a session ID, then we can show them a tutorial (joining via session should not disable tutorials permanently - only for that tab).

mazenemad:
fixes #1309


KallynGowdy:
@mazenemad can you have Claude add a test that verifies that tutorials are skipped if the user joined a session link?

PR #1327, opened 2026-07-04
Task/port twitch transcribe extension

Bored-Wizard:


Bored-Wizard:
@KallynGowdy can you please check why I am getting this error? I don't know what might be wrong in it as the code seems to be fine.

image

KallynGowdy:
@Bored-Wizard it's because the pane management system changed so type is no longer a valid property. You have to use placement instead

packages/transcript-extension/ext_AI_Transcript/main/transcribe.tsx
```diff
@@ -0,0 +1,679 @@
+import { FFmpeg } from "@ffmpeg/ffmpeg";
+import { fetchFile, toBlobURL } from "@ffmpeg/util";
+import { fromByteArray } from "base64-js";
+import axios from "axios";
+
+// --- Config constants -------------------------------------------------------
+
+export const GET_TRANSCRIPTION_KEY_URL =
+  "https://aolab-bible-api.netlify.app/api/ai/getTranscriptionKey";
```
KallynGowdy, line 9: @Bored-Wizard Do we have limits on the OpenAI API Key that we're using here?

KallynGowdy: Like spending limits

Bored-Wizard: rn it doesn't have that. I will add that in a bit.

KallynGowdy (approved):

PR #1331, opened 2026-07-05
fix: prevent profile wipe when profile load fails transiently

Blitzy:
## Summary

Fixes intermittent profile data wipes (reported on mobile). A transient failure reading the profile record was being treated the same as "no profile exists," which let the next config write overwrite the real stored profile with a blank.

## The bug

Every records call — the profile **read** (`getData`) and **write** (`recordData`) — is a `POST` to `…/api/v3/callProcedure`. When the server returned a structured error for the *read* (`server_error` or `not_authorized` — plausible on flaky mobile links, or during a session-refresh/token race), the wipe chain was:

1. `getUserProfile` collapsed **any** failure to a blank `{ name: "" }` — it only distinguished `data_not_found`.
2. The profile-loading effect stored that blank in `profile.value` (an in-memory wipe).
3. The existing `saveProfileConfigValue` guard checks `if (!profile.value)`, but `{ name: "" }` is *truthy*, so it passed.
4. The next auto-save — theme, font size, a tutorial-seen flag, an extension install, etc. — merged its key into the blank base and persisted it, **permanently overwriting the stored name, location, picture, and every other config key.**

This was intermittent and hard to reproduce because it needs a *server-returned* read error at load time followed by any successful write. A plain dropped connection makes `fetch` throw, which was already handled and did **not** trigger the wipe.

## Changes

- **`getUserProfile`** returns a blank default only for `data_not_found` (a genuinely new account). All other failures throw, so a transient read error can no longer masquerade as an empty profile. A malformed stored record is likewise treated as a load failure. (`SessionsManager` already wraps this call in try/catch.)
- **Profile-loading effect** keeps the existing in-memory profile when a reload fails instead of clobbering it, ignores a stale load that resolves after an account switch, and has a passive rejection handler so an unawaited failed load doesn't surface as an unhandled rejection.
- **`updateProfile`** refuses to write while the profile hasn't loaded, rather than merging into a bare `{ name: "" }` base.
- **`profilePromise`** is exposed via a getter so external readers (the `saveProfileConfigValue` wait guard) see the promise assigned by the effect, instead of the `null` captured at construction time (which silently defeated the guard for fresh logins).

Local (logged-out) settings persistence is unaffected — the local `configBot.tags` writes and the `profile ?? local ?? default` read precedence are unchanged.

## Testing

- Added regression tests covering transient-failure handling, the reload-doesn't-clobber path (driven through the effect the way a session refresh does in production), and the `updateProfile` loading guard. Verified these fail against the pre-fix source and pass with the fix.
- Full suite green (2524 tests), `pnpm check:ts` clean.

### Reproduce it yourself

The wipe needs the *server* to return an error for the profile read (a dropped connection won't do it — that throws and is handled). This forces exactly one profile read to fail, no rebuild required:

1. Log into an account and populate the profile so a wipe is obvious: set a display name, upload a picture, change a couple of settings (theme + font size). Note the values.
2. **Log out.**
3. Open DevTools → Console and paste:

    ```js
    // Fail ONLY the first profile read with a server-style error, then behave normally.
    const _origFetch = window.fetch;
    let _failedProfileOnce = false;
    window.fetch = async (input, init) => {
      try {
        if (!_failedProfileOnce && typeof init?.body === "string") {
          const b = JSON.parse(init.body);
          if (b.procedure === "getData" && b.input?.address === "profile") {
            _failedProfileOnce = true;
            console.warn("[repro] Forcing profile read to fail (server_error)");
            return new Response(
              JSON.stringify({ success: false, errorCode: "server_error", errorMessage: "forced repro" }),
              { status: 200, headers: { "Content-Type": "application/json" } }
            );
          }
        }
      } catch {}
      return _origFetch(input, init);
    };
    console.log("[repro] fetch patch installed — now log in");
    ```

4. **Log in** (email code). The app loads your profile — the patch fails that read once.
5. **Change one setting** (font size or theme).
6. **Reload the page** (clears the patch, shows what's actually stored).

**Pre-fix:** name and picture are gone, settings reset. In the Network tab, the `callProcedure` POST with `"procedure":"recordData"` after step 5 has body `"data":{"name":"","config":{…}}` — that blank `name` overwriting the record is the wipe.

**With this fix:** name, picture, and settings are all intact. Console shows `Failed to load user profile; keeping existing profile` and `Cannot save profile config value: profile has not loaded yet`, and no `recordData` write goes out for that setting change. (Expected nuance: the setting changed during the failed-load window won't sync to the cloud until the next successful profile load — it stays in the local cache. That's the intended trade-off: don't write until the base profile is trustworthy.)

The bug is logic-level, not platform-specific, so desktop is enough to prove it. To confirm on a phone, use remote debugging (`chrome://inspect`) and paste the same snippet.

### DIFF
--- a/packages/seed-bible/seed-bible/managers/LoginManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/LoginManager.tsx
@@ -469,6 +469,28 @@ export function createLoginManager({
       return;
     }
 
+    // Make sure the profile has loaded before we upload anything. `updateProfile`
+    // refuses to write while the profile is null (to avoid wiping the account),
+    // so persisting the URL would silently no-op if we ran ahead of the load —
+    // and the caller would see a resolved promise and report a false success.
+    // Failing here, before the (billable) file upload, avoids paying for a file
+    // we couldn't attach to the profile anyway.
+    if (!profile.value) {
+      if (profilePromise) {
+        try {
+          await profilePromise;
+        } catch {
+          // Ignored; the guard below turns a failed load into a thrown error.
+        }
+      }
+
+      if (!profile.value) {
+        throw new Error(
+          "Failed to upload profile picture: profile has not loaded"
+        );
+      }
+    }
+
     const result = await os.recordFile(userId.value, file, {
       mimeType: file.type,
       marker: "publicRead",
--- a/packages/seed-bible/seed-bible/managers/LoginManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/LoginManager.tsx
@@ -158,21 +158,37 @@ export function createLoginManager({
     const data = await os.getData(userId, "profile");
 
     if (!data.success) {
-      console.log("[LoginManager] No profile data found for user:", userId);
-      // Return a default profile
-      return {
-        name: "",
-      };
+      if (data.errorCode === "data_not_found") {
+        // The account genuinely has no profile record yet (new user). A blank
+        // default is the correct, authoritative answer here — the user can
+        // start filling it in and writes should be allowed.
+        console.log("[LoginManager] No profile data found for user:", userId);
+        return {
+          name: "",
+        };
+      }
+
+      // Any other failure (server error, `not_authorized`, network blip — all
+      // common on mobile) is transient: the profile may well exist, we just
+      // couldn't read it right now. We must NOT fall back to a blank profile,
+      // because the caller stores it in `profile.value` and the next config
+      // write (`saveProfileConfigValue` / `updateProfile`) merges into it and
+      // persists it — permanently wiping the real name/location/picture and
+      // every other config key. Surface the failure instead so callers keep
+      // whatever profile they already had.
+      throw new Error(
+        `[LoginManager] Failed to load profile (${data.errorCode}): ${data.errorMessage}`
+      );
     }
 
     const parsed = userProfileSchema.safeParse(data.data);
 
     if (!parsed.success) {
+      // The record exists but doesn't match the expected shape. Returning a
+      // blank default here would also let the next write clobber the stored
+      // record, so treat it as a load failure rather than an empty profile.
       console.warn("Failed to parse user profile data:", parsed.error);
-      // Return a default profile
-      return {
-        name: "",
-      };
+      throw new Error("[LoginManager] Stored profile failed validation");
     }
     return parsed.data;
   };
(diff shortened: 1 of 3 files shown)


Blitzy:
I've continued to run into some strange profile wipe scenarios. I had Claude dive in and analyze scenarios where it could still be occurring and it identified this one. I verified that any error response from the record system that isnt either success or `data_not_found` was causing the app to wipe the profile using the testing method described in "Reproduce it yourself".

This PR does fix that scenario. Also verified that this has no effect on any anonymous user localStorage level settings persistance.

KallynGowdy:
Reviewed the fix — the transient-vs-`data_not_found` distinction and the "keep last-known-good profile" logic look solid and well-tested. One issue I'd flag before merging:

**`uploadProfilePicture` silently drops the picture URL when the profile hasn't loaded yet, while reporting success** ([`LoginManager.tsx:466-483`](``https://github.com/HelloAOLab/seed-bible/blob/d4c0dac9119c788bc4490464c5bef19b0f2839da/packages/seed-bible/seed-bible/managers/LoginManager.tsx#L466-L483``))

If a user opens the profile-picture upload flow shortly after login — before the profile-loading effect's promise has resolved — `os.recordFile(...)` succeeds (the file is actually uploaded to storage, so the cost is real), then `uploadProfilePicture` calls `updateProfile({ pictureUrl: result.url })`. Since `profile.value` is still `null` at that point, the new guard in `updateProfile` just `console.warn`s and returns — it doesn't throw. `uploadProfilePicture` then resolves normally as `Promise<void>`.

The caller in `SettingsPage.tsx` (~line 279) does `await login.uploadProfilePicture(file)` inside a try/catch that never fires here, so the UI treats the upload as successful and closes the modal — but the picture URL was never actually persisted to the profile.

This is a direct side effect of the new "refuse to write instead of risking a wipe" behavior in `updateProfile`. Worth having `uploadProfilePicture` surface that failure back to its caller (e.g. throw, or return a status) so the UI doesn't report success when the picture URL silently didn't save.

Everything else checked out — staleness handling in the profile-loading effect's `.then`/`.catch` branches, the shared `profilePromise` across rapid login/logout, and write-time `userId` checks all correctly re-read live state, so no cross-account merge is possible.

---

Blitzy:
Good catch — fixed in 1e3361c.

`uploadProfilePicture` now waits for the profile to load before doing anything, and throws if it still hasn't loaded — **before** the `recordFile` call, so we also don't pay for a file upload we couldn't attach. In the normal case (uploading a moment after login) the wait just resolves and the flow succeeds as before; only a genuine load failure surfaces as a thrown error, which the `SettingsPage.tsx` caller already propagates. Added a regression test covering the "profile never loads → upload throws and `recordFile` is never called" path.

I left the guard in `updateProfile` itself as a silent no-op on purpose, since `saveProfileConfigValue`/`setAllSettings`/`handleSave` rely on it not throwing — the surfacing belongs at the `uploadProfilePicture` layer where the caller cares about the result.

---

PR #1333, opened 2026-07-05
fix: stack collapsed sidebar tab text vertically instead of in a row

### LINKED ISSUE #1332: Tab text in collapsed sidebar isn't vertically aligned

KallynGowdy:
.sb-collapsed-tab-tile held two lines of text (book abbreviation, chapter
number) meant to stack, but was set to flex-direction: row with a gap
sized for a row layout, plus a leftover margin-top on the chapter span
that only made sense for a column layout. Fixes #1332.


KallynGowdy:
**Before / after** — rendered directly against the real `main.css` at each commit (not mockups), so what you see here is what the collapsed sidebar actually looks like.

**Before:**
before

**After:**
after

`GEN 1`, `1CH 29`, `REV 22` look the same either way — short combos were never broken. The difference shows up on `PSA 150` (book code + 3-digit chapter, the realistic worst case since book IDs are always 3 letters): before, the chapter number sits lower than the book code and the text overflows the rounded tile; after, they're aligned on one line, and when a combo doesn't fit, it wraps to two centered lines instead of spilling outside the tile.

(The screenshot files under `.pr-review-media/` are temporary review assets — removing them in a follow-up commit before merge.)

---

Blitzy (approved):

PR #1347, opened 2026-07-05
feat: move keeping screen awake in mobile settings

mazenemad:


Blitzy (commented):
The change is clean and does what it says. One thing worth thinking a about first:

The wake-lock toggle lives in 3 spots, all driving the same setting — the sidebar ⋯ menu, the Settings page, and (new here) the mobile quick-sheet. This PR takes it off the Settings page and adds it to the mobile sheet.

Net effect: on desktop it still works but is now only reachable via the ⋯ menu — it drops off the Settings page where people would look for it. It's a legit desktop feature (keeps the monitor from sleeping on long reads), so my only hesitation is discoverability, not the feature itself.

Question for @KallynGowdy and @craigbbradley: do we want it off the Settings page, or keep it there and add the mobile toggle?

KallynGowdy:
I think we should keep it in the settings page, but moved to the "Display & Theme" section. So Desktop users can either go `sidebar -> dots menu -> keep screen awake` or `sidebar -> settings -> display & theme -> keep screen awake` while mobile users can go `quick settings -> keep screen awake` or `quick settings -> full settings -> display & theme -> keep screen awake`

Blitzy (changes requested):
@mazenemad per @KallynGowdy's suggestion, lets move the "Keep screen awake" toggle from the settings page to the Display & Theme" page.

Issue #1342, opened 2026-07-05
Improve highlight styling: rounded corners, less blinding highlights in dark theme, more padding


Blitzy:
## What is happening

Highlights are visually harsh. They have hard, square corners, and in dark theme the highlight colors are very bright and saturated — to the point of being blinding to read against. There is also virtually no padding between the text and the highlight edges so it makes it feel very cramped.

_What a contiguous two verse highlight currently looks like in Seed Bible:_

<img height="545" alt="Image" src="https://github.com/user-attachments/assets/9eb248ba-39f8-425a-8371-8382afd25405" />

## What is expected

Highlights should look more polished:

- Rounded corners on highlights. This will likely require changing how highlights are applied/rendered to support rounded corners.
- In dark theme, highlight colors should be muted — less bright and less saturated — so they're comfortable to read rather than blinding.
- A little more padding between highlight edges and text to give the text every so slightly more breathing room and ease readability.

_What a contiguous two verse highlight looks like in YouVersion:_
<img  height="800" alt="Image" src="https://github.com/user-attachments/assets/f2c5eb34-5c01-4d82-b5a6-82e7ccfd7c57" />

PR #1355, opened 2026-07-05
fix: update selected verse underline style to solid and adjust test s…

### LINKED ISSUE #1290: It is difficult to see selected verses with the dark theme and yellow highlight
When viewed in dark mode, it is difficult to see that verses that have a yellow highlight are selected.

Example:

Verse 20 is selected, but the selection indicator is difficult to see.

mazenemad:
This is a proposed option for review — not the final solution.
 
fixes #1290

### DIFF
--- a/packages/seed-bible/seed-bible/app/main.css
+++ b/packages/seed-bible/seed-bible/app/main.css
@@ -80,7 +80,7 @@ html {
   --sb-verse-words-of-jesus-font-family: inherit;
   --sb-verse-words-of-jesus-font-style: inherit;
 
-  --sb-selected-verse-border-bottom: 2px dashed currentColor;
+  --sb-selected-verse-border-bottom: 3px solid currentColor;
   --sb-selected-verse-text-decoration: none;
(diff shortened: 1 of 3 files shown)


KallynGowdy:
For comparison:

Before:
image


After:
image

Before:
image

After: 
image

Before:
image

After:
image

KallynGowdy:
It might be worth trying more different colors instead of making it a solid line. Alternatively, we could adjust the highlight colors we use to make it easier to see white text. YouVersion currently uses a semi-transparent colors for highlights which make it easier to see because the text can stay white

Blitzy:
Yeah id argue the solid underline makes selection visiblity worse. I think the real fix here would be to tweak how the highlights look which ive started documenting: https://github.com/HelloAOLab/seed-bible/issues/1342

In this case if the highlights as just ever so slightly more padding between the edge of a highlight and the text you could see the selection underline much more easily.

Blitzy:
Going to close this in favor of issue #1342 for now

PR #1361, opened 2026-07-07
Fixes 1286

### LINKED ISSUE #1286: Add default Bible translations for each UI language
Currently we only have default translations for some of the UI languages that we support. We should go through each additional UI language that we support and make sure that there is a default translation selected for them. If we don't have a translation for that language, do your best to find a translation that people who speak & read the selected language would be able to understand (e.g. If we didn't have a Spanish translation, we could try showing a portuguese translation).

Default Translations: https://github.com/HelloAOLab/seed-bible/blob/develop/packages/seed-bible/seed-bible/managers/BibleReadingManager.tsx#L234

Kabi4:
Fixes #1286


KallynGowdy:
@Kabi4 the task is to do this for Bible translations, not the UI language. We need to choose a default Bible translation for each UI language that we support

KallynGowdy (approved):

PR #1364, opened 2026-07-07
Add playlists feature

### LINKED ISSUE #1110: [refactor] Port the Playlist extension to the refactor
We need to port the playlist extension to the refactor.

KallynGowdy:
When merged, this PR will add initial support for playlists to the Seed Bible.

## Key features

- Create, edit, play, and share playlists.
- Supports bible verses, chapters, links, and custom text playlist items.
- Mobile view takes over the navigation pills
- Utilizes same data structures as reading plans
- Provides a discover panel that can surface relevant content (like Bible Project content or cross references)
- Provides support for extending reading states to support custom navigation
- Fixes issues with multiple history entries being created when navigating chapters

## Current limitations

- Doesn't entirely match Figma designs because of mobile support
- No annotations yet

#### Discover view:
image

image

#### Creating/Editing view
image

image

#### Playing view
image

image

image

#### Custom text
image

image

#### Videos
image

image

Fixes #1110


KallynGowdy:
@Blitzy This is ready to review now

Blitzy:
@KallynGowdy This is a really great foundational work for playlists. I think there are a lot of UX things we can improve upon but this is still worth merging in and iterating on.

I will leave it up to you on wether you want to address these bugs here before we merge or we can make issues for each and address them later. None of them are highly critical, pretty minor bugs overall.

Here are bugs I found during my testing:
### 1) Scripture ranges exclude the last verse during playback.

If the range is something like Exodus 5:2-5, verses 2-4 will be highlighted but the 5th will not.

image

### 2) Playlist page gets into a weird state if you change tabs while a playlist is active.

Follow these steps:

1. Start a playlist
2. Open a new tab
3. Notice how the playlist sidebar goes back to the main menu but the header keeps the back arrow and now says "Untitled playlist." I expected

*What the playlist sidebar looks like after selecting a non-playlist tab from an active playlist tab*
image

*What i expected it to look like*
image

### 3) Suggestions support cross-chapter ranges but become corrupted when saved.

I'm not really sure if this is even a feature we need/want to support. But the suggestion system does show that you can add references that are cross-chapter. However, when you save them they omit the second chapter number.

For example:

1. Input `Genesis 1:2-3:4`. Notice that `Genesis 1:2-3:4` is a suggestion.
2. Click "Save".
3. Notice that the item in the queue now says `Genesis 1:2-4`. It omitted chapter 3 entirely and now has become a verse range in chapter 1.

It seems to me that we would not support this and actually to implement something like this in a playlist we would just have two scripture items:
1. `Genesis 1:2`
2. `Genesis 3:4`

Thoughts?

image

### 4) During shared sessions, any user can navigate regardless of the "Only host can navigate" flag value of the shared session.

Pretty easy to reproduce this one:

1. Start a shared session. 
2. Open shared session settings.
3. Enable "Only host can navigate".
4. Start a playlist.
5. Open incognito tab and paste the shared session link into it.
6. Move to the next item in the queue as the anonymous user, notice that they are controlling the playlist.

I would have expected that if "Only host can navigate" that only the host would be able to control the playlist.

## UX Ideas - no need for action on this pull request

1. We should have icons for each item type. Each item type in the queue should also have this icon.
2. An indicator should show after clicking Save that shows that a save is inflight.
3. Link items with embed enabled should try to fill a large portion of the screen. People will want websites and videos to appear larger on their screens.
4. Clickable area of queue items should fill more of the button height. It only extends the height of the text on down to the delete button.
5. Clicking "Save" with unsaved content in the "Add Item" section discard the content. Ideally it would ask if you wanted to add the item before saving?

Blitzy (approved):

KallynGowdy:
@Blitzy that is some good feedback. I had Claude look at it and I will integrate those changes tomorrow

Blitzy:
@KallynGowdy I just noticed that the verse highlights in dark mode are barely noticeable compared to light mode.

image

KallynGowdy:
@Blitzy Yeah, we should make that a separate issue because we want to do the diminish effect instead of highlights for those types of things. And it is an issue for search as well

Blitzy:
@KallynGowdy Trying out the cross-chapter ranges and its behaving well, with one weird thing. Im testing with the previous `Genesis 1:2-3:4` example.

1. ✅ Cross chapter range looks good in playlists edit mode now. 
image

2. ✅ Cross chapter ranges appear has seperate entries in play mode. This is kind of cool — so the cross chapter range in edit mode acts like a macro for doing these seperately.
image

3. ❔ Each entry in play mode for the cross-chapter range is displaying technically correctly. However when viewing Genesis 2 in our `Genesis 1:2-3:4` my brain thought oh this is part of a verse selection range so ill see the entirety of Genesis 2 highlighted but it wasn't obviously because its a chapter reference not a verse reference. This is technically find just maybe a UX weirdness?

https://github.com/user-attachments/assets/9c198ce2-a8b5-4acf-8ff2-00f14e8e39b7

KallynGowdy:
Yeah for point 3 it is a weird choice but I think it will feel natural once we're not highlighting the chosen verses and instead diminishing all the other verses.

Blitzy:
Good point, that will make it feel more natural.

PR #1369, opened 2026-07-08
feat: implement book and chapter matching in search functionality

### LINKED ISSUE #1241: Make book / chapter results discoverable via search
Currently search results only include verses but if someone types "Gen" then Genesis should be the top result or Gen 2 then Genesis 2 should be the top result etc.

mazenemad:
fixes #1241


mazenemad:
@craigbbradley @KallynGowdy 
its ready now you can check if this what we need

KallynGowdy (approved):

PR #1372, opened 2026-07-08
fix: rename "UI text size" setting to "UI size" (#1341)

### LINKED ISSUE #1341: Rename "UI text size" to "UI size" and use S/M/L/XL labels on the mobile size shortcut
## What is happening

The setting is currently labelled "UI text size", which is misleading. It doesn't just scale text — it scales the entire user interface. The one thing it doesn't affect is the scripture font size, which is already controlled by its own separate setting.

Separately, the mobile shortcut for changing the size uses a font symbol (the "A" glyph) to represent the size options.

## What is expected

1. The wording should change from "UI text size" to "UI size", to reflect that it scales the whole UI rather than just text.
2. The mobile shortcut for changing the size should use "S", "M", "L", "XL" labels instead of the font symbol.

Blitzy:
## What & why

Will fix #1341 

The setting labeled **"UI text size"** doesn't just size text — it scales the *entire* interface (chrome dimensions, spacing, icons, and fonts) via the `--sb-ui-scale` root variable. The old name misled users into thinking it only affected text. This renames it to **"UI size"** everywhere to signal that it sizes the whole interface. While in the setting, the mobile size buttons were also cleaned up.

## Changes

**Rename "UI text size" → "UI size"**

- User-facing: settings label, `<select>` id (`sb-ui-text-size-select` → `sb-ui-size-select`), and the mobile settings sheet section label.
- Code identifiers in `SettingsManager.tsx` and its consumers: type `UITextSize` → `UISize`; settings field `uiTextSize` → `uiSize` (+ Zod schema); `UI_TEXT_SIZE_OPTIONS` / `UI_TEXT_SIZE_SCALE_MAP` → `UI_SIZE_*`; `parseUITextSize` → `parseUISize`; `setUITextSize` → `setUISize`.
- Persistence keys: config tag `app.uiTextSize` → `app.uiSize`; profile key `uiTextSize` → `uiSize`.
- i18n key `ui-text-size` → `ui-size` (English value: "UI size").
- Doc-comment updates in `app/main.css` and `scripture-map/styles.css`.

**Mobile size buttons**

- Each button now shows its actual size label (**S / M / L / XL**) instead of a generic "A" glyph, so the choice is explicit. Font-size still steps up per button to preview the scale.

## Notes for reviewers

- **Existing users' saved preference resets once.** The stored keys changed (`app.uiTextSize`/`uiTextSize` → `app.uiSize`/`uiSize`) with no migration, so anyone who had picked a non-default size (S/L/XL) will fall back to **M** after this ships — the code now reads a key that was never written. If we want to preserve existing choices, we'd need a one-time migration that reads the old key. Flagging so we can decide whether that matters.

## Testing

- [ ] Desktop Settings → Display: label reads "UI size", changing it scales the UI, selection persists across reload.
- [ ] Mobile settings sheet: buttons read S/M/L/XL and select correctly.


Blitzy:
@KallynGowdy are you good with this change?

KallynGowdy:
Yes

KallynGowdy (approved):

PR #1368, opened 2026-07-08
Extract component CSS into co-located files

Blitzy:
## Summary

Migrates the Seed Bible reader from a single monolithic `app/main.css` (~10,570 lines) to a modular, co-located CSS architecture: each component owns its stylesheet, backed by a small set of shared foundation sheets. Intended as a pure reorganization — no visual or behavioral changes.

## Key changes

**CSS co-location**

- **Removed** `app/main.css` (~10,570 lines) — the single source of all app styling.
- **Added** ~25 co-located component stylesheets, each imported as the first line of its component:
  - `components/Tabs/Tabs.css`
  - `components/SettingsPage/SettingsPage.css`
  - `components/BibleReaderToolbar/BibleReaderToolbar.css`
  - `components/FloatingReaderPanels/FloatingReaderPanels.css`
  - `components/BibleReader/BibleReader.css`
  - `components/PaneLayout/PaneLayout.css`
  - `components/ChatView/ChatView.css`
  - `components/BibleSelector/BibleSelector.css`
  - Plus LoginModal, SidebarSearch, SessionParticipants, Tutorial, ToastHost, MobileSettingsSheet, Avatar, TutorialPrompt, ContextMenu, QuickToolbar, BelowReaderToolbar, MobileSplitPanelWarning, ProfilePictureModal, TermsOfServiceModal, and others.
- **Added** shared foundation sheets in `app/styles/`:
  - `base.css` — `:root` design tokens, html/body reset, and the UI-scaling model.
  - `utilities.css` — utility classes (`.sr-only`, flex helpers, …).
- **Updated** `app/main.tsx` to import the foundation sheets; each component imports its own CSS.

**Helper & component colocation**

- Moved shared helpers into `app/`: `KeyboardNav → app/keyboardNav`, `Utils → app/utils`, `VerseReferenceLink → app/verseReferenceLink`.
- Colocated single-file components into their own folders: `CasualOSApp`, `ModalHost`, `PrivacyPolicyModal`, `CodeOfConductModal`.
- Folded the former global `legacy.css` into its owning components — the book/translation selector styles into `BibleSelector.css`, the translation license/website notice into `BibleReader.css`.

**Lint**

- ESLint now lints the co-located component CSS.
- `css/no-invalid-properties` runs with `{ allowUnknownVariables: true }`: the `--sb-*` design tokens are defined centrally (`base.css` / `ThemeManager`) and referenced across files, which the per-file rule can't resolve — this keeps real property/value validation without false "unknown variable" positives.

[...]

PR #1373, opened 2026-07-08
feat: add book/chapter heading to top of mobile reader content (#1282)

### LINKED ISSUE #1282: Add book title and chapter to the top of the main content in `BibleReader` on mobile.
Currently on mobile we have the book title and chapter visible in the navigation bar and the bottom of the screen. We also want to add the book title and chapter to the top of the main content in `BibleReader`. There is already a header with class name `sb-bible-reader-title` that is rendered in desktop mode but we want the same/similar for mobile as well for consistency.

Blitzy:
## What

Adds a book + chapter heading to the **top of the reading content on mobile**, so the reader shows where you are without relying only on the nav bar and bottom chrome. The desktop reader already renders a title header (`.sb-bible-reader-title`); mobile had no equivalent inside the content itself.

Closes #1282.

## Design

Rather than duplicate the desktop header title, this is a distinct, centered **chapter opener** styled the way a printed Bible opens a chapter:

- **Line 1** — book name (smaller, dimmed).
- **Line 2** — chapter number (large).
- No translation flag and no click target here — it's a heading, not a control. Book/translation switching stays on the nav-bar header.

It scrolls away with the text (the nav-bar header keeps the reference visible once you scroll), and it sizes with the **UI Text Size** setting like the rest of the chrome.

The opener is rendered in all three mobile swipe panels — the current chapter plus the prev/next previews — each using its own book/chapter, so the heading is already in place during a swipe instead of popping in when the swipe lands.

image
image

## Notes

- This branch has `develop` merged in, which includes the recent CSS refactor (#1368) that split `app/main.css` into co-located per-component files. The new styles live in `components/BibleReader/BibleReader.css` accordingly; the deleted `main.css` conflict was resolved by relocating these rules there.

## Testing

- Added a unit test asserting the book name and chapter number render at the top of the mobile content.
- `pnpm vitest run BibleReader.test.tsx` → 37 passed
- `pnpm check:ts` → 0 errors

Visual verification (light/dark, reader size range) done manually in the app.


Blitzy:
@KallynGowdy you think this is good to go?

KallynGowdy (approved):

KallynGowdy:
LGTM

PR #1386, opened 2026-07-09
overhaul mobile navigation: fixed 5-button toolbar, header account avatar, and Tabs button (#1365)

### LINKED ISSUE #1365: Ensure that the "Bible" logo is always centered in the mobile view
Currently, the Bible logo isn't always centered when viewing the toolbar in the mobile view. It is pushed to the left when more tools are present because there are 6 visible tools total. Instead, it should always try to limit the mobile toolbar to 5 total tools so that the bible logo can be centered.

### What it looks like
Image

### What I expected

Image

Blitzy:
## Summary

Fixes #1365 — on mobile the Bible logo drifted left whenever a 6th tool appeared in the bottom toolbar. The toolbar now **always renders exactly 5 buttons**, so the Bible logo is pinned to the center slot regardless of what's installed.

Because the count is now fixed, this also reworks how the extra controls are reached: the account button moves into the reader header, a new **Tabs** button provides a stable 5th slot, and Bookmarks tucks into **More** when it's present.

## Before / After

**Before:** `Today · You · Bible · Search · Bookmarks · More` — up to 6 tabs. With 6 equal-width tabs the center of the row falls *between* tabs 3 and 4, so the Bible logo (tab 3) sat left of center.

**After:** always 5 tabs with Bible in the middle:

- No extension toolbar tools: `Today · Search · Bible · Tabs · Bookmarks`
- With extension toolbar tools: `Today · Search · Bible · Tabs · More ‹Bookmarks, …tools›`

## What changed

**Bottom toolbar** (`BibleReaderToolbar.tsx`)

- Fixed 5-tab layout with Bible always at slot 3 (no spacers / CSS hacks — an odd, fixed count keeps the equal-width flex row centered).
- **Today** is always shown. Tapping it when the `today-screen` extension isn't installed **auto-installs it** (same path as Settings → Extensions, so the install persists) and then opens it. Its active-highlight is derived from the Today pane being open.
- New **Tabs** button opens the sidebar/Tabs drawer (mobile previously had no dedicated way to reach it).
- The 5th slot is **Bookmarks** normally, or **More** when extension toolbar tools exist — in which case Bookmarks moves *inside* the More menu.
- Removed the **You** tab (relocated to the header, below).

**Reader header** (`BibleReader.tsx`, `BibleReader.css`)

- Added an unlabeled account **avatar** between the bookmark and settings buttons; tapping it opens the Account settings view.
- When in a shared session, that slot shows the **session participants** stack instead of the avatar.

**Tabs drawer** (`Tabs.tsx`, `SidebarManager.tsx`, `BibleSelectorManager.tsx`)

- The tabs header now shows a **Close (✕)** when the drawer was opened from the toolbar (just dismisses it), and keeps the **Back arrow** when opened from the book selector (returns there). Tracked via a new `tabsOpenedFromToolbar` flag.

## Screenshots

_New Tabs mobile button and account button in header._
image

_Clicking Tabs mobile button opens directly to the Tabs screen where you can manage tabs more easily and create shared sessions._
image

_With extensions installed, the More mobile button appears and Bookmarks moves in there._
image

_In a shared session, the account button in the header becomes the mobile shared session participants button._
<p>
  image
  image
</p>

## Notes

- Merged latest `develop` (pane refactor): tab navigation calls `panes.closeAll()` to dismiss any fullscreen extension pane, and Today active-state uses develop's pane-based detection.
- Test mocks (`BibleReader.test.tsx`, `TabSlotReader.test.tsx`) gained `login`/`os` since the mobile header now renders the account avatar.


Blitzy:
@KallynGowdy this is a larger change set than i originally set out to make for #1365, but i also think it makes a lot of sense. Specifically moving the "You" account button to the quick toolbar for mobile feels right. Adding a dedicated Tabs button for mobile also feels really nice for mobile UX; it also fulfills the requirement of having 5 core buttons in all situations (no extensions / extensions). Bookmarks are still accessible from the Tabs screen on mobile when it is replaced by the More button, and obviously inside the More button as well.

PR #1409, opened 2026-07-13
feat: add support for highlighted verses with rounded corner

### LINKED ISSUE #1342: Improve highlight styling: rounded corners, less blinding highlights in dark theme, more padding

mazenemad:
fixes #1342

### DIFF
--- a/packages/seed-bible/seed-bible/app/styles/base.css
+++ b/packages/seed-bible/seed-bible/app/styles/base.css
@@ -118,8 +118,8 @@
   --sb-verse-words-of-jesus-font-family: inherit;
   --sb-verse-words-of-jesus-font-style: inherit;
 
-  --sb-selected-verse-border-bottom: 2px dashed currentColor;
-  --sb-selected-verse-text-decoration: none;
+  --sb-selected-verse-border-bottom: none;
+  --sb-selected-verse-text-decoration: underline dotted;
   --sb-selected-verse-text-decoration-color: currentColor;
 
   --sb-hebrew-subtitle-font-family: "Newsreader", serif;
--- a/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
@@ -719,36 +719,45 @@ const DARK_THEME: BibleTheme = {
     selectedTabBackground: "var(--sb-secondary-color)",
     selectedTabFontColor: "var(--sb-primary-color)",
   },
+  // Dark-theme highlights are solid, saturated-but-dark versions of each hue
+  // rather than the bright pastels used in light mode. They read as clearly
+  // colorful and distinct from one another, yet stay dark enough that white
+  // text stays legible on top. The font color is an explicit white (`#ffffff`)
+  // rather than `inherit` — inherit resolved to a non-white color in practice,
+  // so it is pinned here to guarantee the highlighted verse text stays white.
+  // (Translucent tints were tried first but blending low-opacity colors into
+  // the near-black background dragged every hue toward gray — they came out
+  // muddy and washed out, so solid colors are used.)
   highlightColors: {
     yellow: {
-      color: "#fff59d",
-      fontColor: "#333",
-      wordsOfJesusFontColor: "#5c5c5c",
+      color: "#7a6420",
+      fontColor: "#ffffff",
+      wordsOfJesusFontColor: "#ff9e80",
     },
     green: {
-      color: "#a5d6a7",
-      fontColor: "#333",
-      wordsOfJesusFontColor: "#5c5c5c",
+      color: "#2f6d3a",
+      fontColor: "#ffffff",
+      wordsOfJesusFontColor: "#ff9e80",
     },
     blue: {
-      color: "#90caf9",
-      fontColor: "#333",
-      wordsOfJesusFontColor: "#5c5c5c",
+      color: "#2f5f9e",
+      fontColor: "#ffffff",
+      wordsOfJesusFontColor: "#ff9e80",
     },
     pink: {
-      color: "#f48fb1",
-      fontColor: "#333",
-      wordsOfJesusFontColor: "#5c5c5c",
+      color: "#93395c",
+      fontColor: "#ffffff",
+      wordsOfJesusFontColor: "#ff9e80",
     },
     purple: {
-      color: "#ce93d8",
-      fontColor: "#333",
-      wordsOfJesusFontColor: "#5c5c5c",
+      color: "#5b4489",
+      fontColor: "#ffffff",
+      wordsOfJesusFontColor: "#ff9e80",
     },
     orange: {
-      color: "#ffcc80",
-      fontColor: "#333",
-      wordsOfJesusFontColor: "#5c5c5c",
+      color: "#98551c",
+      fontColor: "#ffffff",
+      wordsOfJesusFontColor: "#ff9e80",
     },
   },
 };
--- a/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
@@ -732,7 +732,7 @@ const DARK_THEME: BibleTheme = {
   // muddy and washed out, so solid colors are used.)
   highlightColors: {
     yellow: {
-      color: "#7a6420",
+      color: "#756a0a",
       fontColor: "#ffffff",
       wordsOfJesusFontColor: "#ff9e80",
     },
--- a/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
@@ -454,11 +454,13 @@ export function generateThemeCssVariables(variables: BibleTheme): string {
 }
 
 export function generateThemeCssClasses(theme: BibleTheme): string {
+  // Highlighted text keeps a readable font color; the highlight *background* is
+  // drawn behind the text by the ribbon layer (from `--sb-highlight-<id>-color`),
+  // not as a `background-color` here.
   return Object.entries(theme.highlightColors)
     .map(([colorId]) => {
       return [
         `.sb-highlight-${colorId} {`,
-        `background-color: var(--sb-highlight-${colorId}-color);`,
         `color: var(--sb-highlight-${colorId}-font-color);`,
         `&.sb-words-of-jesus { `,
         `color: var(--sb-highlight-${colorId}-words-of-jesus-font-color);`,
--- a/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
@@ -511,8 +511,8 @@ const LIGHT_THEME: BibleTheme = {
     verseFontColor: "#333",
     verseCursor: "pointer",
 
-    selectedVerseBorderBottom: "2px dashed currentColor",
-    selectedVerseTextDecoration: "none",
+    selectedVerseBorderBottom: "none",
+    selectedVerseTextDecoration: "underline dotted",
     selectedVerseTextDecorationColor: "currentColor",
 
     hebrewSubtitleFontFamily: "Newsreader, serif",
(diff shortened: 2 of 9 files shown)


Blitzy (changes requested):
These are looking so much better than the old highlights! 🥳🥳

After playing around with them and trying out different highlighting scenarios I found a couple of things that should be addressed:

1. When highlighting, the scripture text is shifting to accommodate the highlight. Highlighting should never cause the text to shift. We need to find a way to apply the highlighting without shifting the scripture text.

> [!TIP]
Perhaps we need to look into the use of `::before` selectors or maybe `box-shadow` tricks?

https://github.com/user-attachments/assets/f0e4c3df-98de-4a9c-a195-10e740d13926

2. In situations where there are contiguous verse highlights of the same color, when the highlight wraps it sometimes leaves straight edges instead of rounded corners and the ends.

> [!TIP]
One possible solution to this is looking into using `box-decoration-break`? https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/box-decoration-break

Image

Blitzy:
Fixes #1342

## Update

This PR started out adding rounded highlights via an inline `background-color` on the verse text. Review feedback and testing then surfaced a few things the inline approach fundamentally couldn't do at the same time:

- giving the tint breathing room without either shifting the line or bleeding over and covering the neighboring verse's text (a period was disappearing under the next verse's highlight), and
- making contiguous same-color verses read as one continuous ribbon while still keeping rounded corners on every line.

So the highlight rendering has been reworked: instead of painting a background on the text, highlights are now measured from the live text geometry and drawn as rounded shapes in an SVG layer **behind** the text. Behind the text, a highlight can be inflated and rounded freely; it can never shift a line and never cover a glyph.

## What it does now (behavior)

- **Rounded corners + horizontal/vertical breathing room** instead of hard square edges.
- **Continuous ribbons**: contiguous verses highlighted the same color read as one connected shape (rounded only at the run's outer corners, following the ragged right edge) rather than separate boxes with a seam where verses meet.
- **Never shifts text**: applying or removing a highlight no longer reflows the line.
- **Never covers text**: a highlight can't hide the glyphs of a neighboring differently-colored verse.
- **Muted dark-theme colors**: dark mode uses solid, dark-but-saturated hues instead of the previous blinding pastels.
- **Poetry support**: indented poetic lines get a ribbon that follows the indentation.
- **Selection underline** switched from a `border-bottom` to a `text-decoration` underline so it no longer jumps when a verse is highlighted (a border sits on the padding box; a text underline is anchored to the text).

## How it works

- `app/highlightRibbon.ts` (new): the geometry + measurement core. `collectLineRects()` measures a highlighted run's tight per-line rectangles (recursing past poetry's block-level lines to measure their inline text). `buildRibbonPath()` turns those rectangles into one rounded SVG path — closing the gaps between a run's own lines, and extending the outer edges by half the line's leading so vertically-adjacent runs of different colors meet exactly at the line boundary with no gap at any line height.
- `components/BibleReader/BibleReader.tsx`: each highlighted run is wrapped in a `display: contents` element that carries the font color and a `data-highlight-fill` attribute (a CSS var for preset colors, or the custom hex). `ChapterContent` renders the SVG layer and, in a layout effect, measures those wrappers and paints the ribbons, re-measuring on a `ResizeObserver` so it re-fits when the window, reader font size, or line height changes.
- `managers/ThemeManager.tsx`: `.sb-highlight-<id>` now sets only the (readable) font color; the background comes from the layer via the existing `--sb-highlight-<id>-color` variables.

## Testing

- New unit tests for the path geometry (`highlightRibbon.test.ts`): single line, prose run, poetry indentation, radius clamping, outer-edge slot fill, and two adjacent runs meeting with no gap.
- Updated `BibleReader.test.tsx` to assert the new structure (wrapper carries `data-highlight-fill` + font color; decorators stay clear; contiguous same-color prose verses merge; different colors don't).
- Full suite green: `pnpm check:ts`, `pnpm lint`, `pnpm vitest`, `pnpm format`.
- Manual browser checks: prose + poetry, light + dark, window resize + font-size + line-height changes, single verse / contiguous run / different-color neighbors, select-then-highlight.

Note: the rendered ribbon can't be asserted in jsdom (no layout / `Range.getClientRects`), so the geometry is unit-tested as a pure function and the visual result is verified in the browser.

## Screenshots

> _Typical book highlights:_
>
> image
> image

> _Poetry verses:_
>
> image
> image

> _Scales between smallest <-> largest font and ui scales:_
>
> image
> image

Blitzy (approved):

Blitzy:
@KallynGowdy This is ready for you to review.

This took some exploration but finally settled on something im personally very happy with. It's actually better than the YouVersion highlights in some ways imo.

KallynGowdy:
This is really cool! I wonder do we run into layout performance issues when changing front size or screen size?

KallynGowdy:
One thing I will say is that the yellow highlight on the dark theme looks a little more orange/gold than I would expect.

SEED BIBLE:


YouVersion:

Blitzy:
Going to spend more time iterating on this. I have caught a few small things I want to improve as well as I need to make sure this all works with RTL languages.

Blitzy:
## Ribbon highlight — performance baseline

**Workload:** Genesis 1, all ~31 verses highlighted (alternating colors).
**Devices:** PC (Ryzen AI 7 350 / RTX 5060), Android (Galaxy A54 5G).

Measured via the in-app ribbon perf probe (`?ribbonperf=1`). `max` = worst single `measureRibbons` pass. Available on the branch `fix/highlight-rounded-corners_perf-testing` or at the deployment https://alpha.seedbible.org/b/fix_highlight-rounded-corners_perf-testing?ribbonperf=1

| Scenario | PC max | Android max |
|---|---|---|
| Chapter load | 3.3ms | 10.3ms |
| Toggle highlight off | 3.7ms | 14.0ms |
| Toggle highlight on | 2.7ms | 13.2ms |
| Font size → max | 5.2ms | 12.2ms |
| Font size → min | 5.3ms | 11.1ms |
| Scroll ×5 | 7.5ms | 11.1ms |

**Findings**
- Every pass fits in a 60fps frame budget (16.7ms); worst observed is 14ms on Android. Settled passes are ~0.3–1.5ms (PC) / ~1.7–2.8ms (Android).
- Scrolling does **not** re-measure (`calls 0/s` on both) — the SVG scrolls with the content, geometry is not recomputed.
- The forced-layout read dominates each pass; the `JSON.stringify` signature check is negligible (≤0.1ms).
- Zero long frames (>50ms) anywhere. Android's occasional 33–42ms worst frames appear even when the probe is idle, so they're device baseline jitter, not ribbon work.

**Verdict:** No performance problem at this workload; no optimization warranted yet.

Blitzy:
Added proper support for RTL languages like Arabic.

image

packages/seed-bible/seed-bible/app/styles/base.css
```diff
@@ -118,8 +118,8 @@
   --sb-verse-words-of-jesus-font-family: inherit;
   --sb-verse-words-of-jesus-font-style: inherit;
 
-  --sb-selected-verse-border-bottom: 2px dashed currentColor;
-  --sb-selected-verse-text-decoration: none;
+  --sb-selected-verse-border-bottom: none;
+  --sb-selected-verse-text-decoration: underline dotted;
   --sb-selected-verse-text-decoration-color: currentColor;
```
KallynGowdy, line 123: @Blitzy Did we want the verse selection underline to change?

packages/seed-bible/seed-bible/managers/ThemeManager.tsx
```diff
@@ -511,8 +513,8 @@ const LIGHT_THEME: BibleTheme = {
     verseFontColor: "#333",
     verseCursor: "pointer",
 
-    selectedVerseBorderBottom: "2px dashed currentColor",
-    selectedVerseTextDecoration: "none",
+    selectedVerseBorderBottom: "none",
+    selectedVerseTextDecoration: "underline dotted",
```
KallynGowdy, line 517: @Blitzy Same question here, did we want the selection underline style to change?

packages/seed-bible/seed-bible/managers/ThemeManager.tsx
```diff
@@ -653,8 +655,8 @@ const DARK_THEME: BibleTheme = {
     verseFontColor: "#e6e6e6",
     verseCursor: "pointer",
 
-    selectedVerseBorderBottom: "2px dashed currentColor",
-    selectedVerseTextDecoration: "none",
+    selectedVerseBorderBottom: "none",
+    selectedVerseTextDecoration: "underline dotted",
```
KallynGowdy, line 659: And here

KallynGowdy (commented):
@Blitzy  Just had a question about the verse selection visuals. Also, I believe we used to have a CSS transition for highlight colors, is it possible to bring that back as well?

Blitzy:
> @Blitzy Just had a question about the verse selection visuals. Also, I believe we used to have a CSS transition for highlight colors, is it possible to bring that back as well?

@KallynGowdy Oh yes that was intentional. I forgot to showcase it previously.

I personally think the dotted line is more tasteful. It's worth noting that YouVersion uses the same dotted underline style.

> _New dotted underline in dark and light_
> 
> image
> image

> _Original dashed underline in dark and light_
> 
> image
> image

Blitzy:
> ... Also, I believe we used to have a CSS transition for highlight colors, is it possible to bring that back as well?

@KallynGowdy Brought this back. Highlights fade again, with one compromise worth explaining.

The old highlights were a `background` color on the verse elements, so a CSS `transition` faded each verse independently for free. The new highlights aren't on the text at all: they're SVG `<path>` elements behind it, and one path is a whole run (all its merged verses as one continuous shape in the path's `d`). The two things that make the ribbons look good, one continuous shape (no seams) and a vector path (sits behind text without shifting it), are the same two that make the old per-verse fade hard. You can only fade a whole path, not one verse's slice of it, and a path's `d` isn't practically animatable.

**The compromise:** a highlight fades in when it appears and out when it's removed, but editing one (extend, shrink, split, merge) snaps instead, since we can't fade just the changed part of a single path. Reflow snaps too, and recoloring cross-fades the color. So the fade covers the case that matters most, applying and clearing highlights, while mid edits just appear.

https://github.com/user-attachments/assets/2e15c75f-37bd-4b8c-9aa9-8b60ec4aeeb6

**Future option,** if we want fades everywhere:
- Path-morphing libraries (Flubber, GSAP MorphSVG) can tween one shape into another, but it's a morph, not a fade (the ribbon stretches), it breaks when a run splits into two shapes, and it adds per-frame cost that hurts low-end devices.

Blitzy:
**Chapter-change ghosting**

The fade work surfaced a latent bug: navigating between chapters could leave the previous chapter's highlights fading out on top of the new chapter's text, at their old positions.

This isn't new to the ribbons. The old `background-color` highlights had the samebugt. The reader reuses one `ChapterContent` across navigation (the chapter data swaps in place), so a transition bleeds across the chapter change. With the old highlights that meant a reused verse's background cross-fading from the old chapter's color to the new one's; with the ribbons, the fade bookkeeping carried over, so the layer faded out the old ribbons and, where verse ranges lined up, reused or cross-faded them into the new chapter's.

**Fix**: ribbon keys are now scoped to the chapter (translation + book + chapter, not just the verse range), and the layer resets its bookkeeping whenever the chapter changes. So on navigation the old chapter's ribbons drop instantly and the new chapter's fade in cleanly, with no cross-chapter matching.

Blitzy:
Oh and I also tweaked the color of the yellow highlight in dark mode to look less gold/orange tinted.

image

KallynGowdy:
I think that's a reasonable tradeoff to make for now. In the future we might consider making a new temporary path to fade in and then updating the merged one once the transition is finished. For now though, that's way too much work haha

PR #1417, opened 2026-07-14
Fixes

### LINKED ISSUE #1385: Remove non-toggleable buttons from toolbar customization (Previous Chapter, Next Chapter, Books, Side menu)
## What is happening

The toolbar customization menu lists buttons as toggle options that shouldn't be there:

- **Previous Chapter, Next Chapter, Books** — core navigation. Letting users hide these can leave them stranded, e.g. no obvious way left to change chapter or book.
- **Side menu** — an orphaned entry. This button is coded to appear only on mobile, but nothing actually renders it anymore (the mobile toolbar is hardcoded and excludes it, and desktop hides it by design). So its toggle controls a button that never shows up, which is why toggling it appears to do nothing.

## What is expected

None of these four should appear as toggle options in the toolbar customization menu. The reasons differ slightly:

- **Previous Chapter, Next Chapter, Books** — remove from the toggle list *and* keep them always visible in the toolbar. They should never be hideable.
- **Side menu** — remove from the toggle list because there's no button behind it. It should not resurface as an always-on button either; it's simply dropped from customization.

Every other button stays toggleable as it is today.

## Notes for implementation

- The toggle list is built from `toolsManager.listToolbarTools()` in `ToolbarSettingsView` (`components/SettingsPage.tsx`), which lists every registered tool regardless of whether it can actually render.
- Tool IDs involved, from `getDefaultToolbarTools()` in `managers/BibleToolsManager.tsx`: `previous-chapter`, `next-chapter`, `open-selector` (Books), and `open-sidebar` (Side menu).
- For the three nav tools: exclude them from the customization list and treat them as non-hideable (never added to the `hidden` set).
- For `open-sidebar`: it's currently dead (mobile-only visibility, but not rendered by the hardcoded mobile bar and excluded from the mobile overflow menu). At minimum, drop it from the customization list. Separately, someone should decide whether to wire it up or delete the tool entirely — but that cleanup is out of scope here.
- Any stale saved state should be reconciled: a user who previously hid Previous/Next/Books should have them forced back on.

Kabi4:
Fixes #1385


packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
```diff
@@ -433,13 +433,7 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
...
   );
 
-  const hiddenToolIds = new Set([
-    "previous-chapter",
-    "next-chapter",
-    "open-selector",
-    "open-sidebar",
-    "open-search",
-  ]);
+  const hiddenToolIds = new Set(["open-search"]);
```
KallynGowdy, line 436: @Kabi4 These tools still need to be hidden in the "mobile more menu"

This is what it looks like:
Image

This is what it should be:

Image

packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
```diff
@@ -840,7 +834,7 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
               const prev = previousChapterTool.value;
               const next = nextChapterTool.value;
               const selector = openSelectorTool.value;
-              if (!audio && !prev && !next && !selector) return null;
+              if (!audio) return null;
```
KallynGowdy: @Kabi4 This should probably be reverted too

KallynGowdy (changes requested):

packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
Kabi4: If you allow this it will be not rendering the buttons if has stale data in past

KallynGowdy: I think it will be good the way it was. Plus, right now it will hide all the buttons if there isn't an audio tool, which is not what we want

packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
Kabi4: updated

KallynGowdy (approved):

PR #1433, opened 2026-07-15
Feature/1400

### LINKED ISSUE #1400: Book selector: books should be ordered in columns instead of in rows.
## What is happening

Books in the book selector are ordered by row and wrap to new rows.

For example:
 ```
| Genesis   | Exodus  |
| Leviticus | Numbers | 
...
```

## What is expected

Books should be ordered in columns.

```
| Genesis    |
| Exodus     |
| Leviticus  | 
| Numbers    |
...
```

When the books need to wrap, it starts a new column.

Kabi4:
Fixes #1400


Blitzy:
Hey @Kabi4, the changes dont seem to address the issue. The issue is about making the placement of books in columns instead of in rows, this is really only something you see in the book selector on desktop.

Here is an image to try and describe the issue more clearly. The red marker is showing how books are placed currently. The green marker is showing how we want to try placing the books.

image

Kabi4:
@Blitzy 
image

PR #1444, opened 2026-07-15
Fix language switch revert bug in ConfigManager

### LINKED ISSUE #1443: Language selector is no longer changing the user interface language, only the scripture translation.
## What is happening

Change the language in the setting screen used to set `?lang=` in the URL. It now only sets `?translation=`. The User interface never updates to the selected language.

## What is expected.

Changing the language in the settings screen would primarily change the user interface language via `?lang=` and the secondarily a fitting default translation of scripture for that language via `?translation=`

Blitzy:
## Summary

Fixes a bug (#1443) where changing the UI language in settings would revert to the user's previously saved language. The root cause was that `ConfigManager` was re-applying the profile's language on every URL change, including the URL change caused by the language switch itself.

## Changes

**ConfigManager** (`packages/seed-bible/seed-bible/managers/ConfigManager.tsx`):
- Split the monolithic `syncConfigFromBot` effect into two separate effects with different subscription dependencies
- The profile-language effect now subscribes **only to `login.profile`**, not the URL, so it only runs when the profile loads (e.g., on login) — not on every URL change
- A separate effect re-reads other config (font size, panels, preset) whenever the URL or profile changes
- Added detailed comments explaining why the profile-language effect must not subscribe to URL changes, since `?lang=` is owned by `I18nManager` and is the single source of truth for the URL ↔ language relationship

**I18nManager** (`packages/seed-bible/seed-bible/i18n/I18nManager.tsx`):
- Fixed the `language` signal's setter to route URL-driven `?lang=` changes through `i18n.changeLanguage()` instead of assigning directly
- This ensures the actual i18next translations reload when the URL changes (e.g., deep links, browser back/forward), preventing desync between the `language` signal and what `t()` actually returns

**Tests**:
- Added comprehensive test suite (`test/unit/seed-bible/managers/ConfigManager.test.ts`) covering the regression: verifies that a user's newly selected language is not reverted when the URL changes
- Added tests to `I18nManager.test.ts` verifying that `?lang=` changes in the URL trigger actual i18n reloads and that language selection writes `?lang=` correctly

## Implementation Details

The bug occurred because:
1. User picks a new language → `I18nManager` writes `?lang=fr` to the URL
2. URL change triggers `ConfigManager`'s effect
3. Effect reads the profile's still-unsaved previous language (`lang: "en"`) and applies it
4. User sees the language revert, even though the profile eventually saves the new language

The fix enforces a clear ownership model: `I18nManager` owns the URL ↔ `i18n.language` sync (including reloading translations on URL changes), while `ConfigManager` only applies the profile's language when the profile itself changes, not on URL changes.

https://claude.ai/code/session_012boq6vwxsFMXA98cB1LuRo

### DIFF
--- a/packages/seed-bible/seed-bible/i18n/I18nManager.tsx
+++ b/packages/seed-bible/seed-bible/i18n/I18nManager.tsx
@@ -187,7 +187,16 @@ export function createI18nManager(
         return null;
       },
       set value(newValue: string | null) {
-        language.value = newValue ?? defaultLanguage;
+        // Invoked when the URL's `?lang=` changes on its own — deep links and
+        // browser back/forward. Route through `i18n.changeLanguage` so the
+        // actual translations reload; the `languageChanged` listener above then
+        // moves the `language` signal to match. Assigning the signal directly
+        // here would desync `i18n.language` (and therefore every `t()` call)
+        // from the URL.
+        const next = newValue ?? defaultLanguage;
+        if (next !== i18n.language) {
+          void i18n.changeLanguage(next);
+        }
       },
     },
   });
--- a/packages/seed-bible/seed-bible/i18n/I18nManager.tsx
+++ b/packages/seed-bible/seed-bible/i18n/I18nManager.tsx
@@ -233,6 +233,21 @@ export function createI18nManager(
     ensureTranslationsLoaded = loadTranslations;
   };
 
+  /**
+   * Wired by SeedBibleState to persist the user's chosen UI language (e.g. to
+   * their profile). Invoked ONLY for selector-driven changes via
+   * `requestLanguageChange` — never for URL-driven changes (deep links,
+   * browser back/forward) or profile-applied changes, so opening a shared
+   * `?lang=` link updates the view without overwriting the account's saved
+   * language.
+   */
+  let persistLanguage: ((language: string) => void) | null = null;
+  const setLanguagePersister = (
+    persister: ((language: string) => void) | null
+  ) => {
+    persistLanguage = persister;
+  };
+
   const changeLanguage = i18n.changeLanguage.bind(i18n);
 
   const applyBibleTranslationForUiLanguage = async (uiLanguage: string) => {
@@ -267,6 +282,10 @@ export function createI18nManager(
     if (nextLanguage !== language.value) {
       await changeLanguage(nextLanguage);
     }
+    // Persist the user's explicit selection. Only selector-driven changes reach
+    // this function; URL-driven changes go through the `syncSignalsToUrl`
+    // setter above and are deliberately left un-persisted.
+    persistLanguage?.(nextLanguage);
     await applyBibleTranslationForUiLanguage(nextLanguage);
   };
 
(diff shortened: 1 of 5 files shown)


KallynGowdy:
@Blitzy am I correct in understanding that now URL variables won't cause the user's profile data to be updated? Instead it will only be updated when they change the language via the selector themselves?

KallynGowdy:
@Blitzy There seems to still be an issue where loading with a `lang` parameter causes some weird behavior.

Steps to reproduce:

1. Load Seed Bible, login, and select a language.
2. Refresh just to double check that was saved to your profile
3. Open a new tab in the same browser and load seed bible
  1. Observe that it loads the language saved to your profile
4. Add `lang` parameter with a different language to the URL and refresh
5. Observe that the requested language is loaded before being replaced by your profile language
6. Go to the first tab and refresh
7. Observe that your profile language is now set to the language you added to the URL in step `4`

Here's a quick video:
https://youtu.be/KBZLv1SnATM

Blitzy:
> @Blitzy There seems to still be an issue where loading with a `lang` parameter causes some weird behavior.
> 
> Steps to reproduce:
> 
> 1. Load Seed Bible, login, and select a language.
> 2. Refresh just to double check that was saved to your profile
> 3. Open a new tab in the same browser and load seed bible
> 4. Observe that it loads the language saved to your profile
> 5. Add `lang` parameter with a different language to the URL and refresh
> 6. Observe that the requested language is loaded before being replaced by your profile language
> 7. Go to the first tab and refresh
> 8. Observe that your profile language is now set to the language you added to the URL in step `4`
> 
> Here's a quick video: https://youtu.be/KBZLv1SnATM

@KallynGowdy I just discovered this as well. Looking into it.

A relevant question though, do we actually want navigating to a link with `?lang=` included actually to affect your personal choice given that you may have an actual preference baked into your own user settings?

KallynGowdy:
My gut says that if you go to a URL that has `lang=ar` in it, then you should be seeing the Arabic version, but I can't actually think of any cases where it is really important that the language is displayed as specified in the URL when the user has a preference saved. It is really important that the `lang` URL parameter is obeyed for guest users and for users that don't have a preference (never set via the selector), but beyond that I think we're good to have the user preference override the URL parameter

Blitzy:
Ok I agree with that. 

1. Users with a language set in their preferences should override any language set in the incoming URL.
3. Users **without** a language set in their preferences should use the language set in the incoming URL.
2. Users that are not signed in should use the language set in the incoming URL.

Blitzy:
The overwrite issue is fixed as of `74ea0e4` (live in the branch deploy). Persistence now happens only when you pick a language in the selector — loading a `?lang=` URL is view-only and never writes to your profile.

The agreed precedence is in place:
1. Signed-in user **with** a saved language → profile wins over the URL param.
2. Signed-in user **without** a saved language → URL param is used.
3. Not signed in → URL param is used.

Added tests covering all three cases plus a regression test for the boot-with-`?lang=` scenario (load `?lang=de` with a profile of `en` → shows `en`, and the profile stays `en`).

---

Blitzy:
> 1. Users with a language set in their preferences should override any language set in the incoming URL.
> 2. Users **without** a language set in their preferences should use the language set in the incoming URL.
> 3. Users that are not signed in should use the language set in the incoming URL.

@KallynGowdy 

I've manually tested these three scenarios and verified they are working now.

KallynGowdy (approved):

PR #1453, opened 2026-07-16
Make Bible tab icon respect active theme color

KallynGowdy:
## Summary

The Seed Bible logo on the Bible reader toolbar tab was hard-coded to display in the light theme's orange color, even when the app switched to a dark theme with a different primary color. This PR fixes the icon to dynamically match whatever the active theme's primary color is.

## Changes

- **BibleReaderToolbar.css**: Replaced the hard-coded `filter` chain (which produced a fixed orange tint) with a CSS mask-image approach. When the Bible tab is active, the raster logo image is hidden and a `::after` pseudo-element paints the logo shape using `currentColor`, which automatically inherits the active theme's primary color from the parent button's `color` property.

- **ThemeManager.tsx**: Added four new theme variables to the `BibleThemeVariables` interface and populated them in both `LIGHT_THEME` and `DARK_THEME`:
  - `emptyPaneToolbarButtonBorder` — border color for toolbar buttons in empty panes
  - `emptyPaneToolbarButtonBackground` — background color for those buttons
  - `emptyPaneToolbarButtonColor` — font/icon color for those buttons
  - `emptyPaneToolbarButtonBorderHover` — border color on hover (uses primary color to signal interactivity)

## Implementation Details

The CSS mask-image technique works by:
1. Making the icon container `position: relative` so the `::after` pseudo-element can overlay it
2. Hiding the original raster image with `opacity: 0` when the tab is active
3. Painting a colored rectangle over the same space using the logo PNG as a mask, so only the logo shape shows through in the active theme's primary color

This approach is theme-agnostic — the icon will automatically match any theme's primary color without needing to update the CSS filter values.

https://claude.ai/code/session_01UYyKCk3ayRRmPQXsLCfdEY


KallynGowdy:
this actually works now haha

Blitzy (approved):

PR #1464, opened 2026-07-18
Stop blocking chapter/book navigation on in-flight text requests

### LINKED ISSUE #1414: Changing chapters / books should not be blocked by inflight requests for text content.
## What is happening

When trying to flip through chapters very quickly by tapping the next button in quick succession, navigation gets blocked by the current inflight request for the text content of that book and chapter. You sure unable to change it until the inflight request is finished. 

## What is expected

Navigating books and chapters should not be blocked by inflight requests for text content. You should freely flip through books and the text requests should adjust accordingly.

KallynGowdy:
Tapping next/previous quickly used to be gated by `loading`, so a slow
in-flight chapter fetch would ignore further taps until it resolved.
Navigation methods now queue behind one another instead of racing:
each call always reads fresh state once it is its turn, so rapid taps
advance correctly instead of being dropped or racing a stale response
into the current view. The toolbar chevrons, book/chapter selector,
and keyboard navigation no longer disable/ignore input while a request
is in flight.

Fixes #1414

Claude-Session: https://claude.ai/code/session_01TXzJdVDxjbMzseoAuSb3dS


Blitzy (changes requested):
This needs needs more work. The new queuing mechanism feels slow just in a different way now.

Ideally, navigation should not be tied directly to chapter loading state at all. If I click the next chapter button 10 times from Genesis 1, I should end up on Genesis 11 regardless of the loading state of the chapters in-between. If possible it would not be a queueing system at all and we would be able to handle flipping through chapters rapidly and canceling any `GET` requests for chapter content along the way if the user rapidly moves away from the chapter before it is loaded.

Blitzy:
@claude review

Blitzy:
> [!NOTE]

## Summary

The latest commit ("Compute chapter navigation from book metadata instead of waiting on it") is a rework in response to the earlier CHANGES_REQUESTED review: instead of just queuing navigation calls behind each other (the first fix for issue #1414), every navigation method now computes its target directly from the translation's book metadata and cancels the previous in-flight chapter fetch via a shared generation counter + `AbortController`. That matches what was asked for — repeated "next" taps now advance multiple chapters without waiting on the network, and stale in-flight fetches are aborted instead of racing into the current view. Test coverage for the happy paths (multi-tap advance, book-boundary crossing, cross-method supersession) looks solid.

Digging into the new cancellation plumbing turned up two correctness gaps, detailed below.

## Issues

### 1. (High) Aborting one reader's request can silently kill an unrelated reader's chapter load

**Where:** `packages/seed-bible/seed-bible/managers/FreeUseBibleAPI.tsx`, `_getJson()` (around lines 1105-1137).

**What's happening:** `_getJson` caches in-flight requests by URL so concurrent callers for the same chapter share one `fetch()`. Only the *first* caller for a given URL gets its `AbortSignal` wired into that `fetch()` call — every later caller that hits the cache just gets the same shared `Promise` back (correctly, per the comment on line ~1113, so a joiner's own abort can't affect others). But the reverse isn't handled: if the *original* caller's signal later aborts, the single underlying `fetch()` is cancelled and the shared promise rejects with `AbortError` for **every** caller relying on it — including callers who never asked to cancel anything and whose own navigation is still current.

Since `FreeUseBibleAPI` is instantiated once and shared across the whole app (`SeedBibleStateManager.tsx:410` → `createBibleDataManager(api)`, reused by every tab/reading state created in `TabsManager.tsx` and `SessionsManager.tsx`), this is reachable whenever two independent `BibleReadingState`s request the same chapter URL at close to the same time — e.g. two tabs both landing on the same chapter, or two participants in a shared/multiplayer session. In `BibleReadingManager.tsx`, every navigation method's catch block treats any `AbortError` as "we've been superseded, ignore it" (`if (isAbortError(err) || !isCurrentNavigationTurn(generation)) { return; }`), so the unrelated reader's tap just silently does nothing — no `error.value`, no loaded chapter, `loading` simply flips back to `false`.

**Reproduction (mirrors the existing "unrelated caller" test in `FreeUseBibleAPI.test.ts`, but with the roles reversed):**
1. Caller A requests a URL **with** a signal (cache miss) → `_getJson` starts `fetch(url, { signal: A })` and caches the resulting promise.
2. Caller B requests the same URL before A resolves (cache hit) → gets the exact same promise back, regardless of whether B passed its own signal.
3. Abort A's controller (simulating A's navigation being superseded by another tap in that tab).
4. B's promise also rejects with `AbortError`, even though B's own navigation was never cancelled.

In the app, this shows up as: open two tabs on the same translation, get them both to request the same next/previous chapter around the same moment, then tap "next" again quickly in the first tab before its request resolves — the second tab's pending "next" tap can silently fail to load.

**Suggested fix:** Don't let one subscriber's cancellation tear down the network request other subscribers are relying on. For example, give `_getJson` its own `AbortController` per URL and treat each caller's `options.signal` as "stop waiting for me" rather than "stop the underlying fetch" — e.g. `Promise.race([sharedRequest, rejectWhenAborted(options?.signal)])` per caller — so the shared `fetch()` only actually aborts once every interested caller has walked away (or simply never abort the shared fetch at all, and only shape each caller's own returned promise).

### 2. (Medium) A network failure while resolving next/previous chapter's target escapes as an unhandled rejection instead of setting `state.error`

**Where:** `packages/seed-bible/seed-bible/managers/BibleReadingManager.tsx`, `navigateAdjacent()` (around lines 1628-1637).

**What's happening:** In the non-extension-handled branch of `navigateAdjacent`, the target position is computed with:
```ts
target = await chainPosition(async (base) => {
  if (!base) return null;
  const books = await getBooksForPosition(base.translationId);
  return computeAdjacentPosition(books, base, direction);
});
```
`getBooksForPosition` falls through to `dataManager.getTranslationBooks(translationId)` (no signal, plain network call) whenever the reading state's cached `translationBooks` doesn't already match — e.g. right after switching translations, before that translation's book metadata has been cached. This `await` is **not** wrapped in a try/catch, unlike every other navigation method (`selectTranslation`, `selectTranslationAndChapter` both wrap their equivalent `chainPosition(...)` call and set `error.value` on failure). If that book-metadata fetch fails (server error, offline, etc.), the rejection propagates straight out of `navigateAdjacent`.

Both real call sites drop that rejection on the floor: `TabsLayout.tsx` calls it as `void readingState.loadNextChapter();`, and `BibleToolsManager.tsx` calls `context.readingState.loadPreviousChapter();`/`loadNextChapter();` with no `.catch()` at all. Net effect: the user taps next/previous, nothing happens, no error is shown, and the failure only surfaces as an unhandled promise rejection in the console.

**Reproduction:**
1. Get the reading state into a spot where `translationBooks.value` doesn't match the current `translationId` (e.g. right after `selectTranslation()` on a translation whose books aren't cached yet).
2. Make `dataManager.getTranslationBooks` reject for that translation (e.g. mock a 500 response).
3. Call `loadNextChapter()` (or `loadPreviousChapter()`).
4. Observe the returned promise rejects and `state.error.value` stays `null` — contrast with the existing "catches errors and stores them in state.error" test, which only covers the chapter-fetch failure path inside `selectChapter()`, not this one.

**Suggested fix:** Wrap the target-computation `await chainPosition(...)` in the `else` branch of `navigateAdjacent` in the same try/catch pattern `selectTranslation` already uses, setting `error.value = err instanceof Error ? err.message : "Failed to load ${direction} chapter."` on failure instead of letting it propagate.

## Nit (not a blocker)

`BibleToolsManager.tsx`'s previous/next toolbar buttons now derive `isDisabled` purely from `chapterData.value?.previousChapterApiLink`/`nextChapterApiLink` (the server-provided link on the currently *committed* chapter), while the actual navigation target is computed independently from local book-order metadata (`computeAdjacentPosition`). These two boundary checks are separate sources of truth; if a translation's book catalog (chapter counts/order) and the chapter payload's own next/previous links ever disagree, the button's enabled state could mismatch what tapping it actually does. Probably fine in practice given how the data is generated, just flagging the coupling.


---

KallynGowdy:
Addressed both issues:

1. **Ref-counted cancellation.** `FreeUseBibleAPI` now owns its own `AbortController` per URL, never a caller's own signal. Each caller subscribing to a shared in-flight request is tracked by count; a caller's own abort only rejects their own promise, and the real underlying request is only cancelled once every subscriber (including any with no signal, who can never voluntarily leave) has walked away. Added tests covering: one caller's abort not affecting an unrelated cache-hit subscriber, and the underlying request only being cancelled once all subscribers have aborted.

2. **Unhandled rejection fixed.** Wrapped `navigateAdjacent`'s target-computation step in a try/catch (matching the pattern already used by `selectTranslation`/`selectTranslationAndChapter`), so a failure loading book metadata mid-navigation now sets `state.error` instead of escaping as an unhandled promise rejection. Added a test reproducing this via an extension `navigate` outcome that lands on a translation whose book metadata fetch fails.

All 3104 tests pass, type check and lint are clean.

---

PR #1467, opened 2026-07-18
fix #1455

### LINKED ISSUE #1455: Spacing in settings pages on mobile
These pages could use more space on the edges to separate the content from the sides of the screen.

AmitPant9999:
Fixes:  #1455

### DIFF
--- a/packages/seed-bible/seed-bible/components/Tabs/Tabs.css
+++ b/packages/seed-bible/seed-bible/components/Tabs/Tabs.css
@@ -593,6 +593,11 @@
   flex: 1;
   padding: 5px;
 }
+@media (max-width: 480px) {
+  .sb-sidebar-settings-view {
+    padding: 5px 12px;
+  }
+}
 
 .sb-sidebar-settings-close-button {
   border: none;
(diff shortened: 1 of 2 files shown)


packages/seed-bible/seed-bible/components/Tabs/Tabs.css
```diff
@@ -593,6 +593,11 @@
   flex: 1;
   padding: 5px;
 }
+@media (max-width: 480px) {
+  .sb-sidebar-settings-view {
+    padding: 5px 12px;
```
Blitzy, line 603: @AmitPant9999 make sure we continue to use relative units (`rem`). Relative units will automatically scale with the "UI size" property. Fixed units like `px` should be reserved only for things that we explicitly do not want to scale with "UI size" (which is a pretty small list).

I think in this case `padding: 5x 12px` should become `rem` units.

Blitzy (dismissed):

packages/seed-bible/seed-bible/components/Tabs/Tabs.css
AmitPant9999: will fix it @Blitzy

Blitzy (dismissed):

Blitzy (approved):

PR #1468, opened 2026-07-18
Fix false-positive host-disconnect kick on mobile resume (#1346)

### LINKED ISSUE #1346: Getting kicked out of shared sessions when switching apps on mobile
## What is happening

On mobile, switching away from the app to another app and back kicks you out of the shared session you were in.

_Video showing mobile device joining a shared session via QR code, then switching to another app and then returning:_
https://youtube.com/shorts/IWOocwMrnus

## What is expected

Shared sessions should survive switching between apps on mobile. Returning to the app should restore you to the session you were in, reconnecting automatically if needed rather than dropping you out.

KallynGowdy:
The host-disconnect grace timer judged whether a session's host had
left using only the local client's own connectedUsers view. Right
after a phone resumes from the background, that view can read "host
missing" for a few seconds even though the host never left, so the
timer fired and disposed the participant's own session connection.

Add a live isSynced signal (subscribed to the shared document's
onStatusUpdated for the session's lifetime, not just the first sync)
so the grace timer only arms while this client's own connection is
trustworthy, adds a short post-visibility-resume buffer as further
insurance, and re-verifies both host presence and sync state right
before acting instead of trusting an 8-second-old snapshot. Also add
toast notifications (reconnecting / reconnected / host left) so a
tab closing has a visible explanation instead of silently vanishing.

Claude-Session: https://claude.ai/code/session_01CfQ63s2P1u4WnQWnJ2GBEJ

Fixes #1346


Blitzy (changes requested):
@KallynGowdy These changes did not resolve the core issue. I did a walkthrough video showing two devices simultaneously of what is currently happening as of this change set.

https://www.youtube.com/watch?v=eNMxkUp1hqE

KallynGowdy:
@Blitzy addressed some more issues, including one with CasualOS itself

Blitzy:
> [!NOTE]

## Summary

This PR is the second attempt at fixing issue #1346 (mobile users getting kicked from shared sessions when they background/resume the app). The first attempt (commit `43c9902`) got a CHANGES_REQUESTED review from Blitzy with a two-device video showing the fix didn't actually work — the latest commit (`12ee358`, "Recover session presence after a dropped connection") addresses a second root cause: presence going silently and permanently empty after a reconnect, because the CasualOS client's own peer cache filters out the replayed peer list as duplicates. Kallyn's comment today ("addressed some more issues, including one with CasualOS itself") lines up with that commit, and `pull_request.ready_for_review` firing now is presumably in response to it.

The core logic (`sessionPresenceIsTrustworthy`, `rebuildRemoteClientsSubscription`, `isSynced`) is sound and the new tests in `SeedBibleStateManager.test.ts` and `SessionsManager.test.ts` genuinely exercise the fixed behavior (fake timers advanced to specific boundaries, toast text asserted against the real `en.json` strings via the real i18n module, dispose/teardown paths checked) rather than just asserting things don't throw. I don't see a functional bug in the reconnect/eject logic itself. The items below are smaller process/robustness points.

---

### Medium — `visibilitychange` listener is added with no cleanup, unlike every other DOM listener in this file

**File:** `packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx:1073`

```js
if (typeof document !== "undefined") {
  document.addEventListener("visibilitychange", () => { ... });
}
```

Every other place this file attaches a DOM listener (e.g. the `resize` handler ~30 lines earlier) wraps it in `effect(() => { ...; return () => window.removeEventListener(...) })` so it's torn down. This one is a bare `addEventListener` with no matching `removeEventListener` and no cleanup path at all.

In production this is harmless today because `createSeedBibleState()` is a singleton with no `dispose()`, so it's only ever called once. But `test/unit/seed-bible/managers/SeedBibleStateManager.test.ts` calls `createSeedBibleState()` (via `createState()`/`createStateWithTwoTabs()`) roughly 30 times in this one file, and each call registers another listener on the shared jsdom `document` that never goes away. Every later `document.dispatchEvent(new Event("visibilitychange"))` — including the one added by this PR's own "suppresses arming the timer..." test — fires all previously-accumulated handlers from earlier, already-finished tests, not just the current one.

**Reproduction:** add a `console.log` inside the `visibilitychange` handler and run `pnpm vitest run SeedBibleStateManager.test.ts`; the log count on the last test's dispatch will match the number of `createState*()` calls that ran before it, not one.

This doesn't currently fail an assertion (each test only reads its own `state.app.currentToast`/`state.tabs`), but it's a real leak that will keep growing as more tests are added to this file, and it breaks the file's own convention.

**Suggested fix:** wrap it the same way the resize handler is wrapped:
```js
effect(() => {
  if (typeof document === "undefined") return;
  const handleVisibilityChange = () => { ... };
  document.addEventListener("visibilitychange", handleVisibilityChange);
  return () => document.removeEventListener("visibilitychange", handleVisibilityChange);
});
```

---

### Low — reliance on an undocumented private SDK field has no regression coverage if it silently stops matching

**File:** `packages/seed-bible/seed-bible/managers/OsManager.tsx:212-244` (`clearBranchDeviceCache`)

The fix for the "presence goes empty forever" bug reaches into `InstRecordsClient`'s private `_connectedDevices` map and deletes a key built from `${recordName ?? ""}/${inst}/doc/${docName}` to mirror the SDK's own internal cache key. This is called out in the comment as a workaround for a CasualOS client bug, and it's defensively written (optional chaining, so it quietly no-ops instead of throwing if a future SDK version renames/reshapes the field).

The risk: the unit tests for this (`SessionsManager.test.ts`, "presence recovery after a reconnect") mock `os.clearBranchDeviceCache` itself, so they verify the calling code invokes it with the right arguments, but nothing verifies the private field name/key format still matches reality on a CasualOS SDK bump. If a future `@casual-simulation/*` upgrade renames `_connectedDevices`, this would silently become a no-op (by design, so it won't throw) and issue #1346 would quietly regress with no test failure to flag it.

**Suggested fix:** not blocking, but consider a short comment/TODO linking to an upstream CasualOS issue (if one exists or gets filed) so this is easy to find and re-verify on the next SDK bump, or a small smoke/integration test against the real `InstRecordsClient` shape (even just `expect(new InstRecordsClient(...) as any)._connectedDevices).toBeInstanceOf(Map)`) that would fail loudly if the internal shape changes.

---

### Low — all ~24 non-English locale files were machine-translated for the 3 new keys, which reads as inconsistent with CLAUDE.md

Commit `1167c1d` ("feat: added google trasnlations for new session-host-* keys") adds `session-host-reconnecting` / `session-host-reconnected` / `session-host-left` to every locale file (e.g. `es.json`, `zh.json`, etc.), not just `en.json`. CLAUDE.md for this repo says:

> When adding or updating any translation key, only update `en.json`... don't update other translation files, leave those for the professionals.

This was very likely necessary to satisfy `i18nIncompleteTranslationsRule` (a `type: "problem"` ESLint rule that errors on any locale file missing a key present in `en.json`), and the `build (24.x)` check did pass on this commit, so it's not a CI-breaking issue. But it does mean every non-English reader sees Google-Translate-quality strings for these three keys until a professional pass corrects them — worth a quick confirmation from the team that "auto-translate new keys to unblock lint, professionals fix later" is the intended workflow here rather than an accidental side-effect of chasing a lint error, since it reads as contradicting the documented policy.

---

### Info — PR currently has a merge conflict with `develop`

GitHub reports `mergeable_state: "dirty"` for this PR against `develop` as of the latest push. Not a code-review finding, but worth a rebase/merge before this lands — the `i18n` locale files in particular are likely to conflict given how much unrelated locale churn has landed on `develop` recently.

---

**Testing performed for this review:** read the full diff against the PR's merge-base with `develop` (`627d503`) locally, checked the `OsManager`/`SessionsManager`/`SeedBibleStateManager` changes against the CasualOS SDK usage elsewhere in the codebase, read issue #1346 (including the root-cause writeup) and the prior CHANGES_REQUESTED review/video from Blitzy, and read through both new test files to check they assert on real state transitions rather than trivially passing.

---

KallynGowdy:
@Blitzy This appears fixed to me now

Blitzy:
> [!NOTE]

## Follow-up review

This is a re-review after the previous pass on this PR (Aug 11). Since then: `f7bb8df` addressed the Medium finding from that review, and two merge commits (`11673f9`, `690491b`, `c67a8926`) pulled in `develop` to resolve the merge conflict flagged as an Info item. I diffed the current head against `develop`'s current tip (`d91ce0a4`, which the PR's base already matches — no drift) and re-checked the specific files the conflict touched.

**Medium — `visibilitychange` listener leak: fixed.** `packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx` now wraps the listener in `effect(() => { ...; return () => document.removeEventListener(...) })`, matching the pattern used elsewhere in the file and exactly the suggested fix. Also gained an `!import.meta.env.SSR` guard consistent with the codebase's other SSR-guarded effects.

**Info — merge conflict: resolved.** The conflicting files (`OsManager.tsx`, `SeedBibleStateManager.tsx`, `test/unit/seed-bible/managers/TabsManager.test.ts`) came through the merge clean — I read the full diff on each against `develop` and the resolution kept both sides' logic (the presence-recovery fix and whatever landed on `develop` in the meantime). No regression introduced by the merge.

**Still open, still non-blocking (carried over from the last review):**
- Low — `OsManager.tsx`'s `clearBranchDeviceCache` still reaches into `InstRecordsClient`'s private `_connectedDevices` field with no regression coverage against the SDK's actual shape. Same risk as before: a future `@casual-simulation/*` bump could silently no-op this and reopen No. 1346 with no test failure to flag it.
- Low — the ~24 non-English locale files still carry machine-translated strings for the three new `session-host-*` keys (commit `1167c1d`), which still reads as contradicting CLAUDE.md's "only update en.json" policy, even though it's necessary today to satisfy the `i18nIncompleteTranslationsRule` lint check.

Neither is new, and neither blocks merge — just noting they weren't part of this round's changes.

**Tests:** the new/updated tests in `SessionsManager.test.ts` (`isSynced` tracking, dispose-during-rebuild, no-rebuild-on-initial-sync) and `SeedBibleStateManager.test.ts` (toast text asserted against the real `en.json` strings, grace-timer boundaries with fake timers) are genuine — they assert on specific state transitions and would fail if the fix were reverted, not just "doesn't throw."

**CI:** `build (24.x)` is green on the current head (`c67a8926`).

**One thing worth flagging that isn't a code issue:** there's still an outstanding `CHANGES_REQUESTED` review from Blitzy (2026-07-20, commit `1167c1d`) with a two-device video showing the fix didn't work at that point in time. That review predates the actual root-cause fix (`12ee358`, Aug 3) and Kallyn's "this appears fixed to me now" comment (today). GitHub is still reporting this PR as blocked, which is very likely that stale review rather than anything in the current diff — worth getting Blitzy to re-test and update/dismiss it now that the presence-recovery fix has landed, since a passing CI run won't clear that on its own.

No new bugs found in this round.

---

Blitzy:
This does seem to be working pretty well now. None of the issues i reported before seem to be occurring anymore. Reconnection to the session seems pretty reliable, and I'm not getting kicked from the session anymore.

PR #1469, opened 2026-07-18
feat: add sitemap generator

KallynGowdy:
Generate a sitemap covering every chapter of every book in every Bible
translation, one child sitemap per translation plus a sitemap index and a
robots.txt.

Each chapter URL carries the UI locale (?lang=) that maps to the
translation's Bible language — Spanish translations get the Spanish UI,
English translations the English UI — reusing the existing
UI_TO_BIBLE_LANGUAGE_CODES mapping (now exported) rather than emitting the
full translation × every-UI-language cross product. Translations in a
language with no supported UI locale are still listed without ?lang=;
--mapped-only restricts output to translations that map to a UI locale.

- script/lib/sitemap.ts: pure, unit-tested helpers (language-map inversion,
  URL building, XML rendering, chunking to the 50k-per-file limit, filename
  sanitizing).
- script/generate-sitemap.ts: fetches translations + books via
  FreeUseBibleAPI and writes into the client build output. Best-effort — a
  Bible-API failure warns and exits 0 so a deploy is never blocked
  (--strict to fail). Skips non-main branch builds, whose public files are
  not deployed.
- Wired as "pnpm sitemap" and appended to "pnpm build". Output is served by
  the host's existing .xml/.txt reverse-proxy; no new server route needed.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_017Em949FfugKLX9UiqGj6Kt

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test": "vitest",
     "test:watch": "vitest watch",
     "dev": "bun server/index.ts",
-    "build": "pnpm build:client && pnpm build:ssr && pnpm build:server",
+    "build": "pnpm build:client && pnpm build:ssr && pnpm build:server && pnpm sitemap",
     "build:client": "vite build",
     "build:ssr": "vite build --ssr standalone/entry-ssr.tsx",
     "build:server": "bun build --target=bun --external express --external vite --outfile=server/dist/index.js ./server/index.ts",
@@ -22,6 +22,7 @@
     "build-size-report": "tsx script/build-size-report.ts",
     "i18n": "tsx script/i18n.ts",
     "list-bible-translations": "tsx script/list-bible-translations.ts",
+    "sitemap": "tsx script/generate-sitemap.ts",
     "format": "prettier --write \"packages/**/*.{js,jsx,ts,tsx,css}\"",
     "format:changed": "pretty-quick",
     "prepare": "husky && tsx script/setupMergeDriver.ts"
--- a/packages/seed-bible/seed-bible/managers/BibleReadingManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/BibleReadingManager.tsx
@@ -427,7 +427,7 @@ const FALLBACK_TRANSLATION: TranslationWithLanguage = {
  * Includes aliases so we can match the nearest available text even when the
  * preferred hardcoded ID is missing from the loaded catalog.
  */
-const UI_TO_BIBLE_LANGUAGE_CODES: Record<string, string[]> = {
+export const UI_TO_BIBLE_LANGUAGE_CODES: Record<string, string[]> = {
   am: ["amh"],
   ar: ["arb", "ara"],
   bn: ["ben"],
--- a/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
@@ -887,6 +887,18 @@ export function createSeedBibleState(
       canonicalUrl.searchParams.set("chapter", String(chapter.chapter.number));
     }
 
+    // Preserve an explicit `?lang=` so the canonical is self-referential for
+    // the language-specific URLs the sitemap emits (otherwise search engines
+    // collapse every `?lang=` variant onto the lang-less URL and none of them
+    // index distinctly). Echo only the explicit URL param — deriving it from
+    // the active i18n language would make the canonical vary by
+    // Accept-Language, which must not happen. Set last to match the sitemap's
+    // `translation,book,chapter,lang` ordering.
+    const lang = currentUrl.searchParams.get("lang");
+    if (lang) {
+      canonicalUrl.searchParams.set("lang", lang);
+    }
+
     return `${canonicalUrl.pathname}${canonicalUrl.search}`;
   });
 
--- a/script/generate-sitemap.ts
+++ b/script/generate-sitemap.ts
@@ -1,0 +1,358 @@
+/**
+ * Generates the Seed Bible sitemap: one child sitemap per translation (every
+ * chapter of every book) plus a sitemap index and a robots.txt.
+ *
+ * Each chapter URL carries the UI locale (`?lang=`) that maps to the
+ * translation's Bible language — Spanish translations get the Spanish UI,
+ * English translations the English UI — so we never emit the full
+ * translation × every-UI-language cross product. Translations in a language
+ * with no supported UI locale are still listed, but without `?lang=` (the app
+ * then resolves the visitor's language). Pass `--mapped-only` to emit only
+ * translations that map to a supported UI locale.
+ *
+ * Output lands in the client build directory (`standalone/dist/client/` by
+ * default), which the CD workflow's main-only sync uploads to the bucket root;
+ * the host server already reverse-proxies `.xml`/`.txt` requests to the CDN, so
+ * `/sitemap.xml`, `/sitemaps/*.xml`, and `/robots.txt` are served with no extra
+ * server route.
+ *
+ * The Bible catalog is external, so the run is best-effort: a fetch failure
+ * logs a warning and exits 0 (so a deploy is never blocked) unless `--strict`
+ * is passed.
+ *
+ * Usage:
+ *   pnpm sitemap
+ *   pnpm sitemap --base-url=https://prod.seedbible.org --endpoint=https://bible.helloao.org/
+ *   pnpm sitemap --mapped-only --strict --out=standalone/dist/client
+ */
+import { mkdir, rm, writeFile } from "node:fs/promises";
+import path from "node:path";
+import {
+  FreeUseBibleAPI,
+  getDefaultAPIEndpoint,
+  type Translation,
+} from "@packages/seed-bible/seed-bible/managers/FreeUseBibleAPI";
+import {
+  bibleLanguageToUiLocale,
+  chapterUrlsForTranslation,
+  chunk,
+  renderSitemapIndex,
+  renderUrlset,
+  uniqueSitemapName,
+  MAX_URLS_PER_SITEMAP,
+  type BookChapters,
+  type SitemapIndexEntry,
+} from "./lib/sitemap";
+
+const DEFAULT_ORIGIN = "https://prod.seedbible.org";
+const DEFAULT_OUT_DIR = path.join("standalone", "dist", "client");
+const SITEMAPS_SUBDIR = "sitemaps";
+const DEFAULT_CONCURRENCY = 8;
+
+interface Options {
+  origin: string;
+  endpoint: string;
+  outDir: string;
+  mappedOnly: boolean;
+  strict: boolean;
+  concurrency: number;
+}
+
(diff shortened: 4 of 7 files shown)


Blitzy:
## Code review follow-ups (non-blocking)

The CI-fetch issue is now fixed on this branch (`1ef7984` — the sitemap generator no longer runs on CI builds unless it's the `main` deploy build). Below are the remaining findings from the review, none of which block merge. Severities noted per item.

### 2. Sitemap `?lang=` is stripped by the page's canonical tag — **High (SEO)**
The generator emits language-specific URLs like `…&book=GEN&chapter=1&lang=es`, but the SSR page renders `<link rel="canonical">` from `SeedBibleStateManager.canonicalUrl` (see `standalone/entry-ssr.tsx:85`), which sets only `translation`/`book`/`chapter` and **never `lang`**. Search engines treat the on-page canonical as authoritative and consolidate the `?lang=` variants to the `lang`-less URL, so the language-specific URLs — the headline feature of this PR — won't be indexed as distinct. Fix by either adding `lang` to `canonicalUrl` or dropping it from the sitemap so the two agree.

### 3. `lastmod` is stamped "today" on every run — **Medium (SEO)**
`generate-sitemap.ts` sets `lastmod = new Date().toISOString().slice(0, 10)`, so every child sitemap and the index report "modified today" on each deploy even when a translation's chapters haven't changed. That misleads crawlers into re-crawling unchanged content and wastes crawl budget. A content-derived timestamp (or omitting `lastmod`) would be more honest.

### 4. `translation` param assumes the default endpoint — **Low (latent correctness)**
The sitemap uses the bare `translation.id`, while the app's real/canonical URLs use `data.buildTranslationId(id)` (`BibleDataManager.tsx:590-601`), which returns the raw id **only for the default endpoint** and a full API URL otherwise. For today's production default endpoint these coincide, so it's correct now — but it's an unstated assumption. If `getDefaultAPIEndpoint(origin)` ever resolves to a non-default endpoint, the sitemap URLs would stop matching the site's canonical URLs.

### 5. Dead prioritization pass in `buildBibleLanguageToUiLocale` — **Low (simplification)**
The two-pass split ("`DEFAULT_TRANSLATIONS_BY_LANGUAGE` locales first, then the rest") never changes the output: none of the actual Bible-code collision pairs (`he`/`iw`→`heb`, `fil`/`tl`→`tgl`, `no`/`nb`→`nob`/`nor`) has a member present in `DEFAULT_TRANSLATIONS_BY_LANGUAGE`, so insertion order alone already resolves every tie. Not a bug, but the documented rule #1 is inert and could mislead a future maintainer into thinking it's load-bearing — a single-pass loop would be equivalent given current data.

### 6. Duplicated trailing-slash helpers — **Low (cleanup)**
`trimTrailingSlash` (`generate-sitemap.ts`) and `ensureTrailingSlash` (`lib/sitemap.ts`) are near-identical, and `buildChapterUrl` re-implements the query-param shape already encoded in `SeedBibleStateManager.canonicalUrl`. Low impact; consolidating would keep the URL shape defined in one place.

---

Overall the pure helpers in `lib/sitemap.ts` are well-factored and their test coverage is solid. #2 is the one worth an explicit decision before this ships; #3–#6 are polish.

---

KallynGowdy:
Thanks for the review — all follow-ups addressed in `b421f21`.

**#2 (High, canonical strips `?lang=`)** — Chosen: keep the per-language URLs and make the canonical agree. `SeedBibleStateManager.canonicalUrl` (and thus `<link rel="canonical">` / `og:url`) now echoes the **explicit** `?lang=` from the requested URL. So a crawler landing on a sitemap URL `…&lang=es` gets a self-referential canonical `…&lang=es` and each language variant indexes distinctly; ordinary lang-less pages are unchanged. It reads only the URL param (not `i18n.language`), so the canonical never varies by `Accept-Language`. Covered by two new tests in `SeedBibleStateManager.test.ts`.

**#3 (Medium, `lastmod` = today)** — Dropped. The generator no longer stamps `lastmod`, so unchanged content stops looking modified every deploy. `renderSitemapIndex` keeps optional `lastmod` support for the future.

**#4 (Low, `translation` assumes default endpoint)** — Added a pure `buildTranslationParam(id, endpoint, defaultEndpoint)` mirroring `BibleDataManager.buildTranslationId` (bare id for the default endpoint, full `…/api/{id}/books.json` URL otherwise); the generator now uses it, so sitemap URLs match the site's canonical URLs on any endpoint. Unit-tested both branches.

**#5 (Low, dead prioritization)** — `buildBibleLanguageToUiLocale` is now a single insertion-order pass; dropped the inert `DEFAULT_TRANSLATIONS_BY_LANGUAGE` prioritization and the misleading doc. Output is identical (existing mapping tests still pass).

**#6 (Low, duplicated helpers)** — Consolidated `ensureTrailingSlash`/`trimTrailingSlash` into `lib/sitemap.ts`. I left `buildChapterUrl` as the pure, tested definition of the URL shape (with a comment noting it mirrors `canonicalUrl`) rather than folding in the stateful `buildTranslationId` path — the translation-param derivation genuinely differs (pure vs. manager-backed), so a shared 3-line param-setter wasn't worth the cross-package coupling.

Also fixed the default origin to `https://seedbible.org` (was `prod.seedbible.org`, which is the deploy host, not the public domain).

Verification: `pnpm vitest run` on both suites (76 passing), `pnpm check:ts` (0 errors), `pnpm lint` (0 errors), plus a local mock-API e2e confirming no `<lastmod>`, correct `buildTranslationParam` output, and preserved `?lang=`.

---

Blitzy:
> [!NOTE]

## Summary

This PR adds a sitemap generator (`script/generate-sitemap.ts` + pure helpers in `script/lib/sitemap.ts`) that emits one child sitemap per Bible translation — every chapter of every book — plus a sitemap index and `robots.txt`, wired into `pnpm build` and gated to only run for the `main` deploy build. It also makes the app's canonical-URL logic (`SeedBibleStateManager.canonicalUrl`) echo an explicit `?lang=` so the per-language sitemap URLs stay self-canonical instead of being collapsed by search engines.

## Prior review already addressed

@Blitzy's earlier review on this PR (comment from 2026-07-20) raised five follow-up items (No. 2 High, No. 3–6 Low/Medium), and the author's 2026-07-24 update (`b421f21`) addressed all of them. I re-verified each against the current diff:

- **No. 2 (High — canonical stripped `?lang=`)**: Fixed. `canonicalUrl` now echoes an explicit `?lang=` from the current URL (`SeedBibleStateManager.tsx`), and two new tests cover the echo/omit cases.
- **No. 3 (Medium — `lastmod` stamped "today")**: Fixed. The generator no longer emits `lastmod`; `renderSitemapIndex` still supports it as an optional field for future use.
- **No. 4 (Low — `translation` param assumed default endpoint)**: Fixed via the new `buildTranslationParam()`, which mirrors `BibleDataManager.buildTranslationId` exactly (bare ID for the default endpoint, full `books.json` URL otherwise) and is unit-tested for both branches.
- **No. 5 (Low — dead prioritization pass)**: Fixed. `buildBibleLanguageToUiLocale` is now a single insertion-order pass.
- **No. 6 (Low — duplicated trailing-slash helpers)**: Fixed. Consolidated into `lib/sitemap.ts`.

I also confirmed the CI-vs-deploy gating in `skipReason()` is wired correctly against the actual workflows: `cd.yml`'s `Build` step sets `DEPLOY_BRANCH` explicitly (so the `main` deploy build generates, other branches skip), while `ci.yml`'s plain `pnpm build` has no `DEPLOY_BRANCH` but does run under GitHub Actions' implicit `CI=true`, so it correctly skips the live catalog fetch. CI on the head commit (`b421f21`) is green.

## New (minor, non-blocking) observations

**1. Misleading test name for `buildTranslationParam`'s slash handling — Low, test clarity**
In `test/unit/script/lib/sitemap.test.ts`, the test `"encodes reserved characters in the id for a non-default endpoint"` asserts:
```ts
buildTranslationParam("eng/esv", "https://bible.helloao.org/", DEFAULT)
// => "https://bible.helloao.org/api/eng/esv/books.json"
```
The `/` in `"eng/esv"` is **not** encoded here — `new URL("api/eng/esv/books.json", base)` treats it as an extra path segment, not a literal character in one segment (contrast with `buildChapterUrl`'s query-param case, which genuinely does percent-encode a `/` to `%2F` via `URLSearchParams`). This isn't a functional bug introduced by this PR — `buildTranslationParam` is intentionally a faithful mirror of the pre-existing `BibleDataManager.buildTranslationId` (`BibleDataManager.tsx:590-601`), which has the identical behavior — but the test's name/intent doesn't match what it actually verifies, which could mislead a future reader into thinking slash-encoding is guaranteed here. Consider renaming the test to describe what it actually checks (e.g., "matches buildTranslationId's un-encoded path-segment behavior for non-default endpoints"), or dropping it if this edge case isn't real (translation IDs in the catalog don't appear to contain `/`).

**2. Orphaned per-translation sitemap files can accumulate in S3 over time — Low, operational**
`cd.yml`'s "Sync PWA shell + public files (main only)" step syncs `standalone/dist/client` (where `sitemap.xml`, `robots.txt`, and `sitemaps/*.xml` land) to the bucket root with `aws s3 sync` and no `--delete`. The generator's own `rm(sitemapsDir, { recursive: true, force: true })` only clears the *local* build output before regenerating — it doesn't touch anything already uploaded. So if a translation is later removed from the catalog (or its sanitized filename changes), its old child sitemap XML file stays live in the bucket indefinitely, just unreferenced by the current `sitemap.xml` index. Not a correctness bug in this PR's new code, and not a new pattern (the sync step already worked this way before this PR), but this is the first feature to write a set of files that actually changes shape over time based on an external catalog, so the staleness risk is newly relevant. Worth a follow-up (either add `--delete` scoped to `sitemaps/*`, or accept it as harmless clutter) rather than blocking this PR on it.

## Overall

The pure helpers in `lib/sitemap.ts` are well-factored and thoroughly unit-tested (XML escaping, chunking, name sanitization/dedup, language-code inversion, URL building). The deploy-gating logic is correctly wired to the real CI/CD workflows. Nothing above blocks merge — both notes are polish/follow-up items.

---

KallynGowdy:
Addressed issue **#1** (misleading test name) in `3a7e9c9`.

Renamed the `buildTranslationParam` test from `"encodes reserved characters in the id for a non-default endpoint"` to `"keeps a slash in the id as a path segment (not percent-encoded), mirroring buildTranslationId"`, and added a comment noting the `/` becomes a path separator via `new URL(...)` (matching `BibleDataManager.buildTranslationId`), in contrast to `buildChapterUrl`'s query-param encoding. Kept the assertion — it documents the intentional mirror. No behavior change; 31 tests pass.

On issue **#2** (orphaned `sitemaps/*.xml` in S3 with no `--delete` on the sync) — agreed it's a real but non-blocking staleness risk, and it lives in `cd.yml`'s sync step rather than this PR's code. Leaving it out of this PR as you suggested; happy to do a follow-up that scopes `--delete` to the `sitemaps/` prefix if you'd like.

---

PR #1471, opened 2026-07-19
Cache the user profile locally, and carry anonymous config into new accounts

KallynGowdy:
Show the last-known profile instantly on load (from a localStorage cache)
while LoginManager confirms it over the network, instead of a blank
Account settings page. Also let config set anonymously (font size, theme,
etc.) persist on-device and get adopted as the starting profile.config the
first time a brand-new account logs in on that device.

Claude-Session: https://claude.ai/code/session_01R6ntoC46Us8KVAEUkgZ7VZ


Blitzy:
## Code review

Holding off on merging this for now — the profile-caching and anonymous-config-adoption changes touch the same write paths that have previously caused data-loss, and they need more thorough testing before they're safe to land. Findings below, ranked most severe first.

### 1. Cross-account profile leak: a stale `cachedProfile` survives an account switch that happens before the first load resolves (or after a failed load)
**`LoginManager.tsx:502` / `:524`** — correctness, highest severity

The profile effect only clears `cachedProfile` on switch when `profileUserId !== null` (line 502), but `profileUserId` is set **only** in the success `.then` (line 535). And it only reads a fresh cache when `cachedProfile.peek() === null` (line 524).

Failure path:
1. Log in as user-1 → effect reads user-1's cache into `cachedProfile` (say "Alice"); the network load is still pending, so `profileUserId` is still `null`.
2. Session key is swapped directly to user-2 **before** user-1's load resolves (the "switch without logout" flow the code explicitly supports, line 498–501).
3. Effect re-runs: line 502's guard is `false` (`profileUserId` is `null`) → the clear is skipped; line 524's guard is `false` (`cachedProfile` is non-null "Alice") → user-2's cache is never read.
4. `AccountSettingsView` now displays **user-1's** name/location/picture while the user is signed in as user-2, until user-2's load lands — and *forever* if user-2's load fails transiently.

The same leak triggers if user-1's initial load *failed* (transient/mobile blip, which the comments call common) and the user then switches accounts. The existing "clears the cached profile when switching accounts" test only covers the case where the first load **succeeded**, so this window is untested. Fix likely needs to key the clear/read on `loadingForUserId` rather than gating on `profileUserId` being non-null.

### 2. Save silently discards the user's edits when only the cached profile is shown
**`SettingsPage.tsx:165` + `LoginManager.tsx:596`** — correctness/UX regression

`profile` now falls back to `cachedProfile` (line 165), so the name/location/description inputs render populated even while `login.profile.value` is still `null` (load in flight or failed). But `handleSave` → `updateProfile` bails out at line 596 (`if (!profile.value)`) with only a `console.warn`, and `handleSave` then resets `newName/newLocation/newDescription` to `null` — so the user's typed edits **vanish** and revert to the cached values with no visible feedback. Before this PR the fields were blank in that window, so the user wasn't invited to edit. The Save button (line 390) has no `disabled` gate on `profile.value`. Consider disabling Save (or surfacing "still loading") while `login.profile.value === null`.

### 3. `localConfig` is never cleared when logging into an *existing* account, so anonymous prefs can be adopted by an unrelated future account on a shared device
**`LoginManager.tsx:266`** — correctness/privacy, low-medium

Adoption + clearing of `localConfig` happens only on the `data_not_found` (brand-new account) path. If an anonymous user sets config and then logs into an **existing** account, `localConfig` is left intact. Later, a *different* person who creates a brand-new account on the same device inherits that leftover config. It's only font size/theme, so impact is limited, but it contradicts the "shared device" reasoning in the comment at lines 273–276 (which only guards the adoption path).

### 4. Anonymous config is now persisted in two parallel localStorage stores
**`ProfileConfigSync.tsx:46`** — reuse/altitude, low-medium

`ConfigManager` already mirrors anonymous `fontSize`/`disablePanels` to localStorage via `setLocalTag` (keys `app.fontSize`, `app.disablePanels`; see `ConfigManager.tsx:220,231`). The new anonymous branch writes the *same* values again to a second store (`sb-profile-config-local`, keys `fontSize`/`disablePanels`). Two stores with two key schemes hold the same data, and only one feeds adoption. Worth considering whether adoption could read the existing local cache instead of introducing a parallel one.

### 5. `readLocalConfig` adopts unvalidated data straight into the server profile
**`LoginManager.tsx:161`** — robustness, low

Unlike `readCachedProfile` (which runs `userProfileSchema.safeParse`), `readLocalConfig` returns any non-array object as-is. On first login of a new account that object is written verbatim into the account's `profile.config` via `updateUserProfile`. A corrupt/oversized `sb-profile-config-local` entry becomes durable account state. Consider validating/normalizing before adoption.

---

**Minor:** the anonymous dedup check `existingLocal[key] === value` (`ProfileConfigSync.tsx:43`) is reference-only, inconsistent with the deep-equal check on the logged-in path — harmless for today's primitive-only callers (font size, boolean, lang), but a trap if an object-valued config key is ever saved anonymously. And `effect(() => writeLocalConfig(localConfig.value))` (`LoginManager.tsx:246`) fires once at construction, re-writing the value it just read back to localStorage.

---

Blitzy:
> [!NOTE]

## Code review

This re-review covers the current head (`418f79b`). All five findings from the previous review round (comment above, 2026-07-20) have been verified as fixed by the follow-up commits:

- No. 1 (cross-account profile leak) — fixed in `4f90e46` via the new `cachedProfileUserId` tracking in `LoginManager.tsx`.
- No. 2 (Save silently discarding edits) — fixed in `4f90e46`; `SettingsPage.tsx`'s `AccountSettingsView` now gates `handleSave` and disables the Save button on `isProfileLoaded` (`login.profile.value !== null`), showing "Loading your profile…" instead.
- No. 3 (`localConfig` not cleared on existing-account login) — fixed in `4f90e46`; `LoginManager.tsx:355-357` now clears it on any confirmed profile load, not just the brand-new-account path.
- No. 4 (duplicate anonymous config stores) — fixed in `837d786`, which deletes `ConfigManager.tsx` entirely and folds `fontSize`/`disablePanels` into `SettingsManager`, unifying everything onto `login.localConfig`.
- No. 5 (unvalidated local config adopted into a new account) — fixed in `4edd7a5` with `MAX_LOCAL_CONFIG_JSON_LENGTH`/`MAX_LOCAL_CONFIG_KEYS` caps in `LoginManager.tsx`.

I re-ran `pnpm check:ts`, `pnpm lint`, and the full `pnpm vitest run` suite (3016 tests) against this head — all pass. However, the large `ConfigManager` → `SettingsManager` merge (`837d786`) introduces a new regression that isn't covered by any existing test. Ranked by severity below.

### 1. Anonymous users can never change a setting if the URL carries the matching `app.*` param — the change is silently reverted (High, correctness/UX)

**`managers/SettingsManager.tsx:659-728`**

Before this PR, `SettingsManager`'s settings-derivation effect only re-ran on `login.profile.value` changes, and the URL's query params were captured **once**, at construction, into a plain object (`configBot.tags`) that setters then mutated directly — so a URL param only ever set the *initial* value, and any later edit stuck for the rest of the session.

The merge changed this in two ways at once:
- `readSettings()` now calls `url.searchParams.get(urlKey)` fresh every time it runs, instead of reading a snapshot.
- The wrapping `effect()` now also depends on `login.localConfig.value` (needed to pick up anonymous edits) and `navigation.currentUrl.value`.

The read precedence is `profile ?? url ?? local` (`SettingsManager.tsx:668-671`). For an anonymous user, their own edits only ever land in `local`, which sits **below** the URL in that precedence. Combined with the effect now reactively depending on `login.localConfig.value`, every anonymous edit re-triggers the same effect, which re-reads the (unchanged, still-present) URL param and immediately overwrites the edit back to the URL's value — synchronously, in the same tick as the user's own click.

**Reproduction:**
1. Load the app anonymously with any `app.*` settings param in the URL, e.g. `/reader?app.fontSize=XL` (the realistic case is a partner site embedding Seed Bible in an iframe with a preset config URL — see the "deep-linking" comment at `SettingsManager.tsx:651`).
2. Open Settings and change the font size (or any other setting whose `TAG_*` param is present in the URL) to a different value.
3. The UI briefly shows the new value, then immediately reverts to the URL's original value, with no error or explanation. The user cannot ever change that setting for the rest of the session while anonymous.

I confirmed this with a standalone repro against the actual `createSettings`/`createNavigationManager` code (not just by reading it):
```ts
const login = makeFakeLogin(null); // anonymous
const settings = createSettings(CasualOSManager(), login, navWith("?app.fontSize=XL"));
settings.setFontSize("L");
// login.localConfig.value -> { fontSize: "L" }   (write succeeded)
// settings.settings.value.fontSize -> "XL"       (immediately reverted)
```
This isn't caught by the existing `?app.fontSize=XL` tests in `SettingsManager.test.ts` (lines ~360-402), because those only assert the *initial* read precedence — none of them call a setter afterward and re-check the value.

Note this isn't limited to the two settings that are new to this pattern (`fontSize`, `disablePanels`) — the same `read()` helper and effect apply uniformly to every setting that already lived in `SettingsManager` (`bookOrientation`, `uiSize`, `selectionUI`, `scriptureElements`, `textConfig`, `toolbar`, `keepScreenAwake`, `customHighlightColors`, `scriptureMargin`), so this is a regression for all of them in the anonymous+URL-param case, not just the merged-in fields. Logged-in users are unaffected once they've made an explicit edit, since a write updates `profile.value`, which sits above the URL in the precedence order.

**Suggested fix:** Capture each setting's URL override once at construction (as the old `configBot.tags` did), rather than re-reading `navigation.currentUrl.value` on every `readSettings()` call. That preserves "a deep-link sets the starting value" while letting the effect stay reactive to `profile`/`localConfig` for genuine persistence, without the URL fighting every subsequent user edit.

---

### 2. Minor: redundant recompute on every setter call (Low, no functional impact)

Because the settings effect now depends on `login.localConfig.value` (and `login.profile.value`, which `updateProfile` also writes synchronously), every setter's own `saveProfileConfigValue` call re-triggers the same effect that the setter's optimistic `settings.value = {...}` write just ran. In the non-URL-param case this recomputes to the same value, so it's harmless, but it's worth being aware of as one more symptom of the same root cause as No. 1 above — the effect is doing more reactive work than intended.

---

**Everything else looks solid.** The account-switch cache-ownership tracking (`cachedProfileUserId`), the load-failure handling that keeps a good profile rather than falling back to blank, and the size/shape caps on adopted local config are all well-reasoned and covered by the new `LoginManager.test.ts` (455 lines) and `ProfileConfigSync.test.ts` additions. Given the PR description's own caution about this touching "write paths that have previously caused data-loss," I'd still hold off on merging until No. 1 is addressed, since it's a silent, hard-to-notice regression for the embed/deep-link use case this code explicitly says it's meant to support.

---

KallynGowdy:
## Moved `ThemeManager`'s storage onto `SettingsManager`

`ThemeManager` used to read/write the selected theme (`themeId`), custom color overrides (`customTheme`), and custom highlight overrides (`customHighlights`) directly via `getProfileConfigValue`/`saveProfileConfigValue`, with its own `profile ?? URL param ?? default` read order. That's the same storage mechanism `SettingsManager` already centralizes for its other 14 fields behind a unified `profile > sessionOverrides (URL) > login.localConfig > default` precedence — so `ThemeManager` had the same anonymous-persistence gap `SettingsManager` fixed earlier in this PR: an anonymous user's theme/color choice was written into `login.localConfig` but never read back, so it was lost on refresh.

**Changes:**
- `SettingsManager.tsx`: added `themeId`, `customTheme`, `customHighlights` to `AppSettings` (+ schema, defaults, `readSettings()`, `resetToDefaults()`), and three new setters (`setThemeId`, `setCustomTheme`, `setCustomHighlights`) following the existing setter pattern.
- `ThemeManager.tsx`: `createTheme` now takes a `SettingsManager` instead of `LoginManager`/`NavigationManager`, and reads/writes exclusively through it. `selectedThemeId`/`customOverrides`/`customHighlightOverrides` are now derived `ReadonlySignal`s instead of independently-managed `Signal`s (no consumer impact — nothing outside the manager ever assigned to them directly). Theme-domain validation (which color keys are customizable) stays in `ThemeManager`, layered on top of `SettingsManager`'s generic storage.
- `SeedBibleStateManager.tsx`: `createTheme(login, navigation)` → `createTheme(settings)`.
- Tests: new `ThemeManager.test.ts` coverage for the storage/precedence behavior (profile vs. URL vs. anonymous-local vs. default, including a "survives a simulated refresh" case proving the persistence-gap fix), plus matching `themeId` precedence/reset coverage in `SettingsManager.test.ts`.

No behavior change for logged-in users; anonymous theme/color customization now survives a refresh, matching every other setting in the app.

Verified: full `pnpm vitest run` suite passes, `pnpm check:ts` and `pnpm lint` clean.

---

Blitzy:
> [!NOTE]

## Code review

This re-review covers the current head (`7876102`), triggered by the PR's `ready_for_review` transition. I read through the two prior review rounds on this thread and the full commit history to see what changed since the last automated pass (which reviewed `418f79b` and flagged one regression, "No. 1: anonymous users can never change a setting if the URL carries the matching `app.*` param").

**No new issues found.** Two commits landed since that last review:

1. **`301dad2` — fixes the URL-revert regression (previous finding No. 1).** The fix is correct: `SettingsManager.tsx` now captures the URL's query params once into a per-instance `sessionOverrides` object at construction (`createSettings`, around line 754), instead of re-reading `navigation.currentUrl.value` on every `readSettings()` call. The reactive `effect()` that recomputes `settings.value` (line 843) now only depends on `login.profile.value` / `login.localConfig.value`, so a same-session anonymous edit can no longer be reverted by a stale URL param. I traced through the precedence chain (`profile ?? sessionOverrides ?? local`) and it holds for all 11 pre-existing fields plus the two that previously lacked `sessionOverrides` tracking (`scriptureMargin`, `keepScreenAwake`), which this commit also fixed.

2. **`78761029` — moves `ThemeManager`'s storage onto `SettingsManager`.** This folds `themeId`, `customTheme`, and `customHighlights` into `AppSettings`, giving them the same `profile > sessionOverrides > login.localConfig > default` precedence as every other setting (closing the same anonymous-persistence gap `SettingsManager` fixed earlier for its own fields — an anonymous user's theme/color pick used to be written to `login.localConfig` but never read back). `ThemeManager` now takes a `SettingsManager` instead of `LoginManager`/`NavigationManager` directly, and its own concerns (theme presets, CSS generation, valid-color-key whitelisting via the new `filterValidColorOverrides`) are correctly layered on top. The old `parseThemeId`/`parseCustomTheme`/`parseHighlightOverrides` validation logic moved into `SettingsManager.tsx` intact — I diffed it line-by-line against the pre-move version and it's unchanged. Wiring in `SeedBibleStateManager.tsx` is correct: `settings` is now constructed before `createTheme(settings)`, and I confirmed there are no leftover references anywhere in the tree to the removed `ConfigManager`/`state.config`.

**Verification performed directly** (not just read-through):
- Checked out the PR head in a worktree and ran `pnpm vitest run SettingsManager ThemeManager LoginManager ProfileConfigSync` — **104/104 tests pass**, including the new precedence/refresh-persistence tests for `themeId` (`SettingsManager.test.ts`) and the new `ThemeManager.test.ts` suite (setTheme, custom color/highlight overrides, and an anonymous-choice-survives-refresh test that exercises the exact bug this migration fixes).
- Ran `pnpm check:ts` against the same head — clean, no errors.
- CI's `build`/`deploy` checks on this PR are green.

**Everything from the two earlier review rounds stays fixed on this head:**
- No. 1 (cross-account profile leak) — still fixed via `cachedProfileUserId` tracking.
- No. 2 (Save silently discarding edits) — still gated on `isProfileLoaded`.
- No. 3 (`localConfig` leaking to a new account) — still cleared on any confirmed profile load.
- No. 4 (duplicate anonymous config stores) — still unified onto `login.localConfig` (and now further consolidated with the theme fields too).
- No. 5 (unvalidated local config adopted into a new account) — still capped via `MAX_LOCAL_CONFIG_JSON_LENGTH`/`MAX_LOCAL_CONFIG_KEYS`.
- The URL-revert regression from the second review — fixed as described above.

This looks ready to merge from a correctness standpoint, pending normal human sign-off. Nice work tracking down a subtle reactive-effect bug across two review rounds.

---

KallynGowdy:
@Blitzy Do you think there's anything else blocking this?

Blitzy (dismissed):
Nope I think this is good to go, no blockers.

PR #1474, opened 2026-07-19
Feature/login with translation

### LINKED ISSUE #1445: Save user selected translations to user profile config so that it reopens on returning.
## What is happening

Currently if you select any other book other than our default AAB, it will not open back up upon returning. The only way a translation re-opens is if you are returning using a URL where `?translation=` is baked in.

## What is expected

A user selected translation should save the user's profile config. Upon retrieving signing in, the user's last selected translation should re-open.

AmitPant9999:
fix: #1445


AmitPant9999:
@Blitzy is there any issue with this pr

Issue #1413, opened 2026-07-13
Profile "Save Changes" button should show an indicator while request is inflight.


Blitzy:
## What is happening

In low bandwidth situations, the profile Save Changes button can be pressed but it is unknown when changes are actually saved, and this changes can be lost unintentionally if users close out before the request can finish.

## What is expected

1. The Save Changes button should show an indicator that it is processing the request. 
2. The fields in the profile should not be modifiable while a save request is inflight.

PR #1480, opened 2026-07-20
feat: add loading skeleton and saving indicator for account settings

### LINKED ISSUE #1412: Profile page should show a loading indicator and not be interact able until the profile is fully loaded.
## What is happening 

When on a poor data connection, the profile page is fully usable while silently in the background the profile data is still being fetched.

## What is expected

The profile page should make clear that a data fetch is still occuring in the background. The page should either have an overlay with a loading indicator or even perhaps use skeleton loader placeholders to make clear to the user that the profile is still loading is still loading.

mazenemad:
fixes #1412 & #1413


Blitzy (commented):
Looks good. Im gonna extract this out to a reusable component.

Blitzy (approved):

PR #1483, opened 2026-07-20
Fix/today screen dark theme issues

### LINKED ISSUE #1424: Today screen theming issues
Currently, the Today screen has some issues with the dark theme:

Image

soukigabriel:
Fixes #1424


Blitzy:
> [!NOTE]

## Code Review — PR #1483 (Fixes #1424: Today screen dark‑theme issues)

**Overall: looks good.** This is a focused, coherent fix. The core insight is sound: the Today reading‑history timeline was drawing its cells from `readerToolbarFloatingButtonBackground` → `secondaryColor`, which didn't match the legend and read poorly in dark mode. The PR re‑bases the cells on `dividerColor` → `primaryColor` (matching the legend's `color-mix(... --sb-primary-color, --sb-divider-color)`), and because `dividerColor` is a translucent `rgba()` (`rgba(255,255,255,0.1)` in dark), it correctly adds alpha compositing so `getColorByReadingTime` receives an opaque hex.

I verified locally on `28d1121`:
- `pnpm vitest run` on the two touched test files → **119 passed**.
- Every CSS variable introduced (`--sb-font-color`, `--sb-reader-background`, `--sb-tertiary-color`, `--sb-secondary-font-color`) is emitted by `ThemeManager` (camelCase → `--sb-` kebab), so the style changes are valid in both themes.
- `--text1` (the old value in `readingHistoryTimelineStyles.css`) is **not defined anywhere** in the standalone reader — only in an unrelated Twitch extension — so that label color was effectively unset before. Swapping to `--sb-font-color` is a real fix.
- Removing the debug `console.log`s in `useBook`/`useSocialSection` is a welcome cleanup.

No Critical or High issues. A few Low/Nit items below.

---

### 🟡 Low

**1. The parallel `scripture-map` timeline was not updated — the same dark‑theme bug likely remains there.**
`packages/scripture-map/hooks/useReadingHistoryTimeline.tsx` (lines ~97–103) still uses the old approach:
```ts
baseColor: theme.variables.readerToolbarFloatingButtonBackground ?? "#dfdede",
userColor: theme.variables.secondaryColor,
```
This PR only fixes the `today-screen` copy. If the scripture‑map reading‑history timeline is still reachable, it keeps the mismatched-vs-legend colors that #1424 is about.
- *Repro:* open the scripture‑map reading‑history timeline in dark mode; cell colors still derive from the floating‑button background rather than divider/primary.
- *Suggested fix:* apply the same `ColorParser` + `dividerColor`/`primaryColor` change there, or confirm scripture‑map is deprecated and leave a note. Not a blocker for this PR's stated scope.

**2. `ColorParser` no longer short‑circuits when `sourceType === target`.**
The guard changed from `if (!sourceType || sourceType === target)` to `if (!sourceType)`, so every call now round‑trips through an RGB tuple. This is required for the alpha‑compositing goal, but it is a behavior change to a shared utility with side effects:
- hex output is normalized to **lowercase** (`"#ABCDEF"` → `"#abcdef"`),
- `arrayRGB → arrayRGB` now returns a **new array** rather than the same reference,
- `stringRGB → stringRGB` is reformatted to `rgba(r,g,b,1)`.

I checked every current caller (`today-screen`, and `scripture-map/hooks/useSettings.tsx`) — all feed the result into display/`color-mix`, none rely on identity or exact casing, and existing tests pass, so **no live regression**. Because this is a widely‑importable util, consider adding a regression test that pins the "same source/target still normalizes" contract so a future caller isn't surprised.

---

### 🔵 Nits

**3. `RgbaToRgb` output isn't clamped to `[0,255]`.** Safe today because inputs are constrained (rgb `0–255`, and the regex `([\d.]+)` can't yield a negative alpha), but wrapping the result in the existing `ClampRGBColor` would harden it against malformed theme values at negligible cost.

**4. Malformed/keyword theme colors fall through to black.** If `theme.variables.dividerColor` or `primaryColor` were ever set to something `GetColorType` doesn't recognize (e.g. `"transparent"` or a named color), `ColorParser` returns it unchanged, and `getColorByReadingTime` → `InterpolateHexColors` → `HexToRgb` then warns and returns `[0,0,0]`, painting cells black. The shipped themes all use valid `rgba()`/hex, so this only affects custom/invalid theme config — worth a one‑line guard or a defaulting fallback if custom themes are a supported path.

**5. `.date-label` selector broadened** from `.user-filter-container .date-label` to `.history-card .date-label`. Verified this is correct: `date-label` is rendered as a direct child of the history card (`HistoryCard.tsx:88`), not inside `.user-filter-container`, so the old (empty) rule never matched it and the new one now applies the font color properly. No action needed — just flagging the scope change for the record.

---

Nice, well‑tested change. Addressing item **1** (or explicitly deferring it) is the main thing I'd want resolved before this closes out the dark‑theme timeline story end‑to‑end.

---

Blitzy (approved):

PR #1484, opened 2026-07-21
Feature/1476

### LINKED ISSUE #1476: Bookmark UX improvements
2 issues: 

1: On mobile currently users cannot move a bookmark to a new category (I can't move Luke 20 from My Bookmarks to Family) 

Image

2: Hitting the bookmark icon at the CH level should open the UI that gives the users the opportunity add it to a new category or pick

If users only have 1 category of bookmark (just "My bookmarks" for example) we could potentially skip this step but if they have multiple categories we need to ask them where they want to save the bookmark 

Image

Image

Kabi4:
Fixes #1476 

image
image


Blitzy:
> [!NOTE]

## Code Review — PR #1484 (Feature/1476)

Reviewed against the linked issue [#1476 "Bookmark UX improvements"](https://github.com/HelloAOLab/seed-bible/issues/1476) at commit `03ff2a2`.

### Summary

The PR cleanly addresses both parts of the issue:

1. **Move a bookmark between folders** — adds a "Move to folder" item to each bookmark's kebab menu, a new `moveBookmark()` on `BookmarksManager`, and a `mode: "move"` path in the shared category-picker modal that hides the bookmark's current folder.
2. **Chapter-level bookmark icon opens the picker** — `ReaderBookmarkButton` now opens the category picker when adding (so the user chooses/creates a folder), and removes the chapter-level bookmark when it's already saved.

**Verification:** `pnpm test` (BookmarksManager: 11/11 pass), `pnpm check:ts` (0 errors), and `pnpm lint` (0 errors, i18n keys validated) all pass locally. The chapter-removal path correctly scopes to chapter-level bookmarks only — verse-scoped bookmarks in the same chapter are preserved (`removeBookmarkForLocation` with `verse` undefined matches only `bookmark.verse === undefined`). Nice touch adding a unit test that covers the create-on-move, no-op-when-unchanged, unknown-id, and trimming cases.

No blocking issues found. A few low-severity notes below.

---

### Low

**1. "Move to new folder" persists twice** — `Tabs.tsx` `handleSave` (~line 1300)

When moving a bookmark into a brand-new folder, `handleSave` calls `bookmarks.createCategory(trimmedNew)` and then `bookmarks.moveBookmark(bookmarkId, category)`. `moveBookmark` already creates the target category via `ensureCategory` and persists, so this fires two back-to-back `recordData` writes for one user action.

- *Effect:* extra network round-trip; not incorrect, just wasteful (and a brief window where categories are persisted without the moved bookmark).
- *Suggested fix:* in `move` mode, skip the explicit `createCategory` call and let `moveBookmark` handle category creation:
  ```ts
  if (isMove) {
    if (!bookmarkId) return;
    await bookmarks.moveBookmark(bookmarkId, trimmedNew || selectedCategory.value);
  } else {
    if (isAddingNew.value) { await bookmarks.createCategory(trimmedNew); }
    await bookmarks.addBookmark(/* ... */);
  }
  ```
  (i.e. only `addBookmark` needs the category pre-created, since it takes a category name and does not create folders the way `moveBookmark` does.)

**2. Issue suggests skipping the picker for single-category users; PR always shows it** — `BibleReader.tsx` (~line 81)

Issue #1476 point 2 says: *"If users only have 1 category of bookmark ... we could potentially skip this step but if they have multiple categories we need to ask them where they want to save."* The chapter button always opens the modal, even when the user has exactly one folder ("My Bookmarks").

- This is phrased as optional ("could potentially") in the issue, so it's not a defect — flagging so the decision is explicit.
- *Repro:* fresh account with only the default "My Bookmarks" folder → tap the chapter bookmark icon → modal opens with a single choice rather than saving directly.
- *If you want the shortcut:* in the not-bookmarked branch, when `bookmarks.categories.value.length <= 1`, call `addBookmark(...)` directly instead of `openBookmarkCategoryModal(...)`.

**3. Add-mode default selection changed from the default folder to `categories[0]`** — `Tabs.tsx` (~line 1283)

`BookmarkCategoryPickerContent` used to preselect `DEFAULT_BOOKMARK_CATEGORY`; it now preselects `categories[0]?.name`. Because `normalizeCategories` unshifts the default folder to the front, `categories[0]` is normally the default, so behavior is equivalent in practice. Worth being aware this now relies on that ordering invariant rather than naming the default explicitly — if category ordering ever changes, the preselected folder for the verse-toolbar/tab-row add flow would silently change with it.

---

### Nits

- The `onClick` guard in `ReaderBookmarkButton` (`if (!canBookmark || !translationId || !bookId || !chapterNumber)`) looks redundant with `canBookmark`, but it's actually required for TypeScript to narrow the nullable props to non-null before the call — correct as written, no change needed.
- Per project convention, only `en.json` was updated for the new keys (`move-to-bookmark-category`, `move-bookmark`, `move-to-new`) — 👍 matches CLAUDE.md guidance.

Overall this is a focused, well-tested change that matches the issue's intent. The Low items are optional polish, not merge blockers.

---

Blitzy:
@Kabi4

Have some issues that need addressing.

## Add Bookmark modal needs more feedback while saving.

The Add Bookmark modal is waiting until the save completes. There are a number of UX issues that needs to be resolved around this.

1. Remove the cancel button. We already have a close button that is common across the entire rest of the app.
image

2. The Save button should either show an indicator while saving or the text should change to "Saving...".
image

3. The "New folder" section should be hidden while saving.
image

4. If adding to a newly created folder, that folder should appear selected in the list while saving.
image

## Code review issue number 1

The issue number 1 brought up by the agent code review should be fixed.

## Code review issue number 3

We should continue utilizing `DEFAULT_BOOKMARK_CATEGORY` explicitly as the default bookmark category, even if the ordering changes.

---

I think we can skip the Code review's issue number 2. I think we should bring up the Add Bookmark modal even if there is only one folder, it gives users a chance to make new folders if they want and it just lets them understand the bookmarking process more intuitively.

Blitzy (approved):

Issue #1482, opened 2026-07-20
Extensions are getting polluted


Blitzy:
## What is happening

Extensions are becoming polluted with dependency extensions that only confuse people.

Supporting dependencies between extensions makes total sense on a technical level, and having shared functionality or library-style extensions is definitely useful.

But I don't think we should be surfacing those to users. Users shouldn't have to understand library extensions. They should only care about extensions from the perspective of "How does this change my Seed Bible?" If the answer to that question is unclear or is a no-op on its own, it shouldn't be something users need to see.

The Twitch extensions are a good example. Splitting it into two extensions, one for hosting and one for being a participant, feels very strange from a normal person's perspective. If it were just "Twitch," that would make more sense. The Twitch user extension appears to do nothing if you install it manually, and it's auto-installed anyway if you arrive at the Seed Bible from a Twitch link. Ideally, the Twitch user extension shouldn't even be surfaced or persisted as an extension at all!

<img width="371" height="661" alt="Image" src="https://github.com/user-attachments/assets/f4e363b1-6c67-4f5a-a7dd-d6e3a99adc67" />

## What do we do?

We need to brainstorm how we can cleanup the Extensions screen so that it is user-friendly. 

1. Perhaps we introduce the concept of user-facing extensions and library extensions. Extensions like "Seed bible utils" and "AI Transcript core" do not serve users, they only serve developers and thus should not be seen by users.

Something like a `library` flag could keep it out of the user-facing Extensions screens.

```
{
    ...,
    library: true
}
```

2. I think there is a case to be for some extensions to never be persisted to user configs. Extensions like the "Twitch User Extensions" appear to no-op and do nothing to a user's Seed Bible when installed. This is a special extension where its purpose only becomes somewhat more clear when you come to the Seed Bible through a Twitch link. I cant think of a good reason to have an extension like that persisted to the user's profile.

Something like a `transient` / `ephemeral` flag could allow be implemented and the system updated so that it knows never to persist it to a user config.
```
{
    ...,
    transient: true
}
```

Open to critiques and other ideas! Just putting this out there so we can brainstorm.

Blitzy:
@KallynGowdy I'm curious about your thoughts on this

KallynGowdy:
I like the idea of having a flag to indicate that the extension is a utility extension and doesn't have any UI of it's own. I think it should be called something like `utility` though. It would also be nice to have a section to show all extensions so they can be uninstalled as well

PR #1485, opened 2026-07-21
Feature/1431

### LINKED ISSUE #1431: AI Transcript is showing twice in the extensions page.
## What is happening

AI Transcript is showing twice in the extensions page.

Image

## What is expected

I would expect AI Transcript to appear once.

Kabi4:
Fixes #1431 

image


Blitzy:
> [!NOTE]

## Code Review — PR #1485 (Fixes #1431: "AI Transcript is showing twice")

### Summary

The bug: two entries appeared in **Settings → Extensions** for the AI Transcript feature. There are actually two extension packages — `transcript-extension` (id `ext_AI_Transcript`, title *"AI Transcript core"*) is an internal package that only exists as a dependency, and `transcript-ui-extension` (id `ext_AI_Transcript_UI`, title *"AI Transcript"*) is the one users are meant to see. Both were being listed.

This PR fixes it with three coordinated changes, and they hold together well:

1. **Adds a `hidden` meta flag** and filters hidden extensions out of the Settings list — this removes the `ext_AI_Transcript` "core" entry the user shouldn't see.
2. **Fixes a copy-paste bug** in `transcript-extension/.../init.tsx`: it was registering under `ext_AI_Transcript_UI` (the UI's id) instead of its own id `ext_AI_Transcript`. Now the core registers under the same id its dependents (`ext_AI_Transcript_UI` and `ext_twitchPub`) declare as a dependency.
3. **Makes `registerExtension` idempotent** (skip a duplicate id) instead of the old "replace and re-run init" behavior — defensive hardening against double-registration.
4. Propagates `hidden` through the build-time meta trimming in `vite-plugin-extensions.ts`.

**Verification I ran locally on the branch:**
- `pnpm vitest run ExtensionManager.test.ts` → **64/64 passing**, including the pre-existing `unregisterExtension() allows re-registration of the same extension ID` and `loadExtension() re-invokes ... default export so an extension can be reinstalled after being unloaded` tests.
- `pnpm check:ts` → **0 visible errors** (6 suppressed are pre-existing, unrelated).

I specifically checked the behavior CLAUDE.md flags as sensitive — uninstall→reinstall within the same session. **No regression:** the new skip-if-registered logic is safe because `unregisterExtension` clears the id from `registeredExtensions` (and `initializedExtensionIds`) first, so a later re-registration finds an empty slot and proceeds to init normally. The added test correctly covers the new skip path. 👍

I found no correctness bugs. The items below are all low-severity polish.

---

### Findings

#### 🟡 Low — `hidden` filter only inspects the bundled package, not the registration

`packages/seed-bible/seed-bible/components/SettingsPage/SettingsPage.tsx:1034-1036`

```ts
const extensionsList = extensions.extensions.value.filter(
  (entry) => !entry.extension?.meta.hidden
);
```

Each list entry (`ExtensionListEntry`) can be backed by a bundled package (`entry.extension`) **and/or** a runtime registration (`entry.registration`). This filter only reads `entry.extension?.meta.hidden`. If an extension were ever present *only* as a runtime registration (`entry.extension === null`) — with no bundled `extension.json` — the `hidden` flag would be invisible to this filter and the entry would render with just its bare id and no title.

This is **not a live bug** for `ext_AI_Transcript`: it's always discovered from the bundled set via `virtual:@extensions`, so `entry.extension` is populated and `meta.hidden` is `true`. It's a latent robustness gap rather than something reproducible today.

*Reproduction:* not reproducible with the current extension set (no registration-only hidden extension exists). Flagging for future-proofing only.

*Suggested fix (optional):* also treat `hidden` on the registration path as hiding, e.g. carry `hidden` onto `ExtensionRegistration`/`ExtensionListEntry` and filter on `entry.extension?.meta.hidden || entry.registration?.hidden`, so "hidden" is a property of the extension regardless of how it entered the list.

#### 🟡 Low — leftover `console.log` on an expected code path

`packages/seed-bible/seed-bible/managers/ExtensionManager.tsx:227-229`

```ts
if (this.registeredExtensions.has(extension.id)) {
  console.log(
    `[ExtensionManager] Extension '${extension.id}' already registered, skipping`
  );
  return () => {};
}
```

With the idempotent behavior, hitting this branch is now a normal/expected occurrence, so an unconditional `console.log` will produce recurring noise in the browser console. Consider downgrading to `console.debug` or removing it.

*Suggested fix:* `console.debug(...)` (or drop the log entirely).

---

### Non-blocking observations

- **Behavior change "last-registration-wins" → "first-registration-wins":** worth calling out explicitly since CLAUDE.md flags this area. It's safe here because uninstall clears the registry first (see above), but any future code path that re-registers an id *without* unregistering it first will now be silently ignored rather than replaced. The added test locks in the new contract, which is good.
- **`getAllExtensionsAsSet()` (the "Download extensions" button) still includes hidden extensions.** This is arguably correct — a downloaded set needs to carry dependency packages like `ext_AI_Transcript` so the set is self-contained — but confirm that's the intent, since hidden extensions will appear in the exported JSON.
- **Orphaned dependency on uninstall (pre-existing, not introduced here):** uninstalling `ext_AI_Transcript_UI` leaves the hidden core `ext_AI_Transcript` registered, and since it's now hidden the user can't remove it. Dependencies aren't reference-counted / cascade-uninstalled anywhere in `ExtensionManager`, so this predates the PR — just noting it as a known limitation the `hidden` flag slightly increases the visibility of.

Overall: clean, targeted fix that resolves the reported issue. ✅

---

Blitzy:
I'm actually going to close this pull request only because when the original issue was written both extensions said "AI Transcript". It recently changed to "AI Transcript core" and "AI Transcript" which actually now makes clear that it was **not** a duplicate extension but actually a library extension and a user-facing extension. 

The `hidden` extension flag here is going in the right direction but this may not ultimately be the direction we head after @KallynGowdy and I have had more time to talk about it. This is actually now a larger issue that needs a deeper platform-level technical design for extensions.

See issue for more info #1482

PR #1492, opened 2026-07-21
fix: stop Welcome ↔ Today flash by gating on reading-history status (#1490)

### LINKED ISSUE #1437: Add a loading overlay to the today screen.
## What is happening

The today screen takes a long time to load up its data sets on its first load. As it does this it is showing a "Welcome" screen which incorrectly implies a few things to the user:

1. That the user has no reading data (history, bookmarks, etc).
2. That the website is slow or broken.

## What is expected

1. The today screen should utilize some kind of loading overlay while it loads and caches its data sets. The loading overlay would indicate to users that some data is being loaded and would feel more natural.
2. The today screen's Welcome page should only show if there is nothing to show after data has been loaded.

### LINKED ISSUE #1490: Fix Welcome ↔ Today screen flash on load
# Fix Welcome ↔ Today screen flash on load

## Summary

On a cold load, returning users briefly see the new-user **Welcome** page inside the Today screen before it flips to their personalized **Today** content. The switch is driven by whether reading history has loaded yet, not by whether the user is actually new — so anyone with history gets flashed the wrong screen while their history is still being fetched.

## Current behavior

The Today screen picks Welcome vs. personalized with a single binary gate in `packages/today-screen/infrastructure/presentation/hooks/useTodayContainer.tsx`:

```ts
if (!userId || !userLastReading.value) return { Component: Welcome, ... };
return { Component: TodayContent, ... };
```

`userLastReading` is `signal<UserLastReading>(undefined)` (`packages/today-screen/infrastructure/di/bootstrap.tsx:87`) and `UserLastReading = { bookId; chapter } | undefined` (`packages/today-screen/domain/models/readingHistory.tsx:59`). It is only filled after an async fetch (`bootstrap.tsx:170-195`).

**The problem: `undefined` means both "still loading" and "no history."** So a returning user renders Welcome (which even has a spinner baked into its footer, `components/containers/Welcome.tsx:47`) until the fetch resolves, then flips to `TodayContent`. That flip is the jank.

## Expected behavior

- **New / anonymous users** → Welcome, immediately, with no detour.
- **Returning users (any reading history)** → personalized layout with loading placeholders that fill in. They **never** see the Welcome page.
- No Welcome ↔ Today screen-kind swap at any point on a cold load.
- **Account changes are clean.** Signing out returns to Welcome immediately with no leftover personalized data. Signing in as a different user on the same browser shows that user's own state — never the previous user's resume position, bookmarks, or any cached hint.

## Root cause & the invariant to fix it

Reading history is keyed to `userId`, and `userId` is parsed **synchronously** from the cached session key in `localStorage` at startup (`managers/OsManager.tsx:87`, `managers/LoginManager.tsx:134,138-144`) — it is not async (only `userInfo`/`authBot`/`profile` are). The Today pane mounts well after that read, so `userId` is always known by first paint. And the app can only fetch history for a `userId` it already has (`getUsersIds()` returns `[]` when logged out, `bootstrap.tsx:75-83`).

Therefore "has history" and "userId known at first paint" are the same condition, which gives a clean rule:

> **Gate Welcome on `userId === null`, never on `userLastReading`.** Any known userId → render the personalized layout with loading placeholders until the fetch confirms. A Welcome screen never has history waiting behind it.

## Cross-account correctness (latent bug)

The current effect (`bootstrap.tsx:170-195`) only resets `userLastReading` when there is no userId. On a switch between two logged-in accounts (A → B) it re-runs but does **not** clear the value before fetching, so there is a window where user B sees user A's last-reading position. Worse, A's in-flight `getUserLastReading` promise has no guard against a newer userId, so it can resolve after B's and overwrite B's state with A's. This fix must close that leak, and any added `localStorage` hint must be scoped so one account never reads another's.

## Proposed fix

1. **Replace the binary with a three-state status** for reading history — `loading | empty | ready` — instead of `{...} | undefined`. Update `useTodayContainer.tsx` so `loading` and `ready` both render `TodayContent` (loading shows placeholders), and only `empty` renders `Welcome`.
2. **Derive the first-paint branch from `userId`:** `userId === null` → `empty` → Welcome; `userId !== null` → `loading` → personalized placeholders, then reconcile to `ready`/`empty` when `getUserLastReading` resolves.
3. **Guard `useResumeReadingSection.tsx:25-29`** — it currently throws when `userLastReading` is falsy; in the loading branch it must render a placeholder instead of dereferencing `.value!`.
4. **Reset and guard on every `userId` change.** When `userId` changes, immediately set status back to `loading` (clearing any prior account's value) before fetching, and ignore any async result whose userId is no longer the current one. When `userId` becomes null (sign-out), set status to `empty` (→ Welcome) and drop any cached value. This closes the cross-account leak described above.
5. **Remove the fake loading affordance from the Welcome screen.** Delete the footer spinner and the "Everything begins small" / "No rush" text: the footer block in `components/containers/Welcome.tsx:46-50`, and `footerTitle`/`footerContent` in `hooks/useWelcome.tsx:83-91`. Drop the now-unused `SpinnerIcon` import from `Welcome.tsx`, and the `everything-begins-small` / `no-rush` keys from `en.json` if unused elsewhere. Welcome is now a definite state (shown only when the user is known to be new), so it should not pretend to be loading.
6. **(Optional polish, can be a follow-up)** Persist a lightweight "has-read-before / last-position" hint in `localStorage`, **keyed per `userId`**, so the one remaining ambiguous case — a logged-in user who has never read — resolves to Welcome on first paint instead of showing placeholders that then collapse to Welcome. The hint is only ever read for the current `userId`; a sign-out or account switch must never surface another user's hint. This also lets us paint the real resume card instantly.

## Acceptance criteria

- [ ] A returning user with reading history never renders the Welcome page on a cold load (verify under throttled network).
- [ ] A new/anonymous user (`userId === null`) sees Welcome immediately, with no placeholder detour.
- [ ] No Welcome ↔ personalized screen swap occurs during load.
- [ ] The personalized layout renders a loading state (placeholders) while history is still fetching, rather than the Welcome page.
- [ ] Signing out immediately returns to Welcome with no leftover resume/personalized data from the previous session.
- [ ] Signing in as a different user on the same browser shows that user's own state — never the previous user's resume position, bookmarks, or cached hint.
- [ ] An in-flight history fetch for a previous user cannot overwrite the current user's state.
- [ ] The Welcome screen no longer shows the spinner or the "Everything begins small" footer text.
- [ ] `pnpm check:ts`, `pnpm lint`, and `pnpm test` pass; today-screen container logic has test coverage for the three states and for an account switch.

## Out of scope (separate issues)

- Removing the first-run welcome **modal** (`OnboardingManager` `"welcome"` step).
- Full skeleton designs for each Today section (bookmarks / community / timeline).
- `render()` → `hydrate()` conversion / making SSR paint before the JS bundle.

## Notes

- Reader / gating files: `useTodayContainer.tsx`, `useTodayContent.tsx`, `di/bootstrap.tsx`, `domain/models/readingHistory.tsx`, `components/containers/Welcome.tsx`, `hooks/useWelcome.tsx`.
- The logged-in-but-never-read case is the only one the optional hint fixes; it degrades to placeholder→Welcome (not the jarring Welcome→personalized), so it is safe to ship the core fix first and add the hint later.
- Per project convention, only add or change translation keys in `i18n/en.json`; leave other locale files to the translators.

Blitzy:
## Summary

On a cold load, returning users briefly saw the new-user **Welcome** page inside the Today screen before it flipped to their personalized **Today** content. The switch was driven by whether reading history had loaded yet, not by whether the user is actually new — so anyone with history was flashed the wrong screen while their history was still being fetched.

The root cause was a single binary gate in `useTodayContainer`: `if (!userId || !userLastReading.value) → Welcome`. Because `userLastReading` was `undefined` for **both** "still loading" and "no history," a returning user rendered Welcome (spinner and all) until the fetch resolved, then snapped to `TodayContent`. That snap was the flash.

Closes #1490.

## The fix: a three-state gate

Reading history is now an explicit status — `loading | empty | ready` — instead of `{ bookId, chapter } | undefined` ([readingHistory.tsx](packages/today-screen/domain/models/readingHistory.tsx)):

- **`loading`** — a `userId` is known and the fetch is in flight → renders the personalized layout with placeholders.
- **`empty`** — no `userId` (new/anonymous), or the fetch confirmed no history → renders Welcome.
- **`ready`** — the fetch found a resume position → renders the real content.

Only `empty` renders Welcome. Because `userId` is parsed synchronously from the cached session key at startup, a returning user starts in `loading` and never touches the Welcome screen. "Has history" and "userId known at first paint" are the same condition, which is exactly the invariant called out in the issue.

## Cross-account correctness

The state logic lives in a new, testable factory, [createReadingHistoryState.tsx](packages/today-screen/infrastructure/di/createReadingHistoryState.tsx), which closes the latent A→B leak the issue described:

- On any `userId` **change** it resets to `loading`, clearing the previous account's position before the new fetch resolves.
- Any in-flight fetch whose `userId` is no longer current is **ignored**, so account A's late result can never overwrite account B's state.
- Sign-out (`userId → null`) returns to `empty` (Welcome) immediately, dropping any cached position.
- A same-user refetch (reading progressed) keeps the current card visible while revalidating — no placeholder flicker.

## Loading placeholder

[useResumeReadingSection](packages/today-screen/infrastructure/presentation/hooks/useResumeReadingSection.tsx) no longer throws when there is no value; in the `loading` branch it reports a placeholder, and [ResumeReadingSection](packages/today-screen/infrastructure/presentation/components/containers/ResumeReadingSection.tsx) renders a self-contained shimmer skeleton (local CSS, `rgba`-based to stay baseline-clean). So the wait for reading history now happens behind a skeleton in the real layout instead of behind the Welcome page.

## Welcome cleanup

Removed the fake loading affordance from Welcome, which no longer makes sense now that Welcome is a definite state: the footer spinner and the "Everything begins small" / "No rush" text ([Welcome.tsx](packages/today-screen/infrastructure/presentation/components/containers/Welcome.tsx), [useWelcome.tsx](packages/today-screen/infrastructure/presentation/hooks/useWelcome.tsx)), the now-dead footer CSS, and the `everything-begins-small` / `no-rush` keys from the `en` block of `extension.json` (other locales left for the translators, per convention).

## Testing

- `pnpm check:ts`, `pnpm lint`, and `pnpm test` all pass. `check:ts` and `lint` report 0 errors.
- New unit coverage in [createReadingHistoryState.test.ts](test/unit/today-screen/infrastructure/di/createReadingHistoryState.test.ts) pins the tricky timing cases: the three states, an account switch, the stale-fetch guard, sign-out, and same-user refetch. Container/content/resume hook tests were updated to the status model.
- Manually verified in the browser under throttled network: returning-with-history shows the placeholder and never Welcome; new/anonymous shows Welcome instantly; sign-out returns to Welcome with no leftover data; signing in as a different account shows that account's own state.

## Acceptance criteria

- [x] A returning user with reading history never renders the Welcome page on a cold load (verified under throttled network).
- [x] A new/anonymous user (`userId === null`) sees Welcome immediately, with no placeholder detour.
- [x] No Welcome ↔ personalized screen swap occurs during load.
- [x] The personalized layout renders a loading state (placeholders) while history is still fetching.
- [x] Signing out immediately returns to Welcome with no leftover resume/personalized data.
- [x] Signing in as a different user shows that user's own state — never the previous user's.
- [x] An in-flight history fetch for a previous user cannot overwrite the current user's state.
- [x] The Welcome screen no longer shows the spinner or the footer text.
- [x] `pnpm check:ts`, `pnpm lint`, and `pnpm test` pass; container logic has test coverage for the three states and an account switch.

## Out of scope / follow-up

- The optional per-`userId` `localStorage` hint (issue item 6) is **not** in this PR. Without it, the one remaining ambiguous case — a logged-in user who has never read — degrades to placeholder→Welcome (not the jarring Welcome→personalized), so it is safe to ship the core fix first.
- While investigating, I confirmed the resume card is the slowest section on the screen because "find where you left off" currently means opening a Yjs shared document and scanning a full year of reading events (up to two year-docs), versus already-loaded signal data for the other sections. The `localStorage` hint is the natural follow-up: paint the card instantly from a cached pointer and use the year-scan only to reconcile.
- Still out of scope (separate issues): removing the first-run welcome modal, full skeleton designs for each Today section, and the `render()` → `hydrate()` SSR conversion.

### DIFF
--- a/packages/today-screen/domain/models/readingHistory.tsx
+++ b/packages/today-screen/domain/models/readingHistory.tsx
@@ -56,4 +56,23 @@ export type CommunityReading<T extends string> = {
   [K in T]: FilteredReading;
 };
 
-export type UserLastReading = { bookId: string; chapter: number } | undefined;
+/** A concrete resume position: the last book/chapter a user was reading. */
+export type LastReading = { bookId: string; chapter: number };
+
+/** The result of a last-reading lookup — a position, or `undefined` if none. */
+export type UserLastReading = LastReading | undefined;
+
+/**
+ * The Today screen's reading-history gate as a three-state status, replacing
+ * the old `LastReading | undefined` that conflated "still loading" with "no
+ * history". `loading` and `ready` both render the personalized layout
+ * (`loading` shows placeholders); only `empty` renders the Welcome page.
+ *
+ * - `loading` — a userId is known and the history fetch is in flight.
+ * - `empty`   — no userId (new/anonymous), or the fetch confirmed no history.
+ * - `ready`   — the fetch found a resume position.
+ */
+export type ReadingHistoryState =
+  | { status: "loading" }
+  | { status: "empty" }
+  | { status: "ready"; lastReading: LastReading };
--- a/packages/today-screen/extension.json
+++ b/packages/today-screen/extension.json
@@ -20,6 +20,7 @@
       "wednesday-short": "Wed",
       "friday-short": "Fri",
       "resume-reading": "CONTINUE WHERE YOU LEFT",
+      "resume-reading-loading": "Loading your reading history…",
       "greeting-morning": "Good morning",
       "greeting-afternoon": "Good afternoon",
       "greeting-evening": "Good evening",
--- a/packages/today-screen/extension.json
+++ b/packages/today-screen/extension.json
@@ -5,8 +5,6 @@
     "en": {
       "title": "Today screen",
       "description": "Personalized daily home screen for the reader: time-based greeting, resume-reading shortcut, bookmarks, verse search, a community panel with live sessions and other readers' activity, a reading-history timeline, and chapter recommendations.",
-      "everything-begins-small": "Everything begins small.",
-      "no-rush": "Take your time. — there is no streak to keep, or race to win.",
       "read-first-chapter": "Read the first chapter",
       "open-bible": "Open Bible",
       "anonymous-greeting": "Welcome!",
--- a/packages/today-screen/infrastructure/di/bootstrap.tsx
+++ b/packages/today-screen/infrastructure/di/bootstrap.tsx
@@ -1,6 +1,10 @@
 import { computed, effect, signal } from "@preact/signals";
 import { registerExtension, type SeedBibleState } from "seed-bible";
 import { MaterialIcon } from "@packages/seed-bible/seed-bible/components";
+import {
+  Skeleton,
+  SkeletonContainer,
+} from "@packages/seed-bible/seed-bible/components/Skeleton/Skeleton";
 import { Today } from "../presentation/components/Today";
 import { useI18n } from "@packages/seed-bible/seed-bible/i18n";
 import { TodayReadingHistoryService } from "@packages/today-screen/application/services/TodayReadingHistoryService";
@@ -230,6 +234,8 @@ export const bootstrapExtension = () => {
               config={{
                 ColorParser,
                 MaterialIcon,
+                Skeleton,
+                SkeletonContainer,
                 language,
                 username: context.login.profile.value?.name,
                 userId: context.login.userId.value ?? undefined,
--- a/packages/today-screen/infrastructure/di/bootstrap.tsx
+++ b/packages/today-screen/infrastructure/di/bootstrap.tsx
@@ -5,10 +5,8 @@ import { Today } from "../presentation/components/Today";
 import { useI18n } from "@packages/seed-bible/seed-bible/i18n";
 import { TodayReadingHistoryService } from "@packages/today-screen/application/services/TodayReadingHistoryService";
 import { SubscribedUsersProvider } from "../adapters/subscriptions/SubscribedUsersProvider";
-import type {
-  FilteredReading,
-  UserLastReading,
-} from "@packages/today-screen/domain/models/readingHistory";
+import type { FilteredReading } from "@packages/today-screen/domain/models/readingHistory";
+import { createReadingHistoryState } from "./createReadingHistoryState";
 import type { UtilsAPI } from "@packages/seed-bible-utils/infrastructure/models/seedBible";
 import { getReadingHistoryEvents } from "@packages/seed-bible/seed-bible/managers";
 import { getDefaultTranslationForLanguage } from "@packages/seed-bible/seed-bible/managers";
@@ -85,7 +83,16 @@ export const bootstrapExtension = () => {
         },
       });
 
-      const userLastReading = signal<UserLastReading>(undefined);
+      // Three-state reading-history gate (loading | empty | ready). Derived
+      // from `userId` (known synchronously at startup) so a returning user
+      // never flashes the Welcome page while their history loads.
+      const { readingHistory, dispose: disposeReadingHistory } =
+        createReadingHistoryState({
+          userId: context.login.userId,
+          refetchTrigger: context.app.currentReadingState,
+          getUserLastReading: (userId, range) =>
+            todayReadingHistoryService.getUserLastReading(userId, range),
+        });
 
       /**
        * Fetches the community reading for one exact period. The presentation
@@ -168,33 +175,6 @@ export const bootstrapExtension = () => {
           .trim();
       };
 
-      const cleanupUserLastReading = effect(() => {
-        const userId = context.login.userId.value;
-        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-        context.app.currentReadingState.value;
-
-        if (!userId) {
-          userLastReading.value = undefined;
-          return;
-        }
-
-        const now = Math.floor(Date.now() / 1000);
-        const oneYearAgo = now - 365 * 24 * 60 * 60;
-
-        void todayReadingHistoryService
-          .getUserLastReading(userId, { from: oneYearAgo, to: now })
-          .then((result) => {
-            userLastReading.value = result;
-          })
-          .catch((err) => {
-            console.error(
-              "[Debug] [today-screen] getUserLastReading failed for userId",
-              userId,
-              err
-            );
-          });
-      });
-
       const lastTranslationBooks = signal<{
         books: Array<{
           id: string;
@@ -265,7 +245,7 @@ export const bootstrapExtension = () => {
                       ),
                     }
                   : undefined,
-                userLastReading,
+                readingHistory,
                 getCommunityReading,
                 translate: (key, options) =>
                   t(key, {
@@ -434,7 +414,7 @@ export const bootstrapExtension = () => {
       };
 
       yield () => {
-        cleanupUserLastReading();
+        disposeReadingHistory();
         cleanupTranslationBooks();
         cleanupTranslationId();
         cleanupTodayUrlSync();
--- a/packages/today-screen/infrastructure/di/createReadingHistoryState.tsx
+++ b/packages/today-screen/infrastructure/di/createReadingHistoryState.tsx
@@ -1,0 +1,95 @@
+import { signal, effect, type ReadonlySignal } from "@preact/signals";
+import type {
+  ReadingHistoryState,
+  UserLastReading,
+} from "@packages/today-screen/domain/models/readingHistory";
+
+export interface ReadingHistoryStateDeps {
+  /** The current user id, synchronously known from the cached session key. */
+  userId: ReadonlySignal<string | null>;
+  /**
+   * Any signal whose changes should re-fetch the resume position (the user's
+   * live reading state). Read only for its reactivity — the value is unused.
+   */
+  refetchTrigger: ReadonlySignal<unknown>;
+  getUserLastReading: (
+    userId: string,
+    range: { from: number; to: number }
+  ) => Promise<UserLastReading>;
+}
+
+const ONE_YEAR_SECONDS = 365 * 24 * 60 * 60;
+
+/**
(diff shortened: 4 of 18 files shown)


Blitzy:
> [!NOTE]

## Code review — PR #1492 (fixes #1490)

Overall this is a clean, well-scoped fix that matches the issue's proposed design closely: the three-state `loading | empty | ready` gate, the `userId`-derived first paint, the stale-fetch guard, sign-out handling, and the same-user "no placeholder flicker" refetch are all implemented and directly covered by tests. `Signal` in `Today.tsx` is still used elsewhere, `SpinnerIcon` and the footer CSS are removed cleanly, and the acceptance criteria in the issue are addressed.

There is one behavioral regression worth fixing before merge, plus a few smaller notes.

---

### 🟠 Medium — A failed (or empty) *refetch* throws a returning user back to the Welcome screen

**Where:** `packages/today-screen/infrastructure/di/createReadingHistoryState.tsx` — the `.catch` handler (and the `undefined` branch of `.then`).

**What happens.** The effect refetches whenever `refetchTrigger` fires. That trigger is `context.app.currentReadingState` (`bootstrap.tsx:92`), which changes every time the user reads or navigates a chapter. On such a refetch the code deliberately keeps the current card visible (good — `userChanged` is false, so it does *not* reset to `loading`). But if that refetch **rejects**, the `.catch` unconditionally does:

```ts
readingHistory.value = { status: "empty" };
```

`empty` is the one state that renders `Welcome` (`useTodayContainer.tsx`). So a returning user who already had their resume card showing (`ready`) gets demoted to the **Welcome page** on any transient fetch error — which is exactly the Welcome-flash this PR set out to eliminate, just triggered by a network blip instead of a cold load. The `.then` branch has the same problem if a refetch resolves to `undefined`: a spurious empty result erases a known-good position.

1. Sign in as a user with reading history and open the Today screen — the resume card renders (`ready`).
2. Read / navigate a chapter so `currentReadingState` changes, causing `getUserLastReading` to refetch.
3. Make that one refetch fail — e.g. DevTools Network → Offline for the request window, or temporarily `throw` in `TodayReadingHistoryService.getUserLastReading`.
4. **Observed:** Today flips to the Welcome screen. (A later successful refetch recovers it.)
   **Expected:** the existing resume card stays; a transient refetch error should never surface Welcome to a user who has history.

**Suggested fix.** Only fall to `empty` on a *fresh* load or account switch (where there is genuinely no prior position); on a same-user refetch, keep what's already showing. `userChanged` already distinguishes these two cases — capture it and gate the failure path on it:

```ts
const userChanged = userId !== lastSeenUserId;
// ...
.catch((err) => {
  if (deps.userId.peek() !== requestedUserId) return;
  console.error(
    "[Debug] [today-screen] getUserLastReading failed for userId",
    userId,
    err
  );
  // A fresh load / account switch has no card to fall back to → empty.
  // A same-user refetch keeps the card that's already showing, instead of
  // flashing Welcome on a transient error.
  if (userChanged) readingHistory.value = { status: "empty" };
});
```

Gating on `userChanged` (rather than "is the current state `ready`?") is important so it does **not** reintroduce the cross-account leak: an account switch always has `userChanged === true`, so a failed B-fetch still clears A's card rather than keeping it. Consider applying the same `userChanged` guard to the `undefined` result in the `.then` branch, so a spurious empty refetch result doesn't erase a known position either.

---

### 🟡 Low — Initial-load fetch failure shows Welcome to a returning user (same root cause)

The acceptance criterion *"a returning user with reading history never renders the Welcome page on a cold load"* is violated if the **first** `getUserLastReading` call fails — the `.catch` sets `empty` → Welcome. The test `"falls back to empty when the fetch fails"` (line 184) actually encodes this as intended behavior. This is a more defensible tradeoff than the Medium above (on a cold-load failure there is genuinely no position to show), so it may be acceptable as-is — but it's worth a conscious decision. If you'd rather not show Welcome on a cold-load error, an alternative is to stay in `loading` (placeholder) on failure rather than dropping to `empty`.

---

### 🟡 Low — Orphaned translation keys left in non-`en` locales

`everything-begins-small` and `no-rush` were removed from the `en` block of `packages/today-screen/extension.json` (correct), but they still exist in ~60 other locale blocks (`af`, `am`, `ar`, …). This is consistent with the project convention of "only edit `en`, leave other locales to the translators," so it isn't blocking — but these are now dead keys with no code reference. Worth a note to the translators so they get pruned rather than lingering indefinitely.

---

### ⚪ Nit — Test gap for the refetch-failure case

`createReadingHistoryState.test.ts` covers initial-load failure but not a failure (or `undefined` result) during a *same-user refetch* — which is precisely the Medium scenario above. Suggest adding a case: reach `ready`, bump `refetchTrigger`, `reject` the second fetch, and assert the state remains `ready` (with the fix applied). That would pin the behavior against future regressions.

---

### ⚪ Nit — Skeleton accessibility

`ResumeReadingSection.tsx` marks the loading placeholder `aria-hidden="true"` with no `aria-busy` / live-region hint, so assistive-tech users get silence during the load rather than a "loading" cue. Minor and optional, but an `aria-busy` on the container or an SR-only label would be a small polish.

---

**Verification note:** findings above are from reading the diff and tracing the signal/effect flow; I did not re-run `pnpm test` / `check:ts` (the PR reports both green, and the CI build-size job succeeded). The Medium finding is a logic regression not currently covered by tests, so it would pass CI as-is.

---

Blitzy:
Will also close #1437 when this PR is merge.d

PR #1494, opened 2026-07-22
Remove redundant welcome modal; reorder tutorial and install onboarding

### LINKED ISSUE #1487: Remove the "Welcome" modal
The Welcome modal is no longer needed now that we have the Welcome screen as part of Today.

Blitzy:
## Summary

- Removes the separate first-run "Welcome" onboarding modal — it duplicated the install prompt's job of greeting a new user, so the install prompt now handles that first-run moment on its own.
- Defers the tutorial offer until the reader is actually open to a chapter and visible: no fullscreen pane (e.g. Today) covering it, and not in the brief startup gap before Today auto-opens on a cold load.
- Moves the PWA install prompt so it now appears after the tutorial has been resolved (accepted, declined, skipped, or opted out), instead of before it. It's now offered to any user who hasn't installed or dismissed it, including logged-in users, since by this point their profile has already loaded and there's no risk of flashing a stale prompt.

### Before / after

**Before:** on a fresh visit, the install prompt showed first (computed from localStorage alone, before login/profile even resolved), and the tutorial offer waited behind it. The tutorial could also appear while a fullscreen pane like Today was covering the reader, so its highlighted UI elements weren't even on screen.

**Now:** the tutorial offer only appears once the reader is genuinely visible — chapter loaded, nothing covering it, and (on a cold load into Today) only after Today has actually opened and been closed. Once the tutorial is resolved one way or another, the install prompt appears.

## Test plan

- [x] `pnpm check:ts` — clean
- [x] `pnpm test` — all 3062 tests pass, including updated/new cases in `OnboardingManager.test.tsx` and `TutorialManager.test.tsx`
- [x] `pnpm lint` — no new warnings or errors
- [x] Manual verification in the browser (recommended before merge):
  - [x] Fresh anonymous cold load: Today auto-opens, no tutorial offer flash; close Today → tutorial offer appears → decline/accept → install prompt appears
  - [x] Deep link (`?book=GEN&chapter=1`): Today doesn't auto-open, tutorial offer appears, then install prompt
  - [x] Session link (`?sessionId=...`): neither tutorial offer nor install prompt shows
  - [x] Returning user (tutorial already seen): install prompt appears once the reader is visible
  - [x] Logged-in, not installed: install prompt now appears after the tutorial resolves
  - [x] Settings → "Launch tutorial" and "Install app" replay entries still work

### DIFF
--- a/packages/seed-bible/seed-bible/components/Onboarding/Onboarding.tsx
+++ b/packages/seed-bible/seed-bible/components/Onboarding/Onboarding.tsx
@@ -2,17 +2,13 @@ import "./Onboarding.css";
 import type { ComponentChildren } from "preact";
 import { useI18n } from "../../i18n/I18nManager";
 import { LANG_META } from "../../i18n/languageMeta";
-import {
-  SeedBibleWordmark,
-  InstallAppsIcon,
-  SafariIcon,
-} from "../../components/icons";
+import { InstallAppsIcon, SafariIcon } from "../../components/icons";
 import type { OnboardingManager } from "../../managers/OnboardingManager";
 import type { CasualOSManager } from "../../managers/OsManager";
 
 /**
- * First-run onboarding modals: a welcome notice, then a device-aware prompt to
- * install the app / add it to the home screen.
+ * First-run onboarding modal: a device-aware prompt to install the app / add it
+ * to the home screen.
  *
  * The install affordance differs per platform:
  *  - Android / PC: a real "Install App" button that triggers the native PWA
@@ -32,7 +28,6 @@ export function OnboardingModals({
   toast: (message: string) => void;
   className?: string;
 }) {
-  const { t } = useI18n();
   const step = onboarding.step.value;
 
   if (step === "done") {
@@ -51,39 +46,6 @@ export function OnboardingModals({
       </div>
     </div>
   );
-
-  if (step === "welcome") {
-    return card(
-      <>
-        <div className="sb-onboarding-logo">
-          <SeedBibleWordmark height={52} />
-        </div>
-        <h2 className="sb-onboarding-title">
-          {t("onboarding.welcomeTitle", { defaultValue: "Welcome!" })}
-        </h2>
-        {/* <p className="sb-onboarding-body">
-          {t("onboarding.welcomeBodyPre", {
-            defaultValue: "This scripture session is ",
-          })}
-          <strong>
-            {t("onboarding.welcomeBodyEmphasis", { defaultValue: "temporary" })}
-          </strong>
-          {t("onboarding.welcomeBodyPost", {
-            defaultValue: " and will erase in 12 hours for your privacy.",
-          })}
-        </p> */}
-        <div className="sb-onboarding-actions">
-          <button
-            type="button"
-            className="sb-onboarding-btn sb-onboarding-btn-primary"
-            onClick={onboarding.completeWelcome}
-          >
-            {t("onboarding.continue", { defaultValue: "Continue" })}
-          </button>
-        </div>
-      </>
-    );
-  }
 
   // step === "install" — but never prompt someone who already has the app
   // (e.g. the profile loaded after mount and reported it installed).
--- a/packages/seed-bible/seed-bible/i18n/en.json
+++ b/packages/seed-bible/seed-bible/i18n/en.json
@@ -229,8 +229,6 @@
   "tutorial.back": "Back",
   "tutorial.done": "Done",
   "tutorial.next": "Next",
-  "onboarding.welcomeTitle": "Welcome!",
-  "onboarding.continue": "Continue",
   "onboarding.installTargetDesktop": "desktop",
   "onboarding.installTargetMobile": "home screen",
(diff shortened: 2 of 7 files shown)


Blitzy:
> [!NOTE]

## Code review — #1494 (Remove redundant welcome modal; reorder tutorial and install onboarding)

Reviewed against linked issue **#1487** ("Remove the 'Welcome' modal — no longer needed now that we have the Welcome screen as part of Today"). The change removes the welcome step, gates the tutorial offer on the reader actually being visible, and defers the install prompt until the tutorial is resolved. The direction matches the issue, and the mechanics are sound. Findings below are ranked by severity.

### What I verified locally (all green)
- `pnpm vitest run OnboardingManager TutorialManager` → **11/11 pass**
- `pnpm check:ts` → **0 visible errors** (6 pre-existing suppressed, unrelated)
- `todayWillAutoOpen` (SeedBibleStateManager) is a **faithful copy** of today-screen's `isTodayOpen` predicate in `bootstrap.tsx` — both use `{book, chapter, verse, sessionId}` as the competing-deep-link set with `?today` overriding.
- `coveredByPane` in `readerVisible` is an **exact copy** of `isFullscreenPaneVisible` in `BibleReaderToolbar.tsx` (`placement === "fullscreen" || isMobile`).
- No dangling references to the removed `completeWelcome` / `WELCOME_SEEN_KEY` / `"welcome"` step in code.
- Every tutorial-resolution path (`finish`, `optOut`, `dismissPrompt`/decline) sets `completed` or `optedOut`, so the install-offer effect's `tutorialResolved` guard can never dead-end — the PR's "accepted, declined, skipped, or opted out → install appears" claim holds.

---

### 🟠 Medium — onboarding can silently stall if the Today pane never actually opens

**File:** `packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx` (~L629–669)

`readerVisible` returns `false` while `todayWillAutoOpen && !todayHasOpened.value`. `todayWillAutoOpen` is derived **purely from the initial URL**, but `todayHasOpened` only latches `true` when a pane with id `today-screen-pane` is actually observed:

```ts
const todayHasOpened = signal(false);
effect(() => {
  if (panes.panes.value.some((pane) => pane.id === "today-screen-pane")) {
    todayHasOpened.value = true;
  }
});
```

If the URL predicts Today should auto-open but the `today-screen` pane never opens — the extension is disabled / fails to load, its bootstrap throws, or a future change alters *when* Today opens without the pane appearing — then `todayHasOpened` stays `false` forever, `readerVisible` stays `false` forever, and **both** the tutorial offer and the install prompt silently never appear for that session. There is no timeout or fallback that releases the gate.

In practice `today-screen` is a core extension that reliably opens its pane from the same URL predicate, so today the probability is low — this is a resilience gap rather than a reproducible bug. But it's a silent single-point dependency where a failure in one module disables all onboarding in another, with no diagnostic.

**Reproduction (requires forcing the abnormal condition):** cold-load anonymously with no deep-link params (so `todayWillAutoOpen === true`) while preventing the `today-screen` pane from opening. Expected: once the reader is visible, the tutorial offer (then install prompt) appears. Actual: neither ever appears.

**Suggested fix:** don't wait indefinitely on a signal derived from a different module. Either (a) gate on the same `isTodayOpen` state today-screen actually uses rather than re-deriving it from the URL, or (b) add a short fallback so the "about to be covered by Today" guard is dropped after a bounded delay if the pane never materializes.

---

### 🟡 Low — two hand-copied predicates that will silently drift

**File:** `SeedBibleStateManager.tsx`

1. `todayWillAutoOpen` (~L629) duplicates today-screen's `isTodayOpen` logic in `packages/today-screen/infrastructure/di/bootstrap.tsx` (~L374–383).
2. `coveredByPane` (~L659) duplicates `isFullscreenPaneVisible` in `BibleReaderToolbar.tsx` (~L538–541).

Both are acknowledged in the comments ("mirrors the predicate in…"), and both are currently faithful. The risk is drift: if, say, `playlist` is later added to today-screen's competing-deep-link set (it currently isn't in either), this copy won't follow and `readerVisible` will mis-gate — with no test catching the divergence since each side is tested independently.

**Suggested fix:** extract the auto-open predicate into a small shared helper (e.g. `willTodayAutoOpen(initialUrlParams)`) imported by both, and likewise share the fullscreen-pane-visible predicate. Low urgency; safe to defer.

---

### ⚪ Nit — orphaned i18n keys in non-`en` locales

`onboarding.welcomeTitle` and `onboarding.continue` were correctly removed from `en.json`, but they still exist in ~23 other locale files (`ar`, `da`, `it`, `ko`, …). Per `CLAUDE.md` ("only update en.json … leave [others] for the professionals") this is the expected convention, and no code references them anymore — just flagging them as now-dead keys for the translators to prune later. No action needed in this PR.

---

### Overall
Clean, well-commented change that does what #1487 asks; the removal is complete and the reordering logic is correct on all the paths I traced. The only substantive item is the Medium resilience gap around `readerVisible` depending on the Today pane actually opening — worth a follow-up even if it's not blocking, since the failure mode is total and silent. The two duplication items are maintainability nits.


---

PR #1495, opened 2026-07-22
feat: build info footer + automated release flow

Blitzy:
## Summary

This branch adds two related pieces of work, both centered on the app version: it shows the running build's version and commit in Settings, and it turns "merge `develop` → `main`" into a real release that tags the commit, publishes a GitHub Release, and dates the CHANGELOG.

### 1. Build info in the Settings footer

The client bundle now bakes in its own version and commit at build time (`__APP_VERSION__` from `packages/seed-bible/package.json`, `__GIT_COMMIT__` from CI's `DEPLOY_BUILD_ID`, falling back to `git rev-parse` locally), and the Settings page shows them in a footer that copies the full `v<version> (<commit>)` string on click. Because the value travels inside the JS chunk, a build reports its own version even when a stale copy is being served.

### 2. CHANGELOG + automated releases

Production already ships by merging `develop` → `main` (cd.yml deploys the merged commit to seedbible.org). This adds everything around that deploy: a git tag, a GitHub Release, and a dated CHANGELOG entry.

Releasing is a two-step flow. On `develop`, `pnpm release:prepare <bump>` bumps the version, rewrites the CHANGELOG's `## TBD` heading to `## v<version> — <YYYY-MM-DD>`, and inserts a fresh empty `## TBD` above it so `develop` is ready for the next cycle. After that merges to `main`, the new `release.yml` workflow reads the version, tags the commit `v<version>`, and publishes a GitHub Release whose notes are that version's CHANGELOG section.

Stamping on `develop` (rather than automatically on `main`) keeps the git history clean: no bot commits on the protected branch, the tag points exactly at the deployed commit, and there's no next-cycle CHANGELOG merge conflict. The workflow is idempotent — a push to `main` that didn't bump the version (e.g. a hotfix) deploys but creates no release, and re-runs never duplicate one.

## Files of note

- `.github/workflows/release.yml` — tags + publishes the Release on push to `main`; install/publish steps are gated behind a "does this release already exist?" check so no-op pushes stay cheap.
- `script/lib/changelog.ts` — pure, tested helpers (`bumpVersion`, `stampChangelog`, `extractSection`).
- `script/release.ts` (`pnpm release:prepare`) and `script/changelog.ts` (`pnpm changelog:extract`, used by CI).
- `README.md` — a new "Releasing" section documenting the flow.

## Testing

- `pnpm vitest run changelog.test.ts` — 13 unit tests covering version bumping, TBD stamping, and note extraction (including the round-trip and the missing/empty-section failures).
- `pnpm check:ts` — clean (0 visible errors; the new script files aren't in the tsc-silent suppress list).
- Ran `pnpm release:prepare patch` end-to-end against the real files, confirmed the version bump, the `## v… — <date>` heading (em dash) with a fresh TBD, and that `pnpm changelog:extract` returned exactly that section — then reverted.

## Notes for the first release

The current CHANGELOG `## TBD` is not yet stamped and `packages/seed-bible/package.json` is already at `1.1.0`. To cut `v1.1.0` without a further bump, run `pnpm release:prepare 1.1.0` on `develop` (explicit version) to stamp the existing TBD, then merge as usual. If the changelog is left unstamped, `release.yml` fails the extract step rather than publishing empty notes.

`release.yml` is push-only. To test-run it safely before relying on it, temporarily add a `workflow_dispatch:` trigger (or test on a fork) against a throwaway `vX.Y.Z-test` tag, then remove it.

The root `package.json` version is intentionally left at `1.0.0` — only the app package version (`packages/seed-bible/package.json`) is bumped, matching the existing convention.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,3 +85,19 @@ jobs:
             --target "$GITHUB_SHA" \
             $PRERELEASE
           echo "Published release **v$VERSION**." >> "$GITHUB_STEP_SUMMARY"
+
+      # Repost the release notes to the Discord #announcements channel with a
+      # link to the site. No-ops when DISCORD_ANNOUNCE_WEBHOOK is unset, and
+      # never fails the job (the release is already published). Skipped for
+      # pre-releases (e.g. 1.2.0-rc.1).
+      - name: Announce on Discord
+        if: steps.check.outputs.should_release == 'true' && !contains(steps.check.outputs.version, '-')
+        env:
+          VERSION: ${{ steps.check.outputs.version }}
+          DISCORD_ANNOUNCE_WEBHOOK: ${{ secrets.DISCORD_ANNOUNCE_WEBHOOK }}
+        run: |
+          pnpm notify discord-release \
+            --version "$VERSION" \
+            --notes-file notes.md \
+            --site-url "https://seedbible.org" \
+            --release-url "https://github.com/$GITHUB_REPOSITORY/releases/tag/v$VERSION"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,20 +66,28 @@ jobs:
           corepack enable
           pnpm install
 
-      # Pull the release notes straight from the (already stamped) CHANGELOG. A
-      # missing/empty section is not a hard failure: we log a warning and skip
-      # the rest of the release (no tag, no GitHub Release, no announcement)
-      # rather than failing the job, so an accidental unstamped merge to `main`
-      # still deploys cleanly without a red X.
+      # Pull the release notes straight from the (already stamped) CHANGELOG.
+      # `changelog:extract` exits 2 for the expected "not stamped yet" case
+      # (see EXTRACT_NOT_FOUND_EXIT_CODE in script/lib/changelog.ts) — that's
+      # not a hard failure: we log a warning and skip the rest of the release
+      # (no tag, no GitHub Release, no announcement), so an accidental
+      # unstamped merge to `main` still deploys cleanly without a red X. Any
+      # OTHER non-zero exit is a genuine bug in the script, not an unstamped
+      # changelog, and is left to fail the job with its stack trace intact.
       - name: Extract release notes
         id: extract
         if: steps.check.outputs.should_release == 'true'
         env:
           VERSION: ${{ steps.check.outputs.version }}
         run: |
-          if pnpm --silent changelog:extract "$VERSION" > notes.md 2>extract-error.log; then
+          set +e
+          pnpm --silent changelog:extract "$VERSION" > notes.md 2>extract-error.log
+          EXIT_CODE=$?
+          set -e
+
+          if [ "$EXIT_CODE" -eq 0 ]; then
             echo "notes_available=true" >> "$GITHUB_OUTPUT"
-          else
+          elif [ "$EXIT_CODE" -eq 2 ]; then
             echo "notes_available=false" >> "$GITHUB_OUTPUT"
             cat extract-error.log
             echo "::warning::No stamped CHANGELOG section found for v$VERSION — skipping the release. Run 'pnpm release:prepare' on develop before merging to main next time."
@@ -90,6 +98,10 @@ jobs:
               echo ""
               echo "Run \`pnpm release:prepare\` on \`develop\` before merging to \`main\` next time."
             } >> "$GITHUB_STEP_SUMMARY"
+          else
+            cat extract-error.log
+            echo "::error::changelog:extract failed unexpectedly (exit $EXIT_CODE) for v$VERSION — this looks like a bug in the script, not an unstamped changelog. See the log above."
+            exit "$EXIT_CODE"
           fi
 
       # gh creates the tag `v<version>` at this commit (= the commit cd.yml
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,12 @@
 # It is idempotent: if a release for the current version already exists, it does
 # nothing — so a push to `main` that didn't bump the version (e.g. a hotfix)
 # deploys but creates no release, and re-runs never duplicate one.
+#
+# It is also tolerant of an unstamped CHANGELOG: if `main`'s version has no
+# matching `## v<version>` section (e.g. someone merged without running
+# `pnpm release:prepare` on develop first), the job logs a warning and skips
+# the release rather than failing — a missed release shouldn't put a red X on
+# the protected branch.
 name: Release
 
 on:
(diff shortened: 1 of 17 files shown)


Blitzy:
> [!NOTE]

## Code review — build info footer + automated release flow

Reviewed the diff between `develop` (`77e3beb`) and the PR head (`86ca38a`): 14 files, +602/-1. No linked issue is referenced in the description, and the only PR comment so far is the automated deployment link, so there was no prior discussion to reconcile with.

**Verification I ran locally:**
- `pnpm vitest run changelog.test.ts` → **13/13 pass**.
- `pnpm check:ts` → **0 visible errors** (the new `__APP_VERSION__`/`__GIT_COMMIT__` globals are picked up because `typings/` is in the tsconfig `include`).
- Confirmed `cd.yml` sets `DEPLOY_BUILD_ID=${{ github.sha }}` (full commit SHA), so the footer's commit is correct in production.
- Confirmed `copy`/`copied`/`app-version` i18n keys all exist in `en.json` (satisfies the `translation-missing-keys: error` rule).
- Confirmed `release.yml`'s `setup-bun` + `corepack enable` + `pnpm install` **mirrors the existing `cd.yml`** convention — consistent, not a regression.

Overall this is clean, well-tested, and well-documented work. One operational footgun is worth addressing before this rides to `main`; the rest are minor.

---

### 🟠 Medium

**1. The first release to `main` will fail the Release workflow.**

`packages/seed-bible/package.json` is already at `1.1.0`, but `CHANGELOG.md` still has an unstamped `## TBD` (no `## v1.1.0 …` heading). On the first `develop → main` merge, `release.yml` resolves `VERSION=1.1.0`, finds no existing `v1.1.0` release, so `should_release=true`, and then the **Extract release notes** step throws and fails the job — a red Actions run on the protected `main` branch. `cd.yml` still deploys, but the Release job fails.

The PR's "Notes for the first release" documents the workaround, but the *natural* flow (just merge `develop → main`) triggers the failure, so it's easy to hit.

*Reproduction (verified):*
```bash
$ node -p "require('./packages/seed-bible/package.json').version"
1.1.0
$ pnpm --silent changelog:extract 1.1.0
Error: No changelog section found for version "v1.1.0".
    at extractSection (script/lib/changelog.ts:123:11)
# EXIT CODE: 1
```

*Suggested fixes (pick one):*
- Run `pnpm release:prepare 1.1.0` on `develop` **before** merging (stamps the existing `## TBD` → `## v1.1.0 — <date>`), as the notes describe — and consider doing it as part of this PR so the branch is self-consistent, or
- Make `release.yml` tolerant of the "no section found" case for a genuinely-unstamped version: treat a missing/empty changelog section as *skip the release with a warning* rather than a hard job failure, so an accidental unstamped merge deploys cleanly without a red X.

---

### 🟡 Low

**2. The footer shows "Copied" even when nothing was copied.** (`SettingsPage.tsx`, `SettingsVersionFooter`)

```ts
void navigator.clipboard?.writeText(`v${__APP_VERSION__} (${__GIT_COMMIT__})`);
copied.value = true;
```

Two edge cases:
- In an insecure context (plain `http://`, some embedded webviews) or an older browser, `navigator.clipboard` is `undefined`; the `?.` short-circuits, nothing is copied, yet the UI still flips to "Copied".
- If `writeText` **rejects** (permission denied), the promise is `void`-discarded, producing an unhandled rejection, and "Copied" is still shown.

*Suggested fix:* gate the confirmation on the promise resolving, and handle the missing-API / rejection paths:
```ts
const onCopy = () => {
  navigator.clipboard
    ?.writeText(`v${__APP_VERSION__} (${__GIT_COMMIT__})`)
    .then(() => {
      copied.value = true;
      setTimeout(() => (copied.value = false), 1500);
    })
    .catch(() => {
      /* leave the label unchanged, or surface a failure state */
    });
};
```
*Repro:* open Settings over a non-secure origin (or a browser without the async clipboard API) and click the version footer — it reads "Copied" but the clipboard is unchanged.

---

### ⚪ Nits (optional)

**3. Displayed vs. copied text differ.** The footer renders `v<version> · <7-char commit>` but clicking copies `v<version> (<full commit>)` — different separator *and* commit length. Copying the full SHA is reasonable, but the mismatch is mildly surprising; consider aligning the formats or leaving a comment noting the intent.

**4. `resolveGitCommit()` shells out on every Vite config load.** `const gitCommit = resolveGitCommit();` runs at config-evaluation time, so `git rev-parse HEAD` is invoked on each `pnpm dev` startup too, not only on builds. Harmless (CI sets `DEPLOY_BUILD_ID`, so `git` isn't called there; local falls back to `"unknown"` on failure), but it could be scoped to build-only or lazily resolved.

**5. `setTimeout` isn't cleared on unmount** in the footer. Harmless with signals (no React-style "set state on unmounted component" warning), but a stray timer if the user navigates away within 1.5s.

---

### 👍 What's good
- Release helpers in `script/lib/changelog.ts` are pure and I/O-free, with solid unit coverage (round-trip stamp→extract, missing/empty-section failures, the `-` prerelease-vs-keyword distinction).
- The `extractSection` heading regex correctly avoids false matches (`v1.1.0` won't match `v1.1.0-rc.1` or `v11.1.0`), and the em-dash heading is handled.
- `release.yml` is genuinely idempotent (existing-release check) and gates the expensive install/publish steps behind it, so no-op hotfix pushes to `main` stay cheap.
- Baking the version/commit into the JS chunk (rather than fetching at runtime) is the right call for reporting the true build even behind a stale cache, and it's applied via `define` so client and SSR stay in sync.

---

Blitzy:
I am considering the current main branch to be 1.1.0, so when we do our next release it will become 1.2.0

Blitzy:
> [!NOTE]

## Code review — follow-up on build info footer + automated release flow

Re-reviewed at the current head (`34db6351`) against `develop` (`77e3beb`): 3 new commits landed since the last review here (`613b7803`, `7dee101c`, `34db6351`), on top of the previously-reviewed `86ca38a`. No linked issue in the description; no other discussion threads to reconcile beyond the earlier review comment and Blitzy's follow-up note about `1.1.0` becoming the baseline for the next release.

**Verified locally:**
- `pnpm vitest run changelog.test.ts discord.test.ts` → **18/18 pass**.
- Re-ran `pnpm changelog:extract 1.1.0` → still exits 1 (changelog is genuinely unstamped), confirming the new `release.yml` skip-path is what will actually execute on the first `develop → main` merge.

### Previously-flagged issues — both fixed ✅

- **Medium (release workflow would fail on first release):** `34db6351` now treats a failed `changelog:extract` as "skip the release with a warning" (`notes_available=false`, gated `::warning::` + step-summary message) instead of throwing. Confirmed by reproducing the exact failure above and reading the updated `Extract release notes` step — it no longer propagates the error.
- **Low (footer showed "Copied" even when nothing was copied):** `7dee101c` moved `copied.value = true` inside the `.then()` and added a `.catch()`, so a missing `navigator.clipboard` or a rejected write no longer falsely reports success.

### 🟡 Low

**1. The Discord-announce step swallows all extract failures the same way, including real bugs.**

`Extract release notes` (`.github/workflows/release.yml`) now treats *any* non-zero exit from `pnpm changelog:extract` as "changelog isn't stamped" and quietly skips the release with a generic warning. That's the intended fix for the unstamped-changelog case, but it means a genuine break in the script (a typo in the `pnpm run` name, a Node/tsx crash, a bad import) would produce the exact same "skipping the release, run `pnpm release:prepare`" message — which would send a developer down the wrong path chasing a changelog-stamping problem that isn't the real cause.

*Suggested fix (optional):* have `script/changelog.ts extract` exit with a distinct, documented status/message for "no stamped section found" (its current, expected failure mode) versus letting other exceptions surface with their own stack trace in the workflow log (which they already do via `cat extract-error.log` — so the info is there, just easy to miss under the reassuring warning banner above it).

### ⚪ Nits (carried over from the last review, still unaddressed — optional)

Not required for this PR, just flagging that they weren't part of the latest 3 commits: the copy-vs-display text mismatch (parens+full SHA vs `·`+7-char SHA in `SettingsVersionFooter`), `resolveGitCommit()` shelling out on every `vite.config.ts` evaluation rather than only at build time, and the uncleared `setTimeout` in the footer's copy-confirmation state.

### 👍 What's good

- Both real issues from the first pass were fixed correctly and minimally — no scope creep, no regressions in the fix commits themselves.
- The new Discord announcement path (`script/lib/discord.ts`) is pure/testable like the changelog helpers, has solid truncation-boundary tests, and explicitly suppresses `@mentions` in notes before posting.
- The Discord step is correctly gated behind both `should_release` and `notes_available`, and skips pre-releases (`!contains(version, '-')`), consistent with how the Release-creation step is gated.

---

PR #1496, opened 2026-07-22
Feature/1429

### LINKED ISSUE #1429: Copying verses with poem lines misses some line breaks
For example:

```
The kings of the earth take their standand the rulers gather together,against the LORDand against His Anointed One: (PSA 2:2)
```

Image

It should be:

```
The kings of the earth take their stand and the rulers gather together, against the LORD and against His Anointed One: (PSA 2:2)
```

Kabi4:
Fixes #1429


Blitzy:
> [!NOTE]

## Code review — PR #1496 (Fixes #1429: "Copying verses with poem lines misses some line breaks")

**Scope:** one file changed, `packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx`.

**Short version:** The core idea is right — poem lines were being concatenated with an empty string, so words on adjacent lines touched (`take their stand` + `and the rulers` → `standand`), and joining with a space fixes that. But the change ships a leftover debug `console.log`, and the blanket `.trim()` + `join(" ")` introduces a new whitespace bug for any verse that contains a footnote, an inline line break, or an inline heading. There's also no test that actually reproduces the reported bug.

---

### 🔴 High — leftover `console.log` debug statement

`packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx:792`

```js
console.log(
  "logs copy-verse",
  context.readingState.selectedVerses.value
);
```

This looks like debug scaffolding that was committed by accident. It fires on **every** verse copy and dumps the full selected-verse objects to the browser console. There is no `no-console` ESLint rule in `eslint.config.mts`, so nothing catches it — the build is green and it would ship to production as-is.

**Fix:** delete the statement.

---

### 🟠 Medium — `join(" ")` inserts extra/multiple spaces around non-text parts

`packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx:961-968`

A verse's `content` is an array of `string | FormattedText | InlineHeading | InlineLineBreak | VerseFootnoteReference` (see `FreeUseBibleAPI.tsx:345-352`). The mapper returns `""` for anything that isn't a string or a `{ text }` object — i.e. for inline line breaks (`{ lineBreak: true }`), footnote references (`{ noteId }`), and inline headings (`{ heading }`). When one of those empty results sits between two text parts, `join(" ")` puts a space on **both** sides of the empty string, producing two or more consecutive spaces in the copied text.

**Reproduction (concrete, using the shape the app's own test fixtures already use — see `test/unit/seed-bible/components/BibleReader.test.tsx:95`):**

Verse content:
```js
[
  "In the beginning ",
  { text: "I am the light", wordsOfJesus: true },
  { lineBreak: true },   // → ""
  { noteId: 7 },         // → ""
  "God created.",
]
```
- Copy this verse.
- **Now (this PR):** `"In the beginning I am the light   God created. (…)"` — three spaces between *light* and *God* (one from each join gap around the two empty parts).
- **Expected:** a single space — `"In the beginning I am the light God created. (…)"`.

The same happens for a poem verse whose lines are separated by an explicit `{ lineBreak: true }` part: you get a double space rather than one. (Before this PR these same cases ran the words together with no space — so the change trades one whitespace bug for another rather than fully fixing it.)

**Suggested fix** — keep the raw part text (don't trim per-part), join, then normalize whitespace and drop spaces before closing punctuation:

```js
const text = verse.verse.content
  .map((part) => {
    if (typeof part === "string") return part;
    if (part && typeof part === "object" && "text" in part)
      return (part as { text: string }).text;
    return ""; // headings, line breaks, footnote refs
  })
  .join(" ")
  .replace(/\s+/g, " ")                 // collapse runs of whitespace to one
  .replace(/\s+([,.;:!?’”)\]])/g, "$1") // no space before closing punctuation
  .trim();
return `${text} (${verseReference})`;
```

This preserves the spaces already embedded in prose strings, adds a separator between poem lines, and collapses the extra spaces the empty parts would otherwise create. (The punctuation regex is an English/Latin heuristic — fine here, but worth noting it won't cover every script.)

---

### 🟡 Low — no test reproduces the reported bug or guards the fix

The existing `copy-verse` test (`test/unit/seed-bible/managers/BibleToolsManager.test.tsx:471-528`) builds its fixture from plain strings that already end in a trailing space:

```js
content: [
  "The kings of the earth take their stand ",
  "and the rulers gather together, ",
  ...
]
```

With trailing spaces present, `join("")` (old) and `.trim()` + `join(" ")` (new) produce the *same* string — so this test passes identically before and after the change. It therefore neither reproduces issue #1429 (real poem data arrives as `FormattedText` objects like `{ text: "…", poem: 1 }` **without** trailing spaces) nor locks in the new spacing behavior.

**Suggested fix:** add a case using the real `FormattedText` shape (poem parts with no trailing space, plus a `{ lineBreak: true }` and a `{ noteId }`) and assert single-spaced output. That both proves the fix and guards against the Medium issue above.

---

### Notes
- CI (`build (24.x)`) is green; there is no lint/type gate that would catch the `console.log` or the whitespace behavior, so these need to be caught in review.
- No review comments or prior discussion on the PR or issue #1429 beyond the original bug report, so nothing here is contradicting an earlier decision.

Overall: sound direction, but please remove the debug log and harden the join so verses with footnotes / line breaks / headings don't pick up stray double spaces.

---

Blitzy (approved):

PR #1497, opened 2026-07-22
Feature/1481

### LINKED ISSUE #1481: Automatically copy session link to clipboard when starting a shared session.
It would be convenient to have the shared session link automatically copied to the clipboard when you create one.

Requirements:

- Full shared session url is automatically copied to clipboard on both desktop and mobile.
- Toast that occurs when session is created and explains that a link has been copied to clipboard.

Kabi4:
Fixes #1481


Blitzy:
> [!NOTE]

## Code Review — PR #1497 (Fixes #1481)

### What this PR does

Issue #1481 asks that the shared-session link be copied to the clipboard (with a toast) when you start a session, on **both desktop and mobile**. A maintainer noted in the issue: *"This already happens on desktop, but not mobile for some reason."*

This change explains and fixes that "for some reason": there were two entry points for creating a shared session, and only one of them copied the link.

- **Desktop** used the "More" context menu (`Tabs.tsx:777`), which received the copy-and-toast logic via a `createSharedSession` prop passed down from `Sidebar`. ✅ copied.
- **Mobile** used a dedicated button rendered only inside the `app.isMobile.value` block (`Tabs.tsx:2037`, button at `2106`). Before this PR it called `void state.app.createSharedSession()` **directly** — no clipboard write, no toast. ❌ never copied.

The PR extracts a single helper `createSharedSessionAndCopyLink(state, t)` (`Tabs.tsx:1248`) and calls it from **both** call sites (`:779` and `:2116`), and removes the now-unused `createSharedSession` prop from `TabsHeader`. This correctly makes the mobile button copy the link and show the toast, so the immediate symptom is addressed. The refactor is clean, the removed prop has no lingering references, and the reused i18n key `link-to-join-shared-session-copied` already exists in `en.json`.

Findings below, by severity.

---

### 🟠 Medium–High — Clipboard write happens *after* an async network `await`; may still fail on iOS Safari (the exact target platform)

The helper does:

```ts
const session = await state.app.createSharedSession(); // network round-trip
const url = getSessionUrl(session);
await navigator.clipboard.writeText(url.href);         // runs AFTER the await
```

`state.app.createSharedSession()` → `handleCreateSharedSession` → `sessions.createSession()` → `createBibleReadingSession`, which `await`s `os.getSharedDocument(null, id, "session_data")` (`SessionsManager.tsx:568`) — a network/shared-document setup.

On **iOS Safari**, `navigator.clipboard.writeText()` requires *transient user activation* — it must run within the short window that follows the user's tap. Once you `await` a network operation, that activation is gone, and Safari rejects the write with `NotAllowedError`. Desktop Chrome/Firefox are far more permissive, which is a plausible second reason the old behavior "worked on desktop but not mobile." So consolidating the code is necessary but **may not be sufficient** — the mobile button now *attempts* the copy, but the copy itself can still fail on the platform the issue is about.

**Reproduction (needs a real iOS device or the alpha deploy https://alpha.seedbible.org/b/feature_1481):**
1. Open on iOS Safari.
2. Tap the mobile "New shared session" button.
3. Observe whether the link is actually on the clipboard (paste into another app) and whether the toast appears.

If the paste is empty, the async-then-clipboard ordering is the cause.

**Suggested fix** — tie the clipboard write to the gesture instead of the network result. On Safari, `ClipboardItem` accepts a *promise* value, which preserves the activation across the async gap:

```ts
async function createSharedSessionAndCopyLink(
  state: SeedBibleState,
  t: ReturnType<typeof useI18n>["t"]
) {
  const linkText = state.app
    .createSharedSession()
    .then((session) => getSessionUrl(session).href);

  try {
    if (typeof ClipboardItem !== "undefined") {
      await navigator.clipboard.write([
        new ClipboardItem({
          "text/plain": linkText.then(
            (text) => new Blob([text], { type: "text/plain" })
          ),
        }),
      ]);
    } else {
      await navigator.clipboard.writeText(await linkText);
    }
    state.app.toast(/* copied */);
  } catch {
    // fall back: still surface the link so the session isn't lost — see below
  }
}
```

Given the whole issue is about mobile, please verify on an actual iOS device before merging rather than relying on desktop testing.

---

### 🟡 Medium — No error handling; a failed clipboard write becomes an unhandled rejection and silently drops the toast

Both call sites invoke the helper as `void createSharedSessionAndCopyLink(state, t)`, and the helper `await`s `navigator.clipboard.writeText(...)` with no `try/catch`.

- If the write rejects (Safari activation, an insecure context where `navigator.clipboard` is `undefined`, or a denied permission), the rejection is **unhandled**.
- Because the toast at `Tabs.tsx:1255` runs *after* the awaited write, a failed copy means the session was created but **the user sees no feedback at all** — no "copied" toast, no error.

Note this is a behavior change from the old sidebar implementation, which called `navigator.clipboard.writeText(url.href)` **without `await`** (fire-and-forget), so a failing write there did not block the toast.

**Suggested fix:** wrap the copy in `try/catch`; on failure, still confirm the session was created and give the user the link another way (e.g. a toast containing the URL, or a manual "Copy link" affordance) rather than failing silently.

---

### 🔵 Low / Nits

- **Decouple "session created" feedback from "link copied".** Even with error handling, consider always confirming the session was created (that's the primary action) and treating the copy as best-effort, so a clipboard failure never leaves the user unsure whether anything happened.
- **`t: ReturnType<typeof useI18n>["t"]`** (`Tabs.tsx:1250`) is correct and avoids `any`, but reads awkwardly. If `I18nManager` exports a named `TFunction`/`TranslateFn` type, using it would be a touch cleaner. Purely cosmetic.

---

### ✅ Positives

- Removes real duplication and a fragile prop-drilling path; single source of truth for the create-and-copy flow.
- Correctly identifies and fixes the actual gap (mobile button never ran the copy logic).
- Reuses the existing translation key; no new keys added, consistent with the project's en.json-only convention.

### Recommendation

The refactor is sound and fixes the missing-copy path. **Before merging, validate on a real iOS Safari device** (Medium–High item) and add clipboard error handling (Medium). If the on-device test shows the copy failing, adopt the `ClipboardItem`-with-promise approach above.

*Note: CI status on the head commit is currently `pending` and the PR shows `mergeable_state: blocked` (awaiting required checks/review).*

---

Kabi4:
Working Safari too

image

Blitzy:
> [!NOTE]

## Code Review — PR #1497 (Fixes #1481), commit `c6cdad3`

This is a follow-up review after the two issues raised in the [earlier review](https://github.com/HelloAOLab/seed-bible/pull/1497#issuecomment-5044818430) on this same PR. Both have been addressed in the `4e19db3` ("fixes") and `c6cdad3` ("Handled error") commits:

- **Safari clipboard-timing bug** — now fixed using the `ClipboardItem`-with-promise pattern, which preserves the tap's transient user activation across the async session-creation call. Kabi4 confirmed with a screenshot that it now works on Safari too.
- **Missing error handling** — the clipboard write is now wrapped in `try/catch`, with a fallback that toasts the raw link so it isn't lost if the clipboard write itself fails.

`pnpm check:ts` is clean (0 visible errors) and the removed `createSharedSession` prop has no lingering references.

One new issue surfaced while re-reviewing the updated code, plus a small nit.

---

### 🟡 Medium — If session creation itself fails, the user gets no feedback at all

In `Tabs.tsx:1248-1283`:

```ts
async function createSharedSessionAndCopyLink(state, t) {
  const linkText = state.app
    .createSharedSession()
    .then((session) => getSessionUrl(session).href);

  try {
    // ...clipboard write using linkText...
    state.app.toast(/* copied */);
  } catch {
    try {
      state.app.toast(await linkText);
    } catch {
      // Session creation failed; nothing left to surface.
    }
  }
}
```

The comment on the inner `catch` says it plainly: if `state.app.createSharedSession()` itself rejects (network hiccup, the underlying `os.getSharedDocument` call in `SessionsManager` throwing, etc.), `linkText` rejects, the outer clipboard-write `catch` fires, the fallback `await linkText` re-throws the same rejection, and the inner `catch` swallows it silently. The user sees **no toast, no error, nothing** — just tapped a button that appeared to do nothing.

This is worse than it sounds because `handleCreateSharedSession` (`SeedBibleStateManager.tsx:1225`) calls `closeSidebarAndSettings()` as its very first line, before the `await`. So on failure the sidebar/settings panel has already closed, compounding the "did this button even work?" confusion — there's no session, no link, and no panel to go back and retry from.

**Reproduction (needs a forced failure, e.g. throttle/block the network request `os.getSharedDocument` makes, or temporarily make `sessions.createSession()` throw):**
1. Simulate a failure in shared-document setup (e.g. via devtools network throttling/blocking, or a temporary code change that makes `createSession()` reject).
2. Tap "New shared session" (desktop or mobile).
3. Observe: the sidebar closes, and no toast or error ever appears.

**Suggested fix:** distinguish "session creation failed" from "clipboard write failed" so the failure path still tells the user something happened:

```ts
async function createSharedSessionAndCopyLink(state, t) {
  const linkText = state.app
    .createSharedSession()
    .then((session) => getSessionUrl(session).href);
  let sessionCreationFailed = false;
  linkText.catch(() => {
    sessionCreationFailed = true;
  });

  try {
    // ...same clipboard-write attempt using linkText...
    state.app.toast(/* copied */);
  } catch {
    if (sessionCreationFailed) {
      state.app.toast(
        t("failed-to-create-shared-session", {
          defaultValue: "Couldn't create the shared session. Please try again.",
        })
      );
      return;
    }
    try {
      state.app.toast(await linkText);
    } catch {
      /* unreachable now */
    }
  }
}
```

(New key would need to be added to `en.json` per the project's i18n convention — English only.)

---

### 🔵 Low / Nit — Clipboard-failure fallback toast shows a bare URL with no framing text

`Tabs.tsx:1278`: `state.app.toast(await linkText)` shows the raw session URL (e.g. `https://seedbible.org/...?sessionId=...`) as the entire toast message, with no explanation of what it is or why it's being shown. A user seeing an unexplained URL pop up may not realize it's meant to be manually copied. Consider framing it, e.g. `t("could-not-copy-link", { url, defaultValue: "Couldn't copy automatically — here's your link: {{url}}" })`.

---

### ✅ Positives

- Both prior findings (Safari activation timing, silent clipboard-failure) are properly fixed, and Kabi4 verified on a real Safari device per the issue's requirement.
- The single shared helper now covers both the desktop and mobile entry points, closing the original gap in issue #1481.
- No stray references to the removed `createSharedSession` prop; type check is clean.

### Recommendation

The core fix for #1481 is solid and verified on-device. The one open gap is the silent-failure path when session creation itself errors out — worth a small follow-up before merge so a failed attempt doesn't look identical to a successful one from the user's perspective.

---

Blitzy (approved):

Issue #1403, opened 2026-07-12
Book selector should automatically focus the book that is currently open.


Blitzy:
The book selector, when opened, should automatically highlight and focus the book that is currently open.  Both desktop and mobile need this feature. When book verses are visible the one you are currently on should be obvious.

PR #1498, opened 2026-07-22
Feature/1403

Kabi4:


Blitzy:
> [!NOTE]

## Code review — PR #1498 "Feature/1403"

**What this addresses:** [#1403](https://github.com/HelloAOLab/seed-bible/issues/1403) — when the book selector opens it should auto-highlight and focus the book that is currently open (desktop + mobile), and the current chapter should be obvious.

**Overall:** The approach is sound and reads cleanly. `expandedBookId` becomes the single source of truth for "which book is open", `applyExpandedBookToSidebar` maps it back onto the accordion signals, and a dedicated effect marks the reading-position chapter. Tests are meaningful and all 41 pass locally (`pnpm vitest run BibleSelector.test.tsx BibleSelectorManager.test.ts`). Findings below are edge cases, not blockers.

---

### 🟠 Medium

**1. Auto-focus steals focus from the search box while the user is typing**

`components/BibleSelector/BibleSelector.tsx:754` — the scroll effect now calls `bookTab.focus({ preventScroll: true })`, and its deps are `[openBookId, activeChapter, isOpen.value]`. `openBookId` changes not only on open but also whenever a search narrows the list to a single book (search → `expandedBookId` → `bookData` → `openBookId`). ~50 ms later the effect fires and moves DOM focus off the `<input>` and onto the book row, interrupting typing.

*Repro:*
1. Open the book selector.
2. Click the "Search books…" field and start typing a book name (e.g. `gen`).
3. As soon as exactly one book remains, focus jumps out of the input to the book tile; further keystrokes don't reach the field until you click it again.

The old code only scrolled — it never called `.focus()` — so this is a new regression relative to `develop`.

*Suggested fix:* only move focus when the effect is running because the selector just opened, not because a search re-expanded a book. E.g. bail out if the search field currently holds focus:

```ts
// inside the timeout, before bookTab.focus(...)
const active = document.activeElement as HTMLElement | null;
if (active && active.closest(".searchbar")) return; // don't yank focus mid-search
```

or gate the `.focus()` call on a "first open" flag (reset in `setOpen(false)`), keeping the scroll-into-view behaviour on every change but the focus move only on open.

---

### 🟡 Low

**2. Apocrypha book is not auto-expanded on desktop in the default "All Books" view (and can collide with the NT grid on mobile)**

`managers/BibleSelectorManager.tsx:708-713` — in the fallback branch (`localSelectedTestament === 2`, the default) apocrypha is matched with `cht = 1`, the same testament hint the New Testament grid uses (`BibleSelector.tsx:608`, `:634`). Two consequences when the current reading book is an apocrypha book:

- **Desktop:** in "All Books" view the apocrypha grid is only rendered when `ws <= MOBILE_BREAKPOINT` (`BibleSelector.tsx:610`). On desktop there is no apocrypha grid to expand, so the feature silently no-ops for apocrypha books — the issue's "the one you are currently on should be obvious" isn't met there.
- **Mobile:** the chapter panel's visibility is gated by `chapterPos === index && (chapterHint === undefined || cht === chapterHint)` (`BibleSelector.tsx:511-513`). Since the apocrypha grid and the NT grid share `chapterHint === 1`, an expanded apocrypha book can also satisfy the NT grid's condition, so if `lastBookClicked` lines up with an NT book's chapter-insert index the chapter grid can render under the wrong NT book.

This only affects translations that include apocrypha, and only in the "All Books" filter, hence Low. Worth a follow-up: give the apocrypha section its own distinct hint (and render it in desktop "All Books" if that book should be expandable there), or short-circuit apocrypha through the single-testament path.

**3. Chapter-highlight effect runs even while the selector is closed**

`managers/BibleSelectorManager.tsx:817-830` — this effect writes `highLightedButtonsID` on every reading-position change with no `isOpen` guard, unlike the two effects just above it (`:797`, `:808`). It's cheap and harmless, but for consistency (and to avoid recomputing when nothing is visible) consider an early `if (!isOpen.value) return;`.

---

### Nits (optional)
- `aria-expanded={isSelected}` is applied to a plain clickable `<div>` book row (`BibleSelector.tsx:471`). `aria-expanded` is meaningful on an element with an interactive role; adding `role="button"` (or keeping the value only when the row is an expandable control) would make the ARIA semantics correct. The `aria-current` addition is a nice touch.
- `renderChapterButton` calls the `useLongPress` hook inside a `computed()`/loop (`BibleSelector.tsx:794`, `:826`) — this predates the PR, but the diff touches this function, so flagging it: hooks called conditionally/inside loops violate the rules of hooks and can bite later.

Nice work overall — the finding above worth resolving before merge is **#1 (focus steal on search)**; the rest are safe as follow-ups.

---

Blitzy:
> [!NOTE]

## Follow-up code review — PR #1498 "Feature/1403"

This re-reviews the PR after commit `b31dc158` ("Fixed Typing issue of auto focus + Apocrypha now uses 2 and NT stays 1"), which was pushed in direct response to the review posted earlier today.

### Previously flagged issues — verified fixed

- **Medium No. 1 (focus steal from search box):** `BibleSelector.tsx:794-799` now checks `document.activeElement?.closest(".searchbar")` before calling `bookTab.focus()`. Confirmed the search `<input>` does live inside a `.searchbar` container (`BibleSelector.tsx:233`), so the guard actually matches during typing. Fixed correctly.
- **Low No. 2 (apocrypha/NT chapter-hint collision):** `BibleSelectorManager.tsx:708-712` now assigns apocrypha hint `2` instead of `1` in the "All Books" match logic, and `BibleSelector.tsx:674` renders the mobile apocrypha grid with the same hint `2`. Desktop gained a dedicated short-circuit (`BibleSelector.tsx:582-621`) that renders an apocrypha-only grid when the expanded book is apocrypha, since desktop "All Books" has no apocrypha column to expand otherwise. Traced the index math (`ghostArray` only appends ghost entries after real books, so indices from `findIn(apocrypha, 2)` stay valid against the ghosted render array) — no off-by-one issues found. Fixed correctly.
- **Low No. 3 (highlight effect running while selector closed):** `BibleSelectorManager.tsx:817-820` now has the `if (!isOpen.value) return;` guard, consistent with the neighboring effects. Fixed correctly.

### 🟡 Low — new from this commit

**Test coverage gap for the new desktop-apocrypha branch and the `cht=2` hint**

Neither `BibleSelector.test.tsx` nor `BibleSelectorManager.test.ts` has a case that opens the selector with an apocrypha book as the current reading position on a desktop-width viewport (which is the only path that exercises the new `expandedIsApocrypha` short-circuit at `BibleSelector.tsx:588`), or asserts `chT.value === 2` when an apocrypha book is auto-expanded from "All Books". The existing 41 tests all pass and `pnpm check:ts` is clean, but this specific branch — which is exactly what fixed the collision bug — has no regression coverage, so it could silently regress later.

*Suggested fix:* add a test (mirroring the existing `"auto-expands the current book and highlights the current chapter on open"` case) that sets the reading position to an apocrypha book (e.g. `TOB` or whatever the fixture's translation includes), opens the selector at a desktop viewport width, and asserts both the apocrypha-only grid renders and `chT.value === 2`.

### Still outstanding (optional, not blockers)

- `aria-expanded={isSelected}` on the plain `<div class="sidebar-itm">` book row (`BibleSelector.tsx:471`) is still not paired with an interactive role — same nit as before, cosmetic only.
- `useLongPress` still called inside the `renderChapterButton` closure per-iteration (`BibleSelector.tsx`) — pre-existing pattern, unchanged by this PR.

### Verification

- `pnpm vitest run BibleSelector.test.tsx BibleSelectorManager.test.ts` — 41/41 pass at `b31dc158`.
- `pnpm check:ts` — 0 visible errors.

Nothing here blocks merge — the fix commit correctly resolves everything flagged in the prior review. The only new note is the missing test coverage for the branch that fixed the apocrypha collision.

---

Blitzy:
@Kabi4 on mobile it doesn't seem to be moving the book and chapter into view. It is selecting the book and chapter correctly however.

https://github.com/user-attachments/assets/1dea8d5e-91e5-4a81-a70f-c981767b82be

Kabi4:
image

Blitzy:
> [!NOTE]

## Follow-up code review — PR #1498 "Feature/1403"

This re-reviews the PR after commit `788239f` ("Fix"), pushed in response to the mobile repro you (Blitzy) posted: "on mobile it doesn't seem to be moving the book and chapter into view. It is selecting the book and chapter correctly however."

### What changed and why it fixes the reported bug

`components/BibleSelector/BibleSelector.tsx:819-854` replaces the old hardcoded two-level scroll (`.books-item`, then `.books-container`) with a loop that walks every ancestor between the scroll target and `.sb-selector-panel`, and scrolls whichever ones actually have `overflow-y: auto|scroll|overlay` and overflowing content.

This explains the original mobile bug: the real scrollable container on mobile is `.sidebar-results` (`overflow-y: auto` per `BibleSelector.css:729`), sitting *above* `.books-container` in the DOM — but the old code never touched `.sidebar-results`, only `.books-item`/`.books-container`. So on mobile, the book/chapter could be selected correctly but never scrolled into view, exactly as reported. The new ancestor-walk picks up `.sidebar-results` (and `.books-container`, which is also `overflow: auto` on mobile per the `max-width: 480px` media query) automatically, so this should be fixed for both breakpoints without special-casing either one.

Verified:
- `pnpm vitest run BibleSelector.test.tsx BibleSelectorManager.test.ts` — 41/41 pass at `788239f`.
- `pnpm check:ts` — 0 visible errors.
- Traced the DOM/CSS chain (`.books-item` → `.testament-container` → `.books-container` → `.sidebar-results` → `.sb-selector-panel`) against the loop's break condition — no gaps found; each scrollable ancestor gets a fresh `getBoundingClientRect()` per scroll call, so nested scrolls don't use stale rects.

I can't drive an actual mobile viewport from here, so this is verified by code/CSS tracing and the existing suite, not a manual device check — worth a quick manual confirm on your end before merge.

### 🟡 Low

**1. No regression test for the ancestor-walk fix itself**

This commit only touches `BibleSelector.tsx` — no test was added that would catch a regression in the new `scrollTargetInto` / ancestor-walk logic (e.g. asserting `scrollTo` gets called on `.sidebar-results` when a deeply-nested chapter button is out of view). This is the exact code path that fixed the bug you reported, so it has no safety net if it regresses later. This is the same "fix landed, coverage didn't" pattern as the apocrypha-branch gap flagged in the previous review round — still open too.

**2. Stale comment now contradicts the CSS it's describing**

`BibleSelector.tsx:820-821`:
```ts
// Desktop usually only needs `.books-item`. On mobile `.books-container`
// is `height: auto`, so the real scroller is `.sidebar-results`.
```
Per `BibleSelector.css:1015-1018` (inside the `max-width: 480px` query), `.books-container` actually gets an explicit `height: calc(100dvh - 5rem)` **and** `overflow: auto` on mobile — it's not `height: auto`, and it *is* one of the scrollable ancestors the loop touches (alongside `.sidebar-results`). Worth correcting so the next person reading this doesn't get the wrong mental model of why the ancestor-walk is needed.

### Still outstanding from prior rounds (optional, not blockers)
- Apocrypha desktop auto-expand branch (`cht=2`) still has no dedicated test.
- `aria-expanded={isSelected}` on the plain `<div class="sidebar-itm">` book row is still not paired with an interactive role.
- `useLongPress` still called inside the `renderChapterButton` closure per-iteration — pre-existing pattern, unchanged by this PR.

Nothing here blocks merge — the fix looks correct and well-targeted at the reported bug. The only actionable note is adding test coverage for the ancestor-walk (No. 1) before it's easy to forget.

---

Blitzy:
> [!NOTE]

## Follow-up code review — PR #1498 "Feature/1403"

This re-reviews the PR after the four commits pushed since the last review round (at `788239f`): `d41aa67` ("fix: stop Psalms manual toggle from re-triggering scroll/focus to current chapter"), `17ce6d4` ("refactor: simplify chapter-click toggle and drop unused testament hints"), `8ba84d4` ("fix: stop opening a large book from scrolling its title off-screen"), and `a61de67` ("docs: correct stale comment about mobile scroll ancestors"). Issue [#1403](https://github.com/HelloAOLab/seed-bible/issues/1403) is still the tracking issue — auto-focus/highlight the currently-open book and chapter.

### What each commit does, and whether it holds up

- **`d41aa67` (Psalms manual-toggle bug):** The Psalm-group-auto-expand logic used to live in its own `useEffect` keyed off `[openBookId, activeChapter]`, while the scroll/focus effect re-ran on `[..., currentPsalms.value]` specifically so it would fire again *after* the auto-expand added a section. The bug: `currentPsalms.value` also changes when the user manually opens/closes a Psalms section by hand, so that same dependency caused the scroll/focus effect to fire on every manual toggle too, snapping the view back to the current chapter mid-browse. The fix folds the auto-expand into the same effect and drops `currentPsalms.value` from the dependency array entirely. Since the auto-expand write happens synchronously before the `setTimeout(..., 50)`, the resulting re-render lands well before the timeout fires, so the effect still doesn't need to be re-triggered by the signal write. This is correct and doesn't need the removed dependency.
- **`8ba84d4` (large-book title scroll):** Adds a `targetFits = targetRect.height <= scrollerRect.height` guard so the bottom-edge chase only runs when the target actually fits the scroller. This matters when `visibleChapterButton` is `null` (e.g. opening a book that isn't the current reading position) and the scroll target falls back to the whole chapter panel — for a long book like Psalms that panel is taller than the scroller, so the old unconditional bottom-chase would scroll deep into the grid and push the book's title off-screen. Now that case just leaves the top aligned instead. Reasoning checks out.
- **`17ce6d4` (drop `index`/`cht` from `handleChapterClick`):** `handleChapterClick` now just calls `setExpandedBook(book.id)` instead of manually toggling `bookData`/`chT`/`lastBookClicked`. I checked that `setExpandedBook` (`BibleSelectorManager.tsx:399`) still does the same-book-collapses toggle (`expandedBookId.value = expandedBookId.value === nextBookId ? null : nextBookId`), so clicking an already-expanded book still collapses it — behavior is preserved, this is a clean simplification now that `applyExpandedBookToSidebar` is the single place deriving those signals.
- **`a61de67`:** Just the comment fix; matches the CSS as described.

### Verification

- Checked out `a61de67` locally and ran `pnpm vitest run BibleSelector.test.tsx BibleSelectorManager.test.ts` → **41/41 pass**.
- No test file changed in these four commits, so the counts are unchanged from the last round.

### 🟡 Low — still no regression coverage for what these commits fixed

Same "fix landed, coverage didn't" pattern flagged the last two rounds. None of the four commits added a test, so there's nothing pinning down:
- The Psalms manual-toggle bug itself (e.g. open the selector on a Psalms reading position, manually toggle a different Psalms section open, assert the view doesn't snap back / `currentPsalms` isn't reset).
- The `targetFits` guard (a book whose chapter panel is taller than the scroller shouldn't scroll past its title).

Not a blocker, but worth adding before this logic is touched again — it's already had to be re-fixed twice in this PR's lifetime.

### Still outstanding from prior rounds (optional, not blockers)

- No dedicated test for the desktop-apocrypha auto-expand branch (`cht=2`) — confirmed still absent (`grep -i apocrypha` over both test files returns nothing).
- `aria-expanded={isSelected}` on the plain `<div class="sidebar-itm">` book row still isn't paired with an interactive role — cosmetic only.
- `useLongPress` still called inside the `renderChapterButton` closure per-iteration — pre-existing pattern, unchanged by this PR.

Nothing here blocks merge. The four new commits are correct, reasoned fixes for real bugs (verified by tracing the logic and re-running the full test suite), and the only actionable note is adding regression tests for the two bugs they just fixed.

---

Issue #1394, opened 2026-07-10
Add a unified "Share" modal (link, share sheet, shared session) to mobile More menu, verse toolbar, and tabs options menu


Blitzy:
## What is happening

Right now there's no single place to share from. The verse toolbar's "Share" button, the mobile "More" menu, and the tabs options menu's "Share session" entry don't offer a unified way to share a link, share via the system share sheet, or start a shared session.

## What is expected

A new Seed Bible "Share" modal that asks the user what they want to share, with these options for the first pass:

- **Share a link** - copies a link to the clipboard.
- **Share via...** - opens the operating system's native share sheet.
- **Start and share session** / **Share current session**

If the user is already in a session, the modal offers to share the **current session** instead of starting a new one.

Three entry points open this modal:

- A new "Share" button added to the mobile "More" menu.
- The existing verse "Share" button, which should open this modal instead of its current behavior. Its "Share a link" option carries the selected verses, using the same URL query parameter tracking already in place today.
- The tabs options menu's existing "Share session" entry, which should open this modal instead of its current behavior.
- The mobile participants drawer also should get a share button.

PR #1499, opened 2026-07-22
made the unified share modal

Bored-Wizard:
issue #1394


Blitzy:
> [!NOTE]

## Code review — PR #1499 "made the unified share modal" (issue #1394)

### Summary

This adds a reusable `ShareModal` component (link / share-via / start-or-share-session), wires it into the **verse toolbar's Share button**, and improves `getShareUrl` to include the `chapter` query parameter. Tests pass locally (`pnpm vitest run BibleToolsManager.test.tsx` → 19/19), and the modal renders as a proper Preact component through `ModalHost` (`modal.content({ t })` returns a `<ShareModal/>` VNode), so its hooks are used correctly.

The `getShareUrl` chapter change is a genuine improvement: the reader already reads a `chapter` param (`SeedBibleStateManager.tsx:606`, `TabsManager.tsx:105`), but shared verse links never set it, so a shared link could land on the wrong chapter. `chapterNumber` is typed `Signal<number>` (never null), so `String(chapter)` is safe.

The findings below are ordered by severity. Nothing here is a crash-level blocker — the largest item is feature completeness relative to the issue.

---

### High — Only 1 of the 4 entry points from issue #1394 is implemented

Issue #1394 asks the modal to open from **four** places:
1. the mobile **"More" menu** (new Share button),
2. the **verse toolbar** Share button — ✅ done here,
3. the **tabs options menu** "Share session" entry,
4. the **mobile participants drawer** (new share button).

Only #2 is wired. The component was clearly built to be reusable (good), but the other three entry points still use their old behavior. If this PR is intended as an incremental first pass, that's fine — please say so in the description and track the remaining entry points in a follow-up so #1394 isn't closed prematurely. If it's meant to close the issue, the other three need wiring.

---

### Medium — User-facing toast strings are not internationalized

`BibleToolsManager.tsx:841` (the new `onShareLink` handler) toasts a hardcoded `"Copied!"`:

```tsx
onShareLink={() => {
  navigator.clipboard.writeText(shareUrl.toString());
  context.toast("Copied!");   // not translated
  modals.closeModal(modalId);
}}
```

The app ships 24 locales and everything else in this modal goes through `t(...)`. Non-English users will see an English toast. (Note: the copy-verse tool at `:808` already had this pre-existing `"Copied!"` — worth fixing both while here.)

**Suggested fix:** add a `share-link-copied` key to `en.json` and use `t("share-link-copied", { defaultValue: "Copied!" })`.

---

### Medium — "Share via…" silently does nothing on browsers without the Web Share API

`BibleToolsManager.tsx:844-851`:

```tsx
onShareVia={() => {
  void navigator.share?.({ ... });   // optional-chained
  modals.closeModal(modalId);
}}
```

The `?.` avoids the old unconditional-`navigator.share` throw, but now on any desktop browser without `navigator.share` (Firefox, most desktop Chrome), clicking "Share via…" just closes the modal with no feedback. From the user's perspective the button is broken.

**Repro:** open in desktop Firefox → select a verse → Share → "Share via…" → modal closes, nothing happens, nothing copied.

**Suggested fix:** guard the action — either hide the "via" option when `!navigator.share` (e.g. `typeof navigator.share === "function"`), or fall back to copying the link + a toast when it's unavailable.

---

### Low — `sessionActive` flashes the wrong label on open

`shareModal.tsx:21-39`: `sessionActive` starts `false` and is only ever set to `true` inside an effect after mount. When you open the sheet while already in a shared session, the "session" row renders as **"Start and share session"** for one frame, then flips to "Share current session". It also never resets to `false`.

**Suggested fix:** initialize from current state instead of an effect, e.g.
```tsx
const sessionActive = useComputed(
  () => !!props.app.currentReadingState.value?.tab.sharedSession?.id
);
```
This is reactive, has no initial flash, and drops the effect entirely.

---

### Low — No error handling on clipboard / session-creation in the modal

In `shareModal.tsx` the session actions call `navigator.clipboard.writeText(...)` (unawaited) and immediately toast success; the "start session" branch `await`s `createSharedSession()` with no `try/catch`. If the clipboard write is rejected (permissions / insecure context) or `createSharedSession()` throws, the user still sees the "link copied" toast (or gets an unhandled rejection and a stuck-open modal). The copy-verse tool at `BibleToolsManager.tsx:806-811` already wraps its copy in `try/catch` — mirroring that here would be consistent.

---

### Low — Dead / unused code left behind

Confirmed unused (0 references outside their own declaration):

| Item | Location |
|---|---|
| i18n key `share-link-chapter-subtitle` | `en.json` |
| i18n key `preview-toggle-session` | `en.json` |
| i18n key `preview-toggle-verse` | `en.json` |
| CSS `.sb-share-preview` / `.sb-share-preview-button` | `shareModal.css:96-108` |
| exported type `ShareScope` | `shareModal.tsx:7` |
| prop `hideShareLink` | `shareModal.tsx` (declared + filtered internally, never passed by any caller) |
| `active: false` (hardcoded, so `.sb-share-action--active` never applies) | `shareModal.tsx:62,87` |

These look like leftovers from a fuller design (verse/chapter scope toggle + preview). Removing them keeps the surface clean; if the preview/scope UI is coming in a follow-up, a `// TODO` referencing the issue would make the intent clear.

---

### Low — `getShareUrl` dropped the translation match on selected verses

The verse filter changed from `v.translationId === translation` to `v.chapterNumber === chapter`:

```tsx
.filter((v) => v.bookId === bookId && v.chapterNumber === chapter)
```

Adding the chapter check is correct, but dropping the translation check means that in a multi-translation/parallel selection, verses selected in a *different* translation (same book + chapter) will now be included in the `verse=` param even though the URL's `translation=` is the current one. Likely a rare edge case given how selection works today, but worth a deliberate `&&` on translation if parallel selection is possible.

---

### Nits

- `shareModal.tsx:39` — the `useEffect` dependency reads `props.app.currentReadingState.value?...` directly; switching to the `useComputed` above removes this fragile dependency entirely.
- `getSessionUrl` (`shareModal.tsx:148`) duplicates URL-building logic that also lives in `getShareUrl`/`verseReferenceLink.tsx`; not worth blocking, but a shared helper would reduce drift.

Overall this is a clean, well-structured foundation for the shared modal. The main asks before merge: confirm the entry-point scope (High), internationalize the toast (Medium), and handle the no-Web-Share-API case (Medium).

---

Blitzy:
@Bored-Wizard Some issues that needs to be addressed:

## Per the agent code review

The **High** and **Medium** issues mentioned by the agent code review should be addressed.

1. **High** — This share modal needs to be triggered by the remaining 3 entry points listed in the original issue.
2. **Medium** — As the agent code review above mentioned, the toast should use translation. There is actually already an `copied` key that you can reuse here.
3. **Medium** — We should guard against browsers/devices that don't implement the Web Share API.

## Remove the cancel button

Also lets remove the cancel button at the bottom of the modal. We already have a close button in the upper right corner, and the modal can also be dismissed by clicking/tapping off of it.

Blitzy:
> [!NOTE]

## Code review — PR #1499 "made the unified share modal" (issue #1394), re-review after `fixed things according to the last response`

This re-reviews the new commit (`fixed things`) against the feedback from the earlier review and from the maintainer's follow-up comment. Of the four entry points issue #1394 asks for, three are now wired up (verse toolbar, tabs options menu "Share session", mobile participants drawer). The two "Medium" items (i18n toast, Web Share API guard) and the "cancel button" request are addressed as well. `pnpm check:ts` is clean and `BibleToolsManager.test.tsx` passes (20/20) on this branch.

However, this commit also introduces two new functional regressions that weren't there before, plus it re-adds the exact duplicate-translation-key issue the maintainer explicitly asked to avoid. Listed by severity below.

---

### High — Sharing a background tab's session can share/start the *wrong* session

`Tabs.tsx:641-665`, `openShareSessionModal`:

```tsx
export function openShareSessionModal(
  state: SeedBibleState,
  session: BibleReadingSession
) {
  const shareUrl = getSessionUrl(session);
  ...
  <ShareModal
    app={state.app}
    hideShareLink
    ...
  />
```

`session` (the tab whose "Share session" button was clicked) is only used to build `shareUrl` for the `onShareVia` share-sheet action. It is never passed into `<ShareModal>` — there's no `session` prop on `ShareModalProps` at all. Inside `shareModal.tsx:22-24` and `77-93`, the modal instead derives "is there already a session?" and which session to copy/start from `props.app.currentReadingState.value?.tab.sharedSession` — i.e. the **currently selected/active tab**, not the tab the user actually clicked "Share session" on.

`TabRow` (the row that renders this button) is rendered once per tab with an independent `isSelected` flag (`Tabs.tsx:944-951`), so a background/non-selected tab can absolutely have its own `sharedSession` while the active tab has a different one (or none).

**Repro:**
1. Open two tabs. Start a shared session on tab A, leave tab B without one.
2. Select tab B (make it active), then open tab A's tab-options menu (tab A is not selected) and click "Share session".
3. Expected: share/copy the link to tab A's existing session.
4. Actual: the modal reads `currentReadingState` (tab B, the active one), sees no session, and offers **"Start and share session"** — clicking it creates a brand-new session and shares that, instead of tab A's. If tab B already had its own different session, the modal would instead silently copy/share *that* link under "Share current session", not tab A's.

The same function backs the mobile participants drawer's new share button (`SessionParticipants.tsx`), but that one appears safe in practice — `MobileSessionParticipants` there is only ever given the *active* tab's session (`BibleReader.tsx:1562-1566`), so `currentReadingState` and the passed-in `session` happen to agree. The bug is specific to the Tabs entry point, where rows for non-active tabs exist.

**Suggested fix:** thread the intended session through explicitly instead of re-deriving it from global state — e.g. add a `session: BibleReadingSession` (or `sessionActive: boolean` + a `sessionUrl`) prop to `ShareModalProps` and have `openShareSessionModal` pass the `session` argument it already has, so the modal's session action always operates on the tab that was actually clicked.

---

### High — Mobile verse-toolbar overflow ("More") is now unreachable; extra actions are dropped, not just hidden

`BibleReaderToolbar.tsx:1843-1898`. This commit bumped `COLLAPSED_COUNT` from 3 to 4 (presumably so the new Share entry doesn't need the toggle in the common case), but in the process the "More/Less" toggle button *and* the overflow rendering were commented out entirely rather than adjusted:

```tsx
const COLLAPSED_COUNT = 4;
const needsToggle = actionCards.length > COLLAPSED_COUNT;
const primaryCards = needsToggle
  ? actionCards.slice(0, COLLAPSED_COUNT)
  : actionCards;
// const overflowCards = needsToggle
//   ? actionCards.slice(COLLAPSED_COUNT)
//   : [];

return (
  <>
    {primaryCards}
    {/* {needsToggle && ( ... More/Less button ... )} */}
    {/* {isVerseSheetExpanded.value && overflowCards} */}
  </>
);
```

With the default 4 actions (Highlight, Bookmark, Copy, Share) this is invisible, which is likely why it slipped through — but the moment a 5th action becomes visible, it's just gone, with no button to reveal it:
- **"Add to Playlist"** becomes visible while editing a playlist (`BibleToolsManager.tsx:771`), pushing the count to 5.
- Any extension registering an additional verse-toolbar tool via `applyToolbarCustomization` does the same.

**Repro:** On a small/mobile screen, start editing a playlist, then select a verse. Expected: 5 action cards, with a "More" toggle to reach the 5th. Actual: only the first 4 render; the 5th (and the toggle to get to it) is simply missing — no error, no visual cue anything is hidden.

**Suggested fix:** this looks like leftover debugging from tuning `COLLAPSED_COUNT` — restore the commented-out toggle button and `{isVerseSheetExpanded.value && overflowCards}` block (or delete the dead `overflowCards`/toggle code deliberately if the intent is genuinely to remove the "More" affordance, but then the 5th+ action needs a different way to be reached, e.g. always show `add-to-playlist` in the primary row).

---

### Medium — Maintainer explicitly asked to reuse the `copied` key; a duplicate key was added instead

The maintainer's follow-up comment says: *"the toast should use translation. There is actually already a `copied` key that you can reuse here."* `en.json` already has (`en.json:54`):

```json
"copied": "Copied",
```

This commit instead adds a new, separately-maintained duplicate (`en.json:417`):

```json
"share-link-copied": "Copied!",
```

and wires both toasts (`BibleToolsManager.tsx:808-810` and `843-845`) to `i18n.t("share-link-copied", ...)`. Functionally this does fix the original "not translated" bug, but it re-introduces exactly the duplication the maintainer asked to avoid — now there are two "Copied" strings for translators to keep in sync (and the 23 non-English locales won't have `share-link-copied` at all until someone backfills it, so non-English users will see the English fallback "Copied!" until then, same symptom as before just under a different key).

**Suggested fix:** use `i18n.t("copied", { defaultValue: "Copied" })` at both call sites and drop the new `share-link-copied` key from `en.json`.

---

### Low — Entry point 1 from issue #1394 (mobile "More" menu) is still not wired

Issue #1394 lists four entry points; this PR (across both commits) now wires three of them — verse toolbar, tabs options menu, and the mobile participants drawer. The fourth, a new **Share** button in the mobile "More" menu (`MobileMoreMenu` / `moreTools` in `BibleReaderToolbar.tsx:126,480-485`), still has no share-related tool in `moreTools`. Not a regression from this commit, just flagging it's still outstanding — worth a quick confirmation on whether it's coming in this PR or a follow-up before this closes issue #1394.

---

### Low — Dead CSS left behind after removing the cancel button

Per the maintainer's request, the modal's "Cancel" button was removed from `shareModal.tsx`'s render — but `.sb-share-cancel` and `.sb-share-cancel:hover` (`shareModal.css:65-81`) are still there with no remaining consumer. Safe to delete.

---

### Nits

- `Tabs.tsx:645` / `SessionParticipants.tsx` both build a session URL via `getSessionUrl(session)`/local helpers that duplicate the same URL-building logic already flagged as a nit in the previous review (`shareModal.tsx`'s own `getSessionUrl`). Not blocking, but three near-identical implementations now exist (`Tabs.tsx`, `shareModal.tsx`, and the original `SessionParticipants`/`Tabs` helper) — worth consolidating into one shared helper while this area is already being touched.
- No test coverage exists yet for `openShareSessionModal` or the Tabs "Share session" flow (no `Tabs.test.tsx`), which is how the active-vs-target-tab mismatch above could ship unnoticed — worth a regression test once the fix lands (assert the modal's session action targets the tab that was passed in, not whatever tab happens to be selected).

---

### Summary

Good progress on the three previously-missing entry points and the two Medium items from the last round. The two High findings here are both real, reproducible bugs introduced by this commit (not pre-existing): the Tabs "Share session" button can act on the wrong tab's session, and the mobile verse-toolbar's overflow toggle is currently commented out, so a 5th toolbar action becomes invisible with no way to reach it. Both look like unintended fallout from otherwise-correct fixes (COLLAPSED_COUNT tuning, and session-modal reuse) rather than deliberate changes, and should be straightforward to correct before merge.

---

Bored-Wizard:
@Blitzy I was about to push the fixes lol

Blitzy:
> [!NOTE]

## Code review — PR #1499 "made the unified share modal" (issue #1394), re-review after the latest commits (up to `fb5ff94`)

### Summary

This re-reviews the commits pushed since the last round (`48fe030e` … `fb5ff94c`), which were meant to resolve the two High findings from the previous review plus the outstanding Medium/Low items, and which also add the pinned-header fix for the mobile settings sheet. I checked each previous finding against the current code and ran the suite:

- `pnpm check:ts` → **0 visible errors**
- `pnpm vitest run BibleToolsManager.test.tsx` → **20/20 passing**

**All four previously-flagged High/Medium issues are fixed correctly:**

- **Wrong-tab session sharing** — `ShareModal` now takes an explicit `session: BibleReadingSession | null` prop instead of reading `currentReadingState` from global state. `Tabs.tsx`'s `openShareSessionModal(state, session)` and `BibleToolsManager.tsx`'s `openShareModal` both thread through the session that belongs to the actual surface being acted on (the clicked tab / the tool's own `context.sharedSession`), not whatever tab happens to be selected. Confirmed by reading `Tabs.tsx:641-666` and the `TabRow` share-session `onClick` at `Tabs.tsx:1144-1153`.
- **Mobile verse-toolbar overflow toggle unreachable** — the More/Less toggle button and `{isVerseSheetExpanded.value && overflowCards}` block are back in `BibleReaderToolbar.tsx` (confirmed at lines 1848-1898); only `COLLAPSED_COUNT` changed (3 → 4), so a 5th+ toolbar action (e.g. "Add to Playlist" while editing a playlist) is reachable again.
- **Duplicate i18n key for "Copied"** — both toast call sites now use `i18n.t("copied", { defaultValue: "Copied" })`, and the previously-added `share-link-copied` key is gone from `en.json`.
- **Mobile "More" menu entry point missing** — a `share` tool (priority 130) was added to `getDefaultToolbarTools()` with `isVisible: (context) => !!context.modals && !!context.app`. It defaults to `isControllable: true` (not overridden), so it lands in `moreTools` the same way other optional tools do, surfacing it in `MobileMoreMenu`. That covers all four entry points from issue #1394.
- Also cleaned up as promised: the dead `.sb-share-cancel` CSS, the unused `ShareScope` type, the always-`false` `active` styling, and the unused preview-related i18n keys/CSS are all gone from the current `shareModal.tsx`/`shareModal.css`/`en.json`.

Nothing here blocks correctness. The remaining items are lower severity — mostly the same test-coverage gap called out last round, still not addressed.

---

### Low — No regression test protects either of the two High-severity fixes from the last round

Neither fix has a test that would catch a future regression:

- There is still no `Tabs.test.tsx` (or any test) exercising `openShareSessionModal`/`TabRow`'s share-session button, so nothing asserts the modal acts on the tab that was clicked rather than the active tab — the exact bug that shipped last round.
- `BibleReaderToolbar.test.tsx` (if it exists) doesn't appear to cover the `COLLAPSED_COUNT`/overflow-toggle behavior either — nothing asserts a 5th toolbar action stays reachable via "More".

This was flagged as a nit in the previous review before either bug was known to exist; now that both have actually regressed once, it's worth promoting to something addressed before merge rather than left as a nit. A couple of focused tests (assert `openShareSessionModal` is called with the tab's own session for a non-active tab; assert the toggle renders and reveals overflow cards past `COLLAPSED_COUNT`) would close this gap cheaply.

---

### Nit — `getSessionUrl` is now duplicated in three places

`Tabs.tsx:1260` and `shareModal.tsx:145` each define their own identical `getSessionUrl` (build a URL from `window.location.href`, preserve `pattern`, set `sessionId`). This was flagged in the previous review as a general duplication concern across `Tabs.tsx`/`shareModal.tsx`/the old `SessionParticipants` helper; it's now down to two copies instead of three, but still worth consolidating into one shared helper (e.g. exported from `Tabs.tsx` or a small shared module) while this area is already being touched, rather than three call sites needing the same edit if the URL shape ever changes.

---

### Summary

Solid follow-up round — both High findings and the Medium/Low items from the previous review are genuinely fixed, not just papered over, and I didn't find any new functional regressions in the diff. The only thing I'd want before merge is closing the test-coverage gap on the two fixes that already broke once (Low, above); the `getSessionUrl` duplication is a cheap cleanup but not blocking.

---

PR #1500, opened 2026-07-22
fix: show host in session settings participant list

### LINKED ISSUE #1395: Session settings should show the host in the participant list
## What is happening

The session settings modal's participant list shows all other participants, including any co-hosts, but not the host. The host is missing from this list.

The host does show up correctly in the main mobile session participant drawer, just not in the participant list inside the session settings modal.

## What is expected

The session settings modal's participant list should include the host, matching what's already shown in the mobile session participant drawer.

mazenemad:
fixes #1395

### DIFF
--- a/packages/seed-bible/seed-bible/components/Tabs/Tabs.tsx
+++ b/packages/seed-bible/seed-bible/components/Tabs/Tabs.tsx
@@ -195,7 +195,6 @@ function SessionSettingsModalContent(props: {
     session.updateOptions({ shareTranslation: share });
   };
 
-  const coHostUserIds = options.coHostUserIds ?? [];
   const setCoHost = (coHostKey: string, makeCoHost: boolean) => {
     if (!isHost) return;
     const existing = options.coHostUserIds ?? [];
@@ -207,12 +206,15 @@ function SessionSettingsModalContent(props: {
     session.updateOptions({ coHostUserIds: next });
   };
 
-  // Everyone connected except the immutable host, so the host can promote or
-  // demote co-hosts inline.
-  const participants = session.connectedUsers.value.filter(
-    (user) =>
-      options.hostUserId !== user.userId &&
-      options.hostUserId !== user.connectionId
+  // Everyone connected, host first, then co-hosts, then peers (Array.sort is
+  // stable so peers keep their existing order within a rank — matches the
+  // sidebar tab row and the mobile sheet). The host is listed like any other
+  // user, with a "Host" badge, but can't be promoted, so their row shows no
+  // co-host action.
+  const participants = [...session.connectedUsers.value].sort(
+    (a, b) =>
+      sessionRoleRank(getUserSessionRole(options, a)) -
+      sessionRoleRank(getUserSessionRole(options, b))
   );
 
   return (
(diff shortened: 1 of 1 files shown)


Blitzy:
> [!NOTE]

## Code Review — Show host in session settings participant list

### Summary

This is a clean, correct, and minimal fix for issue #1395. The root cause identified by @Blitzy was that the Session Settings participant list only rendered when there were *non-host* participants: the list explicitly filtered the host out, so a host sitting alone (or a host-only session) saw an empty list and therefore no "Participants" section at all.

The change replaces the host-filtering logic with the same role-based sort already used by the sidebar avatar row and the mobile participant sheet — `[...connectedUsers].sort(sessionRoleRank(getUserSessionRole(...)))` — so the host now appears first with a **Host** badge, co-hosts next, then peers. The host's row correctly omits the "Make co-host" action since the host can't be promoted/demoted.

**I found no correctness, security, or regression issues that block merge.** The notes below are low-severity / for-your-decision only.

### What I verified

- **Consistency:** The new sort/role logic exactly matches the three other render sites — the sidebar avatar row (`Tabs.tsx` ~line 1012), the mobile reader sheet (`Tabs.tsx` ~line 1937), and `SessionParticipants.tsx` (~line 49). The settings modal was the odd one out that still filtered the host; this brings it in line.
- **i18n:** Both `"host"` and `"co-host"` keys already exist in `en.json` (lines 382–383), so the new `t("host", …)` call won't trip the translation-key lint rule.
- **No signal mutation:** `[...session.connectedUsers.value].sort(...)` copies the array before sorting, so the underlying signal isn't mutated in place.
- **Dead variable removed cleanly:** The deleted `const coHostUserIds = …` was only referenced by the old `isCoHost` check; `setCoHost` keeps its own local copy, so nothing else breaks.
- **Badge styling:** `sb-session-participant-badge` is the same class the co-host badge already used in this modal pre-PR, so the Host badge inherits working styles.

### Findings by severity

#### Low — Participant section is still host-only visible (for your decision, likely out of scope)

`Tabs.tsx:397`

```tsx
{isHost && participants.length > 0 && (
```

The section remains gated behind `isHost`, so a non-host participant still sees no participant list in Session Settings. This is **pre-existing behavior, not introduced by this PR**, and it does not regress the reported bug (issue #1395 is specifically about the host being missing from the list the *host* sees, which this fixes).

However, in the linked issue @KallynGowdy asked for "some way to view the list of participants (by name) on desktop" for everyone, since non-hosts can currently only see the icons in the tab. If satisfying that request is in scope for this issue, the section would need to render for non-hosts too (with the co-host action buttons hidden, mirroring how the host row already hides its own action). If it's intentionally host-only because this is primarily the co-host management surface, then no change is needed — just flagging it so it's a conscious call rather than an oversight.

*No reproduction needed — behavioral observation from the gate condition.*

#### Nit — Co-host detection vs. toggle key use slightly different matching

`Tabs.tsx:407` and `Tabs.tsx:450`

The displayed role now comes from `getUserSessionRole(options, user)`, whose co-host check matches on **either** `user.userId` **or** `user.connectionId` (via `isSessionHost`). The toggle action, `setCoHost(coHostKey, …)`, still keys off `getConnectedUserVisualKey(user)` which resolves to `userId ?? connectionId` (userId preferred).

In every normal flow these agree, because promoting a user writes that same `userId`-preferred key into `coHostUserIds`. The only way they diverge is an unusual state where a user's `connectionId` (but not their `userId`) landed in `coHostUserIds` — then the badge would show "Co-host" while the toggle operates on the `userId` key. This is an edge case I don't think is reachable through the current UI, so I'd treat it as informational rather than actionable. If you want them provably identical, deriving the toggle's target from the same role/key source would close the gap.

### Verdict

Approve pending the maintainers' call on the Low note above. The fix does exactly what issue #1395 asks for and does it consistently with the rest of the sessions UI.

---

Blitzy (approved):

PR #1502, opened 2026-07-22
Task/improve today screen mobile layout

soukigabriel:
Today screen mobile layout was not matching the Figma design.

This commit updates the mobile layout to match the intended design.


Blitzy:
> [!NOTE]

## Code Review — Today screen mobile layout (No linked issue found)

I reviewed the diff against `develop` (12 files, +498/-425). The refactor is clean overall: the per-category card is extracted into `BookmarksCategory` + `useCategorizedBookmarks`, the section reuses the shared `TitledSection`, `isMobile` is plumbed through for responsive icon sizing, and the tests were updated to match. I ran the checks locally against this branch:

- `pnpm check:ts` → **0 visible errors**
- The three touched test files → **18/18 passing**

So nothing here blocks correctness in the common case. My findings are mostly about the multi-category layout and cleanup. Listed by severity.

---

### Medium

**No. 1 — Multiple bookmark categories render side‑by‑side (squished) instead of stacked.**

`BookmarksSection` now renders one `<BookmarksCategory>` per category, but the container CSS lays them out in a **row**, not a column:

```css
/* styles.css:797 */
.today-container .bookmarks-section {
  width: 100%;
  display: flex;      /* defaults to flex-direction: row */
  align-items: center;
}
.today-container .bookmarks-section > div {
  width: 100%;        /* each category wants the full width... */
  display: flex;
  align-items: center;
}
```

With a single category (the default `"My Bookmarks"`) there is one child, it takes 100% width, and it looks correct — which is why this passes a quick smoke test. But as soon as there are **two or more categories**, they become two flex items in one row, each asking for `width: 100%`; the default `flex-shrink: 1` then squeezes them to share the row (≈50% each) side‑by‑side, rather than stacking each category on its own line. Since the whole point of this area is grouping bookmarks into categories, the multi-category case is first-class, not an edge case.

*Reproduction:* Add bookmarks assigned to two different categories, then open the Today screen. Expected: each category on its own row (`Label: [scrollable strip]`). Actual: the two categories sit next to each other, each cramped into half the width.

*Suggested fix:*
```css
.today-container .bookmarks-section {
  width: 100%;
  display: flex;
  flex-direction: column;   /* stack categories */
  align-items: stretch;
  gap: 0.5rem;              /* optional spacing between category rows */
}
```

---

### Low

**No. 2 — Missing `key` on the mapped `<BookmarksCategory>`.**

```tsx
// BookmarksSection.tsx:29-37
{Object.entries(categorizedBookmarks.value).map(
  ([category, bookmarksData]) => {
    return (
      <BookmarksCategory
        label={`${category}:`}
        bookmarksData={bookmarksData}
      />   // no key
    );
  }
)}
```

Preact will emit a "key" warning for this list and may mis-reconcile category cards when categories are added/removed/reordered. `category` is a natural unique key: `<BookmarksCategory key={category} ... />`.

**No. 3 — Category display order can jump for numeric-looking category names.**

`categorizedBookmarks` builds a plain object keyed by the category name:

```tsx
// useBookmarksSection.tsx:57
const categorized: CategorizedBookmarks = {};
...
categorized[category].push(data);
```

`Object.entries` iterates **integer-like string keys first, in ascending numeric order**, before insertion-ordered string keys. So a user who names a category `"2024"` would see it forced to the front regardless of when it first appears. Using a `Map<string, BookmarkData[]>` (and `Array.from(map.entries())`) preserves first-appearance order for all names. Edge case, but cheap to make robust — and it would also stop the Today ordering from silently diverging from the BookmarksManager's ordering (which puts the default category first).

**No. 4 — Dead `containerRef` left in `useBookmarksSection`.**

The horizontal-scroll ref moved into `useCategorizedBookmarks`, but `useBookmarksSection` still creates (`useRef`) and returns `containerRef`, and it's still declared in the hook's return type. Nothing consumes it anymore (`BookmarksSection` only destructures `label` and `categorizedBookmarks`). Remove the ref, the return field, and the type entry.

---

### Nits / cleanup

**No. 5 — Large commented-out blocks left behind.** The `MoreButton` in `BookmarksSection.tsx`, the ~40-line `isOverflowing` / `moreButtonData` block (plus the `MoreButtonData` type comment) in `useBookmarksSection.tsx`, and the `/* padding: 0.4rem; */` in `styles.css` are all commented out rather than deleted. Git history preserves them; carrying them as comments adds noise and makes the "more/overflow" feature look half-present. The `VIEW-MORE` translation key is now referenced only inside a comment, so it's effectively dead too.

**No. 6 — `overflow: scroll` and scrollbar hiding on the strip.**

```css
/* styles.css:818 */
.bookmarks-section-container {
  ...
  overflow: scroll;        /* forces scroll on BOTH axes */
  scrollbar-width: none;   /* Firefox only */
}
```

For a single-row strip, `overflow-x: auto; overflow-y: hidden;` expresses the intent more precisely. Also, `scrollbar-width: none` hides the bar only in Firefox — on Chromium/WebKit the scrollbar still shows unless you add `.bookmarks-section-container::-webkit-scrollbar { display: none; }`.

**No. 7 — Category label is shown verbatim and is not localized.** Labels render as `${category}:` with a hardcoded Latin colon. The default category value is the hardcoded English string `"My Bookmarks"` (from `BookmarksManager`), so non-English users will see an untranslated label here now that categories are surfaced on the Today screen. The category value itself is out of this PR's scope, but this is the PR that makes it user-visible, so worth flagging for a follow-up. (Separately: removing the trailing colon from the `BOOKMARKS` key across all locales in `extension.json` is fine/consistent since the colon now lives on the per-category label.)

---

### Summary

No blocking correctness bugs, tests and type-check are green. The one thing I'd want confirmed before merge is **No. 1** — the section stacks categories in a row rather than a column, which will look broken for anyone with more than one bookmark category. The rest are low-risk cleanups.

---

soukigabriel:
@Blitzy Fixed most of the issues flagged above. Also this PR introduces the feature to open the bookmarks from the VIEW MORE button in the today screen.

Blitzy:
@soukigabriel Here are some things i discovered while testing:

1. On desktop, you can't collapse the sidebar if you open the bookmarks page with the "View More" button.

image

2. On mobile, it we should support scrolling the bookmarks side to side exactly like how we do for the Community filter pill buttons. They looks exactly the same and it would be a design consistency to to support it. The figma design shows the bookmark pill buttons overflowing offscreen which implies the ability to scroll them: https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=7957-14106&m=dev

3. Lets remove the horizontal dividers between the bookmark categories. Lets keep the horizontal dividers for seperation between the distinct sections of the Today screen but we can clean up the bookmarks section by not using them for between the categories its visually noisy.

image

Blitzy:
@soukigabriel still need some of these issues addressed

soukigabriel:
Yes, I will start working on this tomorrow, I was trying to leave the Bible Stack as stable as possible before jumping back to this, thank you @Blitzy!

soukigabriel:
@Blitzy The tasks mentioned above should be done now!

Blitzy:
> [!NOTE]

## Code Review — Today screen mobile layout (No linked issue found)

This is a re-review of the latest push (`a701447`). It follows up on @Blitzy's three requests from July 27 — desktop sidebar collapse, mobile horizontal scroll for bookmark pills, and removing dividers between bookmark categories — plus the earlier round of review comments. I checked out the branch and ran the suite directly rather than just reading the diff.

**Good news first:** all three of Blitzy's requests are visibly addressed in the code (`isSidebarCollapsed.value = false` on "View More", `useHorizontalScroll` wired up for mobile pill strips, and category dividers removed from `BookmarksSection`/CSS). The earlier medium/low findings from the first review round (No. 1 stacked-categories layout, No. 2 missing `key`, No. 3 `Map` for category ordering, No. 4 dead `containerRef`) are also resolved.

However, running the actual test suite on this branch turns up real failures that a diff-only read wouldn't show:

---

### High

**No. 1 — Three unit tests currently fail on this branch (`pnpm vitest run test/unit/today-screen`).**

```
❯ useBookmarksSection.test.tsx > moreButtonData > is defined (with a translated label) when any strip wraps to a new line
TypeError: Cannot read properties of undefined (reading 'value')
 ❯ useBookmarksSection.tsx:128:42
     128|     if (!isOverflowing.value || isMobile.value) {

❯ useTodayContent.test.tsx > dividedSectionsIds > lists the standalone sections (bookmarks is not one of them)
AssertionError: expected [ 'bookmarks', 'search', 'social' ] to deeply equal [ 'search', 'social' ]

❯ TodayContent.test.tsx > bookmarks > renders the BookmarksSection when showBookmarks is true
AssertionError: expected null not to be null
```

*Root cause:* this PR adds a `showBookmarks` flag to `useTodayContent` intending to pull `"bookmarks"` out of the generic `dividedSectionsIds` list (so it renders standalone like `ResumeReadingSection`, no longer treated as just another divided tile). The **tests** were rewritten to match that intended end-state, but the implementation was never finished:

- `useTodayContent.tsx` still does `if (showBookmarks) { sectionsData.push("bookmarks"); }` — `"bookmarks"` was never removed from `dividedSectionsIds`.
- `TodayContent.tsx` (the component) was never touched in this PR — it still destructures only `{ dividedSectionsIds, showResumeReading }`, has no `showBookmarks` usage, and still resolves `bookmarks` through `sectionComponentMap` instead of rendering `<BookmarksSection>` directly.
- Separately, `useBookmarksSection.tsx`'s `moreButtonData` computed unconditionally reads `isMobile.value` (added in this PR to hide "View More" on mobile), but the test's mocked `useTodayContext()` return value never includes `isMobile` or `showBookmarksList` — so any test path that reaches that branch throws.

*Suggested fix:* finish the refactor — remove `"bookmarks"` from `dividedSectionsIds`/`DividedSection` and render `showBookmarks && <BookmarksSection />` directly in `TodayContent.tsx` (matching `showResumeReading`) — and add `isMobile`/`showBookmarksList` to the mocked context in `useBookmarksSection.test.tsx`'s `configure()`. Either way, get `pnpm test` green before merge — right now it isn't.

---

### Medium

**No. 2 — `useHorizontalScroll` is called conditionally, violating the Rules of Hooks.**

```tsx
// useBookmarksCategory.tsx
export const useBookmarksCategory: UseBookmarksCategory = () => {
  const { isMobile } = useTodayContext();
  const containerRef = useRef<HTMLDivElement | null>(null);
  if (isMobile.value) {
    useHorizontalScroll(containerRef); // <-- calls useEffect internally
  }
  return { containerRef };
};
```

`isMobile` is `computed(() => viewportWidth.value <= MOBILE_BREAKPOINT)` in `SeedBibleStateManager.tsx` — it's a live signal that flips whenever the viewport crosses the breakpoint (resizing a browser window, rotating/opening a device's on-screen keyboard, etc.), not a value fixed at mount. `useHorizontalScroll` itself calls `useEffect` under the hood, so gating the call on `isMobile.value` means a single mounted `BookmarksCategory` instance can call a different number of hooks across renders. That breaks Preact's hook-slot ordering invariant: the `wheel`-listener cleanup from the previous mobile render isn't guaranteed to run when the condition flips, and toggling back across the breakpoint can leave a duplicate/stale listener attached to the strip.

*Reproduction:* Open the Today screen with bookmarks in a category, at a narrow (mobile) viewport width, then resize the window wider past 480px without navigating away, then resize back narrower. Inspect the bookmark strip's `wheel` listeners (e.g. dev tools "Event Listeners" panel) — the listener added on the first mobile render is not cleanly removed/re-added the way an unconditional hook would.

*Suggested fix:* drop the `if` and always call the hook — `useHorizontalScroll` only intercepts `wheel` when `el.scrollWidth > el.clientWidth`, which is already false on desktop (the strip wraps via `flex-wrap: wrap` there, no overflow), so calling it unconditionally is behaviorally a no-op on desktop and avoids the Rules-of-Hooks issue entirely:

```tsx
export const useBookmarksCategory: UseBookmarksCategory = () => {
  const containerRef = useRef<HTMLDivElement | null>(null);
  useHorizontalScroll(containerRef);
  return { containerRef };
};
```

---

### Low

**No. 3 — `isBookmarksListOpen` is added but never consumed.**

`bootstrap.tsx` now passes `isBookmarksListOpen: context.bookmarks.isFilterActive.value` into `TodayConfig`, and it's declared on the `TodayConfig` interface in `Today.tsx` — but nothing in the diff (or the rest of the codebase, confirmed via grep) reads it. Either wire it up (e.g. to disable/relabel "View More" while the sidebar's bookmarks list is already open) or drop it — right now it's dead config.

**No. 4 — Leftover commented-out prop stubs moved, not removed.**

The last commit's message is "Removed commented code from showBookmarksList" (which only touched `bootstrap.tsx`), but the `iconName`/`MaterialIcon` comment blocks flagged in the first review round (originally in `BookmarksSection.tsx`) are still present — they were relocated verbatim into the new `BookmarksCategory.tsx`:

```tsx
export interface BookmarkProps {
  text: string;
  handleClick: () => void;
  // iconName: string,
  // MaterialIcon: (props: {
  //     children: string;
  //     className?: string;
  // }) => preact.JSX.Element
}
```

Worth deleting for real this time, or implementing, rather than carrying it forward through another refactor.

---

### Nit

**No. 5 — Mobile scrollbar hiding is still Firefox-only.**

```css
.today-container .bookmarks-section .bookmarks-section-container {
  overflow-x: scroll;   /* good — now axis-specific, per the earlier note */
  scrollbar-width: none; /* Firefox only */
}
```

This partially addresses the first-round note (overflow is now correctly `overflow-x` instead of both axes), but on Chromium/WebKit mobile browsers — the large majority of mobile traffic — the scrollbar will still render. Add `.bookmarks-section-container::-webkit-scrollbar { display: none; }` alongside it.

---

### Summary

Type checking is clean and the three requested UX fixes are visibly in place, but **the branch's own test suite is red** (3 failing tests), and the failures point at an unfinished refactor (bookmarks section still living inside `dividedSectionsIds` in the actual component despite the new `showBookmarks` flag) rather than a flaky test. That's the one thing I'd want fixed before merge; the hooks issue (No. 2) is a real bug too but less urgent since it only manifests on breakpoint-crossing resizes. The rest are low-risk cleanups.


---

Blitzy (approved):

PR #1506, opened 2026-07-24
Feature/1415

### LINKED ISSUE #1415: Chapters that are taking long to fetch should show a loading overlay in the reader.
## What is happening 

In low bandwidth situations, chapter requests can take longer than normal to fetch. Currently while a chapter is being fetched the current chapter stays on screen. The mobile swipe feature shows the page slide over but it just shows the currently loaded chapter until the inflight request finishes.

## What is expected

1. For chapter navigation buttons, if the request is taking a long time (let's get 1000ms for now) it should show a blank page with a loading spinner. This is very similar to how you version does it. Alternatively we could allow skeleton loader placeholders for the text which might feel more elegant. 

2. For mobile swiping, the unloaded pages should show the same loading spinner/skeleton placeholder as stated above.

Kabi4:
Fixes #1415


Kabi4:
BibleReader.tsx
Lines 969-969

Controls the time delay in switch

Blitzy:
> [!NOTE]

## Code review — PR No. 1506 (Fixes #1415)

Thanks for this! The core of the feature is in good shape. The button/keyboard navigation path does exactly what the issue and the follow-up discussion asked for: it keeps the previous chapter on screen and only blanks to a spinner after the fetch has been slow for a full second (`CHAPTER_LOADING_DELAY_MS = 1000`), so fast/cached loads never flash — which is the behavior @KallynGowdy and @Blitzy explicitly called out as a priority in the issue comments. The `useDelayedFlag` cleanup on every `active` toggle also correctly avoids a stuck spinner across rapid navigations, and honoring `prefers-reduced-motion` on the spinner is a nice touch.

Below are the issues I found, ordered by severity.

---

### 🟠 Medium

**M1 — Mobile swipe still shows a blank page (no spinner/skeleton) for a chapter that hasn't been prefetched.**

Requirement No. 2 in #1415 is: *"For mobile swiping, the unloaded pages should show the same loading spinner/skeleton placeholder."* This PR relies entirely on the prefetched previews for the swipe path — the new spinner never appears during a swipe.

Why: during a committed swipe the track is translated to a **side** panel (`translateX(sign * PANEL_PCT * 2)` for next), and the side panels render only the prefetched preview:

- `components/BibleReader/BibleReader.tsx:1690-1695` / `:1712-1717` render `renderStaticChapterContent(preview ?? null)`, and `renderStaticChapterContent(null)` returns `null` (`:955`).
- The `ChapterLoadingPlaceholder` lives only inside `renderMainContent()`, which is the **center** panel (`:1701`). During the swipe that panel is translated off-screen, so its spinner is never visible.

The net effect: on a good connection the preview is ready and you see the real next chapter (great), but on the slow/low-bandwidth connection this issue is actually about, the preview may not have arrived yet and the user stares at a fully blank panel with no indicator until the fetch resolves — the exact scenario the issue was filed to fix.

Reproduction:
1. Open the app in mobile layout (or DevTools device mode) and set network throttling to "Slow 3G".
2. Open a chapter and, before the adjacent-chapter prefetch completes, swipe to the next chapter.
3. Observe: the destination panel is blank during the wait — no spinner or skeleton.

Suggested fix: render `ChapterLoadingPlaceholder` in the side panels when their preview is `null`, e.g.

```tsx
<div className="sb-chapter-content">
  {mobileChrome?.nextChapterPreview
    ? renderStaticChapterContent(mobileChrome.nextChapterPreview, scriptureElements)
    : <ChapterLoadingPlaceholder />}
</div>
```

(or overlay the spinner on the track while `swipeNavigatingRef` / `loading` is active). That way the swipe destination shows the same indicator as button navigation.

---

### 🟡 Low

**L2 — Optimistic `scrollPosition = 0` is not restored when a load fails.**

`applyOptimisticChapterLocation` zeroes `scrollPosition` (`managers/BibleReadingManager.tsx:1109`), but the revert paths only restore `bookId` / `chapterNumber`, not `scrollPosition`:
- `revertOptimisticLocation` in `loadNextChapter` (`:1759-1765`) and `loadPreviousChapter` (`:1545-1551`)
- the `catch` in `selectChapter` (`:1727-1728`)

On a fetch failure the old chapter stays on screen (because `chapterData` never changed), but the persisted `scrollPosition` signal is now `0`. The scroller effect re-runs on `chapterData` / `scrollToVerse` changes, so the jump isn't immediate, but the stale `0` can be saved as the reading position or applied the next time that effect fires — silently sending the reader back to the top of a chapter they hadn't left.

Suggested fix: capture `scrollPosition.peek()` alongside book/chapter at entry and restore all three in the revert paths.

**L3 — `selectChapter` has no in-flight guard, so overlapping navigations can clobber each other with a stale location.**

The chevrons (`managers/BibleToolsManager.tsx:602-604`, `:743-745`), keyboard handler (`components/TabsLayout.tsx:483`), and swipe (`:287`) all guard against re-entry while `loading` is true, but `selectChapter` (`managers/BibleReadingManager.tsx:1708`) does not. Each navigation captures its own `previousBookId` / `previousChapterNumber` snapshot and `chapterData` is updated last-write-wins, so a book/chapter jump fired while a next/prev fetch is in flight can interleave. If the older request then fails, `revertOptimisticLocation` restores a snapshot that predates the newer explicit selection, overriding where the user actually wanted to go.

Reproduction (racey, but reachable on a slow connection): start a slow next-chapter load, then open the book selector and jump to a different book before it resolves; if the two fetches resolve out of order, or the first fails, the final location can be wrong.

Suggested fix: serialize navigation with a request/sequence token and ignore resolutions from a superseded request, or at minimum guard `selectChapter` on `loading` for consistency with the other entry points.

**L4 — Micro-race between `loading = true` and the disabling guards.**

In `loadNextChapter` / `loadPreviousChapter`, `loading.value = true` is set only *after* `await runNavigationHooks(...)` (`:1741`+, `:1527`+). The chevron `isDisabled` and keyboard guard key off `loading.value`, so there's a microtask-sized window after the click where a second activation isn't yet blocked. It's hard to trigger with human input, but setting `loading = true` before awaiting the hooks would close the window cleanly.

**L5 — No CHANGELOG entry.** The repo convention (see the `## TBD` section and every recent PR, e.g. `[#1499]`) is to add an entry under `## TBD`. This PR touches user-visible behavior but adds none. Consider an "Added" bullet under `## TBD` referencing this PR.

**L6 — No tests.** `test/unit/seed-bible/managers/BibleReadingManager.test.ts` and `.../components/BibleReader.test.tsx` already exist. The new logic — `resolveAdjacentChapterLocation` (especially its book-boundary crossing via `firstChapterNumber` / `lastChapterNumber`) and the revert-on-failure paths — is pure and cheap to cover. A couple of unit tests here would guard against regressions in the optimistic-navigation math.

---

### Summary

| Severity | Finding |
|----------|---------|
| Medium | M1: Swipe path shows a blank panel (no spinner) for un-prefetched chapters — requirement No. 2 not met on slow connections |
| Low | L2: `scrollPosition` not restored on failed load |
| Low | L3: `selectChapter` unguarded; overlapping navigations can revert to a stale location |
| Low | L4: micro-race on the `loading` guard |
| Low | L5: missing CHANGELOG entry |
| Low | L6: no tests for the new navigation logic |

The one I'd treat as blocking relative to the linked issue is **M1**, since covering the slow-swipe case is half of what #1415 asked for. The rest are polish and robustness.

---

Blitzy:
This is a good start, and will a lot of this work was pulled in by an upcoming PR that will combine the works of multiple branches into a larger feature set of fast navigation, better fetching placeholder (which this one focuses on). Going to close this for now since its work will be included in that one.

PR #1511, opened 2026-07-24
Persist selected Bible translation to user profile (#1445)

### LINKED ISSUE #1445: Save user selected translations to user profile config so that it reopens on returning.

KallynGowdy:
## Summary

Fixes #1445. Bible translation choice used to only survive via `?translation=`/`?translationId=` in the URL — switching translations and then returning without that exact link would silently fall back to the default (AAB). Now the app remembers the last translation a signed-in user explicitly picked and reopens it automatically.

- Picking a translation from the **translation list in the Bible selector** now saves that choice to the user's profile (`profile.config.translationId`), the same place saved language and theme preferences already live.
- On login, once the profile loads, the saved translation is applied to the current tab automatically — mirroring the existing pattern used for saved language (`ConfigManager`) and theme (`ThemeManager`).
- An explicit `?translation=`/`?translationId=` in the URL still wins over the saved profile value, so shared/deep links keep working exactly as before.
- Only an explicit pick from the selector's translation list persists. Programmatic translation changes — opening the selector (which syncs to the current tab's translation), switching the UI language (which auto-applies the nearest matching translation), and adding a custom translation by URL — do **not** overwrite the saved preference, since none of those are a deliberate "I want this translation" choice.

## Changes

- `managers/TabsManager.tsx` — reads `profile.config.translationId` and applies it to the selected tab once the profile loads (no URL translation param wins over it).
- `managers/BibleSelectorManager.tsx` — new `pickTranslation()` action (alongside the existing `selectTranslation()`) that persists the choice via `saveProfileConfigValue`; wired to the selector's translation-list click in `components/BibleSelector/BibleSelector.tsx`.
- `managers/SeedBibleStateManager.tsx` — threads `login` into `createTabs`/`createBibleSelectorState`.
- Test coverage in `TabsManager.test.ts` and `BibleSelectorManager.test.ts` for: saved-translation restore on profile load, URL precedence over the saved value, persistence on an explicit selector pick, and no persistence from any other translation-change path.

## Test plan

- [x] `pnpm check:ts`
- [x] `pnpm lint`
- [x] `pnpm test` (full suite, 2987 tests passing)
- [ ] Manual smoke check: log in, pick a translation from the Bible selector, reload with no `?translation=` in the URL — the picked translation should reopen. Then open a URL with an explicit `?translation=` for a different translation — that one should win instead.


Blitzy:
> [!NOTE]

## Code review — Persist selected Bible translation to user profile (closes #1445)

### Summary

The change does exactly what issue #1445 asks for: an explicit translation pick in the Bible selector now writes `profile.config.translationId`, and that value is applied back to the selected tab once the profile loads. The design is well-scoped and mirrors the existing profile-sync conventions closely:

- **Persist only on a deliberate pick.** A new `pickTranslation` wraps `selectTranslation` and adds the profile write, and only the selector's list-click (`BibleSelector.tsx`) is rewired to it. The many programmatic callers (selector-open sync, the language-driven translation applicator in `SeedBibleStateManager.tsx:1464`, custom-URL addition) correctly stay on `selectTranslation` and don't persist. This matches the `persistLanguage`/`setTheme` split in `ConfigManager`/`ThemeManager`.
- **Restore without fighting the URL.** The apply effect reads `login.profile` (tracked) and everything else `untracked`, so it runs on profile load only — not on URL changes — and bails when an explicit `?translation=`/`?translationId=` is present. This is the same reasoning as the `lang` sync effect it cites.
- **Solid test coverage** for the happy path and the important negative cases (URL precedence, no-persist for reading-state/URL-driven changes). All 50 tests in the two affected files pass locally.

Nothing here is a blocker. One functional gap is worth addressing before merge, plus two minor notes.

---

### Medium

**No. 1 — Restoring a saved translation that doesn't contain the current book fails instead of falling back**

`TabsManager.tsx:468` passes the tab's *current* book/chapter straight into `selectTranslationAndChapter`:

```ts
void readingState.selectTranslationAndChapter(
  savedTranslationId,
  readingState.bookId.peek() ?? DEFAULT_BOOK_ID,
  readingState.chapterNumber.peek(),
  { updateUrl: false }
);
```

`selectTranslationAndChapter` clamps an out-of-range *chapter*, but **throws if the book is not in the target translation** (`BibleReadingManager.tsx:1558`):

```ts
const selectedBook = books.books.find((book) => book.id === nextBookId);
if (!selectedBook) {
  throw new Error(`Book with ID "${nextBookId}" not available for translation "${nextTranslationId}".`);
}
```

That throw is caught and only sets `error.value`; the translation/book/chapter signals are never updated. So restoring a **partial translation** (NT-only, single-book, etc.) while the current book falls outside it leaves the reader on the old translation with an error state, and the saved choice silently fails to apply.

Note the sibling URL path, `syncSelectedTabFromUrl` (`TabsManager.tsx:328-332`), already guards against exactly this:

```ts
const selectedBook = books.find((book) => book.id === requestedBookId) ?? null;
if (!selectedBook) {
  return;
}
```

The profile-restore path is the one place that skips that guard, so the two paths behave inconsistently.

**Reproduction**
1. Sign in.
2. Set the account's `profile.config.translationId` to a translation that does **not** contain Genesis (any NT-only translation in the catalog; the quickest way is to pick such a translation from the selector while reading a book it does contain, which saves it).
3. In a new session, open a deep link to a book outside that translation with **no** translation param, e.g. `/?book=GEN&chapter=1`.
4. When the profile loads, the effect calls `selectTranslationAndChapter(savedTranslation, "GEN", 1)` → the book-not-found branch throws → caught → `error.value` is set and the translation is not applied.

**Suggested fix** — mirror `syncSelectedTabFromUrl`: resolve the saved translation's books first and only restore if the current book exists (otherwise skip, or fall back to the translation's first available book/first chapter). For example:

```ts
untracked(() => {
  // ...existing URL / same-translation guards...
  const readingState = selectedTab.peek()?.readingState;
  if (!readingState || readingState.translationId.peek() === savedTranslationId) {
    return;
  }

  void (async () => {
    const bookId = readingState.bookId.peek() ?? DEFAULT_BOOK_ID;
    const books = (await dataManager.getTranslationBooks(savedTranslationId)).books;
    const targetBook = books.find((b) => b.id === bookId) ?? books[0];
    if (!targetBook) return;
    await readingState.selectTranslationAndChapter(
      savedTranslationId,
      targetBook.id,
      targetBook.id === bookId ? readingState.chapterNumber.peek() : (targetBook.firstChapterNumber ?? DEFAULT_CHAPTER_NUMBER),
      { updateUrl: false }
    );
  })();
});
```

At minimum, ensure a failed restore never surfaces an error banner to the user for something they didn't just do.

---

### Low

**No. 2 — Restored translation isn't in the URL until the first navigation**

The restore uses `{ updateUrl: false }`, so after it applies (say `NIV`) the URL still has no `?translation=`. `getUrlQueryParams` does write a non-default translation on the next navigation (`BibleReadingManager.tsx:1935`), so the URL self-heals once the user turns a page. But in the window *before* that first navigation, an external URL change to a translation-less URL (browser back/forward, a pushed in-app link, a manual edit) triggers `syncSelectedTabFromUrl`, which reads the URL's (absent) translation as the default and reverts the restored choice mid-session.

This is an edge case and consistent with how the `lang` sync already behaves, so it's informational rather than a required change. If you want to close it, having the restore commit the translation to the URL (a `replace`, not a push) would keep the two sources of truth aligned immediately — at the cost of putting `?translation=` in the URL on load, which the current design deliberately avoids. Worth a conscious decision either way.

**No. 3 — Test gap for the book-not-available restore**

The new `TabsManager` tests all restore a translation that shares the book `MAT`, so the failure in No. 1 isn't exercised. Adding a case where the saved translation lacks the current book (asserting a graceful fallback rather than an error) would lock in whatever behavior you choose for No. 1.

---

### Nits (non-blocking)

- `pickTranslation`'s doc comment and the two inline comments explaining the persist/no-persist split are genuinely helpful — thanks for spelling out which call sites intentionally don't persist.
- CI `build (24.x)` was still in progress at review time; assuming it goes green this is otherwise good to merge once No. 1 is addressed.

---

KallynGowdy:
Addressed **No. 1** and **No. 3** in 1c2fa37.

**No. 1** — `applySavedTranslation` (new helper in `TabsManager.tsx`) now resolves the saved translation's own book catalog via `dataManager.getTranslationBooks` before restoring, and falls back to that translation's first book/chapter when the tab's current book isn't in it (instead of calling `selectTranslationAndChapter` with the current book directly, which threw and left an error banner). Chapter is still clamped to the current chapter when the book does match, same as before. A failed books fetch is caught and logged, leaving the tab on its current translation rather than surfacing an error.

**No. 3** — added a regression test ("falls back to the saved translation's first book when it doesn't contain the current book") that restores a saved NIV translation while the tab is on GEN (only in AAB, not in the mocked NIV catalog), asserting it lands on NIV's first book/chapter with no error rather than failing.

Left No. 2 (URL not reflecting the restored translation until the first navigation) as-is per the review's own note that it's informational/a judgment call, not a required fix — happy to revisit if you'd like it closed too.

`pnpm check:ts`, `pnpm lint` (0 errors), and the full suite (2988 tests) all pass.

---

Blitzy:
# PR Review: Persist Selected Bible Translation to User Profile

## Issue

The translation selection saves to the profile correctly (confirmed `translationId` was set to `eng_ulb` in the profile data). The problem is on load: the persisted translation isn't applied when the **Today screen** renders, and it falls back to the default (`AAB`).

There's a brief flicker to `ULB` before it reverts to `AAB`, which suggests the profile value loads momentarily and then gets overwritten by whatever the Today screen initializes with.

The behavior is specifically tied to the Today screen. Any load path that bypasses the Today screen applies the persisted translation correctly:

- Deep link with a book + chapter → works
- Deep link with a book, no chapter → works
- Session ID as the only parameter (bypasses the Today screen) → works
- Cold load with no params, or navigating back to notebook/chapter (both bring up the Today screen) → reverts to `AAB`

## Reproduction Steps

1. Select a non-default translation (e.g. `ULB`).
2. Confirm it persisted: check profile data, `translationId` should be `eng_ulb`.
3. Remove all query params / hashes and do a cold load.
4. Observe the Today screen come up. It briefly flickers to `ULB` then loads back on `AAB` (the default).
5. Confirm `ULB` is still in the profile (it is), so it's a load/apply issue, not a persistence one.
6. To isolate: load a deep link with a chapter (e.g. `?book=EXO&chapter=5`, `?book=EXO&chapter=6`...) or a session ID as the sole param (e.g. `?sessionId=abc123`). These bypass the Today screen and apply `ULB` correctly.
7. Trigger the Today screen manually. It reverts to `AAB`, confirming the Today screen is where the persisted value gets dropped.

KallynGowdy:
Found and fixed the real root cause in 9001f454 — thanks for the detailed repro steps, they narrowed it down fast.

**What was actually happening:** it wasn't the Today screen itself doing anything wrong. A freshly created tab kicks off its own initial chapter load (`loadInitialData()`) synchronously the moment it's constructed — well before the profile has had a chance to load over the network (that's a separate round-trip). `loadInitialData()` has no awareness of anything that happens after it starts, so when it finally finishes, it unconditionally writes back whatever translation/book/chapter it captured at the start (the default, AAB/Genesis 1).

So the sequence was:
1. Tab is created on the default translation (AAB); its initial load starts.
2. Profile loads, sees the saved translation (`eng_ulb`), and switches to it — this is the "brief flicker to ULB."
3. The *original* initial load (still in flight since step 1) finally resolves and unconditionally reasserts AAB/Genesis 1 — this is the "revert to AAB."

The bare cold-load path (no `book`/`chapter`/`translationId`/`sessionId` in the URL) is exactly the one case where nothing short-circuits this: an explicit URL translation skips the restore entirely, and joining a shared session reuses an already-loaded reading state instead of starting a fresh `loadInitialData()`. That's why deep links and `sessionId` links worked fine while a bare cold load (which is also what opens the Today screen) didn't — the Today screen just happened to be the visible witness of a race that already existed underneath it.

**The fix:** the restore now waits for the tab to finish its own initial load before applying the saved translation, so it's always the last write rather than something that can get raced and clobbered. I also made the restore commit the translation to the URL immediately (as a `replace`, not a `push`) instead of waiting for the next real navigation, so it can't get read as "no translation" by anything else that happens to touch the URL in the meantime.

Added a regression test that reproduces the exact ordering (holds the tab's initial chapter fetch open, restores the profile translation while it's still in flight, then lets the initial load finish) and confirmed it fails without the fix and passes with it.

This PR also picked up a `develop` merge along the way (200 commits) — I rebased my fix on top of that cleanly rather than re-merging.

`pnpm check:ts`, `pnpm lint` (0 errors), and the full suite (3353 tests) all pass. Would appreciate a fresh run-through on the deployed branch to confirm the flicker/revert is gone on a bare cold load.

---

Blitzy:
> [!NOTE]
> This review was generated by a Claude Code routine setup by Ryan Cook (ryan@helloao.org). It's a single-pass review done without the multi-agent fan-out/subagent verify pass this routine normally uses, so treat it as one careful read-through rather than the full-strength review.

## Code review — Persist selected Bible translation to user profile (closes #1445)

I read issue #1445, the full PR diff at the current head (`9001f454`), and the whole comment thread — including the first review round (No. 1/No. 2/No. 3 from the earlier pass, addressed in `1c2fa37`) and Blitzy's follow-up bug report about the Today-screen flicker/revert, which `9001f454` fixes by having the restore wait for the tab's own initial load to finish (`waitForIdle`) before applying, and by committing the restored translation to the URL immediately. That fix is solid for the specific race it targets (the tab's *own* initial load racing the restore). The concerns below are about races and gaps that fix doesn't cover.

### Medium

**No. 1 — The restore can silently overwrite a translation the user picks while it's still in flight, and the result is sticky (survives reload)**

`applySavedTranslation` (`TabsManager.tsx:467-538`) captures `savedTranslationId` once, then does two awaits — `waitForIdle` and a `getTranslationBooks` network call — before calling `readingState.selectTranslationAndChapter(savedTranslationId, ...)` at line 518. Nothing between the start of the function and that call re-checks whether the reading state has since moved on for a real reason.

Reproduction:
1. Sign in with a profile whose saved translation is `NIV` (from an earlier session).
2. Cold load with no URL params. The tab starts on the default (`AAB`) and kicks off its own initial load; the profile loads at roughly the same time, so `applySavedTranslation(readingState, "NIV")` starts and is now waiting on `waitForIdle`.
3. While that initial load is still in flight (slow connection, or just normal network latency), the user opens the Bible selector and explicitly picks `ESV`. `pickTranslation("ESV")` runs, updates the reading state to `ESV`, and correctly persists `ESV` to the profile.
4. `waitForIdle` resolves once the tab goes idle (which now includes the `ESV` navigation). The restore, still holding the stale `savedTranslationId = "NIV"` from step 2, fetches NIV's book catalog and calls `selectTranslationAndChapter("NIV", ...)` — silently reverting the reader from `ESV` back to `NIV`, moments after the user picked `ESV`.
5. Because the restore also does `commitSelectedTabToUrl({ replace: true })` right after, the URL now carries `?translation=NIV`. On the *next* load, `getInitialTranslationId` reads the URL first, so the wrong translation sticks even though `profile.config.translationId` correctly says `ESV` — the bug isn't just a momentary flicker, it persists until the user notices and re-picks.

Suggested fix: before the final `selectTranslationAndChapter` call, re-check that the reading state hasn't already moved to a different translation since the restore started (e.g. re-peek `readingState.translationId` right before committing and bail if it no longer matches what it was when the effect captured `savedTranslationId`), or track an "abort if superseded" token the same way the tab's own initial load needed one.

### Low

**No. 2 — Duplicates the exported `resolveChapterInBook` helper instead of calling it**

`applySavedTranslation` re-derives the chapter clamp by hand at `TabsManager.tsx:506-516` (`firstChapterNumber`, `maxChapterNumber`, then the range check) instead of calling `resolveChapterInBook` from `BibleReadingManager.tsx:909`, whose own doc comment says exactly this logic was "previously duplicated in `selectTranslationAndChapter` and `loadInitialData`; this is the single home for it." The hand-rolled version here also drops `resolveChapterInBook`'s fallback to `book.lastChapterNumber` when `numberOfChapters` is missing/non-positive (see `lastChapterOf`, `BibleReadingManager.tsx:889-897`), so a book with malformed chapter-count data resolves differently here than everywhere else in the app. Calling `resolveChapterInBook(targetBook, readingState.chapterNumber.peek())` would remove the duplication and pick up that fallback for free. (Note: the pre-existing `syncSelectedTabFromUrl` a few lines up, `TabsManager.tsx:333-341`, has the same duplication — not introduced by this PR, but this was the moment to consolidate both onto the shared helper instead of adding a third copy.)

**No. 3 — Restore doesn't check whether the tab is still around after its async waits**

`applySavedTranslation` awaits `waitForIdle` and then a `getTranslationBooks` fetch (`TabsManager.tsx:482-497`) with no check that the tab wasn't closed (`removeTab` → `readingState.dispose()`) in the meantime. If the user closes that tab while the restore is waiting, the function still fetches the saved translation's book catalog and calls `selectTranslationAndChapter` on the now-disposed reading state — a wasted network round-trip and a set of signal writes on an instance nothing will ever render again. It won't crash (dispose already makes in-flight content requests settle harmlessly), but it's dead work worth skipping with an early return if `readingState` is no longer any tab's reading state.

**No. 4 — The "no error banner for an unrequested restore" ask from the first review round is only half-covered**

The earlier review's No. 1 asked that a failed restore never surface an error to the user for something they didn't do. The fix added a guarded `dataManager.getTranslationBooks(savedTranslationId)` call at the top of `applySavedTranslation` (caught and logged, `TabsManager.tsx:484-494`), which covers a *missing-book* failure. But `selectTranslationAndChapter` itself (`BibleReadingManager.tsx:2324-2366`) does its *own*, unguarded `getTranslationBooks` call, and any failure there (transient network blip between the two calls, cache eviction, etc.) sets `readingState.error.value` — which is rendered as an error banner. So a restore the user never asked for can still show an error banner for a network hiccup that has nothing to do with anything they just did; the earlier ask is only satisfied for the one failure mode this fix specifically targeted.

---

Nothing above blocks the core feature — persistence and the happy-path restore both work as described, and the fix in `9001f454` genuinely closes the flicker/revert bug Blitzy reported. No. 1 is the one I'd want addressed before merge, since it can leave a user stuck on the wrong translation after a perfectly normal interaction; No. 2-4 are cleanup/robustness notes.


---

KallynGowdy:
Addressed **No. 1**, **No. 2**, and **No. 3** in 14c74fa5 (left No. 4 as-is, per your request to fix 1-3).

**No. 1 — restore silently overwrites a translation picked while it's still in flight, and the wrong one sticks:** `applySavedTranslation` now snapshots `readingState.translationId` before its two awaits (`waitForIdle`, then fetching the saved translation's books). Right before actually committing the restore, it re-checks that snapshot against the current value — if anything changed it in the meantime (the user explicitly picking a different translation, most notably), that deliberate choice wins and the restore bails instead of clobbering it. Added a regression test that has the user pick `BSB` while a gated `NIV` books fetch is still pending, and confirms `BSB` sticks.

**No. 3 — restore doesn't notice the tab was closed:** the same pre-commit check also confirms the tab is still attached (`tabs.peek().some(...)`) before acting, so closing the tab mid-restore is now a clean no-op instead of running `selectTranslationAndChapter` against a disposed reading state. Added a regression test for this too.

**No. 2 — duplicated the chapter clamp instead of calling `resolveChapterInBook`:** replaced the hand-rolled `firstChapterNumber`/`maxChapterNumber` computation with a direct call to `resolveChapterInBook(targetBook, readingState.chapterNumber.peek())`. This also picks up its fallback to `book.lastChapterNumber` for malformed chapter-count data, which the hand-rolled version was missing. (Left `syncSelectedTabFromUrl`'s pre-existing duplication alone, since that's out of scope for this PR per your own note.)

Confirmed both new regression tests fail without their respective fixes and pass with them. `pnpm check:ts`, `pnpm lint` (0 errors), and the full suite (3355 tests) all pass.

---

Blitzy:
> [!NOTE]

## Code review — Persist selected Bible translation to user profile (closes #1445)

This PR has already been through two thorough review rounds (comments above), each followed by a fix commit. I read issue #1445, the full diff at the current head (`14c74fa5`), the entire comment thread (both prior reviews, Blitzy's Today-screen bug report, and both fix write-ups), and re-derived the logic in `TabsManager.tsx`'s `applySavedTranslation`/`waitForIdle` and `BibleSelectorManager.tsx`'s `pickTranslation` from the current file contents rather than just the diff, to check the latest fixes actually hold up.

### Verifying the latest fixes (commit `14c74fa5`)

All three issues from the second review round are correctly fixed:

- **Stale-restore-wins race** (a user's explicit pick could get silently reverted, and the wrong translation would then stick in the URL): `applySavedTranslation` now snapshots `translationIdAtStart` before its two awaits (`waitForIdle`, then `getTranslationBooks`) and re-checks it immediately before calling `selectTranslationAndChapter` — confirmed by tracing `TabsManager.tsx` end to end, this re-check sits right before the only call that would clobber a newer pick, with no further await in between. The new regression test ("does not let the restore overwrite a translation the user explicitly picks while it is still in flight") reproduces exactly the scenario from the previous review.
- **Closed-tab race**: the same pre-commit check now also verifies `tabs.peek().some(tab => tab.readingState === readingState)`, so a tab closed mid-restore is a clean no-op. Covered by its own regression test.
- **Duplicated chapter-clamp logic**: replaced with a direct call to the shared `resolveChapterInBook` (`BibleReadingManager.tsx:909`) — verified its fallback to `book.lastChapterNumber` for malformed chapter-count data is now picked up here, matching every other call site.

The one item left open from that round (an unguarded `getTranslationBooks` inside `selectTranslationAndChapter` itself can still surface an error banner for an unrelated network hiccup during a restore the user didn't ask for) was explicitly deferred at the author's request, not missed — noting it here only for visibility, not as a blocker.

### New finding

**Low — A translation persisted from a custom API endpoint won't restore on a new device**

`applySavedTranslation` calls `dataManager.getTranslationBooks(savedTranslationId)` using whatever endpoint `BibleDataManager` currently has mapped for that ID (`getEndpointForTranslation`, `BibleDataManager.tsx:572`). That mapping is cached in `localStorage` (`BibleDataManager.tsx:788-803`), not part of the user's profile. If a user adds a custom-endpoint translation, picks it from the selector's list (persisting its ID via `pickTranslation`), then logs into a different browser/device, `applySavedTranslation` looks up that ID with no endpoint mapping present, falls back to the default endpoint, gets a 404-style "not found" from `getTranslationBooks`, and the `.catch()` in `applySavedTranslation` (`TabsManager.tsx:484-494`) silently no-ops — the saved translation just never restores on the new device, with only a `console.warn`.

This doesn't affect the catalog-translation case the issue and PR are actually about (No `custom translation URL addition` already stays on `selectTranslation`, per the PR's own design, and never triggers this specifically) — it only applies to *catalog-listed* translations that happen to have been sourced from a non-default endpoint the new device has never seen (e.g. a translation added by URL, then later appearing in the picker while it's the active selection, then explicitly re-picked from the list). Since the click handler in `BibleSelector.tsx` is shared for every entry in that list, that path does reach `pickTranslation`.

Given the failure mode is silent (no error surfaced, reader just stays on the default translation) and the scenario requires a custom-endpoint translation specifically, this is a minor/edge-case gap rather than something that affects the reported bug's users — flagging for awareness rather than as a blocker.

---

### Verdict

No blocking issues from this pass. The two prior rounds already exercised the core races thoroughly, and the latest commit's fixes for those are sound. CI (`build (24.x)`) is green as of `14c74fa5`. Good to merge from a code standpoint, pending maintainer sign-off on whether the custom-endpoint restore gap above is worth a follow-up (or a fast-follow issue) versus leaving as-is.

---

PR #1517, opened 2026-07-26
Feature/1397

### LINKED ISSUE #1397: Footnote scripture references should be navigable
Many footnotes contain references to other books and chapters. We should detect these on the fly and make them clickable so that you can quickly look at the referenced scripture.

Image

Kabi4:
Fixes : #1397 

https://www.loom.com/share/cc50ed7f0c8c4286acfbc2aa452fa512


Blitzy:
> [!NOTE]

## Code review — PR #1517 (Fixes #1397: navigable footnote scripture references)

### Summary

The change adds a `VerseReferenceText` component that scans plain text for scripture references and renders each as a clickable `VerseReferenceLink`, then wires it into the footnote modal so a tapped reference closes the footnote and jumps to that verse. This is a clean, well-targeted implementation that matches the intent of the issue.

**What I verified locally (branch `feature/1397`):**
- `pnpm vitest run test/unit/seed-bible/components/BibleReader.test.tsx` → **41 passed** (including the new test).
- `pnpm check:ts` → **0 visible errors.**
- The parser choice is consistent with the codebase: `VerseReferenceText` reuses the same offset-based `parseVerseReferences` from `BibleDataManager` that `ChatsManager` already uses to detect references in free-form chat text, so the footnote feature follows an established pattern rather than inventing a new one.

No blocking (high-severity) defects found. The findings below are correctness/completeness improvements, most of which are inherited from the shared parser but become more visible now that it is applied to footnote prose.

---

### Medium

**No. 1 — References in non-English translations are never detected.**

The offset parser resolves book names through a hardcoded, English-only map (`BOOK_ID_MAP` / `getBookId` in `managers/BibleDataManager.tsx`). Footnotes are written in the translation's own language, so a Spanish/French/etc. translation whose footnote says `Juan 3:16` or `Génesis 1:1` produces **zero links** — the feature silently does nothing for a large share of the app's content.

Worth contrasting: the other parser in the repo, `managers/parseVerseReference.ts`, matches book names against the *loaded translation's* `books` list (`b.name`), so it already handles localized names — but it parses a single whole-string reference and has no character offsets, so it can't scan free text as-is.

- *Reproduction:* open a footnote in a non-English translation that contains a cross-reference (e.g. a Spanish translation with `Juan 3:16`) → no link is rendered.
- *Suggested fix:* at minimum, note the English-only limitation in the PR/issue. Ideally, teach the free-text scan to resolve book names against the current translation's `books` (reusing the localized matching logic already in `parseVerseReference.ts`) so localized footnotes linkify too. Since this affects chat as well, it may be better tracked as a shared follow-up rather than blocking this PR.

---

### Low

**No. 2 — Ordinary prose can be linkified as a false positive.**

`getBookId` falls back to prefix (`startsWith`) matching, and the free-text regex accepts any word immediately followed by a number. Footnotes are prose and frequently use linguistic abbreviations, so plausible false positives include:
- `divided among 3 groups` → `among 3` matches because `"among".startsWith("amo")` → resolves to **Amos 3**.
- `Heb. 3 ...` / `Gk. 3 ...` — `Heb.` in footnotes almost always abbreviates *Hebrew*, not the book of Hebrews, but `Heb 3` resolves to **Hebrews 3**.

This exists in chat today too, but footnote text is longer prose and uses these abbreviations far more, so the exposure is higher here.

- *Reproduction:* open (or construct) a footnote whose text contains `among 3` or `Heb. 3` → the phrase becomes a clickable link to the wrong book/chapter.
- *Suggested fix:* tighten `getBookId` so the prefix fallback only fires for reasonably long tokens, or require the matched book+chapter to actually exist in the current translation before linking. Validating against the translation's `books` would also naturally address this.

**No. 3 — Duplicated slice-and-link logic + re-parse on every render.**

`VerseReferenceText` re-implements essentially the same "walk matches → push text slices and link elements" loop that `ChatsManager` already has (`parseChatText`), and it calls `parseVerseReferences(text)` on every render instead of memoizing.

- *Suggested fix:* extract a shared helper for the slice-and-link walk (chat could then drop its private copy), and wrap the parse in `useMemo(() => parseVerseReferences(text), [text])`. Purely maintainability/perf; the footnote modal re-renders rarely, so the perf impact is negligible.

---

### Nit

**No. 4 — No CHANGELOG entry.** The repo keeps a `## TBD` section in `CHANGELOG.md` and each feature PR adds a line under `### ✨ Added` referencing the PR. This PR adds none — consider adding one (e.g. "Make scripture references inside footnotes clickable so they jump to the referenced passage. ([#1517](https://github.com/HelloAOLab/seed-bible/pull/1517))").

---

### Nice touches
- The `onReferenceClick`-optional design keeps `VerseReferenceText` reusable (the doc comment already mentions chat messages as a future consumer).
- Keyboard access is preserved for free — the links render as real `<a>` elements with an `href`, so focus + Enter works, and middle/cmd-click still opens the passage in a new tab.
- Good test coverage for both the multi-reference detection and the click → `openVerseReference` + close-footnote behavior.

---

Kabi4:
References in non-English translations are never detected. It works for english only as per parseVerseReferences

Kabi4:
Also i noticed other langauge don't have footnotes as of now

KallynGowdy:
Yeah for some reason I haven't been able to find a non-english translation that has verse references in footnotes. Additionally, I found an issue where references to 1 Corinthians weren't detected, but that appears to be an issue with `parseVerseReferences()` (which this PR doesn't touch).

image

Overall, this looks good to merge to me. I'll make another separate issue for `parseVerseReferences()`.

KallynGowdy (approved):

Kabi4:
can we use a translator to convert any language detected into English? Then
use this sparse reference to have the reference related and link those
reference to the relative places that will be easy because I don’t think we
should cover every language and every translation via code


Kushagra Singh Karki

KallynGowdy:
That will be yet another issue for improving `parseVerseReferences()`. It currently doesn't support multiple languages, so that change is out of scope for this pull request. Right now, English-only is fine

PR #1518, opened 2026-07-26
update reading plans

### LINKED ISSUE #1284: Implement reading plans
There should be designs for this in Figma (in progress section): https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=503-4846&t=Dz1RqRc0ZokfvKEa-1

mazenemad:
fixes #1284 

https://www.loom.com/share/d29f60b4c8f24728a0aa35754f965ac5?from_recorder=1&focus_title=1

image


Blitzy:
> [!NOTE]

## Code review — "update reading plans" (fixes #1284)

This PR builds out the reading‑plans feature: a create‑plan wizard, a plans list with "Today"/Active/Not‑started/Completed sections, a per‑day detail view, an in‑reader "This reading belongs to" card, and a verse‑toolbar "Add to plan" action. Nice, cohesive UX and the pure helpers (`summarizeCalendar`, `sessionMatchesPassage`, `estimateReadingMinutes`) are well‑documented.

CI `build (24.x)` is green, so this is not about type/lint failures — the notes below are logic and efficiency issues I found by reading the diff. Nothing here is a security or crash risk; the most serious are data‑integrity problems in the create flow.

---

### High

**H1 — Reducing the duration after adding readings can silently create an empty plan (and drops readings).**
`CreateReadingPlanWizard.tsx`

- `totalReadings` counts readings across **every** day key in `readingsByDay` (`Object.values(...).reduce(...)`, line ~99), but `handleCreate` only iterates `for (day = 0; day < durationDays; day++)` (line ~165). Any reading stored on a day index `>= durationDays` is counted toward the "can submit" gate but never written.
- Because the submit guard is `totalReadings === 0`, you can pass the guard while every actual reading lives on a now‑out‑of‑range day, producing a plan with **zero sessions**.

Reproduction:
1. On the schedule step set Duration = 30.
2. On the scripture step, select Day 20 and add a reading.
3. Go **Back** to the schedule step and set Duration = 5.
4. Return to scripture and press **Create plan**.

Result: the button is enabled (it thinks there is 1 reading), but the created plan has no sessions; the Day‑20 reading is silently discarded.

Suggested fix: when `durationDays` shrinks, prune `readingsByDay` keys `>= durationDays` (and clamp `selectedDay`), or compute `totalReadings`/the create loop over the same in‑range set so the gate and the write agree.

**H2 — A mid‑flight failure during create leaves an orphan plan, and retrying creates a duplicate.**
`CreateReadingPlanWizard.tsx` `handleCreate` + `ReadingPlansManager.tsx` `createNewReadingPlan` / `addSessionToReadingPlan`

`createNewReadingPlan` persists the plan record immediately (`await saveReadingPlan(plan)`), then the wizard appends sessions one at a time. If any `addSessionToReadingPlan` rejects (e.g. a network blip on day 10 of a longer plan), the catch shows the error and re‑enables the button — but the plan is already saved with a partial set of sessions. Pressing **Create plan** again calls `createNewReadingPlan` a second time, producing a **duplicate, orphaned** plan.

Reproduction: force a rejection on the 3rd `addSessionToReadingPlan` call → the record already exists with 2 sessions; retry → a second plan appears in the list.

Suggested fix: build all sessions locally and write the plan once (e.g. let `createNewReadingPlan` accept the full `sessions` array, or add a batch `addSessions`), so creation is a single atomic save. This also fixes M1 below. Failing that, remember the created plan on the error path and resume/clean up instead of re‑creating.

---

### Medium

**M1 — Creating a plan is O(days) sequential writes, and each one re‑loads *every* plan.**
`ReadingPlansManager.tsx` `addSessionToReadingPlan` + the new `syncFullReadingPlans` effect

- Each `addSessionToReadingPlan` re‑serializes and writes the **entire** growing plan twice (`saveReadingPlan` records both the full doc and the metadata doc). The wizard calls it once per day, so a D‑day plan performs ~D sequential double‑writes of ever‑larger payloads.
- Worse, each call replaces `userReadingPlans.value` with a new array, and the newly added `effect(() => void syncFullReadingPlans())` follows that signal — so **every** append triggers a `Promise.all` that reloads the full content of *all* of the user's plans over the network. Creating a 30‑day plan can fan out into ~30 full‑library reloads on top of the 30 writes.

Reproduction: create a plan with readings spread across many days and watch the network tab — you'll see a burst of PUTs plus repeated bulk GETs of all plans.

Suggested fix: the atomic single‑save from H2 removes most of this. Separately, consider debouncing/guarding `syncFullReadingPlans` (or updating the cache in place from the just‑saved plan) so a metadata mutation doesn't refetch the whole set.

**M2 — "Add to plan" mis‑saves a selection that spans more than one chapter.**
`BibleToolsManager.tsx` `add-to-reading-plan` → `onSelect`

The handler takes the chapter from the **first** selected verse (`first.chapterNumber`) but computes `lowVerse`/`highVerse` as the min/max verse number across **all** selected verses, regardless of their chapter. A comment says it assumes a single‑chapter selection, but nothing enforces it. For a cross‑chapter selection the saved `ref` is wrong.

Reproduction: select a range crossing a chapter boundary (e.g. Genesis 1:30 through 2:2) and add it to a plan → the reading is stored as roughly "Genesis 1:2–30" instead of the intended span.

Suggested fix: restrict the collapse to verses whose `chapterNumber === first.chapterNumber`, or build a proper cross‑chapter `ref` using `endChapter`/`endVerse`.

---

### Low

**L1 — "Restart" on a completed plan doesn't restart.**
`ReadingPlansPane.tsx` (completed section, ~line 321) — the **Restart** button just calls `onOpen(row.meta)`, opening the detail view. Since the plan already has progress, the detail view shows the day list, not a start/restart action, so the label overstates what happens. Either wire it to `startReadingPlan` (which already creates a fresh progress) or relabel it (e.g. "View").

**L2 — Cross‑time‑zone edge in `summarizeCalendar`.**
`ReadingPlansManager.tsx` `summarizeCalendar` compares `day.date` (documented as local midnight *in the plan's time zone*) against `nowStart = DateTime.fromMillis(nowMs).startOf("day")` (the **device** zone). For a user whose device zone differs from the plan's zone, `behind`/`streak` can be off by a day around midnight. Deriving `nowStart` in the same zone as the calendar days would make it robust.

**L3 — No unit tests for the new pure helpers.**
`summarizeCalendar` (non‑trivial streak/behind logic), `sessionMatchesPassage`, and `estimateReadingMinutes` are pure and easy to test, but `ReadingPlansManager.test.ts` doesn't cover them. The streak/behind edge cases (today pending, gaps, all‑complete) in particular would benefit from a few table‑driven cases.

---

### Questions / context

- Issue #1284 references an in‑progress Figma design and mazenemad asked on 2026‑07‑13 whether the design was final — that thread doesn't appear to be resolved yet. Worth confirming the wizard's flow (name → type → schedule → scripture) and the "belongs to" card match the intended design before merge.
- The `note` ("Reflect" card) and `planType`/`suggestedStartDateMs` fields are persisted and read in the detail view, but the wizard never lets a user set a `note`. If that's intentionally author‑only for now, ignore this.

Happy to dig deeper on any of these.

---

KallynGowdy:
@mazenemad In addition to the Claude review notes, I would add these:

1. The "Add to plans" button should only be shown when the user is editing a plan and it should say "Add to plan". Clicking it should add the selection to the plan that is being edited.
2. Users should be able to add all types of playlist items to a reading plan (not just scripture, but also links and text just like playlists)
3. The "This reading belongs to" section shouldn't be displayed over the scripture text on mobile, Instead, it should be displayed after the user scrolls all the way to the bottom of the page.

KallynGowdy:
@mazenemad  Here's a bunch of feedback

- When viewing a reading plan:
    1. The back button, title, and create button should be in the header of the pane 
       image
    2. The icon should be the same as the one used in the reading plan list
        not this one:
        image
        this one:
       image
    3. The "Start Plan" button can be hidden by the toolbar
        image
        image
- When creating a reading plan
    1. The "Next" button can be hidden by the toolbar
        image
    2. There should not be any "Reading plan types". Reading plans are just a list of reading sessions and suggested cadences. A user can choose to read any reading plan however they want (flexible/at their own pace or using a suggested cadence). One reading plan might have multiple paces for reading it (e.g. The Bible in a Year vs The Bible in Two years). This means that when a users goes to start a reading plan, then they are presented with how they want to read it (e.g. choose one of these cadences or go at your own pace). It also means that there's no single duration for a reading plan. Instead, it should be calculated based on the selected cadence.
    3. There should not be a "Select duration, start & date" page. Reading plans can be started at any time and have any duration (as mentioned above). Instead, there should be a "Reading cadences" page that the user can use to setup the available cadences for the reading plan. There should be some default options that can be checked to include them in the reading plan, such as:
        - one session a day
        - two sessions a day
        - three sessions a day
        - one session every other day
    4. In the "Select readings" page:
        - "days" should be called "sessions" and should be a vertical list instead of a horizontal list.
        - Users should be able to add and remove sessions
    5. We should automatically save the reading plan to the user's account with a status property set to "draft". When they finish all the steps, we can set the status to "complete". We should save whenever the user changes the reading plan so that they don't lose any progress if they leave for any reason (e.g. save when they add a reading, save when they change available cadences, etc.)

Blitzy:
> [!NOTE]

## Code review — "update reading plans" (fixes #1284), head `e600425`

This follow-up covers the current head commit, which reworks the feature per @KallynGowdy's redesign feedback: cadences instead of a fixed duration, sessions instead of days, and autosaved drafts. Good news first — every issue flagged in the earlier review on this PR is now fixed, and all seven of the maintainer's redesign requests are addressed, with one caveat below. One new bug turned up in the refactor.

### Previous findings — now fixed

- Duration/day mismatch that could silently create an empty plan: fixed — sessions are now a flat array with duration derived from `cadenceDurationDays(cadence, sessionCount)`, so there's no separate duration value left to drift out of sync.
- Orphan/duplicate plan on a mid-create failure: fixed — drafts upsert to the same address on every autosave, and `finishEditingReadingPlan` chains after the in-flight autosave, so a retry overwrites rather than duplicates.
- O(days) sequential writes plus a full-library refetch on every append: fixed — plan creation is a single `saveReadingPlan` call, and mutations seed the `fullReadingPlans`/`userReadingPlans` caches in the same `batch()` so `syncFullReadingPlans` sees a cache hit instead of refetching everything.
- Cross-chapter "add to plan" mis-saving the verse span: fixed — `BibleToolsManager.tsx` now sorts selected verses by `(chapterNumber, verse.number)` before deriving the start/end and emits `endChapter`/`endVerse` for spans crossing a chapter.
- "Restart" on a completed plan not actually restarting: fixed — it now calls `startReadingPlan`, which creates a fresh `ReadingPlanProgress`.
- Timezone mismatch in `summarizeCalendar` ("today" computed in the device zone instead of the plan's zone): fixed, with a Tokyo/LA test case.
- No unit tests for the pure helpers: fixed — `summarizeCalendar`, `sessionMatchesPassage`, and `estimateReadingMinutes` each have dedicated test blocks, plus new coverage for cadence math, session add/remove, and autosave.

Nice work closing all of that out.

### New issue found in this revision

**Medium — removing an earlier session can silently redirect new readings into the wrong session**
`packages/seed-bible/seed-bible/managers/ReadingPlansManager.tsx`, `removeSessionFromEditingPlan` (~line 1690)

```js
selectedSessionIndex: Math.min(
  current.selectedSessionIndex,
  sessions.length - 1
)
```

This clamps the *old* selected index into the new array's bounds, but doesn't account for the fact that removing a session **before** the selected one shifts every later session down by one.

Reproduction:
1. Create a draft with 4 sessions (A, B, C, D).
2. Select session C (index 2, shown as "Session 3").
3. Remove session A (index 0). The array becomes [B, C, D] (length 3).
4. `selectedSessionIndex` is computed as `min(2, 2) = 2`, which now points at D — but the UI still shows "Session 3" selected, with no visual cue that anything changed.
5. Add a reading. It's added to D's session instead of the C session the user was looking at.

This is a new bug from the session add/remove feature — the existing removal tests only cover removing the selected session itself or the sole remaining session, so this path isn't exercised.

Suggested fix:
```js
let next = current.selectedSessionIndex;
if (index < next) next -= 1;
next = Math.min(Math.max(0, next), sessions.length - 1);
```

### Low / informational

- **`saveReadingPlan` writes the full plan and its `_metadata` companion via `Promise.all`** (`ReadingPlansManager.tsx` ~line 1129) — if one write succeeds and the other rejects, the two records can diverge (e.g. the plan says "complete" but the metadata list still shows "draft"). This pattern predates this PR, but autosave now calls `saveReadingPlan` far more often, raising the odds of hitting it. It's self-healing on the draft path since the next autosave overwrites both records, so this is low priority — just worth keeping in mind if it shows up in the wild.
- **`createNewReadingPlan` and `addSessionToReadingPlan` are now dead code** — both are still exported and unit-tested, but no component calls them anymore; the wizard exclusively uses the draft path (`startEditingReadingPlan` → `mutateDraft` → `finishEditingReadingPlan`). Worth removing in a cleanup pass rather than carrying two parallel, untested-in-practice creation paths.

### Maintainer redesign requests — status

1. "Add to plan" shown only while editing a plan, and adds to that plan — addressed.
2. All playlist item types (not just scripture) addable to a plan — addressed.
3. "This reading belongs to" card no longer overlays scripture text on mobile — addressed.
4. No more plan "types"; cadences instead of a fixed duration — addressed.
5. Duration/start-date step removed, replaced with a "Reading cadences" step — addressed.
6. "Days" renamed to "sessions," shown as a vertical list, addable/removable — addressed, but see the Medium finding above: removal has a selection-tracking bug.
7. Draft autosaved on every change, flips to "complete" on finish — addressed.

No i18n violations (only `en.json` touched), no new `any` usage, and no security concerns turned up in this pass.


---

Blitzy:
## Reading plans — hands-on review

I went through the plans feature on the deploy preview mostly cold, as a user would. The shape is good and the redesign feedback from earlier in this thread clearly landed — cadences, drafts, and the vertical session list all work.

Here's what broke or got in my way, roughly in priority order. Code-level detail is tucked into the collapsed blocks so this stays skimmable.

| # | Issue | Kind |
|---|---|---|
| 1 | "Mark as complete" in the reader completes the whole session, not the chapter you read | 🔴 Bug |
| 2 | Multi-chapter readings have no per-chapter progress | 🟠 Gap |
| 3 | You can't open scripture from inside a plan | 🟠 Gap |
| 4 | No way to edit or delete a plan once it's created | 🟠 Gap |
| 5 | A plan that fails to load lands you on a blank "Untitled plan" | 🔴 Bug |
| 6 | Opening a plan is slow and gives no feedback | 🟡 Polish |
| 7 | Checkbox placement differs between reading types | 🟡 Polish |
| 8 | "Session" collides with our shared sessions | 🔵 Decision |
| 9 | The cadence step's wording doesn't explain what it's for | 🟡 Polish |
| 10 | The create wizard could be a single-pane editor | 🔵 Direction |

---

### 🔴 1. "Mark as complete" completes far more than you read

My plan's first reading is John 1–10. I read to the bottom of **John 4**, where the "This reading belongs to" card appears, and tapped the plan row.

**What happened:** the entire reading was marked done — all of John 1–10, *plus* the unrelated text and link readings sitting in the same session. I'd expect credit for John 4 and nothing else.

A reader can silently lose 9 chapters of a plan by tapping one button on the chapter they actually read, which is why I'd call this a correctness bug rather than a rough edge.

**Suggested direction:** track completion per chapter for scripture readings (see item 2), and have the card complete only the chapter in view.

<details><summary>Where it is in the code</summary>

`ReadingPlanBelongsCard.tsx:155-171` — both the row toggle and the "Mark as complete" button call `setSessionCompleteForProgress` for every session matching the current passage, and that marks the whole session complete (`ReadingPlansManager.tsx:1343`). There's no notion of "this chapter, out of the range."

Same code path is why the session's text and link readings get marked too — and those aren't even reachable from the reader.

</details>

### 🟠 2. Multi-chapter readings need per-chapter granularity

A reading of John 1–10 is a single atomic checkbox today. That's what makes item 1 possible, and it also means a reader who is 6 chapters in has no way to record it.

**The playlist code already solves this.** `PlaylistManager.tsx:145` has `expandCrossChapterItem`, which splits a `bible-verse` item spanning chapters into one item per chapter — specifically so playback "advances chapter by chapter instead of only ever visiting the first chapter." Reading plans should reuse it (or the same idea) when building the calendar, so progress is stored per chapter.

### 🟠 3. You can't get to scripture from inside a plan

In an active plan, tapping a scripture reading only toggles its checkbox — there's no way to actually go read it. Text and link readings open in a modal; scripture does nothing.

This is the most basic thing a reading plan needs to do: take me to the reading. Right now the only path to the passage is to navigate there manually and hope the "belongs to" card shows up.

Two things I'd want:

1. **Tapping a scripture reading navigates to it** — for a range, to the first chapter you haven't completed, not always chapter 1.
2. **Playlist-style next/back stepping** through the plan's readings. `PlaylistManager` already has `next` / `previous` / `jumpTo` (`PlaylistManager.tsx:280-302`) driving exactly this over an expanded queue. Talking with @KallynGowdy, we both landed on the same conclusion: a reading plan is essentially a chunked playlist with tracking, and should reuse the playlist's playback controls rather than grow a parallel implementation.

<details><summary>Where it is in the code</summary>

`ReadingPlanDetail.tsx:429` — `canPreviewPlaylistItem` returns false for `bible-verse` by design (scripture is meant to be read in the reader, not previewed in a modal), and nothing else handles the click, so the card falls through to being a pure toggle.

</details>

### 🟠 4. No way to edit or delete a plan once it's created

Once a plan leaves draft status there is no edit and no delete anywhere in the UI. The only affordances on a finished plan are "open" and, if completed, "Restart." I created two test plans and had to go into the auth portal and erase the records by hand to get rid of them.

Whether editing means edit-in-place or duplicate-then-edit is worth a decision — in-place changes a plan other people may be reading, which matters once plans are shareable. But **delete should just exist.**

<details><summary>Where it is in the code</summary>

The manager is already most of the way there:

- `deleteReadingPlan` exists, but is only wired to draft cards (`ReadingPlansPane.tsx:424`).
- `canEditSelectedPlan` is computed and exported (`ReadingPlansManager.tsx:1079`) but no component ever reads it — the permission check for editing was anticipated and then never used.
- Completed plans get "Restart" only (`ReadingPlansPane.tsx:539-545`); active and not-started plans get nothing.

</details>

### 🔴 5. A plan that fails to load shows a blank "Untitled plan"

I deleted a plan's record but left its metadata — which is exactly what a partial write or a stale list would look like in the wild. Opening it dropped me on an empty screen titled "Untitled plan," with no indication anything went wrong.

Failure should keep you on the list and show an error ("Couldn't load this plan"), not navigate you to a dead screen.

<details><summary>Where it is in the code</summary>

`selectReadingPlan` catches the error, logs to console, and leaves `selectedReadingPlan` at `null` (`ReadingPlansManager.tsx:1258`) — but the view still flips to `detail`, so the header falls back to "Untitled plan" (`ReadingPlansPane.tsx:128`) and the body renders nothing (`ReadingPlanDetail.tsx:74`).

</details>

### 🟡 6. Opening a plan is slow and gives no feedback

Tapping a plan (or the "Today" resume banner) hangs for a noticeable beat with no spinner, no disabled state, nothing. Either fix would help, ideally both:

- Render from the cached full plan immediately and revalidate in the background.
- Show a pending state on the tapped card while it loads.

<details><summary>Where it is in the code</summary>

`ReadingPlansPane.tsx:59-71` awaits `selectReadingPlan` before switching the view, and `selectReadingPlan` (`ReadingPlansManager.tsx:1249-1261`) always does a fresh `os.getData` round-trip — even though the full plan is usually already in memory in `fullReadingPlans`, which the same list component already has on hand as `fullById`.

</details>

### 🟡 7. Inconsistent checkbox placement between reading types

In the plan detail, scripture readings put the radio/check **inside** the card, while text and link readings put it **outside**, in the left margin.

They should look the same. My preference is the text/link style — check on the outside left, card as its own target — applied to scripture too. That also falls out naturally if scripture readings become tappable per item 3.

<details><summary>Where it is in the code</summary>

`ReadingPlanDetail.tsx:486-534`. The split exists for a reason: a scripture card is one big toggle target, while text/link cards need the card itself to open the content. Making scripture navigable removes the reason.

</details>

### 🔵 8. "Session" collides with shared sessions — a decision, not just a rename

Throughout the plan UI, "session" means a chunk of reading. Everywhere else in Seed Bible it means a shared multiplayer session — "Share session," "End Session," "New shared session," and a dozen more keys in `en.json`. A user reading "Three sessions a day" or "Add session" has no way to know these are unrelated concepts.

My preference is to reserve "session" for shared sessions and call these **readings** — you build a plan by breaking it into readings.

> [!NOTE]
> @KallynGowdy specifically asked for days → "sessions" earlier in this thread ([comment](https://github.com/HelloAOLab/seed-bible/pull/1518#issuecomment-5119213989), item 4), and that's what shipped in this revision. So this needs the two of us to agree before @mazenemad changes anything.

**Cost of the change, if we make it:** the user-facing half is cheap (~20 keys in `en.json`, lines 271-297). The data model half is not — `ReadingPlanSession`, `plan.sessions`, and `sessionId` inside persisted progress records are already live in users' accounts, so renaming those is a migration. Renaming just the copy and leaving the schema alone is fine, but if we ever want them aligned, doing it before more plans exist in the wild is much cheaper.

### 🟡 9. The cadence step's wording doesn't explain what it's for

The "Reading cadences" step says (`en.json:270`):

> Pick the paces readers can choose from. They can also read at their own pace.

Coming in cold I read that as a contradiction — "at their own pace" isn't one of the four checkboxes in front of me. It's actually saying self-pacing is always available and I don't need to provide it. It also wasn't obvious that this step is me authoring options *for other readers*, not choosing my own pace. Something closer to:

> Choose the reading paces you want to offer. Readers can always choose to go at their own pace instead.

Two smaller things in the same area:

- Being asked for cadence *immediately* after naming the plan is backwards — I don't know what the plan contains yet, so I can't judge the paces. It belongs after the readings.
- With a one-reading plan, "Three sessions a day" gives "Finishes in 1 day." Correct, slightly odd, fine for now.

### 🔵 10. I don't think this needs to be a wizard

Three steps with a progress bar sets an expectation of complexity that isn't there — a plan is a title, some cadences, and a list of readings, and the list is 95% of the work. Meanwhile the wizard actively costs us the thing I most want in item 4: a wizard is a create-only flow, so there's no natural home for editing an existing plan.

Playlists already have the right pattern — one pane where the title and the items are edited together (`CreatePlaylistForm`). I'd collapse the plan editor into the same shape: name, cadences, and readings on one screen. Then "create" and "edit" become the same screen, which makes item 4 nearly free.

This is a bigger change than the rest of the list and may be better as a follow-up than a change to this PR. I'm flagging it now because the wizard is the reason several of these gaps exist.

---

### Minor notes

- **Reading time estimates are flat.** `estimateReadingMinutes` (`ReadingPlansManager.tsx:967`) charges 3 minutes per chapter regardless of length, so John 1–10 shows as 30 minutes. Fine as a first pass, but it'll read as wrong for long chapters (Psalm 119 vs. Psalm 117). A verse- or word-count basis would be more honest if we keep the number visible.
- **The record layout looks right.** I poked at the stored data in the auth portal: a `readingPlan` record, a `readingPlan_metadata` companion, and separate progress records. Splitting metadata out so the list renders without loading every plan's contents is the right call — which makes it more frustrating that opening a plan refetches it anyway (item 6).

KallynGowdy:
A couple more things: 

1. The "Read" button is obscured by the toolbar on mobile. 


2. When reading a session via the read button, navigating to the last chapter in the session doesn't load it.
https://github.com/user-attachments/assets/ea67a905-67be-47ae-bf85-6925c047ecfa

3. Sometimes navigation is completely broken when trying to read a reading plan session.

See this video: https://youtube.com/shorts/IfX1sukXYaw?is=DHduYRiznNua7poS

KallynGowdy (dismissed):
This looks really good now

Blitzy (dismissed):
Agreed, this is in pretty good shape.

PR #1524, opened 2026-07-26
feat: implement tab state persistence and restoration on reload

### LINKED ISSUE #1383: Save the user's current tabs to local storage
Currently, when a user refreshes the screen, their tabs are recreated from the initial ones. This task is to try to preserve their current tab state in local storage so that it is preserved when the refresh or revisit seedbible.org. 

It should follow the following logic:

- Save the current tabs state in local storage (Translation, book ID, chapter number, selected tab, and tabs layout - omit ephemeral values like highlights, decorations, selected verses, etc)
- On initial load, load the tabs state like this:
  1. First, check if tabs were stored in local storage and load them.
  2. Second, check the query parameters for reading state (translation, book ID, chapter number) and compare them with the loaded tabs:
      1. If no query parameters were specified, then do nothing and keep the loaded tabs state.
      2. Otherwise, find the first tab that matches the query translation (or the default one if none was specified in the query). If it exists then update it to match the query parameters and select it.
      3. If there is no tab that matches the query translation, then
          1. If there is only one saved tab, update the saved tab to match the query.
          2. Otherwise, create a new tab matching the query and select it.

mazenemad:
fixes #1383


Blitzy:
> [!NOTE]

## Code Review — Tab state persistence and restoration on reload

Reviewed against the spec in #1383 (and Blitzy's comment there about always being thrown back to Genesis 1). Overall this is a careful, well-documented implementation that follows the issue's restore algorithm faithfully. The split into a pure `TabsPersistence` module with strong runtime validation, the versioned storage key, the "skip redundant writes" guard, and the id-normalization to keep `TabsManager`/`TabsLayoutManager` in agreement are all nicely done. `pnpm check:ts` is clean and the new tests pass (54 tests across the two files).

I verified a few things that turned out **not** to be problems, so reviewers can skip them:
- **SSR/localStorage divergence is a non-issue** — the client `render`s rather than `hydrate`s (`app/init.tsx`), so the server markup is discarded and there's no hydration mismatch from restoring a different tab count on the client.
- **Selection/slot mismatch after reconcile self-heals** — when a deep link selects a tab that the restored `selectedSlotIndex` doesn't point at, the existing "keep selection and slots aligned" effect in `SeedBibleStateManager` (around line 859) re-selects the correct slot.
- **No conflict with an existing reading-position restore** — there is no profile/`lastReadingLocation`-based tab-position restore that would fight the localStorage restore.

Nothing here rises to a blocking (High) severity. Findings below are Medium and Low.

---

### Medium

**No. 1 — The most complex new logic (multi-slot layout restoration) has no direct test coverage.**
`TabsPersistence.test.ts` thoroughly covers `reconcileStoredTabs`, `normalizeStoredTabsState`, and read/write. But the restore branch in `TabsLayoutManager.createTabsLayout` (`canRestoreLayout`, rebuilding `slots` from `slotTabIds`, the `isSelectedShown` injection of a freshly-appended selected tab, and `removeUnreferencedSlotOnlyTabs`) is entirely untested. That's the intricate part where a regression is most likely to slip in unnoticed.

Suggested coverage (unit tests on `createTabsLayout` with a mock `TabsManager`, seeding `localStorage["sb-tabs-state"]`):
- A stored `split-2v` with two visible tabs restores two slots referencing the right tabs and the right `selectedSlotId`.
- A stored layout whose `slotCount` disagrees with `slotTabIds.length` (corrupt) falls back to a single slot without throwing.
- Restore with `panelsEnabled = false` collapses to a single slot and disposes the `slotOnly` clones (assert `removeTab` was called for the orphans).
- A `slotOnly` clone referenced by a restored slot is **kept**, while an unreferenced one is removed.

---

### Low

**No. 2 — Debug `console.log` runs on every app load.**
`TabsManager` logs `"Restoring TabsManager from stored state:"` (new) and still logs `"Creating TabsManager with initial URL parameters:"` (pre-existing) on every cold load, including in production. This is console noise and leaks reading state into the log.
*Suggested fix:* remove these or gate them behind an existing debug flag.

**No. 3 — A split layout built with panels enabled is silently flattened once the app is opened with panels disabled.**
When `panelsEnabled` is false (e.g. `disablePanels`), `canRestoreLayout` is false, so restore collapses to a single slot and the persistence effect then rewrites storage as `layout: "single"` with one slot. Returning to a panels-enabled context won't bring the split back — it's been overwritten.
*Reproduction:* on desktop, open two panes (`split-2v`) → enable "disable panels" (or open on a device where panels are off) and reload → the split is gone → re-enable panels and reload → still a single pane.
This is a reasonable limitation rather than a defect, but worth a decision: either accept and document it, or avoid overwriting the stored multi-slot layout while panels are disabled (persist the layout, just don't render it).

**No. 4 — A deep link with `book`/`chapter` but no translation retargets the *default*-translation tab, not the tab you were reading.**
Because `targetTranslation = query.translationId ?? defaultTranslationId`, a link like `?book=JHN&chapter=3` matches the first tab whose translation equals the default. If you were last reading a non-default-translation tab, restore switches you to (and mutates) a different tab.
*Reproduction:* save two tabs — Tab A = default translation (AAB), Tab B = NIV — while selected on Tab B → open `/?book=JHN&chapter=3` (no `translation`) → you land on Tab A (AAB) at John 3 rather than Tab B at John 3.
This is **exactly what the issue's step 2.2 specifies** ("...or the default one if none was specified in the query"), so it's correct-as-designed — flagging only to confirm the UX is intended, since a translation-less share link changing which translation you're reading can surprise users.

---

### Nit

**No. 5 — Inconsistent URL source for two startup reads.** `buildRestoredTab` reads `?verse=` highlights from `navigation.currentUrl.value`, while the reconcile query params are read from `navigation.initialUrl`. They're almost certainly equivalent at cold start, but reading both from `initialUrl` (the frozen pre-echo snapshot, as the code comment already argues for the query params) would be more obviously correct and consistent.

---

packages/seed-bible/seed-bible/managers/TabsPersistence.tsx
```diff
@@ -0,0 +1,286 @@
...
+    state.tabs.every(isPersistedTab) &&
+    typeof state.selectedTabId === "string" &&
+    typeof state.layout === "string" &&
+    Array.isArray(state.slotTabIds) &&
+    state.slotTabIds.every((id) => id === null || typeof id === "string") &&
+    (state.selectedSlotIndex === null ||
+      (typeof state.selectedSlotIndex === "number" &&
+        Number.isFinite(state.selectedSlotIndex)))
+  );
+}
```
KallynGowdy: @mazenemad Can you refactor these functions to use zod schemas instead of manually verifying?

packages/seed-bible/seed-bible/managers/TabsPersistence.tsx
```diff
@@ -0,0 +1,286 @@
...
+  }
+  try {
+    const raw = window.localStorage.getItem(TABS_STORAGE_KEY);
+    if (!raw) {
+      return null;
+    }
+    const parsed: unknown = JSON.parse(raw);
+    if (!isPersistedTabsState(parsed)) {
+      return null;
+    }
```
KallynGowdy: @mazenemad Can this use zod's schema.safeParse() instead of checking `isPersistedTabsState()`?

KallynGowdy:
@mazenemad Can you fix the things I commented from my review?

mazenemad:
@KallynGowdy I don't see any review comments on the PR?

KallynGowdy:
@mazenemad Sorry I must have forgotten to submit them. You should see them now

KallynGowdy (dismissed):

KallynGowdy:
@mazenemad can you fix the merge conflicts for this?

mazenemad:
On it

PR #1526, opened 2026-07-27
Feature/1392

### LINKED ISSUE #1392: Scripture map: use a masonry layout for books
## What is happening

The Scripture map's book layout is a rigid grid. Because books have different heights (number of chapters), the grid leaves a lot of empty space between books on different rows.

## What to try

Move to a masonry layout, where books pack together based on their actual height instead of being locked to uniform grid rows. This should remove the empty space between books and make the layout feel less rigid.

> [!NOTE]
> This is somewhat experimental in concept. We want to improve the look and feel of the scripture map and this is just one idea.

Image
Image

Kabi4:
image
Fixes #1392


Blitzy:
> [!NOTE]

## Code review — PR #1526 (Feature/1392)

Reviewed against the goal in issue #1392: move the Scripture map's rigid book grid to a **masonry** layout so books pack by their real height and remove the empty gaps between rows. The approach here — a `scripture-map-books-container-masonry` modifier that switches the container to CSS multi-column (`column-width` / `column-gap` with `break-inside: avoid` on each `.book-container`) — is a reasonable, dependency-free way to get that effect, and the packing goal is met. The notes below are about the details.

The diff is small (2 files, +30/-3): a `masonry` prop on `BooksContainer` and the matching CSS.

---

### High

**No. 1 — The "toggle" is never wired to anything; masonry is effectively always-on.**
The PR (and commit "Added boolean to toggle mansory layout") frames this as a toggle, but both render sites pass no prop, so `masonry` always falls back to its `true` default:

- `packages/scripture-map/components/containers/TestamentContent.tsx:61` — `<BooksContainer key={...}>`
- `packages/scripture-map/components/containers/Container.tsx:24` — `<BooksContainer>`

Nothing in the diff reads a setting, context value, or user preference to drive the prop, so there is no way — for a user or for the code — to turn masonry off. Given issue #1392 explicitly calls this "somewhat experimental," a real on/off switch is probably what's wanted.

Suggested fix — pick one:
- **If a toggle is intended:** thread a flag from `ScriptureMapContext` (where `showTestamentLabels` / `bookWidth` etc. already live) down to both call sites and pass it as `masonry={...}`, so it can be flipped from a setting or feature flag.
- **If it should always be on for now:** drop the prop entirely and apply the `scripture-map-books-container-masonry` class unconditionally. That removes a dead parameter and makes the intent obvious, and the toggle can be reintroduced when there's something to bind it to.

---

### Medium

**No. 2 — CSS multi-column reorders books from row-major to column-major; confirm canonical order still reads correctly.**
With `column-width`, content flows top-to-bottom down column 1, then continues at the top of column 2. So Genesis, Exodus, Leviticus… now read **down** the first column rather than **across** the first row. In a Bible app, canonical book order is meaningful, so this is a deliberate behavior change worth a conscious sign-off. The mock-ups in #1392 look consistent with column-major flow, so this may be exactly what's wanted — but note that true masonry (CSS Grid `masonry`, or a JS masonry lib) preserves left-to-right reading order, whereas CSS columns does not. Please confirm against the intended reading order (and RTL locales, if the map is shown in those).

Reproduction: open the Scripture map at a width that shows 3+ columns and read the book sequence — it will run down-then-across rather than across-then-down.

**No. 3 — Columns grow wider than `--book-width`, stretching each book box.**
`column-width: var(--book-width)` is a *minimum*, not a fixed width — the browser widens the columns to fill the container. Combined with `.book-container { width: 100% }` (styles.css:511-512), each bordered book box stretches to that wider column while the chapter grid inside stays a fixed size and centered (`align-items: center`), which can leave extra horizontal space inside each book's border compared to the current fixed-width grid.

Reproduction: open the Scripture map at a viewport width that is **not** an exact multiple of `--book-width` and compare book-box width against `develop` — the boxes should look wider with the cover floating in the middle. If that's undesired, cap the box (e.g. keep the border/background on an inner element sized to `--book-width`, or constrain `.book-container` `max-width`) rather than letting it fill the column. Worth a quick visual check against the design either way.

---

### Low

**No. 4 — No CHANGELOG entry.**
`CHANGELOG.md` is actively maintained in this repo (several recent `chore: Updated CHANGELOG` commits on `develop`). Consider adding a short line for this change to match convention.

**No. 5 — Selector-scoping is asymmetric.**
The masonry container rule is scoped under `.testament-content` (styles.css:503), but the masonry `.book-container` rule is not (styles.css:509). Both current call sites sit inside `.testament-content`, so this works today. If `BooksContainer` with `masonry` is ever reused outside `.testament-content`, it would keep the class but lose the `display:block; column-width` styling and render as the default flex container. Scoping both the same way (or neither) avoids that latent surprise. Also minor: the base `.testament-content .scripture-map-books-container` flex declarations (`flex-wrap`, `justify-content`, `align-items`) still apply to the masonry container as inert leftovers — harmless, but could be tidied.

---

### Summary

The masonry effect works and matches the issue's intent. The one thing I'd resolve before merge is **No. 1** — as written there's no way to toggle the layout the PR is named for, so either wire the prop to a real setting/context or drop it and apply the class unconditionally. **No. 2** and **No. 3** are visual/behavioral confirmations best checked on the deploy preview (https://alpha.seedbible.org/b/feature_1392) against the design and the desired book order. The rest are polish.

---

Kabi4:
image

Blitzy:
@Kabi4 I'm not really sure what happened but the books look a bit worse no then they did originally.

The primary issue seems to be padding? The books are thinner than they used to be and the chapter buttons are now right up against the outer edges of the books with no padding at all. Also the dropdown arrow for some books with longer names (like Deuteronomy) now are touching the edge.

image

Here is what it looks like before. Notice the chapter buttons have padding from the outer edge of the books. The book covers when closed dont have the chapter buttons sticking out.

image

The other thing i noticed was that the ordering of the books seems almost random?

image

Here is the ordering when the books originally:

image

Kabi4:
The wrong order came from shortest-column packing.
Fixed now

image

Blitzy:
> [!NOTE]

## Code review — PR #1526 (Feature/1392), re-reviewed at "ready for review"

This re-reviews the current head (`dbb8267`) against issue #1392 ("use a masonry layout for books") and the discussion that happened since the last review — including the visual regression flagged in the thread (thinner books, missing padding, near-random order) and the fix pushed for it.

**Good news: both concerns raised in that thread are resolved in the current code.**

- **Reading order.** The layout now assigns `column = index % columnCount` (`useMasonryLayout.tsx`), which is a strict round-robin across columns — Genesis, Exodus, Leviticus… still flow left-to-right in canonical order, they just stack per-column by real height. I checked this against the current diff, not just the "fixed now" screenshot.
- **Thinner books / missing padding / chapter buttons touching the edges.** The whole approach changed from CSS `column-width` (which forced `.book-container` to stretch to fill a column) to a JS-computed `columnWidth` with `width: max-content`. I pulled the current `styles.css` from the `feature/1392` branch and confirmed the base `.book-container` rule (`padding: calc(8px * var(--scale-factor))`, border, etc.) is untouched by this PR — nothing in the diff removes that padding, so this class of regression shouldn't be present in the current head.

I also ran the existing suite against this branch: `BooksContainer`, `Container`, and `TestamentContent` tests (22 tests) pass, `pnpm check:ts` is clean, and `pnpm eslint` on the four changed/added files reports nothing.

---

### Medium

**No. 1 — New masonry layout algorithm has no test coverage.**
`useMasonryLayout.tsx` is 168 lines of real layout logic (column-count derivation from container width, gap scaling via `--scale-factor`, a memoized "signature" to skip redundant re-layouts, ResizeObserver + MutationObserver wiring, and a no-ResizeObserver fallback). Every other hook in `packages/scripture-map/hooks/` has a matching file under `test/unit/scripture-map/hooks/` (e.g. `useResizeObserver.test.tsx`, `useContainer.test.tsx`) — this one doesn't, and `BooksContainer.test.tsx` wasn't extended to cover `masonry={true}` either (it still only exercises the default/no-masonry path).

Worth at least covering: column-count derivation at a few container widths, that round-robin column assignment preserves item order, and that disabling `masonry` clears the inline styles it applied (the `clearItemStyles`/`clearContainerStyles` path).

---

### Low

**No. 2 — The masonry "toggle" still isn't wired to anything (carried over from the earlier review).**
Both call sites now literally write `<BooksContainer masonry>` (`Container.tsx:24`, `TestamentContent.tsx:61`) — a hardcoded `true`, not a real toggle. That's a valid way to resolve the earlier note ("if it should always be on for now, drop the prop and apply the class unconditionally"), except the prop — and the whole disable/cleanup branch in `useMasonryLayout` — is still there, just never exercised. Since issue #1392 explicitly calls this layout "somewhat experimental," it'd be worth either actually wiring it to a setting/flag so it can be turned off if the experiment doesn't pan out, or dropping the now-dead `masonry` prop and always-class-on branch to keep the code honest about there being no toggle today. Not blocking either way.

**No. 3 — Still no CHANGELOG entry.**
`CHANGELOG.md` is untouched across all 6 commits on this branch; this repo's convention (per recent `develop` history) is to add a line under the `## TBD` section for user-facing changes like this one.

**No. 4 — Layout runs before webfonts settle.**
`useMasonryLayout` measures `offsetHeight`/`offsetWidth` in a `useLayoutEffect`, which can run before `document.fonts.ready` resolves. If a custom font shifts a book's title/chapter-grid height after the initial pack, the masonry columns won't reflow until something else (a resize, a DOM mutation) triggers it. Minor and cosmetic — likely only affects the very first paint — but worth an eye if book covers ever look slightly misaligned right after a cold load.

---

### Verdict

The two things that mattered most from the earlier round of feedback — reading order and the box-stretching/padding regression — are fixed in the current head, and the change is otherwise well-isolated (no other components affected, tests/types/lint all green locally). I'd like to see test coverage for the new masonry algorithm (No. 1) before merge given how much logic it carries, but I don't see anything that should block shipping the visual behavior itself.

---

Blitzy (dismissed):
I think this looks good now. @KallynGowdy Since the masonry layout idea originally came from you, what do you think? I do think its visually more pleasing seeing the books backed together like this.

KallynGowdy (commented):
looks good to me as well, however this section does look a bit strange to me:

Image
(also there's a bug where Nahum isn't rendered correctly)

I wonder if we grouped them by their traditional groupings if that would help make it look nicer? (Torah, Songs, Wisdom, Major Prophets, Minor Prophets, etc.). The problem with the masonry layout is that while it is more space efficient, the order can be difficult to discern, especially when there's a mix of large items and small items.

KallynGowdy:
Actually I think that is our solution. We need to group books beyond just Old Testament and New Testament. This way, we can do the masonry layout on the groups instead of on individual books. This way, we're able to keep some of the "map" idea while also being space efficient. In this case for the map, the order of each book doesn't matter as much as long as it remains easy to find them (books near each other remain near each other).

I think this is good to merge, and we can write up something else for this next idea

Blitzy (approved):

PR #1531, opened 2026-07-28
fix: enhance chapter load error handling with user-friendly retry option

### LINKED ISSUE #1377: Improve the display for when a chapter is unable to be fetched over the network
Currently, when a chapter fails to load, we just display an ugly "Failed to fetch" error message in bright red colors:

Image

Instead, we should display a nice looking error message along with a button giving the user the option to try loading the chapter data again. Something like: 

Icon: `cloud_off`
Header: "Chapter unavailable"
Sub header: "We were unable to load the data for this chapter. Please check your internet connection and try again"
Button: "Reload"

mazenemad:
fixes #1377


Blitzy:
> [!NOTE]

## Code Review — PR No. 1531: Chapter load error handling with retry

### Summary

This cleanly implements what issue #1377 asked for: instead of a raw red `Failed to fetch` string, a failed chapter load now shows a centered `cloud_off` panel with a **"Chapter unavailable"** heading, a friendly description, and a **Reload** button that re-runs the exact operation that failed. The design matches the issue spec (icon, header, sub-header, and button copy all line up).

The approach is sound. `BibleReadingManager` records the last load operation in a per-instance `lastLoadAttempt` closure, `retryLoad()` replays it, and the component keeps the panel on screen while the retry is in flight (`retrying` state) so it doesn't flash the empty body. Nice touches worth calling out: `role="alert"`, `aria-hidden` on decorative icons, `aria-busy` on the button, a `prefers-reduced-motion` guard on the spinner, `rem`-based sizing with a comment explaining why it tracks UI scale rather than the reader glyph knob, and translation keys added to `en.json` only (per the repo convention).

I ran the two affected unit test files locally — **123 tests pass**, including the new coverage for the retryable failure state, the in-flight panel persistence, and both `retryLoad()` manager paths (chapter-selection replay and initial-load replay). The `build (24.x)` CI job was still in progress at review time; I could not confirm the full suite / lint / typecheck are green, only these two files.

Nothing here is blocking. Findings below are low severity / minor.

---

### Low severity

**L1 — Retrying next/previous navigation re-runs extension navigation hooks, which may have side effects.**

`loadNextChapter` and `loadPreviousChapter` call `runNavigationHooks(...)` *before* they assign `lastLoadAttempt` (`BibleReadingManager.tsx:1475`/`1483` and `1670`/`1678`). Because `lastLoadAttempt` stores the bare function reference (`lastLoadAttempt = loadNextChapter`), pressing **Reload** re-invokes `runNavigationHooks("next"/"previous")` from scratch. Those hooks are extension-provided (`runtime.instance.navigateNext` / `navigatePrevious`, `BibleReadingManager.tsx:977`) and can perform navigation as a side effect, not just return a predicate. If an extension's hook advances internal state and returns a non-`default` outcome, a retry would apply that side effect a second time. There's also an edge case where the panel can get stuck: if the hook returns `handled` or `prevent` on the retry, the function returns *before* clearing `error.value`, so `retrying` flips back to false with the error still set and the panel stays up with nothing having happened.

- **Realistic?** Only when an installed extension implements `navigateNext`/`navigatePrevious` with side effects *and* the failure occurred on a next/previous navigation. The far more common failure paths (`selectChapter`, `selectTranslationAndChapter`, `loadInitialData`) store an argument-bound closure and don't touch the hooks, so they replay cleanly. That's why this is Low, not higher.
- **Reproduction (requires such an extension):** install an extension whose `navigateNext` mutates state and returns `{ type: "navigate", ... }`; go offline; tap next so the fetch fails and the panel appears; go back online; press Reload → the hook side effect runs twice.
- **Suggested fix:** capture the resolved next/previous chapter target *outside* the hook replay, or store a replay closure that skips re-running the hooks (e.g. have `loadNextChapter`/`loadPreviousChapter` set `lastLoadAttempt` to a closure that resumes from the post-hook fetch step). At minimum, move the `lastLoadAttempt =` assignment and `error.value = null` so a `handled`/`prevent` retry can't leave a stuck panel.

---

### Minor / nits

**M1 — `setRetrying(false)` can run after unmount.** In `retryChapterLoad` (`BibleReader.tsx:1449`), the `finally` block calls `setRetrying(false)` unconditionally. If the user switches tabs (unmounting this `BibleReader`) while a retry is still in flight, the state update lands on an unmounted component. Preact tolerates this, so it's harmless in practice — flagging only for completeness. A mounted-ref guard would silence it if you care.

**M2 — Reload button has no dark-mode-specific hover.** `.sb-reader-error-retry:hover` mixes the primary color with `black 6%` (`BibleReader.css`). On a dark primary that darkens rather than lightens the button, which reads slightly oddly in dark mode. Cosmetic; consider `color-mix(... white 8%)` under a dark selector if you want the hover to feel consistent across themes.

**M3 — CHANGELOG placement.** The entry lands under `### 🔧 Changed`. Given the issue is labeled `enhancement`, that's defensible, but if the team treats this as a UX fix for a broken state, `### 🐛 Fixed` might fit better. Purely a categorization call.

---

### Verification checklist

- [x] Matches issue #1377 spec (icon `cloud_off`, header, sub-header, Reload button)
- [x] Raw error string no longer surfaced to the reader (asserted in tests)
- [x] `retryLoad` added to both test fixtures (`BibleReader.test.tsx`, `TabSlotReader.test.tsx`) so the `BibleReadingState` interface stays satisfied
- [x] Translation keys added to `en.json` only (repo convention)
- [x] Two affected unit test files pass locally (123 tests)
- [x] New `useState` at `BibleReader.tsx:1448` sits alongside the existing `useRef` at `1366`, both after the SSR-gated `throw` at `1317`; since `import.meta.env.SSR` is a build-time constant (always false on the client), no new Rules-of-Hooks violation is introduced
- [ ] Full CI (build / lint / typecheck) — in progress at review time, not confirmed

---

KallynGowdy:
@mazenemad can you fix issue L1 from the code review above? One scenario where this would be an issue is when a user is playing a playlist. They click next to go to the next item in the playlist but the chapter fails to load. Clicking "Reload" would then cause the playlist to advance yet again instead of trying to reload the correct chapter data.

KallynGowdy (dismissed):

PR #1533, opened 2026-07-28
feat: implement OfflineTranslationsManager for downloading and managing Bible translations

### LINKED ISSUE #1375: Add the ability to download entire translations for offline use
Currently, translations are dynamically loaded chapter-by-chapter as they are viewed. This is great for first time load speed, but some users will want access to the entire Bible on their device even when offline. The Free Use Bible API supports this use-case by providing the [`/api/{translation}/complete.json` endpoint](https://bible.helloao.org/docs/reference/#get-an-entire-translation).

In the translation selector dialog, we should give users the ability to download the entire translation to their device. This would download the complete JSON file to the user's device and store it in IndexedDB (not local storage because of size limitations). BibleDataManager would then need to check IndexedDB first for a locally downloaded version before trying to fetch it from the Free Use Bible API. Additionally, the BibleDataManager should support checking for updates to downloaded translations by comparing the downloaded translation `sha256` property against the one from the `/api/available_translations.json` list (this property tracks the content hash of the entire translation in the API). Of course, this should only happen if the user has network connectivity. Finally, the user should be able to delete an already downloaded translation from their device.

mazenemad:
fixes #1375 

- Added OfflineTranslationsManager to handle downloading entire translations for offline use.
- Implemented methods for downloading, deleting, and checking for updates on translations.
- Introduced IndexedDB storage for managing downloaded translations and chapters.
- Created unit tests for OfflineTranslationStore and OfflineTranslationsManager to ensure functionality.
- Added support for progress tracking during downloads and error handling for failed downloads.


Blitzy:
> [!NOTE]

## Code review — Offline translation downloads (#1533)

Thanks for this — it's a thorough, well-documented implementation that maps cleanly onto what #1375 asked for: download a whole translation to IndexedDB, serve chapters back in the same shape as the network, detect stale copies by `sha256`, and delete on demand. The manager/store split is clean, the SSR/no-IndexedDB fallback is handled, and there's good test coverage (73 passing across the four new suites). `pnpm check:ts` and the offline test suites are green on my end.

The notes below are mostly edge cases and polish — nothing that blocks the core feature working.

---

### Medium

**M1 — Reading a downloaded translation's books can hide the "update available" badge**

`updateAvailable` is derived by comparing the stored copy's hash against the hash currently in `availableTranslations` (`OfflineTranslationsManager.tsx`, the `downloaded` computed). The problem is that stale metadata gets written *back* into `availableTranslations`:

- `publishRecordTranslations()` merges each stored (possibly stale) `record.translation` into the list on startup, and after every download.
- `BibleDataManager.getTranslationBooks()` → `cacheBooks()` calls `mergeTranslations(endpoint, [record.translation])` the first time a downloaded translation's books are loaded.

`mergeTranslations` is last-write-wins (`BibleDataManager.tsx:566-569`), so any of these publishes that happens *after* `checkForUpdates()` fetched the fresh list will overwrite the newer `sha256` with the download-time one. Once that happens, `latestHash === record.sha256`, so `updateAvailable` flips back to `false` and the update button disappears until the next `checkForUpdates()` runs.

Reproduction:
1. Download a translation (hash A).
2. Publish a new version server-side (hash B).
3. Open the selector → `checkForUpdates()` runs → update badge appears (this path is covered by the test at `OfflineTranslationsManager.test.ts:288`).
4. Open/read that translation so its books load → `getTranslationBooks()` merges the stored translation (hash A) → the badge disappears even though B is still available.

The existing update tests never call `getTranslationBooks`, which is why this isn't caught. Suggested fixes (either works):
- Have `downloaded` compare against a dedicated "latest known hash" source that stored metadata can't clobber, rather than the mutable `availableTranslations`; or
- Make the stored-translation publishes fill in only translations *not already present* (don't overwrite an existing, potentially-fresher entry) — e.g. a `mergeTranslations(..., { fillOnly: true })` variant used by `publishRecordTranslations`/`cacheBooks`.

---

### Low

**L2 — Cancelling during the "saving" phase doesn't actually stop the save**

`store.save()` (`OfflineTranslationStore.ts:305`) doesn't receive the `AbortSignal`, so it writes every chunk regardless. In the manager (`OfflineTranslationsManager.tsx:610-625`) the abort only short-circuits the *progress callback* — the underlying writes keep going, `setRecord(record)` runs, and `downloadTranslation` returns `true`. Since the cancel button is still shown during the saving phase (any `progress` entry renders the cancel button), a user who taps cancel while it's writing to disk ends up with a fully-stored translation and no indication their cancel did anything. Saving is fast relative to the download, so impact is small, but the button is misleading. Consider passing `controller.signal` into `save()` and bailing out of the chunk loop (and rolling back) when it's aborted, or hide/disable the cancel affordance once the phase is `"saving"`.

**L3 — `getCompleteTranslation` progress path holds several copies of the payload in memory at once**

In the `onProgress` branch (`FreeUseBibleAPI.tsx:1199+`) the code accumulates all chunks in an array, allocates a second merged `Uint8Array`, decodes that to a JS string, then `JSON.parse`s it — so peak memory is roughly 4–5× the decompressed payload (a complete translation is ~7 MB of JSON, more decompressed). That's the opposite of ideal for the low-bandwidth / older-mobile audience this feature (labels `low-bandwidth`, `offline-resilience`) is aimed at, and could OOM a low-end device. The non-progress branch uses `response.json()` and is fine. Options: drop each chunk after copying into the merged buffer, decode incrementally with a streaming `TextDecoder`, or accept the estimate-only progress and keep using `response.json()`. At minimum worth a comment acknowledging the tradeoff.

**L4 — `online`/`offline` listeners are never removed**

`OfflineTranslationsManager.tsx:505-513` adds `window` listeners with no teardown. Harmless for the app-lifetime singleton, but it leaks a listener each time a manager is constructed (e.g. repeated `createBibleDataManager` in tests). A returned cleanup / `dispose()` would make it tidy.

**L5 — End-of-Bible neighbours fall through to the network even when fully downloaded**

`getAdjacentChapter` returns `null` at the first/last chapter of the Bible (`adjacentChapterRef` has no neighbour), so `getNextChapter`/`getPreviousChapter` in `BibleDataManager` fall back to `api.getNextChapter`/`getPreviousChapter`. For a fully-downloaded translation read offline at Revelation 22 (or Genesis 1 going back), that network call will fail rather than cleanly reporting "no next chapter". Minor, but it's a spot where an offline-complete translation still touches the network.

---

### Nits

- **N6 (a11y):** the download / downloaded / update buttons expose their label only via `title` (`OfflineTranslationControls`), which screen readers don't reliably announce — the in-progress button correctly adds `aria-label`, so mirroring that on the others would be consistent. The `material-symbols-outlined` glyph spans could also be `aria-hidden`.
- **N7:** no CHANGELOG entry — recent history (`chore: Updated CHANGELOG`) suggests that's a repo convention worth following here.

---

Overall this is close. M1 is the one I'd want addressed before merge since it undermines the update-detection that #1375 explicitly called for; the rest are polish and can be judged on their merits.

---

mazenemad:
Thanks for the thorough review. All of it is addressed in d59a3bcba except L5, which I believe is a false positive — details below.

### M1 — update badge hidden by reading a downloaded translation ✅

You were right about the mechanism, and it's slightly worse than just the badge: stale download-time metadata was overwriting *everything* the app had since learned about the translation (name, website, licence URL), not only `sha256`.

Went with your second suggestion. `mergeTranslations` now takes `{ fillOnly: true }`, which fills in translations that aren't already known and leaves the rest alone. Two callers pass it:

- the startup publish in `ready`, and
- `cacheBooks` when the books came from storage.

The publish *after a completed download* deliberately does **not** pass it, and that turned out to matter: that record's metadata came from the `complete.json` payload we just downloaded, so it's the freshest thing anyone has. Had I made every stored-metadata publish fill-only, downloading while the cached list was stale would have left `record.sha256` ≠ `latestHash` and shown "update available" immediately after a successful download.

Two tests, both of which fail without the fix:

- `keeps the flag raised after the downloaded translation is read` — your exact reproduction, ending in `getTranslationBooks("AAB")`.
- `keeps the flag raised when the stored copy loads after the API list` — a fresh page load where the API already publishes the newer version, with a gated store so the storage read deterministically lands *after* the list request. On a real device those two race.

### L2 — cancelling during "saving" ✅

`store.save()` now takes the `AbortSignal`, checks it between chunks (a chunk is one transaction, so that's the only point where stopping is clean), deletes what it wrote, and rejects with an `AbortError` — which the manager already treats as "cancelled, not failed". Removed the `signal.aborted` guard from the progress callback, since the save itself now stops.

Since the cancel button is now honest, I left it visible during the saving phase and added "tap to cancel" to the saving label instead of hiding it.

One consequence worth flagging: a save clears the previous copy before writing the new one (that's what makes "metadata written last" mean "not downloaded" rather than "downloaded with holes"), so **cancelling an update costs the user the copy they had**. I think that's the right trade against a translation with chapters from two different versions in it, but it's a judgement call — say the word if you'd rather it kept the old copy readable.

Rollback also exposed a pre-existing gap: any failure during the save left `records` still advertising a download the device no longer had. `downloadTranslation` now re-reads storage on failure so the signal matches reality — covered by `stops claiming a translation is downloaded when replacing it fails`.

### L3 — peak memory in the progress path ✅

Now decodes each chunk as it arrives with a streaming `TextDecoder` and drops it, so the chunk array and the merged `Uint8Array` are both gone. Peak is roughly the decoded string plus the parsed object (~2×) instead of ~4–5×.

Added `decodes text whose characters are split across chunk boundaries` — one byte per chunk, so every non-ASCII character arrives in pieces. It fails without `{ stream: true }`, which is the one thing that can go quietly wrong with this approach.

### L4 — listener teardown ✅

Added `dispose()` on the offline manager: removes both `window` listeners and aborts any in-flight download. Two tests cover it.

### L5 — end-of-Bible neighbours ⚠️ I don't think this one is real

`getAdjacentChapter` does return null at the edges, but the fallback doesn't reach the network. `buildChapter` synthesizes `nextChapterApiLink: null` when there's no neighbour, and `api.getNextChapter` returns null on a missing link without calling `fetch`:

```ts
async getNextChapter(chapter, endpoint?) {
  if (!chapter.nextChapterApiLink) {
    return null;
  }
  ...
```

So a fully-downloaded translation read offline at Revelation 22 gets a clean null. The existing `walks next/previous chapters locally` test already relies on this — unmocked URLs reject in that suite, so a request would have failed it — but that was implicit, so I added `answers the ends of the Bible without touching the network`, which asserts the fetch count stays flat across both edges.

I did implement the short-circuit first, then reverted it: it added an `await` to every chapter navigation and would have removed the network fallback for a chapter that's genuinely missing from a download (which `falls back to the API when a chapter is missing from the download` covers, and which is the behaviour we want). Happy to reinstate if you're seeing something I'm not.

### N6 — a11y ✅

Every offline button now carries its label as both `title` and `aria-label`, and the `material-symbols-outlined` glyph spans are `aria-hidden="true"`.

### N7 — CHANGELOG ✅

Added under `## TBD` → `✨ Added`.

---

`pnpm check:ts` clean, `pnpm lint` 0 errors, full suite green (3132 tests, 185 files) — 11 new tests, 4 of which fail against the pre-fix code.

KallynGowdy (approved):

PR #1534, opened 2026-07-28
Feature/offline app loading

### LINKED ISSUE #1376: Support offline loading
Currently, while the Seed Bible has a service worker that enables better caching, it does not have support for loading the Seed Bible when the user's device is offline.

It should do the following to support loading offline:

- Cache the index.html file use network-first caching strategy. It should use a short timeout (like 3 seconds) so that user's don't have to wait a long time for it to load when offline.
  - It should respond with the index.html file on the following paths:
    - Anything that does not start with `/b/`
    - Anything that does not match another cache key (like JS, CSS, etc.)
  - It should cache and respond with the index.html regardless of query parameters. Query parameters like `?book=GEN&chapter=10` should also match queries like `?book=JHN&chapter=9&lang=ar`.
- Cache JS/CSS, images, fonts, etc. using a cache-first strategy. It should only cache files that are not stored under a custom branch path (like `/branches/*`).
- It should utilize precaching for core assets (index.js, vendor.js, CSS, fonts, images) and runtime caching for non-core assets (extra language bundles, extensions, etc.). The index.html file should utilize runtime caching so that the most recent response is used from the cache.

mazenemad:
fixes #1376


Blitzy:
> [!NOTE]

## Code review — offline app loading (`fixes #1376`)

### Scope of this review

This PR's diff shows **22 files / +3557**, but that spans two commits:

- **`e9d0bbe`** — `OfflineTranslationsManager` + `OfflineTranslationStore` + `BibleSelector` changes. This is the same work as the separate open PR #1533 (which targets `develop`). It is not part of issue #1376 and is being reviewed on its own PR, so I've left it out of scope here.
- **`26d47b6`** — the actual offline-app-loading work for #1376: the service worker and its build wiring. **7 files:** `standalone/sw.ts`, `standalone/tsconfig.sw.json`, `vite.config.ts`, `tsconfig.json`, `package.json`, `pnpm-lock.yaml`, `CHANGELOG.md`.

My review focuses on commit `26d47b6`. One housekeeping note up front: because the diff carries all of #1533's work too, whoever merges should either land #1533 first and rebase this branch, or review the service-worker commit in isolation — otherwise the two PRs will show a confusingly overlapping diff.

### Overall

This is a careful, well-documented implementation that maps cleanly onto every requirement in #1376:

- HTML served **NetworkFirst with a 3s timeout**, cached under a single key so `?book=GEN&chapter=10` and `?book=JHN&chapter=9&lang=ar` share one copy — matches the issue exactly.
- App shell served for any same-origin navigation that isn't a `/b/` branch preview and isn't a file request.
- JS/CSS/images/fonts served **CacheFirst**, scoped to *this* build's asset prefix so another branch's chunks are never cached.
- **Precache** for boot assets, **runtime cache** for the on-demand pieces (the other 23 locales, extensions).

The reasoning about the single shared app-shell key holds up: I traced the client and confirmed language and layout are re-derived client-side (`getUrlLanguage(url) ?? getInitialLanguage(navigatorLanguages())` in `I18nManager`, `window.innerWidth` for mobile in `SeedBibleStateManager`), and the entry uses `render()` not `hydrate()`, so the SSR markup and header-derived config in a cached shell are discarded on every offline boot. Good.

The findings below are mostly about **verification risk** and polish — I did not find a definite functional bug in the worker logic.

---

### Medium

**M1. The whole feature is unexercised by this PR's preview deploy — verify it on a root-scoped build before merging.**

The service worker only ships on the **root** build:

```ts
// vite.config.ts
const isRootBuild = !deployBranch || deployBranch === "main";
...
plugins: [ ..., ...(isRootBuild ? [VitePWA({ ... })] : []) ]
```

The preview for this PR is deployed at `https://alpha.seedbible.org/b/feature_offline-app-loading` — a `/b/` **branch** build, where `isRootBuild` is `false`. So on that preview the `VitePWA` plugin is never added: no `registerSW.js` injection, no `sw.js`, no precache. **None of this code path runs on the preview URL**, which means the parts most likely to break in production have not been observable yet:

1. **Cross-origin precache install.** `selectAndRelocateCoreAssets` rewrites each core asset to its absolute CDN URL (`https://<assetHost>/branches/main/<buildId>/assets/...`). Workbox then `fetch`es those absolute URLs during `install`. A single failed precache fetch aborts the whole install and the worker never activates. This works only if the CDN returns `Access-Control-Allow-Origin` for those requests (the app already loads them as `crossorigin` module scripts, so ACAO is *probably* present — but it has never been tested through the precache path).
2. **Registration injection through SSR.** There's no explicit `virtual:pwa-register` import or `navigator.serviceWorker.register(...)` call; registration relies on vite-plugin-pwa's default `injectRegister: 'auto'` writing a `<script>` into the built `index.html`, which the host then serves as its template (`server/store.ts` → `readFile(.../client/index.html)`). This is plausible on a root build but, again, unverified.

*Suggested verification (no code change needed if it passes):* deploy to a root/`main`-scoped environment (or a staging root), then in the browser:
- DevTools → Application → Service Workers: confirm the worker reaches **activated**.
- DevTools → Application → Cache Storage: confirm a `workbox-precache-*` cache exists and is populated with the **absolute** asset URLs, and that the vendor chunk is present (this is why `maximumFileSizeToCacheInBytes` was raised).
- Check the Network/Console for any CORS or 404 errors during install.
- Then go offline (DevTools → Network → Offline) and hard-reload `/?book=GEN&chapter=10` — it should open the app, not the browser's "no connection" page.

If the CDN turns out not to send CORS headers for precache fetches, the fix is either to add `crossorigin`-friendly CORS on the asset host or to fetch those precache entries with an appropriate mode — but confirm the symptom first.

---

### Low

**L1. No automated tests for the service-worker routing/manifest logic.**

`isAppShellNavigation`, `isCacheableStaticAsset` (in `standalone/sw.ts`) and `selectAndRelocateCoreAssets` (in `vite.config.ts`) carry the load-bearing decisions, and a silent regression in any of them breaks offline boot without any visible signal at build time. `selectAndRelocateCoreAssets` in particular is a pure function that's easy to cover: given a sample Vite manifest plus a set of globbed entries, assert that (a) only core chunks + their CSS + images/fonts survive, (b) `assets/*` URLs are rewritten to the absolute base, and (c) root-served entries like the web manifest pass through untouched. Consider extracting the two `sw.ts` predicates so they can be unit-tested the same way.

**L2. `selectAndRelocateCoreAssets` precaches *every* image and font in `assets/`, not just the core ones.**

```ts
if (!core.has(entry.url) && !IMAGE_OR_FONT_RE.test(entry.url)) continue;
```

The `IMAGE_OR_FONT_RE` branch keeps any image/font regardless of whether a boot chunk actually references it. This does satisfy the issue's "precache … fonts, images" wording, but it also pulls decorative or feature-specific images into the install payload. You already have the precise set of images a core chunk references via `chunk.assets` in `readCoreAssetFiles()`, so if install weight becomes a concern you can tighten this to core-referenced assets only. Flagging as a conscious trade-off rather than a bug.

---

### Nits / informational

**N1. Imprecise comment on the app-shell cache key.** The block comment above `APP_SHELL_CACHE_KEY` says the only URL-specific thing the shell carries is the server-injected `basePath`. The injected config actually also carries `acceptedLanguages` and `renderedAsMobile` (both header-derived, see `renderAndRespond` in `server/index.ts`). It's still safe — those are only read during SSR and the client re-derives them — but the claim as written could mislead a future maintainer into adding header/URL-specific data to the shell without realizing every cached navigation shares one copy. Worth tightening the wording to "…and the client re-derives language and layout, so header-derived config in the cached shell is discarded."

**N2. `selectAndRelocateCoreAssets` always returns `warnings: []`.** The `warnings` field is never populated. Either wire it up to something useful (e.g. warn when an expected core entry is missing from the glob output, or when a core chunk exceeds `maximumFileSizeToCacheInBytes` and would be dropped) or drop the field to avoid implying a diagnostic that doesn't exist.

**N3. `warmAppShellCache` bypasses the route's `CacheableResponsePlugin` and writes straight to the cache.** It gates on `response.ok` (200–299) while the NetworkFirst route only accepts `200`. Functionally equivalent for the `/` shell, but for consistency you could gate on `response.status === 200` to match the route exactly.

---

I've reviewed the linked issue #1376 and the PR conversation — the only existing comment is the automated deployment link, so there was no prior discussion to reconcile against. The implementation tracks the issue's acceptance criteria faithfully; the main ask before merge is exercising it on a root build (M1).

---

PR #1535, opened 2026-07-28
fix: enhance verse toolbar behavior with fullscreen panes

### LINKED ISSUE #1515: Mobile: verse toolbar interfering with fullscreen panes
Fullscreen panes like the Location tool's are being covered in large part by the mobile verse toolbar.

I expected the fullscreen panes to either hide or cover the mobile verse toolbar.

mazenemad:
fixes #1515


Blitzy:
> [!NOTE]

## Summary

This PR does a clean, well-scoped job on issue #1515. It fixes **both** halves of the reported problem:

1. **Toolbar covering the pane** — the verse toolbar is now *suspended* (not dismissed) whenever a pane covers the reader, so it no longer floats on top of the Locations map (and other fullscreen panes). The selection is kept, so the toolbar returns unchanged when the pane closes.
2. **Close/clear also closing the pane** (the follow-up reported in the [issue comment](https://github.com/HelloAOLab/seed-bible/issues/1515#issuecomment-5080358325)) — `?verse` was being counted as a "navigation" in the fullscreen-pane effect, so selecting or clearing a verse looked like a move and closed the pane the user had just opened. Dropping `?verse` from the location key fixes this.

The change is small (7 files, +225/−21), the intent is documented well in the comments, and the geometry cleanup (moving `position`/`inset`/`z-index` out of the inline style into `PaneLayout.css` so the mobile bottom inset can actually override it) is a genuine improvement over the old inline `bottom: 0` that silently defeated the media query.

I ran the new test file locally — all 3 tests pass (`test/unit/seed-bible/components/BibleReaderToolbar.test.tsx`). I also verified the mobile breakpoint used by the code (`MOBILE_BREAKPOINT = 480`, `isMobile = width <= 480`) matches the CSS `@media (max-width: 480px)`, and that the reserved `4.375rem` equals the documented toolbar box height (`0.5 + 3.25 + 0.625rem`). Both line up.

---

## Issues by severity

### 🟡 Low — a same-chapter sidebar-search result leaves the covering pane open, and a comment overstates the behavior

**Where:** `managers/SeedBibleStateManager.tsx` (the comment at lines 606–608) and `components/SidebarSearch/SidebarSearch.tsx` (`openSearchResult`, ~line 181).

The new comment on the fullscreen-pane effect says:

> Verse-level navigation that really should reveal the reader (a verse reference link, a search result) calls `closeFullscreenPanes` explicitly on its own path.

That's true for verse-reference links (`handleOpenVerseReference` calls `panes.closeFullscreenPanes()` at `SeedBibleStateManager.tsx:1265`) and for the desktop floating panels (`FloatingReaderPanels.tsx` calls it at lines 363/372/392/405). But it is **not** true for the sidebar search path: `SidebarSearch.tsx`'s `openSearchResult` calls `selectTranslationAndChapter(...)` **without** an explicit `closeFullscreenPanes()`, relying entirely on the book/chapter-change effect to reveal the reader.

Because the location key was narrowed from `book|chapter|verse` to `book|chapter`, that effect no longer fires when only the verse changes. So in the specific case where a search result lands in the **same book + chapter** the reader is already showing, the covering pane is not closed and the result stays hidden behind it.

Before this PR the old `book|chapter|verse` key would have caught that (the verse differed → location changed → pane closed), so this is a small behavior regression for that narrow case, not just a pre-existing gap.

**Reproduction (mobile viewport, ≤480px):**
1. Open a fullscreen pane over the reader (e.g. the Locations map) while the reader is on, say, John 3.
2. With that pane still open, open the sidebar search.
3. Search for and tap a result that is also in **John 3** (same book and chapter, different verse).
4. Observe: the reader loads/scrolls to that verse *behind* the pane, but the pane stays on top, so you don't see the result. (Tapping a result in a *different* chapter closes the pane correctly, because book/chapter changes.)

**Suggested fix** — make the sidebar-search path do what the comment already claims, so it doesn't depend on the book/chapter key changing:

```tsx
// components/SidebarSearch/SidebarSearch.tsx, in openSearchResult(...)
const targetTab = getOrCreateSearchTargetTab(state);
state.panes.closeFullscreenPanes(); // reveal the reader even for a same-chapter result
await targetTab.readingState.selectTranslationAndChapter(
  result.translationId,
  result.bookId,
  result.chapterNumber,
  { scrollToVerse: result.verseNumber ?? undefined }
);
```

If you'd rather not touch the search path, at minimum tighten the comment in `SeedBibleStateManager.tsx` so it doesn't claim the search path closes panes explicitly when the sidebar-search variant actually leans on the effect.

---

## Nits / optional

- **Manual visual check on mobile still worth doing.** The `4.375rem` bottom inset assumes `.sb-content-row` spans the full viewport height on mobile so that an `absolute` `bottom: 4.375rem` lines up with the `fixed` bottom bar. The math checks out and the unit tests cover the show/hide logic, but the tests don't assert pixel geometry — a quick look at the deployed branch on a real phone (the map pane's own footer/close controls sitting just above the bottom bar, no gap, no overlap) would confirm the reserved height is exact. Deploy link is in the PR's auto-comment.
- **No test for the same-chapter search case above.** If you take the suggested fix, a small test asserting a covering pane is closed when a search result resolves to the same book/chapter would lock it in.

Nothing here is blocking on its own — the core fix is correct and the low-severity item is a narrow edge case. Addressing it (or at least correcting the comment) would make the change airtight.

---

PR #1536, opened 2026-07-28
fix: close mobile More menu on outside tap or Escape key press

### LINKED ISSUE #1509: "More" menu on mobile needs to be dismissible by tapping outside of it.
The mobile "More" menu needs to support being closed when tapping off of it. We should not block any selection of elements thought. For example we should still be able to select a verse, or select anything in the top quick toolbar, while the More menu is open.

Image

### LINKED ISSUE #1510: Mobile search bar does not dismiss when opening the "More" menu.
## What is happening

If you open the search tool on mobile and then tap the More button, the search bar stays on screen.

Image

## What I expected

Tapping the More button should have dismissed the search bar.

mazenemad:
fixes #1509 
fixes #1510

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
@@ -545,6 +545,9 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
     )
   );
   const isMoreMenuOpen = useSignal(false);
+  // The mobile More button, so dismissing its menu with Escape can hand focus
+  // back to it instead of dropping it on the removed popover.
+  const moreButtonRef = useRef<HTMLButtonElement>(null);
   const selectedToolbarToolId = useSignal<string | null>(null);
   const selectedVerseToolId = useSignal<string | null>(null);
   // Whether the mobile verse sheet shows its overflow actions (the "More" /
@@ -839,6 +842,13 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
   // toolbar button works normally while the menu is open, it just also
   // dismisses the menu. Capture phase so we still see the tap even if the
   // target stops propagation.
+  //
+  // `pointerdown` (rather than `click`) means a touch-scroll that starts while
+  // the menu is open also dismisses it, since a scroll gesture begins with a
+  // pointerdown. That is intended: it matches how dropdowns usually behave, and
+  // dismissing as the gesture starts feels more responsive than waiting for it
+  // to finish. Scrolling the menu's own list is unaffected — those touches land
+  // inside the anchor and return early below.
   useEffect(() => {
     if (!isMoreMenuOpen.value) return;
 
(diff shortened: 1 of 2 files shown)


Blitzy:
> [!NOTE]

## Code Review — close mobile More menu on outside tap / Escape

**Scope:** one file, `components/BibleReaderToolbar/BibleReaderToolbar.tsx` — a single `useEffect` that, while the mobile More menu is open, listens on `document` for a capture-phase `pointerdown` (closing the menu when the tap lands outside `.sb-reader-toolbar-more-anchor`) and for `Escape` on `keydown`.

### Verdict

This looks correct and it satisfies both halves of issue No. 1509:

- **Closes on outside tap** — a `pointerdown` whose target is not inside `.sb-reader-toolbar-more-anchor` sets `isMoreMenuOpen.value = false`.
- **Does not block other selections** — because the listener never calls `preventDefault()`/`stopPropagation()`, the same tap still reaches its target. I verified the top quick-toolbar buttons and the chapter verses are **not** descendants of `.sb-reader-toolbar-more-anchor`, so tapping them both closes the menu *and* performs the button action / verse selection, exactly as the issue asks.

I also confirmed a few things that could have been landmines but are fine:

- `MobileMoreMenu` is rendered inline as a child of the anchor `<div>` (not portaled), and its popover is `position: absolute` inside the `position: relative` anchor — so `target.closest(".sb-reader-toolbar-more-anchor")` correctly treats taps on the popover and its items as "inside." Menu-item taps don't self-dismiss via this handler; the items' own `onClick` handles closing.
- Tapping the More button to **close** while open: the capture `pointerdown` early-returns (button is inside the anchor), then the button's `onClick` toggles it shut — a single, clean close, no double-toggle/reopen.
- The listener is attached after the open (open happens on `click`, which follows the opening `pointerdown`), so the interaction that opens the menu can't immediately re-close it.
- The `[isMoreMenuOpen.value]` dependency and capture-phase document listener mirror the existing `hasVerseSelection` outside-click effect a few lines above, so the pattern is consistent with the file.

No critical, high, or medium-severity issues found. The items below are minor.

---

### Low

**No. 1 — Escape doesn't return focus to the More button (accessibility).**
When the menu is dismissed with `Escape`, keyboard focus is left wherever it was rather than moving back to the trigger. For a keyboard/screen-reader user this means focus can end up "lost" after closing. Note the button already exposes `aria-expanded`, so it's clearly intended to be keyboard-operable.

*Reproduction:* focus the More button with the keyboard, press Enter/Space to open, press Escape — focus is not restored to the More button.

*Suggested fix:* stash the trigger element when opening and restore focus on close. Roughly:

```tsx
const handleKeyDown = (event: KeyboardEvent) => {
  if (event.key === "Escape") {
    isMoreMenuOpen.value = false;
    moreButtonRef.current?.focus(); // ref on the More <button>
  }
};
```

This is a nice-to-have polish, not a blocker for a mobile-first menu.

**No. 2 — On touch devices the menu closes when a scroll gesture starts.**
`pointerdown` fires at the beginning of a touch-scroll, not just on a discrete tap. So if a user starts scrolling the chapter content while the More menu is open, the menu dismisses on the first touch. This is standard dropdown behavior and is very likely the desired outcome — flagging it only so it's a conscious decision. Worth a quick check on the deployed branch (https://alpha.seedbible.org/b/fix_mobile-more-menu-tap-off-close) that dismiss-on-scroll-start feels right; if not, switching the outside handler to `click` (which won't fire on a scroll) would be the alternative, at the cost of not dismissing until the gesture completes.

### Nits (optional)

**No. 3 — Capture vs. bubble inconsistency with the sibling listener.** This new outside-click effect uses the capture phase (`, true`) while the adjacent `hasVerseSelection` outside-click effect uses the bubble phase. The inline comment explains capture is intentional (so a target that stops propagation is still seen), so this is fine — just noting the two similar handlers now differ, in case you want them uniform.

**No. 4 — No automated test.** There's currently no test file for `BibleReaderToolbar`, so I'm not treating this as a gap, but if you later add DOM tests for the toolbar, "outside `pointerdown` closes the menu" and "tap on a quick-toolbar button both closes the menu and fires its action" would be the two cases worth locking in.

---

PR #1537, opened 2026-07-28
Fix/missing space from beginning of verse

### LINKED ISSUE #1514: Missing space from beginning of verse sentence when verse numbers are disabled.
When the verse numbers are disabled, the beginning of the verse is always mashed up against the punctuation of the previous verse with a space to separate them like you would expect. 

The two images below illustrate this. In the first image the blue circles are examples of verses with the verse number on. The second image the red circles are those same verses but with the numbers disabled; notice that they are up against the end of the previous verse and are missing a space.

mazenemad:
fixes #1514

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.css
@@ -552,7 +552,7 @@
   vertical-align: middle;
   /* Phase 2 (UI-scaling): gutters in em so they track the reader glyph knob.
      Color/opacity adopted from develop. */
-  margin-inline-start: 0.625em;
+  /* margin-inline-start: 0.625em; */
   margin-inline-end: 0.3125em;
   color: inherit;
   opacity: 0.7;
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
@@ -833,6 +833,41 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
     };
   }, [hasVerseSelection.value]);
 
+  // Tapping anywhere outside the mobile More menu closes it. Deliberately done
+  // with a document listener rather than a backdrop element so the tap still
+  // reaches whatever was tapped — selecting a verse or hitting a top quick
+  // toolbar button works normally while the menu is open, it just also
+  // dismisses the menu. Capture phase so we still see the tap even if the
+  // target stops propagation.
+  useEffect(() => {
+    if (!isMoreMenuOpen.value) return;
+
+    const handleDocumentPointerDown = (event: PointerEvent) => {
+      const target = event.target as HTMLElement | null;
+      // The anchor wraps both the More button and the popover, so this covers
+      // taps on either. The button's own click handler does the toggling.
+      if (target?.closest(".sb-reader-toolbar-more-anchor")) return;
+      isMoreMenuOpen.value = false;
+    };
+
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "Escape") {
+        isMoreMenuOpen.value = false;
+      }
+    };
+
+    document.addEventListener("pointerdown", handleDocumentPointerDown, true);
+    document.addEventListener("keydown", handleKeyDown);
+    return () => {
+      document.removeEventListener(
+        "pointerdown",
+        handleDocumentPointerDown,
+        true
+      );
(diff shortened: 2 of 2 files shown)


Blitzy:
> [!NOTE]

## Code Review — PR #1537 (fixes #1514)

Thanks for taking this on! I reviewed the diff against `develop`, the linked issue #1514, and the surrounding reader code. There are two high-severity issues that I think need attention before this can merge, plus a couple of minor cleanups.

---

### 🔴 High — This PR includes unrelated commits from PR #1536

This PR contains **three** commits, but only the last one relates to #1514:

| Commit | Message | Belongs to |
| --- | --- | --- |
| `9052751` | fix: close mobile More menu on outside tap or Escape key press | #1536 |
| `4487013` | fix: enhance More menu behavior to close overlapping panels on open | #1536 |
| `44d8628` | fix: remove margin-inline-start from BibleReader.css to correct spacing issue | **#1514 (this PR)** |

The branch `fix/missing-space-from-beginning-of-verse` was cut from `fix/mobile-more-menu-tap-off-close` instead of from `develop`, so the entire diff of PR #1536 (all the `BibleReaderToolbar.tsx` changes) is dragged into this PR. That means merging this PR would silently merge the More-menu work under a PR whose title and linked issue are about verse spacing.

**Suggested fix:** rebase this branch onto the latest `develop` so only the CSS commit remains:
```bash
git fetch origin develop
git rebase --onto origin/develop origin/fix/mobile-more-menu-tap-off-close fix/missing-space-from-beginning-of-verse
git push --force-with-lease
```
(Or wait for #1536 to merge first, then rebase — the toolbar commits will drop out on their own.) The end state should be a diff that touches only `BibleReader.css`.

---

### 🔴 High — The CSS change does not fix the reported bug, and regresses the numbers-on layout

Issue #1514 is specifically about missing space between verses **when verse numbers are disabled**. The change comments out `margin-inline-start` on `.sb-verse-number` (`BibleReader.css:555`). Based on reading the render code, this can't address the off-case:

- The verse-number element is only rendered when numbers are **on**: `scriptureElements.showVerseNumbers && <sup className="sb-verse-number">…</sup>` (`BibleReader.tsx:686`, `:721`, `:759`). When numbers are **off**, that `<sup>` — and therefore both `margin-inline-start` and `margin-inline-end` — is absent from the DOM entirely.
- Prose verses are emitted as adjacent `<span class="sb-verse">` nodes with no whitespace or separator between them (`renderVerseNode` results are pushed into an array at `BibleReader.tsx:848`/`:905`; Preact inserts no text node between array children), and there is no numbers-off styling anywhere in `BibleReader.css`.

So the net effect is the opposite of what's wanted:

- **Numbers off (the bug case):** no change at all — the element doesn't exist, so toggling its margin does nothing. Verses still run together.
- **Numbers on:** removes the intended leading gutter, so the verse number now abuts the previous verse's ending punctuation. That gutter looks deliberate — see the comment at `BibleReader.css:553-554` and the ribbon logic that relies on it at `BibleReader.tsx:1088` ("leaves the verse-number margin as a clean gutter").

**Reproduction (please double-check on the deployment):**
1. Open any chapter.
2. Settings → turn **off** verse numbers.
3. Observe consecutive prose verses running together with no separating space (matches the issue screenshots).
4. On the deployed branch, this is still present with the change applied.

**Suggested direction:** the spacing needs to come from something that survives when the number is hidden. Two options:
- Add trailing space to the verse box itself, e.g. `.sb-verse { margin-inline-end: 0.3125em; }` (or `padding-inline-end`), which applies whether or not the number renders; or
- Emit an explicit separator (a space) between verses when `showVerseNumbers` is `false`.

Either way, please verify against both the numbers-on and numbers-off states, and — since #1514's comment notes this affects non-English translations too — spot-check at least one RTL/non-English translation.

---

### 🟡 Low — Commented-out CSS left in place

`BibleReader.css:555` leaves the rule as `/* margin-inline-start: 0.625em; */` rather than deleting it. Even once the real fix lands, prefer removing the line outright (git history preserves it); the explanatory comment just above it (`:553-554`) now describes a rule that's disabled, which is misleading to the next reader.

### 🟡 Low — Sparse PR description

The description is just `fixes #1514`. A short note on what was changed and what was tested (which translations, numbers on vs. off) would make review faster and give reviewers a way to confirm the fix.

---

### Note on the More-menu changes

I didn't review the `BibleReaderToolbar.tsx` changes in depth here since they belong to PR #1536 — they should be reviewed there. They did compile (the `sidebar.close*` methods all exist and are used elsewhere in the file), so no blocker from a build standpoint; the point is only that they don't belong in this PR.

---

**Summary:** the branch-hygiene issue and the fix not actually resolving #1514 are the two blockers. Happy to re-review once the branch is rebased onto `develop` and the spacing approach is adjusted.

---

PR #1538, opened 2026-07-28
Fix/missing space from beginning of verse

### LINKED ISSUE #1514: Missing space from beginning of verse sentence when verse numbers are disabled.

mazenemad:
fixes #1514


Blitzy:
> [!NOTE]

## Code review — PR #1538 (fixes #1514)

**What the PR does:** When verse numbers are turned off, adjacent verses used to render with no gap between them ("...had your fill.Do not work..."). This inserts a real space text node between verse spans (and between verses inside a single highlight run) whenever numbers are hidden, so they read as "...had your fill. Do not work...". The core `BibleReader.tsx` change and its three new tests are a clean, well-reasoned fix for the reported issue — the separator sits *between* spans, so highlight ribbons and verse selection still stop at each verse's own glyphs. 

Tests pass locally (43/43 in `BibleReader.test.tsx`) and CI build is green.

However, two things in this PR need attention before merge — one is a likely visual regression, the other is scope.

---

### 🔴 High — Removing `margin-inline-start` re-creates the same "mashed" gap when verse numbers ARE shown

`BibleReader.css` comments out `margin-inline-start: 0.625em` on `.sb-verse-number`:

```css
.sb-verse-number {
  ...
  /* margin-inline-start: 0.625em; */   /* <-- removed */
  margin-inline-end: 0.3125em;
}
```

That leading margin was the *only* horizontal gap between the end of one prose verse and the start of the next verse's number. I confirmed nothing else fills that gap:

- Prose verses render inline with no separating text node when numbers are shown (`needsVerseSpacing` is `false`, so the new `" "` is **not** emitted in numbers-on mode — see `BibleReader.tsx:860`).
- `.sb-verse` uses `--text-verse-margin-left/right`, both of which default to `0` (`app/styles/base.css:80-81`).
- The only other `.sb-verse-number` rule (`BibleReader.css:533`) sets its own `margin-inline-start: -1.2em` and applies to poetry (`.sb-verse-line`) only.

So with the base margin gone, in the **default** view (numbers on) each prose verse number now butts directly against the previous verse's ending punctuation — e.g. `...had your fill.²Do not work...` — which is the mirror image of the very bug #1514 is fixing.

The margin removal doesn't help the numbers-off case at all (the superscript isn't rendered then), so it appears to be an unrelated change that only affects, and regresses, the numbers-on case.

**Note:** the new test *"does not add a separating space when verse numbers are shown"* asserts `textContent` equals `"First verse.2Second verse."`. `textContent` ignores CSS margins, so this test passes either way and cannot catch this regression — it arguably even codifies the mashed layout as expected.

**Reproduction:**
1. Open the reader with verse numbers enabled (the default) on the deploy preview.
2. View any prose chapter and look at the start of each verse.
3. The verse number now hugs the previous verse's final punctuation with no gap; compare against `develop`, which shows a small gap.

**Suggested fix:** Revert this one CSS line (restore `margin-inline-start: 0.625em`). The `BibleReader.tsx` space-insertion fully handles the numbers-off case on its own, and the numbers-on case never needed changing. If there was a genuine issue this was meant to address (e.g. an unwanted indent on the first verse of a paragraph), target that specific case with a first-verse selector rather than dropping the margin globally. Please also confirm visually on the deploy preview — this finding is inferred from the CSS/DOM rather than a screenshot.

---

### 🟠 Medium — Unrelated "More menu" toolbar changes are bundled into a PR scoped to #1514

The PR body says only `fixes #1514`, but the branch also carries two commits that modify `BibleReaderToolbar.tsx` and are unrelated to verse spacing:

- `9052751` — "close mobile More menu on outside tap or Escape key press"
- `4487013` — "close overlapping panels (search / chat / settings) when opening the More menu"

These match the work in PR #1536 (which targets issues #1509 / #1510). I verified the code is genuinely new relative to the PR's base (`develop` @ `6c75f8b` does not contain the More-menu `pointerdown`/`keydown` effect). The changes themselves look fine — the new `useEffect` is a distinct closure from the pre-existing verse-selection `pointerdown` handler, so there's no duplicate-listener collision — but they don't belong in this PR.

Risks of leaving them here:
- Reviewers of a "spacing" PR won't expect toolbar behavior changes.
- Duplicated/competing history with PR #1536 and possible merge conflicts depending on merge order.
- The `CHANGELOG.md` entry only documents the spacing fix, so the More-menu behavior changes would ship undocumented if this PR merges first.

**Suggested fix:** Rebase this branch onto `develop` and drop the two More-menu commits (let PR #1536 own that work), keeping #1538 limited to `BibleReader.tsx` / `BibleReader.css` / the tests / the CHANGELOG. If the intent is deliberately to stack on #1536, please say so in the PR body and coordinate the merge order.

---

### 🟡 Low / nit — Separator strings are unkeyed siblings of keyed elements

The `" "` nodes pushed into the `nodes` array (and the `[" ", verseNode]` returned from the highlight-run `flatMap`) are unkeyed, while their sibling verse/heading/break elements all carry keys. Preact tolerates mixed keyed/unkeyed children, so this is functionally fine, but it can produce dev-mode key warnings and marginally less stable reconciliation when `showVerseNumbers` is toggled (the whole list rebuilds anyway). Optional: emit the separator as a keyed fragment, or leave as-is if you'd rather not add noise.

---

**Summary:** The primary fix for #1514 is solid and well-tested. Please (1) revert the `margin-inline-start` CSS change (or scope it to the first verse) to avoid regressing the default numbers-on view, and (2) split out the unrelated More-menu commits. The Low item is optional.

---

Blitzy:
@mazenemad need to address the code review issues. 

The big ones are that with verse numbers enabled, they are now pushing up against the text of the previous verse where before they were spaced.

Now: 
image

Before: 
image

The second issue is that this branch seems like it was branched off of your "More" menu work? It includes work from it that should not be included in this PR. Those commits should be dropped from this this branch so that this PR is isolated to the missing space issue.

Blitzy:
> [!NOTE]

## Code review — PR No. 1538 (fixes issue No. 1514)

**What this PR does:** With verse numbers turned off, adjacent verses used to run straight into each other ("...had your fill.Do not work..."). This inserts a real space between verse spans (and between verses inside one highlight run) whenever numbers are hidden, leaving the numbers-on rendering untouched. The core `BibleReader.tsx` logic and its three new tests are solid, and CI (lint, `check:ts`, full test suite) is green on the current head commit (`a60bb31`).

This PR was already reviewed once (see the earlier comment on this thread) and a maintainer confirmed a real regression from that review with screenshots. I re-checked the current head commit against that discussion — good news on two of those points, one new blocking issue below.

---

### 🔴 High — Branch has real merge conflicts against `develop`; can't be merged as-is

GitHub reports `mergeable_state: dirty` for this PR. I confirmed it locally by fetching `develop` and this branch's head and doing a merge:

```
git fetch origin develop
git fetch origin refs/pull/1538/head
git merge-base origin/develop <head-sha>   # finds the common ancestor
git merge-tree <merge-base> origin/develop <head-sha>   # simulates the merge
```

Two files conflict:

1. **`CHANGELOG.md`** — this branch's new "Fixed" entry and a since-merged entry from PR No. 1587 ("Fix highlights from a previous account staying visible...") both got added at the same spot in the list.
2. **`packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx`, the import block at the top of the file** — `develop` now imports `RefObject` and `useEffect` (from unrelated work merged after this branch was cut) on the same line this PR changes to add `Fragment`. Git can't reconcile the two import lists automatically.

**Suggested fix:** merge (or rebase onto) `origin/develop`, keep both `CHANGELOG.md` entries, and combine the import list to include everything both sides need (`Fragment`, `RefObject`, `useEffect`, etc.). Neither conflict is hard to resolve by hand, but someone has to do it before this can merge.

---

### 🟠 Medium — Two commits duplicate the now-merged PR No. 1536; should be dropped

This branch still carries two commits — `9052751` ("close mobile More menu on outside tap or Escape key press") and `4487013` ("enhance More menu behavior to close overlapping panels on open") — that are unrelated to issue No. 1514. A maintainer already asked for these to be dropped in the previous review round on this PR ("this branch seems like it was branched off of your 'More' menu work... those commits should be dropped from this branch").

That request wasn't acted on, and in the meantime the identical work shipped separately via PR No. 1536 (merged 2026-07-30). Right now these two commits are a no-op against `develop` — my merge simulation shows `BibleReaderToolbar.tsx` merges cleanly with zero conflict, because both sides already have the same content — but they're still dead weight in this branch's history: they inflate the diff, and if `BibleReaderToolbar.tsx` changes again upstream before this merges, they become a third conflict.

**Suggested fix:** `git rebase -i` onto `develop` and drop those two commits, so this PR's history only touches `BibleReader.css`, `BibleReader.tsx`, the test file, and `CHANGELOG.md` — the files actually relevant to issue No. 1514.

---

### ✅ Already fixed — no action needed

Two things flagged in the earlier review round are resolved as of the latest commit (`a60bb31`, 2026-08-01) and don't need further changes:

- **The `margin-inline-start: 0.625em` CSS regression is gone.** I checked `BibleReader.css` at the current head — the rule is intact, only a new explanatory comment was added above it. The numbers-on rendering path is unaffected by this PR (confirmed both by reading the code and by the fact `needsVerseSpacing` is `false` whenever `showVerseNumbers` is `true`, so no new separator is ever emitted in that mode).
- **The separator nodes are keyed.** `verseSeparator` returns `<Fragment key={key}>{" "}</Fragment>`, so it's a proper keyed sibling of the verse spans around it, not a bare unkeyed string.

---

**Summary:** The fix for issue No. 1514 itself looks correct and well-tested. Before this can merge: (1) resolve the two real conflicts against `develop` (CHANGELOG.md, and the import line in BibleReader.tsx), and (2) drop the two leftover More-menu commits that duplicate already-merged PR No. 1536, per the maintainer's earlier request.


---

KallynGowdy (dismissed):

Blitzy (approved):

PR #1539, opened 2026-07-28
feat: add drag-and-drop support for mobile verse sheet handle

### LINKED ISSUE #1519: Users should be able to drag the verse toolbar up on mobile
Currently, mobile users can only expand the verse toolbar by clicking the "more" button. There is a grab handle for it, but dragging it does nothing. Users should be able to drag it up on order to expand the verse toolbar on mobile.

mazenemad:
fixes #1519


Blitzy:
> [!NOTE]

## Code Review — PR #1539: drag-and-drop support for the mobile verse sheet handle

### Summary

The change adds a taller drag target (`.sb-verse-toolbar-handle-area`) around the existing 4px grab pill and wires pointer handlers to it. Dragging the handle up past a 24px threshold expands the mobile verse sheet, dragging down past the threshold collapses it, and a tap (movement under a 6px slop) toggles the expanded state. `touch-action: none` and `event.preventDefault()` correctly keep the gesture from scrolling the chapter behind the sheet, and pointer capture keeps coordinate math stable even as the sheet resizes mid-drag.

The core "drag up to expand" behaviour from the issue and the "tap to expand" ask from Blitzy's issue comment are both implemented. The gesture logic itself is sound — I could not find a correctness bug in the threshold/tap/cancel state machine. My findings are one unmet requirement plus a few smaller quality points.

---

### Medium — Dragging down does not close the sheet (unmet requirement)

In the issue, KallynGowdy commented:

> Also users should be able to drag the toolbar down to the bottom of the screen to close it.

This PR does not implement that. Dragging down only flips `isVerseSheetExpanded` to `false` (collapse). The sheet is only ever dismissed by the close (X) button, which calls `clearSelectedVerses()`. When the sheet is already collapsed, dragging the handle down does nothing at all — `isVerseSheetExpanded` is already `false`, so the drag is a no-op and the sheet stays open.

**Reproduction:**
1. On a mobile viewport, select a verse so the bottom verse sheet appears (collapsed).
2. Drag the grab handle downward toward the bottom of the screen.
3. Expected (per the issue comment): the sheet dismisses. Actual: nothing happens; the sheet stays open.

**Suggested fix:** in the pointer-up handler, when the sheet is collapsed and the net travel was a downward drag past a close threshold, dismiss the selection instead of no-opping. Because the current move handler doesn't retain the final `dy`, you'd store it (or the last direction) on the drag ref. Roughly:

```tsx
const handleVerseSheetHandlePointerUp = (event: PointerEvent) => {
  const drag = verseSheetDrag.current;
  if (!drag || drag.pointerId !== event.pointerId) return;
  const handle = event.currentTarget as HTMLElement;
  handle.releasePointerCapture?.(event.pointerId);
  verseSheetDrag.current = null;

  const dy = event.clientY - drag.startY;
  if (drag.maxTravel <= VERSE_SHEET_TAP_SLOP) {
    isVerseSheetExpanded.value = !drag.startExpanded; // tap toggles
  } else if (dy >= VERSE_SHEET_CLOSE_THRESHOLD && !isVerseSheetExpanded.value) {
    readingState.value?.clearSelectedVerses(); // drag down while collapsed → close
  }
};
```

This is a judgment call for the maintainers on scope — the issue title is specifically about dragging *up*, so it's reasonable to land expand/collapse now and track close-on-drag-down separately. Flagging it because it was an explicit request from the issue author on the linked issue and isn't mentioned in the PR description.

---

### Low — No automated tests for the new gesture logic

The repo has a strong test culture (thousands of tests), and this PR adds a small pointer state machine (threshold, tap-slop, reversibility, cancel-restores) that is a natural fit for unit testing — yet no test file was added or updated. The threshold and tap-vs-drag branches in particular are easy to regress silently. Consider adding coverage for: tap toggles, drag-up-past-threshold expands, drag-back-under-threshold reverts to the start state, and pointer-cancel restores `startExpanded`.

### Low — Dead zone between the tap slop and the drag threshold

`VERSE_SHEET_TAP_SLOP` is 6px and `VERSE_SHEET_DRAG_THRESHOLD` is 24px. A gesture that travels between 7px and 23px and is then released does nothing: it's too far to count as a tap (so no toggle in pointer-up), but never crossed the drag threshold (so pointer-move left the state at `startExpanded`). To a user, a small-but-deliberate drag can feel unresponsive. This is minor and arguably intentional, but worth a conscious decision — e.g. treating any release past the tap slop as a commit in the dominant drag direction would remove the dead zone.

### Low / Info — Interactive drag area is `aria-hidden="true"`

`.sb-verse-toolbar-handle-area` carries the pointer handlers but is `aria-hidden="true"`. The comment justifies this (the "More"/"Less" card exposes the same toggle to assistive tech), which is a fair call since the handle is a redundant visual affordance and keyboard/AT users retain the More button. No action needed — noting it so the accessibility trade-off is explicit and intentional.

---

### Verdict

The implemented gesture logic is correct and the CasualOS/Preact-signals conventions are followed cleanly. The main open question is scope: whether drag-down-to-close (the Medium item) should land in this PR or be tracked separately, since it was explicitly requested on the linked issue. The remaining items are quality nits. No blocking correctness defects found.

---

KallynGowdy:
This is gonna need some more work. Dragging on the grab handle doesn't produce any visual changes, and with the default height of the verse toolbar, there often isn't a need for the "more" button anyways.

I think we should have @amir2iqbal447 come up with a new design for us that doesn't use the "more" button and uses less vertical height by default before we continue. 

### Example video
https://github.com/user-attachments/assets/338f4a8b-3f31-4375-99bd-d6ac70d062ae

Blitzy:
I dont think we necessarily need more design work here, the implementation is just wrong.

The drag handle needs to actually move the drawer. Instead of a remove button there needs to be a little bit of text at the bottom that says "Swipe up to see more", we had this a little while ago but it was removed until we could support dragging the drawer which is what this issue is for.

Here are designs on figma showing this: https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=8598-3490&t=zujIgDnT7J7bvyYt-4

Blitzy:
@mazenemad can you take another attempt at this?

mazenemad:
on it

Blitzy:
> [!NOTE]

## Code Review — PR #1539 (latest commit `d9d33b41`)

This review covers the rewrite pushed after the earlier feedback thread (Kallyn's "needs more work" and Blitzy's "the drag handle needs to actually move the drawer" / Figma link). The new implementation replaces the old "More"/"Less" toggle button entirely: the overflow row is now revealed by an actual finger-tracked drag on the handle, a "Swipe up to see more" hint sits at the foot of the sheet instead of a button, and a downward drag on the collapsed sheet slides the whole sheet away to dismiss the selection. That matches what was asked for in the thread. CI (`build 24.x`) is green on this commit, and a substantial new test file (`test/unit/seed-bible/components/BibleReaderToolbar.test.tsx`) covers the tap/drag/settle/cancel state machine — a good improvement over the previous version, which shipped with no tests for this logic.

I did not find a correctness bug in the core drag state machine itself (tap-vs-drag, settle-at-halfway, cancel-restores-start). My findings are about a stale-measurement edge case, an accessibility gap, and a scope question carried over from the original issue thread.

---

### Medium — Overflow row height is measured once and never re-measured, so it can go stale

`measureVerseSheetOverflow` (packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx, around the `verseSheetOverflowHeight` signal and its ref callback) sets `verseSheetOverflowHeight.value = element.scrollHeight` from a ref callback, which only fires when the DOM node is created or destroyed — not on every render. Two consequences:

1. **Rotation/resize while a selection is open**: the overflow row is `flex-wrap: wrap`, so its natural height depends on the sheet's width. Rotating the phone (or resizing the window) changes how many actions wrap per row, which changes the row's true height, but `verseSheetOverflowHeight` keeps the old value since the div isn't remounted. Dragging the handle after rotating will reveal a row whose CSS height target no longer matches its actual content height — either clipping the last row of actions or leaving dead space.
2. **The signal is never reset to 0**: the `useEffect` that resets `isVerseSheetExpanded`/`verseSheetDragReveal`/`verseSheetDismissOffset` when `hasVerseSelection` goes false does *not* reset `verseSheetOverflowHeight`. If the overflow row's content ever renders at a different height between selections (e.g. a translation change that makes the highlight/bookmark card labels wrap differently, or an extension registering/unregistering a verse-toolbar tool mid-session), the stale height persists into the next selection.

**Reproduction**: On a mobile viewport, select a verse with enough registered tools to produce an overflow row of, say, 2 lines. Rotate the device to landscape (now the row fits in 1 line). Drag the handle open — the revealed height still targets the old 2-line measurement, leaving a visible gap under the last action row.

**Suggested fix**: re-measure on more than just mount — e.g. observe the row with a `ResizeObserver` and update `verseSheetOverflowHeight` on changes, or at minimum re-measure in the existing `window.resize` handling this component already has for `isSmallScreen` (search for where that signal is set) and in the "selection cleared" effect (reset to 0 there too, so a new selection starts from a clean measurement).

---

### Low — Handle only exposes pointer + keyboard activation, not `click`

`.sb-verse-toolbar-handle-area` (BibleReaderToolbar.tsx, the div with `role="button"`, `aria-expanded`, `onPointerDown/Move/Up/Cancel`, and `onKeyDown`) has no `onClick` handler. Pointer events cover mouse and touch, and `onKeyDown` covers physical keyboards, but some assistive-tech activation paths (notably certain VoiceOver/TalkBack "activate" gestures on custom `role="button"` elements) dispatch a synthetic `click` rather than a full pointerdown/pointerup pair. On those paths this control would not toggle at all, silently. Given the element explicitly opts into button semantics (`role="button"`, `aria-expanded`, a focus-visible outline), it's worth adding a defensive `onClick` that mirrors the tap-toggle behavior (guarded by `hasVerseSheetOverflow.value`, same as the keyboard handler), so activation is guaranteed. Low confidence this is currently observable — flagging for defense-in-depth rather than as a confirmed break.

---

### Low / Info — Drag-down-to-dismiss only works from the collapsed state, not mid-gesture from expanded

Kallyn's comment on the linked issue (No. 1519) says: "Also users should be able to drag the toolbar down to the bottom of the screen to close it." In the current logic (`handleVerseSheetHandlePointerMove`), the dismiss offset only ever activates when `drag.startReveal === 0` — i.e., the drag has to *begin* while the sheet is already collapsed. If a user starts the drag while the sheet is expanded and drags all the way down in one continuous motion, the gesture closes the overflow row and stops there; it does not continue into a dismiss even if the finger keeps moving past where the row is fully closed. The user has to release and start a second, separate downward drag to dismiss.

This may be an intentional, acceptable simplification (many bottom-sheet patterns work this way — collapse first, dismiss second), but since it was an explicit, named request in the issue thread and the PR doesn't call out this scoping decision, it's worth the maintainers confirming whether one continuous top-to-bottom drag (expanded → collapsed → dismissed) is expected, or whether the current two-gesture behavior is fine.

---

### Verdict

No blocking correctness defects in the gesture state machine, and this version directly addresses the two rounds of prior feedback (drawer actually moves with the finger; "Swipe up to see more" replaces the More button). The stale-measurement issue (Medium) is a real, reproducible visual bug worth fixing before merge; the other two are lower-confidence/scope items for the maintainers to weigh in on.

---

KallynGowdy:
This is a lot better! The only thing I'd mention as far as UX is that after dragging the toolbar up, it snaps back down afterwards. Ideally it wouldn't snap back down but would just stay where it is.

Blitzy:
@mazenemad This is definitely in the right direction.

I found some issues while testing:

1. If you tap the handle to open it, it will select verse text that is behind the handle.

https://github.com/user-attachments/assets/d156409d-00e9-4eea-9c5f-f7e240790d6a

2. When tapping the handle to open or close the drawer, the transition appears to snap the drawer slightly above or below before moving it to its new position. This makes the transition animation appear a little broken/janky,

https://github.com/user-attachments/assets/528410db-1ff7-4679-9970-aca0dbd3bd34

3. When dragging the handle to close the drawer, you should be able to drag it all the way to the bottom of the screen to completely close it. Right now its two stages while expanded, you have to drag to collapse the extra buttons, and then you have to drag again to close it.

https://github.com/user-attachments/assets/d837019f-271a-41b4-9322-f8c3c763230b

Blitzy (commented):
> [!NOTE]

## Code Review — PR #1539 (latest commit `c5016a8`)

This is the third round of review on this PR. I compared the latest commit against the two prior automated reviews and the human feedback from KallynGowdy and Blitzy (July 29–30), and re-verified each open item by reading the actual diff rather than assuming prior comments still apply. I also ran the unit tests, `tsc`, and `eslint` for the touched files — all pass, and CI is green on `c5016a8`.

### Good news: all three requirements from the linked issue are now implemented, and every human-tested bug from the last round appears fixed

Issue No. 1519 asked for drag-up-to-expand, tap-to-expand (Blitzy's comment), and drag-down-to-close (KallynGowdy's comment). All three are present in `BibleReaderToolbar.tsx`. The three bugs Blitzy filmed on July 30 (tapping the handle selects the verse text behind it; the toggle transition looked "broken/janky"; closing was a two-stage drag) all look resolved by this rewrite:

- Text-selection-behind-handle → `user-select: none` (with `-webkit-*` variants) was added to `.sb-verse-toolbar-handle-area` in `BibleReaderToolbar.css`, with a comment explaining exactly this failure mode.
- Janky snap on tap → the sheet no longer moves at all until the drag exceeds `VERSE_SHEET_TAP_SLOP` (6px), so a tap animates from its actual resting position instead of a nudged one.
- Two-stage close → `verseSheetDismissOffset` is now derived from the same continuous `dy` as the reveal (`Math.max(0, dy - drag.startReveal)`), so one downward drag from the expanded state collapses the overflow row and then carries into a dismiss, per Kallyn's July 30 video.

KallynGowdy's "it snaps back down after I drag it up" comment is also addressed: `settleVerseSheet` now settles in the direction the drag traveled rather than checking a midpoint, so any upward lift stays open.

### Medium — Overflow row height is still measured once and never re-measured (carried over from the previous review, unresolved)

`measureVerseSheetOverflow` (`BibleReaderToolbar.tsx`, the `ref` on the overflow row, around line 2156) still only runs from a ref callback, which fires on mount/unmount, not on every render. I checked specifically for a `ResizeObserver` or a resize-driven re-measure and found none. This is the same issue flagged in the last review and it doesn't look like it was addressed in this commit.

**Reproduction:**
1. On a mobile viewport, select a verse with enough registered tools that the overflow row wraps onto 2 lines.
2. Rotate the device to landscape (or resize the window wider), so the same cards would now fit on 1 line.
3. Drag the handle open. The revealed height still targets the old 2-line measurement, leaving a visible gap under the last row of actions (or, in the opposite rotation direction, clipping the last row).

**Suggested fix:** watch the row with a `ResizeObserver` and update `verseSheetOverflowHeight` on change, or at minimum re-measure in the "selection cleared" `useEffect` (around line 1041) so a new selection never starts from a stale measurement, and hook into whatever resize handling already exists for `isSmallScreen` in this component.

### Low — Handle still has no `onClick` fallback (carried over, unresolved)

`.sb-verse-toolbar-handle-area` (the `role="button"` div) wires `onPointerDown/Move/Up/Cancel` and `onKeyDown`, but still no `onClick`. Some assistive-tech "activate" gestures on a custom `role="button"` dispatch a synthetic `click` rather than a full pointer sequence, which this element wouldn't respond to. Low confidence this is currently observable in the field, but since the element already opts into full button semantics (`role`, `aria-expanded`, focus-visible outline), a defensive `onClick` mirroring the tap-toggle (gated on `hasVerseSheetOverflow.value`, same as the keyboard handler) would close the gap.

### Low — Orphaned translation key

`i18n/en.json` still has `"less": "Less"`, but the "More"/"Less" toggle button that used it was removed in this rewrite. I grepped the codebase and found no remaining `t("less", ...)` call anywhere (`"more"` is still used elsewhere, e.g. `Tabs.tsx` and the mobile nav's "more" tab, so that one's fine — it's specifically `"less"` that's now dead). Not a functional bug, just worth deleting from `en.json` per the repo's normal cleanup of unused keys.

### Low — A second pointer on the handle mid-drag hijacks the gesture

`handleVerseSheetHandlePointerDown` unconditionally overwrites `verseSheetDrag.current` with the new pointer's data, even if `verseSheetDrag.current` is already set from an in-progress drag with a different `pointerId`. If a second finger touches the handle while the first is still dragging, the first pointer's capture is silently orphaned (its `pointerup`/`pointercancel` will no longer match `drag.pointerId` and will be ignored, so it never gets released or settled). This is an unlikely gesture on a ~1.5rem-tall handle, so I'd treat this as a nice-to-have rather than a blocker — a simple guard (`if (verseSheetDrag.current) return;` at the top of the pointerdown handler) would close it if you want the extra robustness.

---

### On the diff size (220 files, +3445/-23744) shown on this PR

That number is not from this feature — I checked directly. This branch (`fix/drag-verse-toolbar-handle-to-expand`) forked from an old point in the tree (commit `6c75f8b`, "Fix missing space from beginning of verse", July 28) and hasn't been rebased or merge-updated against `develop` since. `develop` has since moved on with a large set of unrelated commits, so GitHub's PR diff (head vs. current `develop`) shows all of that as if it were being deleted by this branch — it isn't. The actual commits on this branch touch exactly 4 files: `BibleReaderToolbar.tsx`, `BibleReaderToolbar.css`, `i18n/en.json`, and the new test file (`+914/-72` combined). GitHub's own `mergeable_state` for this PR is currently `dirty`, consistent with this. **Recommend merging/rebasing onto current `develop` before this lands**, both so the PR diff reviewers see matches the actual change, and so merging doesn't risk reverting unrelated work that landed on `develop` in the meantime.

### Verdict

No blocking correctness defects in the gesture state machine itself, and this revision closes out every prior human-reported bug and both requirements from the linked issue. The stale-measurement issue (Medium) is a real, reproducible visual bug carried over from the last review and still worth fixing before merge. The remaining items are low-severity/defense-in-depth. Separately (not a code issue, but blocking a clean merge): **this branch needs to be updated against `develop`** before merging, given how far it's diverged.

---

KallynGowdy:
@mazenemad Looks like c5016a88897c5b1151486998016e8ab5b935cc9e actually undoes a bunch of changes on `develop`. Was there some sort of git issue?

KallynGowdy:
This will be merged and implemented when https://github.com/HelloAOLab/seed-bible/pull/1593 is merged

PR #1542, opened 2026-07-28
Prevent deployments from using special characters in branch names

KallynGowdy:
Currently, all special characters are allowed for the generated deployments. After this PR is merged, only alphanumeric (A-Z, a-z, 0-9), hyphen (-), and underscore (_) are allowed. All other symbols will be converted to an underscore (_).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -70,7 +70,7 @@ jobs:
         id: ids
         run: |
           RAW="${{ github.head_ref || github.ref_name }}"
-          BRANCH=$(echo "$RAW" | sed 's/\//_/g')
+          BRANCH=$(echo "$RAW" | sed 's/[^a-zA-Z0-9]/_/g')
           echo "branch=$BRANCH" >> "$GITHUB_OUTPUT"
           echo "build_id=${{ github.sha }}" >> "$GITHUB_OUTPUT"
 
@@ -252,7 +252,7 @@ jobs:
         with:
           script: |
             const commentDeployment = require("./.github/scripts/comment-deployment.cjs");
-            const branch = "${{ github.head_ref }}".replace(/\//g, "_");
+            const branch = "${{ github.head_ref }}".replace(/[^a-zA-Z0-9]/g, "_");
             await commentDeployment({
               github,
               context,
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -249,10 +249,12 @@ jobs:
       - name: Comment on pull request with deployment link
         continue-on-error: true
         uses: actions/github-script@v7
+        env:
+          HEAD_REF: ${{ github.head_ref }}
         with:
           script: |
             const commentDeployment = require("./.github/scripts/comment-deployment.cjs");
-            const branch = "${{ github.head_ref }}".replace(/[^a-zA-Z0-9-]/g, "_");
+            const branch = process.env.HEAD_REF.replace(/[^a-zA-Z0-9-]/g, "_");
             await commentDeployment({
               github,
               context,
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -70,7 +70,7 @@ jobs:
         id: ids
         run: |
           RAW="${{ github.head_ref || github.ref_name }}"
-          BRANCH=$(echo "$RAW" | sed 's/[^a-zA-Z0-9]/_/g')
+          BRANCH=$(echo "$RAW" | sed 's/[^a-zA-Z0-9-]/_/g')
           echo "branch=$BRANCH" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


Blitzy:
> [!NOTE]

## Code Review — Prevent deployments from using special characters in branch names

### Summary

Small, well-targeted change (2 files? no — 1 file, +2/-2). It swaps the branch-name sanitizer in `.github/workflows/cd.yml` from "replace only `/`" to "replace anything that isn't `A–Z a–z 0–9 -`" in the two places the sanitized branch name is produced:

- The `deploy` job's `Compute branch + build id` step (`sed 's/[^a-zA-Z0-9-]/_/g'`)
- The `comment-existing-deployment` job (`.replace(/[^a-zA-Z0-9-]/g, "_")`)

**Overall this looks correct and safe to merge.** The two regexes are now identical in character class, the trailing `-` is a literal in both a `sed` bracket expression and a JS character class (so hyphens really are preserved, not treated as a range), and the common case is unaffected — a name like `feature/foo-bar` still resolves to `feature_foo-bar` exactly as before. The live deployment comment on this PR confirms the behavior end-to-end: `claude/branch-name-special-chars-3ctwgp` → `claude_branch-name-special-chars-3ctwgp` (slash converted, hyphens kept).

I also checked that these are the **only** two places that need to change. The host (`server/index.ts`) reads the branch straight from the URL path segment via `decodeURIComponent(branchSeg)` and does **not** re-sanitize, so it just trusts whatever key the workflow wrote to S3 and put in the comment link. The `notify` step and the `__invalidate` ping both consume the already-sanitized `$BRANCH` value. So there is no third sanitizer that could drift out of sync. Good.

Nothing here rises to a blocking issue. The items below are all **Low / informational**.

---

### Low — the `pull_request` job still interpolates untrusted input directly into the script body

`.github/workflows/cd.yml:255`

```js
const branch = "${{ github.head_ref }}".replace(/[^a-zA-Z0-9-]/g, "_");
```

`${{ github.head_ref }}` is expanded by the Actions runner into the `script:` text *before* the JavaScript runs, so a branch name is being embedded directly inside a `"..."` string literal. This is the classic GitHub Actions script-injection pattern: if a branch name contains a character that terminates the string literal, the remainder is interpreted as code. Git's ref-name rules block many metacharacters (spaces, `~ ^ : ? * [ \`), which makes this hard to exploit in practice and it requires push access to create the branch — hence Low, not High — but it is avoidable defense-in-depth.

Notably, the sibling `deploy` job already does this the safe way (`.github/workflows/cd.yml:198-204`): it passes the value through `env:` and reads `process.env.BRANCH` inside the script. Since this PR is editing this exact line anyway, it's a natural moment to bring the two jobs in line.

**Suggested fix** — move `head_ref` into the step env and read it from `process.env`, then sanitize in JS:

```yaml
      - name: Comment on pull request with deployment link
        continue-on-error: true
        env:
          HEAD_REF: ${{ github.head_ref }}
        uses: actions/github-script@v7
        with:
          script: |
            const commentDeployment = require("./.github/scripts/comment-deployment.cjs");
            const branch = process.env.HEAD_REF.replace(/[^a-zA-Z0-9-]/g, "_");
            ...
```

This keeps the branch name entirely in data-space and matches the pattern already used a few lines up.

---

### Low / informational — the sanitizer widens the branch-name collision surface

Because every disallowed character now collapses to `_`, distinct branch names can map to the same deploy key. For example `feat/x`, `feat.x`, and `feat_x` all sanitize to `feat_x`, so they would share the same S3 prefix (`branches/feat_x/...`), the same `current.json` pointer, and the same `/b/feat_x` URL — the last deploy wins.

This is not a regression introduced by this PR in kind — the old `/`-only replacement already collided `feat/x` with `feat_x` — but the change broadens the set of names that can collide. In practice branch-naming conventions make this unlikely, so I'd treat it as a known limitation rather than something to fix here. Worth a one-line comment in the workflow (or the PR description) so the trade-off is on record.

**Reproduction (illustrative):** push `feat/x`, let it deploy, then push `feat.x`; both resolve to `branches/feat_x/` and the second overwrites the first's `current.json`.

---

### Informational — no automated coverage; manual verification stands in

Workflow-shell logic isn't unit-testable in this repo, so there's no test to add here. The behavior is instead demonstrated by the deployment-link comment on this PR (see above). If you want a cheap regression guard for the JS half, the `comment-deployment` path or the sanitize expression could be factored into `.github/scripts/` and given a tiny Vitest case, but that's optional and out of scope for a change this size.

---

### Nits

- `github.ref_name` (the `deploy` job's fallback on `push`) and `github.head_ref` (the `pull_request` job) are both pre-existing and correct for their respective triggers — no action needed, just noting I checked they weren't affected.

### Verdict

Approve-level change on correctness. The only thing I'd genuinely encourage before/after merge is the Low item above (env-passing the branch name in the `pull_request` job) so both jobs handle the branch name the same, safe way.

---

PR #1543, opened 2026-07-28
Fix feature flags

KallynGowdy:
Previously, every feature flag behaved as if it was disabled. Now, after this fix, feature flags that are configured in PostHog are detected correctly.

### DIFF
--- a/packages/seed-bible/seed-bible/app/initPostHog.tsx
+++ b/packages/seed-bible/seed-bible/app/initPostHog.tsx
@@ -14,6 +14,7 @@ if (
   (globalThis as any).posthog = {
     init: () => {},
     capture: () => {},
+    onFeatureFlags: () => {},
     identify: () => {},
     setPersonProperties: () => {},
   };
@@ -24,10 +25,13 @@ if (
   !(function (t, e) {
     var o, n, p, r;
     e.__SV ||
-      (((window as any).posthog = e),
+      // @ts-ignore
+      (window.posthog && window.posthog.__loaded) ||
+      // @ts-ignore
+      ((window.posthog = e),
       (e._i = []),
-      (e.init = function (i: any, s: any, a: any) {
-        function g(t: any, e: any) {
+      (e.init = function (i, s, a) {
+        function g(t, e) {
           var o = e.split(".");
           (2 == o.length && ((t = t[o[0]]), (e = o[1])),
             (t[e] = function () {
@@ -50,7 +54,7 @@ if (
         for (
           void 0 !== a ? (u = e[a] = []) : (a = "posthog"),
             u.people = u.people || [],
-            u.toString = function (t: any) {
+            u.toString = function (t) {
               var e = "posthog";
               return (
                 "posthog" !== a && (e += "." + a),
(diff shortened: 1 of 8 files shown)


Blitzy:
> [!NOTE]

## Summary

This PR fixes feature flags that were previously always reported as disabled. The fix has two parts:

1. **`FeaturesManager`** now subscribes to PostHog's `onFeatureFlags` push and exposes each flag as a reactive `ReadonlySignal<boolean>` instead of a plain synchronous `boolean`. The old code called `posthog.isFeatureEnabled()` synchronously at a time when flags weren't loaded yet, so it always fell back to "off".
2. **`initPostHog`** is bumped to the newer PostHog snippet (`defaults: "2026-05-30"`), adds a `__loaded` re-init guard, and stubs `onFeatureFlags` in the SSR/dev-disabled path.

**Overall this looks correct and I could not find a functional bug that breaks the fix.** Tests are added, `build (24.x)` and the deploy checks are green, and the alpha deployment succeeded. The findings below are all low-severity / polish. No linked issue is referenced in the PR description — if there's a tracking issue for "feature flags always disabled", linking it would help traceability.

### One thing I checked and want to put to rest

`getToolbarTools()` runs inside a `useComputed` in `BibleReaderToolbar.tsx`, and `isFeatureEnabled()` writes to a signal (`s.value = …`) on every call. Writing to a signal while a computed is evaluating *looks* dangerous. I pulled the `@preact/signals-core` source to confirm: the `Signal` value setter has **no** "side-effect in computed" guard and only notifies targets when `value !== this._value`. Because the `onFeatureFlags` callback keeps each per-key signal in sync with `flags`, the write inside `isFeatureEnabled` is always a no-op in steady state — so there is **no throw and no recompute cycle**. The write-on-read is in fact load-bearing: it seeds a feature signal that gets created *after* a flags push (otherwise it would be stuck at its initial `false` until the next push). So the design is sound, just non-obvious.

---

## Findings

### 🟡 Low — DEV / SSR override is not honored by the `onFeatureFlags` callback

`isFeatureEnabled` forces `true` in DEV and `false` in SSR / no-posthog:

```ts
if (import.meta.env.DEV) {
  s.value = true;
} else if (import.meta.env.SSR || !posthog) {
  s.value = false;
} else {
  s.value = flags.peek().includes(featureKey);
}
```

…but the `onFeatureFlags` callback ignores those cases and unconditionally overwrites the signal with the raw flag value:

```ts
posthog.onFeatureFlags((f) => {
  flags.value = f;
  for (const [featureKey, s] of signals.entries()) {
    const featureEnabled = flags.peek().includes(featureKey);
    s.value = featureEnabled ?? false; // ignores DEV/SSR
  }
});
```

So in DEV *with a real PostHog client* (`VITE_ENABLE_POSTHOG` enabled), a flags push flips a dev-forced `true` back to whatever PostHog actually says, and the final value ends up depending on whether `isFeatureEnabled` or the callback ran last. The "DEV enables everything" contract silently breaks.

**Why it's only Low:** the default dev setup stubs PostHog with a no-op `onFeatureFlags` (see `initPostHog.tsx`), so the callback never fires and DEV stays `true`. It only bites when PostHog is deliberately turned on in dev.

**Repro (requires `VITE_ENABLE_POSTHOG`):**
1. `pnpm dev` with `VITE_ENABLE_POSTHOG=true`, pointing at a PostHog project where `reading-plans` is **off**.
2. On load (before flags arrive) the Reading Plans toolbar button shows (DEV forces on).
3. Once PostHog pushes flags, the callback sets the signal to `false` and the button disappears — contradicting the dev-always-on behavior.

**Suggested fix:** centralize the DEV/SSR logic so it can't be bypassed. A per-key `computed` derived from a single `flags` signal removes the manual writes, the manual update loop, and the redundant coalescing all at once — and it still passes every test in the new `FeaturesManager.test.ts`:

```ts
export function createFeaturesManager(posthog: PostHog | null) {
  const flags = signal<string[]>([]);
  const signals = new Map<string, ReadonlySignal<boolean>>();

  const isFeatureEnabled = (featureKey: string): ReadonlySignal<boolean> => {
    if (!signals.has(featureKey)) {
      signals.set(
        featureKey,
        computed(() => {
          if (import.meta.env.DEV) return true;
          if (import.meta.env.SSR || !posthog) return false;
          return flags.value.includes(featureKey);
        })
      );
    }
    return signals.get(featureKey)!;
  };

  if (posthog) {
    posthog.onFeatureFlags((f) => {
      flags.value = f;
    });
  }

  return { isFeatureEnabled };
}
```

This is fully reactive (the computeds re-derive when `flags` changes), read-only (no write-on-read side effect — addressing the "put to rest" note above at the source), and keeps DEV/SSR handling in exactly one place.

### 🟡 Low — two component test mocks now return the wrong type

`isFeatureEnabled` used to return `boolean` and now returns `ReadonlySignal<boolean>`. The mock in `BibleToolsManager.test.tsx` was correctly updated to `vi.fn(() => signal(true))`, but three call sites were missed and still return a raw boolean:

- `test/unit/seed-bible/components/BibleReader.test.tsx:242` — `isFeatureEnabled: vi.fn(() => true)`
- `test/integration/seed-bible/components/TabSlotReader.test.tsx:230` — `isFeatureEnabled: vi.fn(() => true)`
- `test/integration/seed-bible/components/TabSlotReader.test.tsx:260` — `isFeatureEnabled: vi.fn(() => true)`

Production reads `.value` on the result. With a raw boolean, `true.value` is `undefined` (JS auto-boxes the primitive, no error), which is falsy — so under these mocks the Reading Plans tool is treated as **hidden**, the opposite of what `() => true` implies. The tests still pass only because they don't assert on that tool's visibility, so the mocks are silently masking their own intent and won't catch a regression there.

**Suggested fix:** match the real signature, mirroring the `BibleToolsManager.test.tsx` change:

```ts
isFeatureEnabled: vi.fn(() => signal(true)),
```

### ⚪ Nit — redundant `?? false`

In the `onFeatureFlags` callback (`FeaturesManager.tsx:39-40`):

```ts
const featureEnabled = flags.peek().includes(featureKey);
s.value = featureEnabled ?? false;
```

`Array.prototype.includes` always returns a `boolean`, so `featureEnabled` can never be nullish and `?? false` is dead code. It disappears entirely under the `computed`-based refactor suggested above.

### ⚪ Nit — no CHANGELOG entry

The repo keeps a `CHANGELOG.md` with a `## TBD` section, and its `### 🐛 Fixed` list is currently empty. Since this is a user-visible fix (every feature flag was effectively off), consider adding an entry referencing this PR, consistent with recent history.

---

## Note on `initPostHog.tsx`

I verified the `getFeatureFlagPayload` → `getFeatureFlagResult` change in the snippet's stub-method list: this is just the method list from the newer official PostHog snippet (matching `defaults: "2026-05-30"`), and nothing in the codebase calls `getFeatureFlagPayload`, so dropping that pre-load stub is safe. The `window.posthog && window.posthog.__loaded` re-init guard is also standard for the newer snippet. No concerns here.

---

KallynGowdy:
Issue 1 can be safely ignored because enabling PostHog in dev should actually enable all PostHog features

PR #1545, opened 2026-07-28
Add test coverage reporting to CI workflow

KallynGowdy:
## Summary

This PR adds automated test coverage reporting to the CI workflow. When tests run, a markdown summary of coverage metrics is now generated and posted to the GitHub workflow summary, making coverage regressions visible without opening the full HTML report.

## Key Changes

- **New coverage library** (`script/lib/coverageSummary.ts`): Exports `renderCoverageSummary()`, which transforms vitest's `json-summary` coverage report into a markdown table. The output includes:
  - An overall metrics table (statements, branches, functions, lines)
  - A per-package breakdown sorted by worst line coverage first, so regressions are immediately visible
  - Graceful handling of files with zero coverable lines (renders "n/a" instead of crashing)

- **New CLI tool** (`script/coverage-summary.ts`): A commander-based script that reads a `coverage-summary.json` file and writes markdown to stdout or a file. Handles missing reports gracefully (writes a note instead of failing).

- **Vitest configuration** (`vite.config.ts`): Added coverage config with v8 provider, targeting `packages/**/*.{ts,tsx}` and excluding test files, type definitions, obsolete code, and patterns.

- **CI workflow** (`.github/workflows/ci.yml`):
  - Changed test step to run `pnpm test:coverage` instead of `pnpm test`
  - Added a new "Coverage summary" step that runs the CLI tool and appends output to `$GITHUB_STEP_SUMMARY`
  - Updated the PR comment step to include the coverage summary alongside the build size report (with fallback if coverage step fails)

- **Package scripts** (`package.json`): Added `test:coverage` script that runs vitest with `--coverage` flag.

- **Tests** (`test/unit/script/lib/coverageSummary.test.ts`): Full test suite covering metric rendering, package grouping, sorting, path normalization, and edge cases (empty files, missing reports).

## Implementation Details

- **Package grouping**: Files are grouped under `packages/<name>` (the workspace unit). Anything outside `packages/` falls back to its top-level directory or "other" for bare filenames.
- **Sorting**: Packages are sorted by line coverage percentage (worst first), with "Unknown" (zero coverable lines) sorted to the top.
- **Root prefix stripping**: Absolute paths are relativized before grouping, so the markdown is readable regardless of where the workflow runs.
- **Graceful degradation**: If the coverage report is missing, the workflow doesn't fail—it writes a note to the summary instead. If the coverage summary markdown is missing when posting the PR comment, only the build size report is included.

https://claude.ai/code/session_01TWNf6a659cdpmyBKHVrVCJ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,13 @@ jobs:
           pnpm check:ts
       - name: Test
         run: |
-          pnpm test
+          pnpm test:coverage
+      - name: Coverage summary
+        if: always()
+        continue-on-error: true
+        run: |
+          pnpm coverage-summary --summary coverage/coverage-summary.json --out coverage-summary.md
+          cat coverage-summary.md >> "$GITHUB_STEP_SUMMARY"
       - name: Build
         run: |
           pnpm build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,16 @@ jobs:
             const fs = require("fs");
             const marker = "<!-- build-size-report -->";
             try {
-              const body = `${marker}\n## 📦 Build Size Report\n\n${fs.readFileSync("size-report.md", "utf8")}`;
+              // Read independently of the try/catch below so a missing
+              // coverage-summary.md (e.g. the coverage step failed) only
+              // drops this section instead of skipping the whole comment.
+              let coverageSection = "";
+              try {
+                coverageSection = `\n\n## 🧪 Coverage Summary\n\n${fs.readFileSync("coverage-summary.md", "utf8")}`;
+              } catch (error) {
+                core.warning(`Failed to read coverage summary: ${error}`);
+              }
+              const body = `${marker}\n## 📦 Build Size Report\n\n${fs.readFileSync("size-report.md", "utf8")}${coverageSection}`;
               const { data: comments } = await github.rest.issues.listComments({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ pattern-dist
 size-head.json
 size-base.json
 size-report.md
+coverage-summary.md
 .DS_Store
 package-lock.json
 .claude
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "test": "vitest",
     "test:watch": "vitest watch",
+    "test:coverage": "vitest run --coverage",
     "dev": "bun server/index.ts",
     "build": "pnpm build:client && pnpm build:ssr && pnpm build:server",
     "build:client": "vite build",
@@ -76,6 +77,7 @@
     "@types/prompts": "^2.4.9",
     "@typescript-eslint/utils": "^8.58.0",
     "@vitejs/plugin-react": "^6.0.2",
+    "@vitest/coverage-v8": "^4.1.8",
     "axios": "catalog:",
     "babel-jest": "^30.1.2",
     "casualos": "4.2.4-alpha.25223847785",
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "@types/prompts": "^2.4.9",
     "@typescript-eslint/utils": "^8.58.0",
     "@vitejs/plugin-react": "^6.0.2",
-    "@vitest/coverage-v8": "^4.1.8",
+    "@vitest/coverage-v8": "4.1.8",
     "axios": "catalog:",
     "babel-jest": "^30.1.2",
     "casualos": "4.2.4-alpha.25223847785",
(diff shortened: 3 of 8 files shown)


Blitzy:
> [!NOTE]

## Code Review — Add test coverage reporting to CI workflow

**Summary:** This is a clean, well-tested, and well-documented change. The `coverageSummary.ts` library has good unit coverage, the CI wiring degrades gracefully when a report is missing, and the comment/summary steps are correctly placed in the same job so `coverage-summary.md` is available where it's read. CI is currently green. No correctness bugs were found in the rendering logic. The notes below are mostly polish and one dependency-hygiene item.

There is no linked issue on this PR and no prior review discussion (only the deployment bot comment), so this review is against the PR description alone.

---

### Medium

**1. `@vitest/coverage-v8` and `vitest` versions don't match → unmet peer dependency**

`package.json` pins `"@vitest/coverage-v8": "^4.1.8"`, which resolves to **4.1.10** in the lockfile, while `vitest` is locked at **4.1.8**:

```
'@vitest/coverage-v8@4.1.10(vitest@4.1.8)':
    peerDependencies:
      '@vitest/browser': 4.1.10
      vitest: 4.1.10          # <-- exact match required, but 4.1.8 is installed
```

`@vitest/coverage-v8` declares an **exact** peer dependency on `vitest` (`4.1.10`), so installing it against `vitest@4.1.8` leaves an unmet peer. It happens to work today (CI is green), but the vitest team ships coverage-v8 and vitest as a matched set, and a future minor bump on one side without the other can produce real breakage rather than just a warning.

_Reproduction:_ run `pnpm install` on the branch and watch for an unmet-peer warning for `@vitest/coverage-v8` (expecting `vitest@4.1.10`, found `4.1.8`).

_Suggested fix:_ pin coverage-v8 to the same version vitest actually resolves to so the two always move together, e.g.

```jsonc
"@vitest/coverage-v8": "4.1.8",  // match vitest exactly (drop the caret)
```

or bump `vitest` to `4.1.10` in the same PR. Either keeps the pair aligned.

---

### Low

**2. `coverage-summary.md` is not gitignored (inconsistent with `size-report.md`)**

`.gitignore` ignores `coverage/` (line 16) and `size-report.md` (line 7), but not `coverage-summary.md`. The new CI step writes `coverage-summary.md` to the repo root, and anyone running `pnpm coverage-summary --out coverage-summary.md` locally will get an untracked file at the root that's easy to commit by accident.

_Reproduction:_ run `pnpm test:coverage && pnpm coverage-summary --summary coverage/coverage-summary.json --out coverage-summary.md`, then `git status` — `coverage-summary.md` shows as untracked.

_Suggested fix:_ add `coverage-summary.md` to `.gitignore` next to `size-report.md`.

**3. `program.parse()` doesn't await the async action in `coverage-summary.ts`**

The `.action()` handler is `async`, but the script calls `program.parse()` rather than `program.parseAsync()`. Commander ignores the promise returned by the action, so the async work isn't awaited and any rejection (e.g. malformed JSON in the summary file) surfaces as an unhandled promise rejection instead of a clean error. It works in practice because pending fs I/O keeps the event loop alive, but it's fragile.

_Suggested fix:_ replace `program.parse();` with `await program.parseAsync();` (top-level `await` is fine here, or wrap in an async main). This guarantees the action completes and lets errors propagate through commander's normal error path.

---

### Nits (optional)

**4. `patterns/**` in the coverage `exclude` is a no-op.** The coverage `include` is `packages/**/*.{ts,tsx}`, and `patterns/` lives at the repo root (there are no `patterns/` directories under `packages/`), so nothing the include glob matches can ever hit that exclude. Harmless, but it can mislead a future reader into thinking pattern code lives under `packages/`. Consider dropping it, or add a comment noting it's defensive.

**5. The CLI wrapper `script/coverage-summary.ts` has no direct test.** `coverageSummary.ts` is thoroughly tested, but the CLI's own behavior — specifically the "no report found" branch that writes the fallback note — isn't exercised. It's thin glue, so this is low priority, but a small test around the missing-file path would lock in the graceful-degradation behavior the CI relies on.

**6. Packages with no coverable lines sort to the top as "worst".** In `renderCoverageSummary`, an `"Unknown"` line percentage (zero coverable lines) sorts before the lowest real percentage. A package with nothing to cover isn't a regression, so surfacing it above genuinely low-coverage packages could draw the eye to the wrong row. This is a deliberate, documented choice — flagging only in case sorting `"Unknown"` to the bottom reads better for the "worst first" intent.

---

### Note on scope (informational, not a defect)

The `pnpm-lock.yaml` diff adds catalog entries for `@casual-simulation/aux-common`, `@casual-simulation/aux-websocket`, `@casual-simulation/websocket`, `base64-js`, and `luxon` that are unrelated to coverage. These are **not** new dependencies — they're already declared in `pnpm-workspace.yaml`'s `catalog:` block, so this is just the lockfile catching up to a previously stale state during `pnpm install`. Benign, but worth a glance to confirm it's expected churn rather than something to split out.

---

PR #1546, opened 2026-07-28
Add automatic sign-out when session expires or account is suspended

KallynGowdy:
## Summary

Adds automatic session invalidation detection to the app. When the CasualOS API reports that a session key is dead (expired, unrecognised, or the account is banned), the app now signs the user out immediately without waiting for the user to notice or manually sign out. The user sees a toast explaining why they were signed out.

## Key Changes

- **SessionGuard** (`packages/seed-bible/seed-bible/managers/SessionGuard.ts`): New module that wraps the records client with a Proxy. Every API call result is checked for three fatal error codes (`session_expired`, `invalid_key`, `user_is_banned`). When one is detected, a callback fires to notify the app that the session is dead. The guard is careful not to sign out the wrong session — it compares the session key at request time to the current key, so if the user signed out or signed back in while a request was in flight, that stale failure is ignored. A burst of simultaneous failures (e.g., two init requests both hitting an expired key) collapses into a single sign-out.

- **OsManager** (`packages/seed-bible/seed-bible/managers/OsManager.tsx`): Wraps the raw CasualOS records client with `guardRecordsClient`. Exposes a `sessionInvalidated` signal that fires when the guard detects a dead session.

- **LoginManager** (`packages/seed-bible/seed-bible/managers/LoginManager.tsx`): Watches `os.sessionInvalidated` and calls `forceLogout()` when it fires. `forceLogout()` clears the session from memory and storage, sets a `sessionEnded` signal with a reason (`"session_expired"` or `"account_suspended"`), and does **not** call `revokeSession` (the session is already gone server-side). Also improved the refresh logic to check if the session changed while the request was in flight, preventing a successful refresh from resurrecting a session that was just dropped.

- **SeedBibleStateManager** (`packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx`): Watches `login.sessionEnded` and shows a toast with a user-friendly message explaining the sign-out reason.

- **Tests**: Added comprehensive test suites for `SessionGuard` (logic against a fake client) and `SessionGuardRecordsClient` (integration with the real CasualOS client). Extended `LoginManager` tests to cover automatic sign-out scenarios, including the case where multiple requests fail simultaneously. Updated `SeedBibleStateManager` tests to verify toast messages.

- **Network failures don't sign anyone out**: Only `.then()` is hooked, never `.catch()`. A rejection (offline, DNS, CORS, a 500 with no body) passes through unobserved and never triggers a sign-out. This is intentional — a blip on a mobile connection is not the same as a dead session.

- **`revokeSession` is exempt**: Failures from `revokeSession` are never reported as a dead session. A deliberate sign-out often comes back `session_expired` (the key was already dead, which is often why the user is signing out), and reacting to that would tell someone who just pressed "Sign out" that their session had expired.

- **Monotonic IDs prevent silent failures**: Both `SessionInvalidatedEvent` and `SessionEndedEvent` carry an incrementing `id`. Preact signals skip notification when the new value is `===` the old one, so without the id a second `session_expired` (after signing back in, say) would silently fail to notify anyone.

- **Mid-flight checks prevent resurrection**: `refreshSession()` and `loadUserInfo()` both peek at the session key before and after their request. If the session changed (the user signed out, or this very request reported the key dead), the success branch is skipped, preventing a successful refresh from resurrecting a session we just dropped.

- **Transient failures are survivable**: The refresh no longer clears the session on failure. A refresh fails for transient reasons far more often than real ones (a mobile dead spot, a 500, a rate limit), and the key is usually still good for another week. The three codes that

https://claude.ai/code/session_01AnL8i8W2DPdbnfUwdK9TjR

### DIFF
--- a/packages/seed-bible/seed-bible/i18n/en.json
+++ b/packages/seed-bible/seed-bible/i18n/en.json
@@ -420,6 +420,6 @@
   "start-share-session-subtitle": "Invite others to read along live",
   "share-current-session": "Share current session",
   "share-current-session-subtitle": "Copy a link to invite others to read along live",
-  "session-expired-message": "Your session has expired. Please sign in again.",
+  "signed-out-message": "You've been signed out. Please sign in again.",
   "account-suspended-message": "Your account has been suspended."
 }
--- a/packages/seed-bible/seed-bible/i18n/en.json
+++ b/packages/seed-bible/seed-bible/i18n/en.json
@@ -419,5 +419,7 @@
   "start-share-session": "Start and share session",
   "start-share-session-subtitle": "Invite others to read along live",
   "share-current-session": "Share current session",
-  "share-current-session-subtitle": "Copy a link to invite others to read along live"
+  "share-current-session-subtitle": "Copy a link to invite others to read along live",
+  "session-expired-message": "Your session has expired. Please sign in again.",
+  "account-suspended-message": "Your account has been suspended."
 }
--- a/packages/seed-bible/seed-bible/managers/LoginManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/LoginManager.tsx
@@ -1,11 +1,31 @@
 import { batch, computed, effect, signal, type Signal } from "@preact/signals";
 import * as z from "zod/v4";
 import type { CasualOSManager, UserInfo } from "./OsManager";
+import type { FatalSessionErrorCode } from "./SessionGuard";
 import type {
   CompleteLoginResult,
   LoginRequestResult,
   LoginRequestSuccess,
 } from "@casual-simulation/aux-records/AuthController";
+
+/**
+ * Why a session ended without the user asking it to.
+ *
+ * The three server error codes collapse to two cases the UI has to explain, so the
+ * mapping happens here rather than in the view — a view only ever needs two messages,
+ * and widening the set of fatal codes later changes no view code.
+ */
+export type SessionEndedReason = "session_expired" | "account_suspended";
+
+export interface SessionEndedEvent {
+  reason: SessionEndedReason;
+
+  /**
+   * Monotonically increasing id, so two events with the same reason still notify
+   * subscribers (signals skip notification when the new value is `===` the old one).
+   */
+  id: number;
+}
 
 export interface LoginManager {
   /**
@@ -27,6 +47,16 @@ export interface LoginManager {
    * The current auth bot. Null if not authenticated or if background auth has not completed yet.
    */
   authBot: Signal<UserInfo | null>;
+
+  /**
+   * Fires when the user was signed out without asking — because the server reported
+   * their session key dead, or their account suspended. Null until that happens.
+   *
+   * Only set when a forced sign-out actually took place, so a sign-out the user asked
+   * for and a request that merely failed never produce an event. The UI reads `reason`
+   * to pick which message to show.
+   */
+  sessionEnded: Signal<SessionEndedEvent | null>;
 
   /**
    * The user's profile information. Null if the user is not logged in or if the profile has not loaded yet.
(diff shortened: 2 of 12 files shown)


Blitzy:
> [!NOTE]

## Code review — Add automatic sign-out when session expires or account is suspended

**Overall:** This is a strong, carefully-reasoned change. The central design decision — put one `Proxy` in front of the records client so every current and future call site is covered, instead of sprinkling checks around — is the right call, and the guard is conservative in all the ways that matter: only `.then` is hooked (never `.catch`), only three fatal codes count, `revokeSession` is exempt, anonymous requests are ignored, and a "which session was this for?" comparison collapses simultaneous failures and prevents signing out the wrong session. The mid-flight checks in `refreshSession`/`loadUserInfo` and the monotonic-id trick on both events are correct and genuinely necessary. Test coverage is excellent, including an integration suite against the real SDK client.

I traced the main flows (dead-session detection, burst-collapse, deliberate logout during in-flight failures, refresh-after-forced-signout resurrection) and found the logic sound. No linked issue is attached to the PR, so I could not cross-check intended scope against a spec. The findings below are all **low severity / advisory** — nothing here blocks the design.

---

### Low — Auto sign-out is scoped to HTTP records-client calls only

`guardRecordsClient` wraps the HTTP records client. The websocket/inst path (`getInstClient` → `AuthenticatedConnectionClient`, and `RemoteYjsSharedDocument` for multiplayer/Yjs) shadows `client` with the inst client (`OsManager.tsx:224`, `:133–136`) and is **not** guarded. This is a clean separation and is probably the right scope — but it means a session that dies while the user is sitting in a live shared session, making no HTTP records calls, won't auto-sign-out until the next records call fires.

- **Impact:** minor and likely acceptable; a records call almost always follows soon.
- **Suggestion:** add a one-line note to the `guardRecordsClient` doc-comment stating the guard covers HTTP records calls only, so the websocket gap is a documented decision rather than a surprise later.

### Low — `invalid_key` is shown to the user as "session expired"

`forceLogout` maps everything except `user_is_banned` to `session_expired`, whose toast reads *"Your session has expired. Please sign in again."* (`LoginManager.tsx:215–221`, `en.json`). `invalid_key` also covers *"the server no longer recognises this key"* — e.g. the session was revoked from another device — where "expired" is slightly inaccurate.

- **Impact:** cosmetic; the user still ends up correctly signed out.
- **Suggestion:** either accept the generalization (reasonable) or use slightly more generic wording such as *"Your session has ended. Please sign in again."*

### Low / Informational — Fragile i18n lint workaround

The toast destructures `const { t } = i18n` specifically because the translation lint rule only recognizes calls to a bare `t`, and `translation-unused-keys` is auto-fixable (`SeedBibleStateManager.tsx`, comment at the effect). This is well-documented, but it's a booby-trap: a future "cleanup" that rewrites this to `i18n.t("...")` would let the auto-fixer **silently delete** the two new keys from `en.json`.

- **Suggestion:** consider a tiny `// eslint-disable-next-line <rule>`-style anchor or a shared helper so the coupling is enforced rather than relying on the next editor reading the comment. The inline `defaultValue` fallbacks already soften the blast radius, which is good.

### Nit — Guard returns a fresh wrapper function per property access

Each `get` on a method builds a new arrow function, so `wrapped.getData !== wrapped.getData`. Harmless in current usage (every caller invokes the method immediately), but worth being aware of if anything ever memoizes or identity-compares a client method reference.

---

### Verification notes for the human reviewer

- **CI:** the standalone `deploy (24.x)` job shows a failure, but that appears to be the deploy/secrets step rather than the code — the `build (24.x)` job (lint + type-check + tests) was still **in progress** at review time on the head commit (`a0eea13`). Please confirm build goes green before merge; `mergeable_state` is currently `blocked`.
- **No linked issue** on the PR, so scope was reviewed against the PR description only.
- The `deploy (24.x)` failure is worth a glance to confirm it's infra, not a regression, but I did not see anything in the diff that would break the build.

Nice work — the comments-as-rationale style throughout `SessionGuard.ts` and `LoginManager.tsx` made this easy to verify, and the "why not `.catch`", "why exempt `revokeSession`", and "why the monotonic id" reasoning are all spot-on.

---

PR #1547, opened 2026-07-28
Add path-based URL routing for Bible reading positions

KallynGowdy:
## Summary

Replaces query-parameter-only URLs with a canonical path-based routing scheme for Bible reading positions. URLs now follow the pattern `[/{lang}]/{translationId}/{bookSlug}/{chapter}` (e.g., `/NIV/matthew/1` or `/es/spa_onbv/john/3`), with automatic 301 redirects from legacy query-param and old path formats. Adds fuzzy book-name matching to correct close typos in URLs, and displays a "book not found" UI when a requested book isn't available in the selected translation.

## Key Changes

- **New `ReadingUrlPath` module** (`packages/seed-bible/seed-bible/managers/ReadingUrlPath.ts`): Parses and builds the canonical reading path format. Handles 3-segment (translation/book/chapter, default language omitted) and 4-segment (explicit language) forms. Includes `parseReadingPath()` to extract reading state from a URL and `buildReadingPath()` to construct canonical URLs.

- **Book slug mapping and fuzzy matching** (`BibleDataManager.tsx`):
  - Added `BOOK_SLUGS` record mapping each `BookId` to its canonical URL slug (e.g., `"SNG"` → `"song-of-solomon"`)
  - Added `getBookSlug()` to retrieve the slug for a book ID
  - Added `levenshteinDistance()` and `findClosestBookId()` for typo correction (e.g., "senesis" → "GEN")
  - Updated `getBookId()` to accept hyphenated slugs (e.g., "song-of-solomon")

- **SSR redirect logic** (`standalone/entry-ssr.tsx`): New `legacyReadingUrlRedirect()` function detects obsolete URL shapes (bare root with `?book=`, old `/{book}/{chapter}` paths, or canonical paths with fuzzy book matches) and returns a 301 redirect target. Handles both exact and fuzzy book resolution.

- **TabsManager URL sync** (`packages/seed-bible/seed-bible/managers/TabsManager.tsx`):
  - Integrated `parseReadingPath()` and `buildReadingPath()` into initial state extraction and external URL change handling
  - Added `selfHealFuzzyBookMatch()` to correct the address bar when a book only resolved via fuzzy matching
  - Coordinates language, translation, book, and chapter into a single path-based commit (no longer separate query params)
  - Passes `basePath` (deployment prefix) through all URL parsing

- **Book not found UI** (`BibleReader.tsx` and `.css`):
  - Added `bookNotFound` computed signal that detects when a requested book isn't in the translation's book list
  - Renders a centered "Book not found" message with an icon and a button to jump to the translation's first available book
  - Prevents chapter content from rendering alongside the not-found state

- **NavigationManager enhancement** (`NavigationManager.tsx`): New `updatePathAndQueryParams()` method sets pathname and query parameters in a single history entry, avoiding separate pushes for path and params.

- **I18nManager URL language resolution** (`i18n/I18nManager.tsx`): Updated `getUrlLanguage()` to prioritize the reading path's language segment over the legacy `?lang=` query param, with the 3-segment form canonically implying `DEFAULT_UI_LANGUAGE` ("en").

- **Test coverage**: Added comprehensive tests for `parseReadingPath()` and `buildReadingPath()` in new `ReadingUrlPath.test.ts`. Updated existing tests in `TabsManager.test.ts`, `BibleDataManager.test.ts`, `BibleReader.test.tsx`, and others to work with the new path-based routing and handle the book-not-found state.

- **Language segment omission**: The 3-segment form (`/AAB/john/3`) canonically means the default language ("en"), not "detect from browser." Browser detection only applies to a bare `/` with no reading path. This makes the canonical URL for the same content consistent across visitors.

- **Fuzzy matching is conservative

https://claude.ai/code/session_01RfqGZ5UcWKgT6hzBYRqdpG


Blitzy:
> [!NOTE]

I checked out this branch locally and traced the new routing logic end-to-end (`ReadingUrlPath.ts`, the `BOOK_SLUGS`/`findClosestBookId` additions in `BibleDataManager.tsx`, and the SSR redirect in `entry-ssr.tsx`). There's no linked issue on this PR to cross-reference against, so this review is based purely on reading the code and its tests.

## High severity

**Every Apocrypha/Deuterocanonical book's "canonical" URL infinitely redirects to itself, making that content permanently unreachable — the opposite of the SEO goal this PR is for.**

- File: `packages/seed-bible/seed-bible/managers/BibleDataManager.tsx` (`BOOK_ID_MAP`, `findClosestBookId`) and `standalone/entry-ssr.tsx` (`legacyReadingUrlRedirect`)

**What's happening:** `BOOK_ID_MAP` (used by `getBookId()` for *exact* name/slug resolution) only has entries for the 66 standard Protestant books. The ~48 Apocrypha/Deuterocanonical/Pseudepigrapha book IDs (`TOB`, `JDT`, `WIS`, `SIR`, `1MA`...`4MA`, etc.) have no entry there — they only exist in the newly added `BOOK_SLUGS` map. `findClosestBookId()` builds its candidate list from *both* `BOOK_ID_MAP` and `BOOK_SLUGS`, so when a URL's book segment is literally `"tob"` (Tobit's own canonical slug), `getBookId("tob")` returns `null` (no exact entry exists to match), but `findClosestBookId("tob")` finds a distance-0 "fuzzy" match against its own slug. `parseReadingPath()` then classifies this as `bookMatch: "fuzzy"` even though the URL is already 100% canonical.

`legacyReadingUrlRedirect()` treats any `"fuzzy"` match on an already-canonical-shaped path as needing a 301 to the "corrected" path — but the "corrected" path is built from the exact same resolved `bookId`/`chapter`/`translationId`, so:
- If the requested language/translation is the site default, the redirect target is byte-for-byte identical to the request URL → the server 301s a URL to itself, forever.
- Otherwise, it can ping-pong between the 3-segment and 4-segment forms (e.g. `/AAB/tob/1` → `/en/AAB/tob/1` → `/AAB/tob/1` → ...), since both forms independently re-derive the same "fuzzy" classification on every request.

Either way, the response is an infinite redirect loop (`ERR_TOO_MANY_REDIRECTS` in a browser). The same misclassification also feeds `TabsManager.tsx`'s `selfHealFuzzyBookMatch()` client-side, and — worse — `SeedBibleStateManager.tsx`'s `canonicalUrl` computed signal, which emits this exact broken URL as the page's `<link rel="canonical">` for search engines to index.

**Reproduction steps:**
1. On this branch, run `pnpm dev` and open a translation that has Apocrypha available (`apocryphaAvailable` in `BibleSelector.tsx` — any Catholic/Orthodox-inclusive translation).
2. Either click an Apocrypha book (e.g. Tobit) in the Bible selector, or directly request a path like `/AAB/tob/1` (adjust the translation ID to whatever your local default is).
3. Observe the request 301-redirecting back to an equivalent URL indefinitely instead of rendering the chapter.

You can also confirm this without a browser, purely at the unit level: `findClosestBookId("tob")` returns `"TOB"` (not `null`), and `parseReadingPath("/AAB/tob/1", "")?.bookMatch` comes back `"fuzzy"` instead of `"exact"` — neither is exercised by the current test suite (`ReadingUrlPath.test.ts` and `BibleDataManager.test.ts`'s `findClosestBookId()`/roundtrip tests only use canonical 66-book examples like `"senesis"` → `GEN` and `getBookId(getBookSlug("REV"))`).

**Suggested fix:** Make an exact match against a book's own slug count as `"exact"`, not `"fuzzy"`. The simplest fix is in `getBookId()`: build a reverse lookup from `BOOK_SLUGS` (hyphens stripped, same normalization as the existing map) and check it before falling through to the `startsWith` loop, so every `BookId` — Apocrypha included — has an exact-match path, not just the 66 with a `BOOK_ID_MAP` entry. Alternatively, special-case distance-0 results in `findClosestBookId()`/`parseReadingPath()` as `"exact"`. Either way, please add a regression test that round-trips an Apocrypha book through `parseReadingPath()` and asserts `bookMatch === "exact"`, and consider adding a test file for `entry-ssr.tsx`'s redirect logic — there currently isn't one, which is likely why this slipped through.

## Notes (not blocking, just flagging)

- Given the above, it'd be worth double-checking whether any other non-standard `BookId` values (anything added in the future without a matching `BOOK_ID_MAP` entry) could hit the same gap — the fix above generalizes to all of them, not just the current Apocrypha set.
- Nice touch keeping `ReadingUrlPath.ts` framework-agnostic (no `BibleReadingManager` import) and the conservative ambiguity handling in `findClosestBookId` (tie distances return `null` rather than guessing) — the core parsing/building logic itself looks solid and is well covered by `ReadingUrlPath.test.ts`.

---

Blitzy:
@KallynGowdy Will this break the path based branch deployments?

Blitzy:
# Review: PR #1547 — Add path-based URL routing for Bible reading positions

**Branch:** `claude/seed-bible-seo-x0mw01` → `develop`
**Head:** `4fab6cba397891cf9460ab37c2511d176c19795e`
**CI:** green (build 24.x ×2, deploy 24.x)

## High — non-canonical book segments resolve as `exact`, so they never redirect

`getBookId()` falls back to a `startsWith` scan over `BOOK_ID_MAP` in insertion order (`packages/seed-bible/seed-bible/managers/BibleDataManager.tsx:572-577`). Anything with a book alias as a prefix matches. Verified output:

| URL | Resolves to | `bookMatch` | Canonical slug |
| --- | --- | --- | --- |
| `/AAB/luke-skywalker/1` | `LUK` | `exact` | `luke` |
| `/AAB/genocide/1` | `GEN` | `exact` | `genesis` |
| `/AAB/mark-twain/1` | `MRK` | `exact` | `mark` |
| `/AAB/acts-of-congress/1` | `ACT` | `exact` | `acts` |
| `/AAB/gen/1` | `GEN` | `exact` | `genesis` |
| `/AAB/Genesis/1` | `GEN` | `exact` | `genesis` |

Because `legacyReadingUrlRedirect()` only redirects on `bookMatch === "fuzzy"`, every one of these is a 200 with no correction.

The behavior is backwards from the intent: a typo (`senesis`) gets a clean 301, but `luke-skywalker` is served as Luke at a permanent, indexable URL. Every alias and every casing variant is a live duplicate too.

**Suggested fix:** one condition covers all three cases — redirect whenever the request path differs from `buildReadingPath()` of the resolved state, rather than keying on `bookMatch`.

---

## Medium — crawlers get bounced off the canonical short URL

`acceptLanguageRedirect()` 302s to the explicit `/en/...` form when no `Accept-Language` header was sent (`standalone/entry-ssr.tsx:214-216`).

That is exactly what Googlebot sends: nothing. So the 3-segment form — the canonical shape this whole scheme is built around — is never served to the crawler that matters.

**Suggested fix:** render the no-header case as English at the same URL, and only negotiate when a header is actually present.

---

## Medium — the sitemap still emits the old URLs

`script/lib/sitemap.ts:105-109` is untouched and still writes:

```
<origin>/?translation=<id>&book=<BOOK>&chapter=<n>[&lang=<locale>]
```

Those now all 301. Every sitemap entry is a redirect, and each one disagrees with its target page's own `rel=canonical`.

---

## Medium (pre-existing, but relevant here) — `rel=canonical` is `/` on every server-rendered page

`canonicalUrl` needs `chapterData`, and SSR only awaits `state.i18n.ready` — no chapter fetch happens server-side. So the HTML a crawler receives carries `<link rel="canonical" href="/">` and `og:url="/"` for every reading URL.

The PR's own test comment confirms the mechanism:

> The default endpoint has no mocked chapter data, so the reading state never settles a chapter and `canonicalUrl` falls back to `basePath`.

Not introduced by this PR, but it caps what the PR can achieve, and it is the natural companion fix to the `exact`-match finding above.

---

## Low — `legacyReadingUrlRedirect()` has no direct tests

`test/unit/standalone/entry-ssr.test.ts` covers `acceptLanguageRedirect` and the `render()` wiring. `legacyReadingUrlRedirect` is the branchiest new function — three URL shapes crossed with exact/fuzzy/unresolved — and it is neither exported nor tested directly.

---

## Low — a book missing from a translation is still a soft 404

`/AAB/tob/1` resolves exactly, so `notFound` is false → 200 with the "Book not found" UI. The SSR check is URL-level only and cannot see what is actually in a translation.

That is the same soft 404 the code comment says it is avoiding, one layer down. Worth an explicit note in that comment, since Deuterocanon URLs across most translations land here.

---

## Worth verifying — service worker plus redirects

`standalone/sw.ts` uses `NetworkFirst` for navigations and caches the result under a single shell key. When a navigation now 302s, the response Workbox caches has `redirected: true`, and returning a redirected response for a navigation request is a known browser failure mode.

This only affects `main` — the SW is not emitted for branch builds, and `/b/` is excluded from the route — so previews will not surface it.

The doc comment at `standalone/sw.ts:101` also still uses `/?book=GEN&chapter=10` as its example, which this PR makes obsolete.

---

## Resolved — the earlier Apocrypha finding

The previous review's high-severity report of an infinite redirect loop on Apocrypha URLs was valid when posted, and is now fixed. Commit `6e6e7b60` ("fix: Add extra USFM book IDs to `BOOK_ID_MAP`") added the ~34 USFM apocrypha IDs five minutes after that review landed.

Confirmed: `parseReadingPath("/AAB/tob/1", "")?.bookMatch === "exact"`, and TOB / JDT / WIS / SIR / 1MA / LAO all round-trip through `getBookSlug()` → `getBookId()`.

No regression test was added for it, though — worth one, since the fix is a data table that is easy to fall out of sync again.

---

## Summary

| Severity | Finding |
| --- | --- |
| **High** | Prefix matching makes junk and alias URLs resolve as `exact`, so they never canonicalize |
| Medium | No-`Accept-Language` 302 bounces crawlers off the canonical short URL |
| Medium | Sitemap still emits the now-redirected query-param URLs |
| Medium | `rel=canonical` is `/` on every SSR'd page (pre-existing) |
| Low | `legacyReadingUrlRedirect()` untested |
| Low | Book absent from a translation returns 200 + "not found" UI (soft 404) |
| Verify | Workbox `NetworkFirst` caching a `redirected: true` navigation response |
| Resolved | Apocrypha infinite-redirect loop — fixed in `6e6e7b60`, no regression test |

KallynGowdy:
@Blitzy no it won't

KallynGowdy:
@Blitzy these new changes could use another review pass

Blitzy:
> [!NOTE]

Re-reviewed after the fixes pushed following Blitzy's second pass (head now `658806c`, since `4fab6cb`). I diffed just that range and traced the logic by hand rather than trusting the commit messages, then cross-checked each of the earlier findings against the current code.

## Previously-reported issues — status

All of these are now fixed and covered by new tests, verified against the current code (not just the commit messages):

- **High — prefix-matched junk/alias URLs served as "exact" (`luke-skywalker` → Luke, no redirect).** Fixed: `legacyReadingUrlRedirect`, `acceptLanguageRedirect`, and the client-side `selfHealNonCanonicalPath` all now redirect whenever the request path differs from `buildReadingPath()` of what it resolved to, instead of keying off `bookMatch`. I confirmed by hand that `/en/AAB/luke-skywalker/1` resolves to `LUK`, rebuilds to `/en/AAB/luke/1`, and those differ — so it 301s.
- **Medium — crawlers bounced off the canonical short URL when `Accept-Language` is absent.** The underlying design changed rather than patching around it: `buildReadingPath()` now always emits the explicit `/{lang}/...` form (the 3-segment form is purely a redirect entry point, never written to the address bar, sitemap, or `canonical` tag). A header-less request still lands on the explicit English URL. This is a more thorough fix than the original finding asked for.
- **Medium — sitemap still emitted the old `?translation=&book=&chapter=` URLs.** `script/lib/sitemap.ts`'s `buildChapterUrl` now goes through the same `buildReadingPath()` the app uses, so sitemap entries and `rel=canonical` agree.
- **Medium — `rel=canonical` was `/` on every SSR'd page.** `canonicalUrl` is now derived from the reading *position* signals instead of the loaded chapter, so an API failure or load timeout no longer collapses it to the front page.
- **Low — `legacyReadingUrlRedirect` untested.** Now exported and has direct test coverage (`entry-ssr.test.ts`).
- **Verify — Workbox caching a `redirected: true` navigation response.** `sw.ts` now copies the response via `new Response(response.body, response)` before caching, to strip the `redirected` flag.
- **Resolved (Apocrypha infinite loop)** — still holds; unaffected by this range.

One earlier finding was **deliberately left as-is, with a clear comment explaining why**: a book absent from a specific translation (most Deuterocanon in most translations) is still a soft 404 rather than a real one, because closing it would mean a network round trip before every render. That's a reasonable trade-off, not an oversight — just flagging that it's a conscious choice, not a miss.

## New finding

**Low — malformed percent-encoding in the URL path throws, and is silently swallowed into a degraded (but non-crashing) response.**

- File: `packages/seed-bible/seed-bible/managers/ReadingUrlPath.ts` (`parseReadingPath`), and the duplicate segment-splitting logic in `standalone/entry-ssr.tsx`'s `legacyReadingUrlRedirect`/`acceptLanguageRedirect`.

`parseReadingPath` splits the path and runs `.map((segment) => decodeURIComponent(segment))` with no `try`/`catch`. `decodeURIComponent` throws a `URIError` on a truncated or invalid percent-escape (e.g. a lone `%`, or `%E0` with no continuation bytes) — this is a routine occurrence on the open web (bots probing odd paths, a copy-pasted link with a stray `%`), not a contrived edge case.

**What actually happens today:** `render()` in `entry-ssr.tsx` wraps its work in `renderAndRespond`'s `try`/`catch` (`server/index.ts`), so this doesn't crash the process — it logs an error and falls back to serving the unrendered HTML shell with a plain 200. So there's no outage, but for a URL shape the rest of this PR is specifically designed to handle cleanly (garbage → 404, typo → 301, etc.), a malformed-percent-encoding URL instead gets: no SSR content, no redirect, a 200 instead of a 404, and a logged server error that looks like a real bug every time it happens.

**Reproduction:** request a path like `/en/AAB/%/1` (a lone `%` is enough — `decodeURIComponent("%")` throws by itself). Locally: `pnpm dev`, then `curl -i http://localhost:<port>/en/AAB/%/1` — expect a plain 200 with no rendered chapter content, and a stack trace in the server log ("SSR render() failed…").

**Suggested fix:** wrap the `decodeURIComponent` call in `parseReadingPath` (and the two inline copies in `entry-ssr.tsx`) in a `try`/`catch`, treating a decode failure the same as "garbage" — return `null` from `parseReadingPath` (or `bookMatch: "unresolved"` where a raw segment is still needed), so it falls through to the existing 404 handling instead of an uncaught exception.

## Minor / non-blocking

- **PR description is now stale.** The top-level summary still describes the URL scheme as `[/{lang}]/{translationId}/{bookSlug}/{chapter}` with the language segment optional for the default case. That was true at `4fab6cb`, but `39e917e` ("Always redirect to the canonical URL for a page") changed the design so the language segment is *always* present — the 3-segment form is now only ever a transient redirect source, never a real destination. Worth updating the description before merge so anyone reading the PR history later (or writing the changelog entry) doesn't describe the old scheme.
- **Inconsistent fallback-language priority order.** `BibleSelector.tsx`'s `shareTranslatation` resolves a fallback language as `uiLocaleForDefaultTranslation(translationId) ?? bibleLanguageToUiLocale(translation.language)`, while `SeedBibleStateManager.tsx`'s `canonicalUrl` resolves the same pair in the opposite order. In practice these should almost always agree (a translation's hardcoded default-language mapping and its catalog `language` field shouldn't disagree), so this is unlikely to cause a real bug, but it's easy to overlook when one of them is touched later. Consider extracting one shared "resolve a UI locale for a translation" helper so the priority only has to be decided once.
- GitHub reports this PR's `mergeable_state` as "dirty," but a local `git merge-tree` against the current `develop` HEAD found no actual conflicts — likely just a stale mergeability cache on GitHub's side rather than a real conflict. Worth a fresh look before merging, just in case.
- CI is green on the current head (`658806c`): build and deploy both succeeded.

---

PR #1551, opened 2026-07-29
fix: split reading position from chapter content so navigation never waits on a download

Blitzy:
## The problem

The reader's position and the reader's content were the same commit. `bookId`, `chapterNumber`, `translationId` and `chapterData` were all written together in one `batch()`, and that batch only ran once the chapter's text had finished downloading. Three consequences, all of them the issue:

- The "next chapter" target was read off the chapter you were still looking at, so tapping next five times quickly computed the same target five times and advanced you one chapter.
- The chevrons, the arrow keys and the book selector were switched off while a request was open.
- Nothing was ever cancelled, so on a slow connection eight unwanted chapter downloads sat ahead of the one you were actually waiting for.

## The fix

Split **position** from **content**. `translationId` / `bookId` / `chapterNumber` are the reader's intent and are written synchronously the instant you tap, with the target computed from the book catalog already in memory rather than from links on the loaded chapter. `chapterData` follows behind, loaded by a single `effect()` that watches the position, cancels the request it supersedes, and uses a generation counter so only the newest result may commit.

Centralising the loading in one effect also fixes paths that previously wrote a position and loaded nothing: shared sessions moved the header and left the text behind.

## Video demo

https://www.youtube.com/watch?v=GWKGxW3CIbs

## Reconciling with #1464

Worth flagging up front, because it shapes the diff: `develop` merged #1464, an earlier independent attempt at this same issue, while this branch was in progress. #1464 advances the logical position through a promise chain, but still commits the position signals only after the fetch resolves, so the header continues to lag the tap. This branch keeps its own implementation for that reason.

`BibleReadingManager.tsx` was therefore resolved wholesale rather than hunk by hunk: the two sides are different designs, not different edits, and a tool-merged result would have been something neither side tested. I verified beforehand that nothing unrelated to #1464 had touched that file on `develop`.

Kept from #1464, gratefully:

- Its ref-counted request cancellation in `FreeUseBibleAPI`, which is the same design this branch had planned. It arrived without conflict and saved writing that layer from scratch.
- `ApiRequestOptions` threaded through `BibleDataManager`.
- Nearly all of its tests, which pass against this implementation and are useful independent coverage of the same requirement. Two had to be rewritten and both are called out below.

## Bugs found along the way

Some of these were pre-existing on `develop` and are fixed here because this work either exposed them or made them reachable:

- **Shared sessions could freeze and crash the tab with an out-of-memory error.** Remote sync relaunched each superseded application on completion, which two overlapping updates turned into a perpetual loop that never yielded to rendering. Remote positions are now applied one at a time toward the newest known target, with no retry. The regression test measures, idles, and measures again; on the old code it fails by a growing margin.
- **A failed first chapter load hung the server.** `chapterDataPromise` resolved only on the first non-null `chapterData`, never rejected, and had no deadline, so `renderToStringAsync` awaited it forever. It now settles on the first terminal outcome and always resolves rather than rejecting, since a rejected thrown promise loses the whole document.
- **Cancelling a request could poison an unrelated caller.** Aborting only starts the teardown; the doomed promise stayed advertised in the response cache until the rejection landed a microtask later, so a caller arriving in that window inherited an `AbortError` it never asked for. Reachable by navigating back to a chapter whose request was just superseded.
- **An enabled chevron that did nothing.** `hasNext`/`hasPrevious` fall back to the loaded chapter's adjacency links when a translation's catalog is missing, but navigation computed its target only from the catalog and returned silently. Found by one of #1464's tests. Navigation now follows the same links the control trusts.
- **`?chapter=99999` left the address pointing at a chapter you were not on**, so pressing Back returned you to it and bounced forward again.
- **`?chapter=0.5&verse=3` silently lost the verse highlight.** The chapter param floored after its range check, producing chapter 0. The reading state normalises the chapter again and so never saw it, but the deep-link verse decoration is keyed off the raw value, where nothing corrects it.

## What reviewers should look at

The diff is large but concentrated. `BibleReadingManager.tsx` is most of it, and I'd suggest reading it in this order:

1. `applyPosition` / `applyChapterContent`, the two halves the old single commit was split into.
2. The content loader effect and `requestContent`. The effect's only tracked reads are the three position signals; everything else goes through `.peek()` deliberately, because subscribing to the catalog there would abort and re-issue the request it just started on every translation switch.
3. `whenContentSettled`, which keeps the navigation methods' promises meaning what they always meant, so callers that await a navigation and then read `chapterData` are unchanged.
4. `SessionsManager.tsx`, which is self-contained and can be reviewed on its own.

The tests worth reading rather than skimming:

- `"advances one chapter per press while the text is still downloading"` asserts the position and title synchronously, before any await, with the content still on the old chapter.
- `"stops syncing once a burst of remote navigations has been applied"` is the guard for the crash.
- `"a fast skim costs one history entry, not one per chapter"` spies on the real `window.history`.

## Deliberate trade-offs

- **Adjacency now walks into apocryphal books** wherever a translation's catalog lists them, where link-chasing previously followed whatever the API decided. Pinned by a test so it is deliberate rather than incidental. Reversible as a one-line filter if that is wrong for some translation.
- **Taps during an unresolved translation switch are not chained behind it.** #1464 chains them; this branch lets the switch land on top. Chaining costs the instant position write, and picking a translation from a menu is not the rapid-fire path this issue is about.
- **The loading state is two-stage.** The chapter you left dims immediately and the placeholder only takes over after 500ms, so a fast connection never flashes a skeleton. A cold start skips the delay, since there is nothing to dim.
- **The placeholder's bar count is static and overshoots** rather than being measured. It remounts on every navigation, so a measure-then-render pass would reflow on every chapter of a skim. Falling short is visible dead space; overshooting spills below the fold.
- **Adjacent-chapter prefetch stays mobile-only.**

## Public API changes

- `loading` and `translationBooks` on `BibleReadingState` narrow to `ReadonlySignal`.
- New `isChapterContentStale`, which is what a loading placeholder should key off, and new `initialChapterLoadSettled`.
- `chapterDataPromise` now settles on failure instead of hanging, and always resolves.
- Navigation hooks should return synchronously for a `default` outcome. Returning a promise puts a microtask in front of every press, which is enough to make two quick presses compute the same target.

## Testing

`pnpm test` 3165 passed, `pnpm check:ts` 0 visible errors, `pnpm lint` 0 errors, `pnpm build` clean across the client, SSR and server bundles and both `.aux` patterns.

Verified in the browser at each step, throttled to Slow 3G: rapid navigation by press, arrow key and mobile swipe; the dim-then-placeholder sequence across desktop, tablet and phone widths in both themes; two windows in one shared session in both directions; cancelled requests showing as `(canceled)` in the Network tab with a revisited chapter still served from cache; `?verse=` deep links; Back after a skim; and a failing endpoint returning the shell with an error rather than hanging.

## Follow-ups, not in this PR

- Navigation methods cannot be cancelled, so a local navigation made within one network round trip of a remote shared-session update can still be clobbered. Narrowed from ten in-flight applications to one, but not closed.
- The response cache in `FreeUseBibleAPI` is still unbounded. Pre-existing and not made worse here, but a real memory question for a long reading session.
- `applyPosition` emits a navigation even when the position is unchanged, so a redundant apply costs a Back entry. Harmless now that the sync loop is fixed.


Blitzy:
> [!NOTE]

## Code review — split reading position from chapter content (#1414)

Reviewed `feature/1414-fast-navigation` at `69a1f83` against `develop` (merge base `20c4397`). This is a strong, well-documented PR: the position/content split is the right shape for #1414, the generation-counter + single-loader design is clean, and the write-up of the reconciliation with #1464, the deliberate trade-offs, and the along-the-way bug fixes is genuinely helpful for review. No blockers — everything below is a Medium and a few Lows, and I agree with the follow-ups already called out in the description.

The core mechanic I checked most closely — position signals move synchronously, the loader effect is the single owner of fetching, generation counter + `abortOpenContentRequest` guard commits, `whenContentSettled` preserves the navigation-promise contract, and the SSR `chapterDataPromise` now always resolves — all holds up. The shared-session crash fix (`queueRemoteSync` draining toward the newest target with no relaunch-on-completion) and the `resolveSessionDataToPublish` "read all deps before any early return" guard are both correct and address real hazards.

---

### Medium

**M1 — Re-applying the *current* position after a failed load never retries, and the returned promise never resolves.**

The content-loader effect is keyed only on the `translationId` / `bookId` / `chapterNumber` signals. When a load fails, `error.value` is set but the position signals stay put and `chapterData` stays stale (or `null`). If a caller then applies the *identical* position again, `applyPosition` writes the same values, so no signal changes, so the effect does not re-run — no new request is issued. Worse, the `await whenContentSettled(target)` in that caller registers a waiter for a position that has no in-flight request and no future content commit, so the promise it returns never settles.

This is a behavior change from `develop`, where `selectChapter` / `selectTranslationAndChapter` always issued a fresh fetch regardless of current state.

- Reachable via the book/chapter selector: if the current chapter failed to load, selecting the same book+chapter again does nothing (no spinner, no retry).
- Also reachable in a shared session: `applyRemoteSessionData` awaits `selectTranslationAndChapter`; if a peer re-broadcasts the same position that just failed locally, that await hangs and stalls the `remoteSyncDrain` loop.
- Relevant to the in-flight chapter-load-retry work in #1531, whose natural implementation is "retry the current position."

Reproduction:
1. Throttle the network / point the chapter endpoint at a failing URL so a chapter fetch rejects.
2. Land on a chapter and let it error (error banner shows).
3. Open the selector and pick the exact same book + chapter.
4. Observe: no new network request, banner stays, and the `selectChapter` promise never resolves.

Suggested fix: give the effect a reason to re-run for an identical position when there is no content for it — e.g. bump `loadGeneration` and call `requestContent` from `applyPosition` when `chapterData` doesn't already match `next` and no request is open, or track a retry nonce signal that the effect reads. At minimum, guard `whenContentSettled` so it resolves (rather than hangs) when the target is current but nothing is or will be loading it.

---

### Low

**L1 — A stale error banner persists through the whole of a subsequent navigation.**

Navigation methods no longer clear `error.value` up front. It is now cleared only in `applyChapterContent` (i.e. when new content actually commits). Because `BibleReader` renders content/skeleton only under `!error.value`, after a failed load the error banner stays on screen for the entire duration of loading the *next* chapter — no dimmed text, no skeleton — and then content pops in. On `develop` the error cleared the instant navigation started.

Suggested fix: clear `error.value` in `applyPosition` (for a genuine position change) or when a request begins, so the placeholder/dimming path is visible during the recovery navigation.

**L2 — `applySessionDataToReadingState` writes the three position signals unbatched, so the loader effect can fire up to three times for one incomplete remote update.**

`applySessionDataToReadingState` (`SessionsManager.tsx`) assigns `translationId`, `bookId`, and `chapterNumber` in separate statements. Now that the loader effect reacts to each, an incomplete remote update can run the effect several times and start/abort several requests for a single change (and can transiently fetch a `newTranslation + oldBook + oldChapter` combination). Wrapping the assignments in `batch()` collapses them to one effect run. Low impact because this is the `!canLoadSessionData` (partial-data) branch, but it's a one-line tidy-up.

**L3 — `applyPosition` emits a navigation even when the position is unchanged.**

Already listed as a follow-up in the PR description; flagging only so it's tracked. A no-op `applyPosition` still calls `emitPositionNavigate`, costing a Back entry. The 400 ms coalescing hides it for bursts, but an isolated redundant apply (e.g. re-selecting the current chapter) still pushes/replaces an entry. Fine to defer, but pairing this with the M1 fix (both stem from "apply the same position again") would be natural.

---

### Things I verified and am satisfied with

- The initial speculative fetch (when `initialBookId` is set, the loader effect runs once at creation before `loadInitialData`) is superseded cleanly — the later `applyPosition` bumps the generation and `abortOpenContentRequest` cancels the speculative request, whose `AbortError` is swallowed rather than surfaced.
- `translationBooks` as a `computed` over `dataManager.translationBooks` recomputes correctly (the data manager assigns a fresh `Map` instance on each catalog load).
- `sortBy` is still used elsewhere in `BibleReadingManager.tsx`, so removing `getAdjacentBook`'s use of it didn't orphan the import.
- The `getInitialFirstTabChapter` fix (`Math.floor` *before* the range check) correctly turns `?chapter=0.5` into the default chapter rather than chapter 0.
- Only `en.json` was touched for i18n, per the repo convention.
- SSR suspense keys off `initialChapterLoadSettled` in `BibleReader`, `TabRow`, and `ChapterContent`, and `chapterDataPromise` resolves (never rejects) once the latch flips — so a failed first load can no longer hang `renderToStringAsync`.

I did not run the suite for this review; the description reports `pnpm test` (3165), `check:ts`, `lint`, and `build` all clean, and the M1/L1 cases above are edge paths I would not expect the existing tests to cover. Worth adding a regression test for M1 (re-select the same position after a failed load and assert a fresh request is issued and the promise resolves).

---

Blitzy:
Thanks — all four findings are fixed in bfa4b7f8. Details and the one place I deviated from the suggestion:

**M1 — failed load could not be retried, and the promise never settled.** Confirmed, and it reproduced exactly as described. Took the retry-nonce option rather than calling `requestContent` from `applyPosition`: the second option would have added a second call site for fetching, and "the loader effect is the single owner of fetching" is the invariant the rest of this PR leans on. So `applyPosition` now bumps a `contentRetryNonce` signal that the effect subscribes to, and the effect stays the only thing that issues requests.

One condition beyond what the review suggested: the nonce is bumped only when the position is unchanged, its text is missing, **and** no request for that position is already open. Without the last part, re-picking a chapter you are currently waiting on would abort the in-flight download and start it again from zero. That needed a new `openContentRequestPosition` tracked alongside the abort controller, so the code can tell "nothing is fetching this" from "this is already on its way".

I did not add the defensive `whenContentSettled` guard. With the fix at the source, every path that leaves a position current with no content now issues a request, so a registered waiter is always settled — the guard would be unreachable. Happy to add it if you would rather have the belt and braces.

**L1 — stale error banner persisted through the recovery navigation.** Fixed by clearing `error.value` as `requestContent` opens rather than in `applyPosition`. `requestContent` is the single point every load passes through, so one clear covers both a genuine move and an M1-style retry; clearing in `applyPosition` would have missed the retry, since a retry writes no position.

**L2 — unbatched position writes in `applySessionDataToReadingState`.** Wrapped in `batch()`.

**L3 — no-op `applyPosition` costing a Back entry.** Fixed, but not by skipping the emit, which was my first instinct and is wrong. The URL can be out of step with the position for reasons other than a move — a clamped chapter to correct, a `verse` param to drop, an extension's params to re-derive — so skipping would leave it stale and regress the `?chapter=99999` fix. Instead a redundant apply still emits, forced to `replace: true`.

It also goes straight to `emitNavigate` rather than `emitPositionNavigate`, because a no-op is not a gesture: stamping the coalescing clock would make the reader's *next* real navigation look like the continuation of a press that never happened. That turned out to be load-bearing rather than tidiness — see the mutation table below. A verse change within the current chapter still pushes, since that is somewhere new.

## Testing

**3172 tests passing** (up from 3165), `check:ts` 0 visible errors, `lint` 0 errors, prettier clean.

Six new tests. I mutated each fix to confirm they discriminate rather than pass vacuously:

| Mutation | Result |
|---|---|
| Remove the nonce bump | `expected 'hung' to be 'settled'` — reproduces M1's hang exactly |
| Remove the in-flight guard | 5 requests instead of 4 |
| Remove the error clear | banner still on screen mid-load |
| Remove the forced replace | `replace: true` → `false`, **and** the next real navigation gets dragged into the coalescing window |
| Remove the `batch()` | loader runs twice instead of once |

## CHANGELOG

Added one bullet, for L3 only. I checked develop and it emits `{ replace: false }` on a same-chapter navigation too, so that one is a genuine improvement over v1.2.1.

M1, L1 and L2 get no entry on purpose: all three are regressions this branch introduced and this commit removes, so the net effect versus the last release is zero and an entry would read as a fix for something users never saw.

The nine existing `## TBD` bullets now carry the `#1551` reference, matching the rest of the file.

Blitzy:
> [!NOTE]

## Re-review — fixes for the first review round (commit bfa4b7f8)

Reviewing just the new commit (`bfa4b7f8`, "fix: make a failed chapter load retryable again"), pushed in response to the earlier review on this PR. Current head still matches this commit, so there's nothing newer to look at.

**All four findings (M1, L1, L2, L3) are correctly fixed, and I found no new issues.** Details below for anyone who wants to follow the reasoning; nothing here blocks merge.

### M1 (failed load couldn't be retried) — fixed correctly

The `contentRetryNonce` signal gives the content-loader effect a reason to re-run when a position is re-applied unchanged. I traced the three-way guard on when it bumps (`BibleReadingManager.tsx`, inside `applyPosition`):

- Position unchanged (`!didPositionChange`)
- No content for it (`!chapterMatchesPosition(chapterData.peek(), next)`)
- Nothing already fetching it (`!positionsEqual(openContentRequestPosition, next)`)

The new `openContentRequestPosition` tracker is what makes the third condition possible — it distinguishes "nothing is fetching this" from "this is already on its way," which is exactly the distinction No. 2 in the mutation table needed. I confirmed `positionsEqual(null, x)` returns `false` (via the `a === b` fallback when either side is `null`), so a retry is allowed whenever no request is open, and blocked while one is — matching both new tests ("re-requests a position that is already current when its last load failed" and "does not restart the download when the same position is re-picked mid-flight").

I also confirmed the shared-session case from the commit message actually routes through this fix: `applyRemoteSessionData`'s full-data path calls `selectTranslationAndChapter`, which calls `applyPosition` — so a peer re-broadcasting a position that just failed locally does hit the retry-nonce logic, not just a local re-pick.

The decision to skip the defensive `whenContentSettled` guard (mentioned in the reply) is reasonable — with the fix at the source, I don't see a path that leaves a position current with no content and no request in flight without a nonce bump, so the guard would indeed be unreachable dead code.

### L1 (stale error banner through recovery) — fixed correctly

`error.value = null` moved from `applyChapterContent` (commit) to the top of `requestContent` (request open). Since `requestContent` is the single fetch entry point for both a genuine navigation and an M1-style retry, one clear point covers both — no gap for the retry case, which wouldn't have been covered by clearing in `applyPosition` alone (a retry writes no position).

### L2 (unbatched partial remote writes) — fixed correctly

`applySessionDataToReadingState` now wraps its three conditional signal writes in `batch()`. Confirmed via the new SessionsManager test that instruments a stand-in effect with the same three tracked reads and asserts it runs once instead of up to three times.

### L3 (no-op apply costing a Back entry) — fixed correctly, and the deviation from my original suggestion is the right call

Skipping the emit entirely (my first-instinct fix) would have been wrong, and the PR reply's reasoning holds up: the URL can be out of step with the position for non-navigation reasons (a clamped chapter, a dropped `verse` param, extension-derived params), so a no-op still needs to *rewrite* the URL — it just shouldn't *push*. The fix routes a same-position apply through `emitNavigate({ replace: true })` directly rather than `emitPositionNavigate`, which is important: `emitPositionNavigate` is what stamps `lastNavigateAt` for the 400ms coalescing window, and a no-op apply stamping that clock would make the *next* real navigation look like a continuation of a gesture that never happened. I checked `emitPositionNavigate`'s implementation and confirmed `lastNavigateAt` is only touched there, not in `emitNavigate` — so this path genuinely doesn't affect the coalescing window, matching the new test "does not let a no-op apply drag the next real navigation into its history entry."

### Verification performed

- Checked out `bfa4b7f8` and ran the two affected test files directly: **168/168 passing**.
- Ran `pnpm check:ts`: **0 visible errors**.
- Read `positionsEqual`, `chapterMatchesPosition`, `emitNavigate`/`emitPositionNavigate`, and `selectTranslationAndChapter` to verify the control flow described above rather than taking the commit message at its word.
- Compared the new CHANGELOG bullet against the diff: matches (one new bullet for L3 only; M1/L1/L2 correctly get no entry since they're regressions introduced and removed within this same PR, net-zero against the last release).

Nothing further from me — this is ready from a correctness standpoint.

---

Blitzy:
@KallynGowdy I think this is ready

KallynGowdy:
Looks really good. I found one thing: when offline and a chapter load fails, navigating to the previous chapter (that is cached) fails.

Steps to reproduce:

1. Load Seed Bible at Genesis 1
2. Go to the next chapter (Genesis 2)
3. Go offline (in dev tools network tab)
4. Go to the next chapter (Genesis 3)
5. See that it fails to load
6. Go to the previous chapter (Genesis 2)
7. See that it also fails to load even though we just loaded it fine
8. Go to the previous chapter again (Genesis 1)
9. See that it loads fine

Blitzy:
Good find. Fixed in 18f0c9df. Genesis 2's text was never actually lost — it was still in `chapterData`, hidden behind Genesis 3's error banner, because I was clearing the error where a request opens and going back to a chapter you already hold opens no request (which is also why Genesis 1 worked: it needed a request, served from cache). Moved the clear into the loader effect that every navigation passes through, and added a test that walks your exact steps including going offline partway.

KallynGowdy (approved):

PR #1557, opened 2026-07-30
Add OpenTelemetry tracing and metrics instrumentation

KallynGowdy:
## Summary

Adds comprehensive OpenTelemetry tracing and metrics to the SSR server, enabling observability of request handling, rendering performance, cache behavior, and store operations. Telemetry is opt-in via the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable and has zero overhead when disabled.

## Key Changes

- **New `server/telemetry.ts` module** — Hand-written OpenTelemetry instrumentation (not auto-instrumentation) because the server runs under Bun and ships as a pre-bundled file, making module-load interception impossible. Provides:
  - Tracer and meter initialization with configurable OTLP exporters
  - Span helpers (`withSpan`, `recordError`, `extractContext`) for explicit span creation at call sites
  - Metric recording functions for HTTP requests, SSR render duration/failures, cache lookups, store operations, and asset proxy calls
  - Label helpers (`routeLabel`, `branchLabel`) that collapse unbounded values (branch names, build IDs) into fixed metric attributes to prevent cardinality explosion
  - Express middleware for dev server request tracing
  - Store wrapper that instruments all artifact reads with spans and metrics

- **Updated `server/index.ts`** — Integrated telemetry throughout the request lifecycle:
  - Opens a server span for every request (except `/healthz` liveness probes)
  - Records HTTP request metrics with method, status code, route, and branch
  - Traces SSR render calls and records render duration/failures
  - Records cache hits/misses for pointer, module, and HTML caches
  - Traces store operations (readPointer, fetchArtifacts, fetchHtml) with duration and outcome
  - Traces asset proxy upstream fetches
  - Installs graceful shutdown handlers that flush telemetry before exit (critical for capturing the batch that prompted a deploy)
  - Refactored request handling into `handle()` and `dispatch()` to open the server span before routing

- **Test coverage** (`test/unit/server/telemetry.test.ts`) — Unit tests for label helpers, initialization logic, span behavior when disabled, and store instrumentation

- **Documentation** — Updated `server/README.md` with `SHUTDOWN_GRACE_MS` and `INVALIDATION_SECRET` environment variables

- **Dependencies** — Added OpenTelemetry SDK packages (`@opentelemetry/api`, `@opentelemetry/sdk-trace-node`, `@opentelemetry/sdk-metrics`, exporters for OTLP HTTP, and semantic conventions)

## Implementation Details

**Why hand-written spans:** Auto-instrumentation via `@opentelemetry/auto-instrumentations-node` requires `require-in-the-middle` hooks, which Bun does not support. The production bundle is pre-built with `bun build --outfile`, so all dependencies are inlined at build time — there are no module loads left to intercept at runtime.

**Telemetry is off by default:** When `OTEL_EXPORTER_OTLP_ENDPOINT` is not set (or `OTEL_SDK_DISABLED` is truthy), `initTelemetry()` returns a disabled handle. Every helper takes a fast path that allocates no spans, so `pnpm dev` and the test suite are unaffected.

**Metric cardinality protection:** Branch names and build IDs are unbounded (anyone can deploy a branch), so they cannot be used directly as metric attributes — they would create infinite time series. `branchLabel()` collapses unknown branches to `"other"`, and `routeLabel()` maps raw paths to fixed route patterns. Spans are not aggregated, so they keep the real values for debugging.

**Graceful shutdown:** The server now installs SIGTERM/SIGINT handlers that close the listener, wait up to `SHUTDOWN_GRACE_MS` for in-flight requests, then flush telemetry before exit. This ensures the batch covering the deploy event is not lost.

**Trace context propagation:** Extracts `traceparent` headers from inbound requests so traces started by a CDN or load balancer continue through the server instead of starting over.

https://claude.ai/code/session_01K4TQtgWGKNiShVELYmuu13


Blitzy:
> [!NOTE]

## Summary

This PR adds hand-written OpenTelemetry tracing and metrics to the SSR host server (`server/telemetry.ts`, wired into `server/index.ts`), off by default unless `OTEL_EXPORTER_OTLP_ENDPOINT` is set. The rationale for hand-rolled spans (Bun doesn't support the `require-in-the-middle` hooks that auto-instrumentation needs, and the server ships as a single pre-bundled file) is sound and well explained in both the code comments and `server/README.md`.

I pulled the branch, ran a full `pnpm install`, `tsc --noEmit`, `eslint`, and the new `test/unit/server/telemetry.test.ts` suite locally:
- **Type check**: no new errors — the handful of pre-existing errors (in `script/lib/browser.ts`, `script/lib/extension.ts`, and a vendored `fast-json-stable-stringify` patch) reproduce identically on `develop`, so they predate this PR.
- **Lint**: clean on all three changed/added files.
- **Tests**: all 16 tests in `telemetry.test.ts` pass.
- **CI**: build (24.x) is green on the PR.

## Issues by severity

### Medium — asset-proxy requests report inaccurate span/metric duration
`proxyAsset()` (`server/index.ts`) starts streaming the response body without awaiting it:

```ts
if (upstream.body) {
  Readable.fromWeb(upstream.body as NodeReadableStream<Uint8Array>).pipe(res);
} else {
  res.end();
}
```

`.pipe()` returns synchronously; nothing here waits for the `finish`/`close`/`error` events on `res`. Because `handle()` now wraps the whole `dispatch()` call in a `SERVER` span and records `http.server.request.duration` in a `finally` block right after `dispatch()` resolves, both the span and the metric close as soon as piping *starts*, not once the body has actually finished sending. For the exact request class this path exists for (root-scoped PWA-shell files like `sw.js`/`registerSW.js`/the web manifest), this means:
- `http.server.request.duration` will read as "instant" regardless of file size or a slow/throttled client, undercutting the value this PR is adding.
- Any error that happens mid-stream (client disconnect, broken pipe) never reaches the span or shows up as an error status, since it fires after the span already ended.

**Repro:** point `ASSET_HOST` at something slow (or throttle the client), request `/sw.js`, and compare wall-clock time to the recorded `http.server.request.duration` value for that request — the metric will be far shorter than the actual transfer time.

**Suggested fix:** have `proxyAsset()` return a promise that resolves after the response stream finishes (mirroring the pattern `expressSpanMiddleware` already uses for the dev-server case — listening for `res.once("finish", ...)` / `res.once("close", ...)`), and `await` it before returning from `dispatch()`.

### Medium — the "telemetry enabled" code path has no test coverage
Every test in `telemetry.test.ts` either exercises the disabled/no-op fast path or pure helper functions (`routeLabel`, `branchLabel`). None of them actually call `initTelemetry()` with an OTLP endpoint set, so the real behavior this PR adds — provider/exporter registration, `withSpan` creating a real span with the right name/attributes, metrics actually recording numbers — is entirely unexercised by CI. A regression here (e.g. a bad option passed to `NodeTracerProvider`/`MeterProvider`, or an attribute key typo) would only surface once telemetry is actually turned on in a live environment.

**Suggestion:** add at least one test that enables telemetry against an in-process exporter (OpenTelemetry ships `InMemorySpanExporter`, or a small fake `SpanProcessor`/`MetricReader`) and asserts a span produced via `withSpan`/`recordHttpRequest` has the expected name and attributes. Doesn't need a real OTLP collector.

### Low — `resolveRoute(url)` is called twice per request
In `handle()`, `resolveRoute(url).branch` is computed just to derive the metric's branch label, and then `dispatch()` immediately calls `resolveRoute(url)` again for actual routing. It's cheap (a string split), so not a real perf problem, but it's duplicate work on every single request and a chance for the two call sites to drift. Simplest fix: compute `route` once in `handle()` and pass it into `dispatch()` instead of re-deriving it.

### Low — `expressSpanMiddleware` doesn't handle a rejected span promise
```ts
void withSpan(`${method} ${route}`, { ... }, async (span) => { ... next(); await finished; ... });
```
`withSpan`'s returned promise is fired with `void` and has no `.catch()`. If `fn` ever rejects for a reason other than the normal request lifecycle (e.g. a downstream middleware throws synchronously instead of going through Express's `next(err)` convention), that becomes an unhandled promise rejection in the dev server process. Low severity since this only runs outside production, but a defensive `.catch((err) => console.error("expressSpanMiddleware:", err))` would be cheap insurance.

## Other observations (not blocking)

- The CI build-size bot already flagged this itself, but worth calling out in review too: the server bundle grew **+610KB (1.3MB → 1.9MB, ~47%)** from the OTel SDK packages. Given telemetry is opt-in and this is a server-only bundle (not shipped to browsers), this is probably an acceptable tradeoff, but worth a conscious sign-off from whoever owns deploy/cold-start budgets.
- `server/README.md` is updated thoroughly and clearly — the "why hand-written" and cardinality-protection explanations mirror the code comments well, which will help the next person who touches this.
- No linked issue was found on this PR to cross-reference for additional requirements/discussion.


---

KallynGowdy:
All four points addressed in 1bd03c4. Thanks — the asset-proxy one turned out to be hiding a worse bug.

### Medium — asset-proxy duration

Fixed, with one clarification: the review was written against `9739963`, before the merge from `develop`. That merge had already replaced `.pipe()` with `pipeline()`, and made the **request** span wait for the body to flush — so `http.server.request.duration` was already correct. What was still wrong was the `asset.proxy` span itself, which closed as soon as the upstream headers arrived.

The span now stays open across the body transfer. Measured with your repro (a stub upstream trickling out a 7.7KB file):

| | `asset.proxy` | actual wall-clock |
|---|---|---|
| before | 78ms | 761ms |
| after | 747ms | 762ms |

Stream errors are now recorded on the span too.

**The part worth flagging:** writing the disconnect test surfaced a bug the merge had introduced. Under Bun, a client that aborts mid-transfer emits *nothing* on the response — no `close`, and `res.closed`/`res.destroyed` stay `false`/`undefined` — while `stream/promises` `pipeline` never settles either. Only the request emits `aborted` and the socket closes. So waiting on the response alone hung forever: the span leaked and the request vanished from telemetry entirely. `responseFinished()` now also watches the request's `aborted` and the socket's `close`, and the proxy races the stream against that. An aborted download now records a correctly-timed span and is deliberately *not* marked as an error — a user navigating away isn't a server fault.

### Medium — no coverage of the enabled path

Agreed, that was the real gap. `initTelemetry()` now accepts in-process exporters and `Telemetry` gained `forceFlush()`. Ten new tests (26 total) assert real spans and metrics: name, kind, semconv attributes, parent/child nesting, error recording, the degraded-render flag, `traceparent` continuation, and the branch-label cardinality clamp on actual recorded metric data.

I mutation-checked them rather than trusting green: removing the cardinality clamp fails 2 tests, and typo'ing `http.route` → `http.rout` fails 1.

This also required `shutdown()` to unregister the global providers. The OTel API silently ignores repeat global registration, so a second `initTelemetry()` kept using the dead providers — which is exactly why the first version of these tests failed.

### Low — `resolveRoute` called twice

Resolved once in `handle()` and passed into `dispatch()`.

### Low — unhandled rejection in `expressSpanMiddleware`

`.catch()` added, logging rather than crashing the dev server.

### On the +610KB bundle

Flagging rather than deciding, as you suggested — this needs a call from whoever owns the deploy/cold-start budget. It's server-only (nothing reaches browsers) and telemetry is opt-in, but the SDK is bundled unconditionally, so the cost is paid whether or not it's switched on. If that's not acceptable, the OTel packages could be marked `--external` and loaded from `node_modules`, which trades bundle size for a runtime dependency. Happy to do that if preferred.

Verified after rebasing onto `9739963`: typecheck fully clean (your "Typescript errors in tests" commit cleared the 9 pre-existing `entry-ssr.test.ts` errors), lint 0 errors, 3724 tests passing, and the end-to-end checks re-run against the bun-bundled production binary.

---

Issue #1558, opened 2026-07-30
Chapter + highlights are flickering on page swipe


Blitzy:
On mobile, when swiping to change chapters there is a flicker of what appears to be chapter content and the highlights before settling on the new chapter that was swiped to.

If I had to guess, I would say this was introduced with the fast navigation changes?

https://github.com/user-attachments/assets/e469ee08-cb57-4aca-a1b7-4a1b44779eab

PR #1559, opened 2026-07-31
fix: keep the swiped-to chapter on screen until its text arrives

Blitzy:
Swiping between chapters on mobile flashed the chapter the reader had
just left. The gesture slid the track over to the neighbouring panel —
a static preview of the chapter being swiped to — and then, at the end
of the animation, snapped straight back to the centre panel *before*
navigating. Because navigation no longer waits on the download, the
centre panel still held the outgoing chapter at that moment, so it was
painted for at least a frame before the new text replaced it.

The track now rests on the neighbouring panel until the navigation
settles, then recentres. In the usual case that panel is the prefetched
preview of exactly the chapter being fetched, so the wait is a few
milliseconds. A settle budget caps it, so a slow download falls back to
the existing dimmed-text path rather than stranding the reader on a
static, unscrollable preview.

A commit token tracks who owns the track's transform, so a swipe still
waiting on its chapter cannot yank the track away from a gesture that
has since taken over, or write to a track the effect no longer owns.

Refs #1558

Claude-Session: https://claude.ai/code/session_01Ez9cyKrfsnpuiaGt8cxiMD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 
 ### 🐛 Fixed
 
+- Fix swiping between chapters on mobile flashing the chapter you just left. The swipe used to slide over to the next chapter's preview and then immediately snap back to the centre of the reader — which was still showing the previous chapter — while the new text downloaded. It now stays on the preview until the new chapter's text is ready, and only gives up and snaps back if the download is genuinely slow. ([#1558](https://github.com/HelloAOLab/seed-bible/issues/1558))
 - Changing chapter or book no longer waits on an in-flight text request. The position updates the moment you press, and repeated presses advance a chapter each, instead of the chevrons and arrow keys switching off until the download finished. ([#1551](https://github.com/HelloAOLab/seed-bible/pull/1551))
 - Fix shared sessions freezing and eventually crashing the tab with an out-of-memory error when another participant moved through chapters quickly. ([#1551](https://github.com/HelloAOLab/seed-bible/pull/1551))
 - Fix an out-of-range chapter in the address, such as `?chapter=99999`, leaving the address pointing at a chapter you are not on, so pressing Back returned you to it and bounced straight forward again. ([#1551](https://github.com/HelloAOLab/seed-bible/pull/1551))
--- a/packages/seed-bible/seed-bible/components/TabsLayout.tsx
+++ b/packages/seed-bible/seed-bible/components/TabsLayout.tsx
@@ -36,6 +36,14 @@ export function TabSlotReader(props: TabSlotReaderProps) {
   const swipeTouchStartY = useRef<number | null>(null);
   const swipeDirectionLocked = useRef<"h" | "v" | null>(null);
   const swipeCurrentDx = useRef(0);
+  // Identifies whoever currently owns the track's transform. Bumped by every
+  // new gesture, so a committed swipe still waiting to recentre knows it has
+  // been superseded and leaves the track to the gesture in progress.
+  const swipeCommitToken = useRef(0);
+  // True while a committed swipe is resting on a neighbouring panel waiting for
+  // the chapter it navigated to. The track is off-centre in that window, which
+  // a new gesture has to correct before it starts following the finger.
+  const swipeParked = useRef(false);
   const currentChapterRef = useRef(readingState.chapterData.value);
   const lastScrollTopRef = useRef(0);
   const scrollerCleanupRef = useRef<(() => void) | null>(null);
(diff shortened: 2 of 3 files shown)


Blitzy:
> [!NOTE]

## Code review — keep the swiped-to chapter on screen until its text arrives

**Short version:** This is a clean, well-reasoned fix for the flicker reported in issue No. 1558. The diagnosis is correct — because fast navigation (No. 1551) no longer waits on the download, recentring the track immediately snapped back onto the centre panel while it still held the *outgoing* chapter, so it painted for a frame. Parking on the neighbouring preview panel until the new text settles (with a budget cap) is the right shape for the fix. The comments are excellent and the three new tests cover the main paths well. I found no correctness bugs that would block merge — only two low-severity edge cases and a couple of notes.

I reviewed this statically (the container has no `node_modules`, so I did not run the suite). I did verify the surrounding facts the change relies on: `loadNextChapter()` / `loadPreviousChapter()` resolve only after `chapterData` actually updates to the target (via `whenContentSettled`), and `CHAPTER_SKELETON_DELAY_MS` is `500`, so the "settle budget is half the skeleton delay" comment is accurate.

---

### Low — pending swipe timeouts are not cleared on effect teardown

`components/TabsLayout.tsx` (`commitSwipe`, around lines 398–432 and the cleanup at 481–489)

The commit is animated with `window.setTimeout(..., SWIPE_ANIMATION_MS)`, and the id is never stored. The effect cleanup bumps `swipeCommitToken` and resets `swipeParked`, but it does not clear that pending timeout. When the timeout fires it calls `navigate()` **before** it checks the commit token:

```js
window.setTimeout(() => {
  const settled = navigate().catch(() => undefined); // always runs
  if (swipeCommitToken.current !== commit) {
    return;                                           // token check is after navigate()
  }
  ...
}, SWIPE_ANIMATION_MS);
```

So if the effect tears down within the ~250 ms animation window, `loadNextChapter()` / `loadPreviousChapter()` still runs and advances the reading position for a reader who has left. The effect's deps are `[isMobile, readingState]`, so the realistic triggers are a viewport resize/rotation that crosses the mobile breakpoint, or the reader slot/tab unmounting, in the moment right after a committed swipe.

Note this is deliberately *different* from the supersede-by-a-new-gesture case, where you explicitly want the navigation to still run ("the reader completed the swipe that asked for it") and the token only guards the transform. The token can't tell "superseded by gesture" from "torn down" apart, so the clean fix is to cancel the pending timeouts **only** in the cleanup:

```js
const swipeTimers = new Set<number>();
// inside commitSwipe:
const animId = window.setTimeout(() => {
  swipeTimers.delete(animId);
  ...
}, SWIPE_ANIMATION_MS);
swipeTimers.add(animId);
// (do the same for the SWIPE_SETTLE_BUDGET_MS timer)

// cleanup:
return () => {
  swipeTimers.forEach((id) => window.clearTimeout(id));
  swipeTimers.clear();
  swipeCommitToken.current += 1;
  swipeParked.current = false;
  ...
};
```

That leaves the gesture-supersede behaviour untouched (a new gesture doesn't run cleanup) while stopping a torn-down reader from navigating.

**Reproduction:** hard to hit reliably by hand because of the ~250 ms window, but: swipe to the next chapter on mobile and, before the slide finishes, rotate/resize across the mobile breakpoint (or close the tab). The chapter still advances underneath.

---

### Low — a new touch during the parked wait recentres before direction is known

`components/TabsLayout.tsx` (`onTouchStart`, lines 320–329)

While a commit is parked on the preview panel waiting for its chapter, the very next `touchstart` unconditionally un-parks and recentres the track onto the live centre panel:

```js
if (swipeParked.current) {
  swipeParked.current = false;
  track.style.transform = centreTransform();
}
```

If the new touch turns out to be a vertical scroll (or a tap) *and* the chapter hasn't loaded yet — i.e. the reader touched the screen during the sub-250 ms parked window on a slowish connection — the centre panel still holds the outgoing chapter (dimmed), so this can momentarily reveal exactly the frame the PR is trying to suppress.

This is genuinely a trade-off rather than a clear defect: recentring at `touchstart` is what lets a scroll or a follow-up swipe work immediately, and the exposure is a narrow, dimmed flash on slow loads only. If you want to close it, un-park inside `onTouchMove` at the point the horizontal direction locks (recentre there instead of at `touchstart`), leaving the parked preview in place when the gesture resolves to vertical. I'd lean toward leaving it as-is unless the flash proves noticeable in practice — flagging it so the choice is deliberate.

**Reproduction:** throttle the network so a chapter takes ~150–250 ms; swipe to the next chapter and, during the parked wait, place a finger on the screen. On a slow load you may see a brief flash of the previous chapter before the new text lands.

---

### Notes / nits (no action required)

- **`swipeCommitToken` is incremented twice per committed swipe** — once in `onTouchStart` (line 318) and again in `commitSwipe` (line 403). This is harmless (the commit captures the post-increment value), and the `onTouchStart` bump is meaningful on the *non*-committing path (it invalidates a parked commit when a gesture ends below threshold). Worth a one-line comment noting the two bumps are intentional, since it reads oddly otherwise.
- **`swipeTimers`/settle timer growth** — if you adopt the Set-based cleanup above, remember to `delete` each id when its callback fires (shown above) so the Set doesn't accumulate stale ids over a long-lived reader.
- **Test coverage gap (optional):** all three new tests mock `loadNextChapter`, so they never exercise the real re-render where `chapterData` updates and the preview panels shift (`next` becomes `next+1`). The design handles this — the preview recompute is async and the recentre is a synchronous microtask — but a test that drives an actual chapter update through the fixture would lock in that the recentre lands on the freshly-rendered current panel, which is the exact behaviour the fix promises.

---

### What's good

- Correct root-cause diagnosis, and the fix targets it precisely rather than papering over the symptom.
- The commit-token ownership model cleanly separates "the navigation the reader asked for" from "who owns the transform," and the supersede test (`hands the track to a new gesture...`) proves it.
- The budget cap is the right safety valve, tuned to the existing skeleton delay, so a slow download degrades to the established dimmed-text path instead of stranding the reader on a static preview.
- Comments explain the *why* throughout, which makes this area much easier to maintain.

---

PR #1560, opened 2026-07-31
perf: cut the initial bundle download roughly in half

KallynGowdy:
The client shipped 2,162 KB (602 KB gzipped) before it could paint, most of
it a single eager `vendor.js` chunk at 1,286 KB / 367 KB gzipped. Over half
of that chunk was code no first-time visitor ever runs.

Initial download is now 1,124 KB / 312 KB gzipped, with `vendor.js` at
230 KB / 69 KB. No features were removed — the deferred pieces load on
first use.

The vendor group defeated every lazy import. `codeSplitting.groups` put all
of node_modules into one chunk, and rolldown applied that literally: a
package landed in `vendor` even when the only route to it was a dynamic
`import()`. Since the entry statically needs part of vendor, the chunk
loads eagerly and dragged the rest along. TipTap already loaded through
`lazy()`, yet its emitted chunk was 3.8 KB while ~500 KB of ProseMirror sat
in vendor; the transcript extension's reference parser (121 KB) and
dompurify (26 KB) were there for the same reason. Tagging the group
`$initial` restricts it to boot-path modules and keeps one long-cacheable
vendor chunk.

Five packages were also bundled twice; three of them are collapsed here —
prosemirror-view (94 KB wasted), zod (54 KB), and prosemirror-transform
(30 KB). Vite's `resolve.dedupe` cannot fix this, because it resolves to the
project root `node_modules` and pnpm does not hoist transitive dependencies
there; the prosemirror entries listed there had never had any effect and are
removed. Pinning through `pnpm-workspace.yaml` `overrides` is what collapses
them. chromium-bidi keeps zod 3, which its internals require, via a more
specific selector; it is dev-only tooling and never reaches the client.

The other two, lib0 and uuid, are left duplicated on purpose. aux-common
depends on both directly, at older majors than yjs and the app use, so an
override would force it onto APIs it was not built against — the same hazard
that carved chromium-bidi out of the zod override. The two stale copies come
to about 1 KB (lib0 0.2.49 at 1.1 KB, uuid 10.0.0 at 0.8 KB) and neither is
on the initial download: both sit in the lazily loaded shared-documents
chunk, and vendor carries only uuid 13.

Four things that only matter after some user action now load on demand:
Typesense search (with the copy of axios it carries, ~154 KB), Yjs shared
documents (~97 KB), and the avatar editor (12 KB).

The Yjs split needed one extra piece. A dynamic import alone did nothing,
because a static path to Yjs also existed: aux-websocket's
WebsocketConnectionClient, which the boot path needs, imports the aux-common
package root, and that barrel re-exports ./partitions, which reaches the Yjs
document classes. aux-common ships no `sideEffects` field, so none of it
could be shaken out and Yjs counted as statically reachable. A tree-shaking
hint marking aux-common's `documents`/`partitions` directories
side-effect-free is what actually moves it.

luxon (69 KB) is gone. Relative timestamps use `Intl.RelativeTimeFormat`,
which matches luxon's output across all 77 translated locales — with one
deliberate exception: CLDR carries no relative-time data for Guaraní, so
both luxon and Intl silently rendered English inside an otherwise
translated UI. Guaraní now maps to Spanish, its co-official language in
Paraguay, and the mapping applies only while the platform actually lacks
the locale.
The reading-plan calendar needed real IANA-time-zone-aware arithmetic, so
it moves to `civilDate.ts`, which resolves an instant to the calendar date a
person in that zone sees and then adds and subtracts whole days on the
calendar. Millisecond arithmetic would drift by an hour at each daylight
saving change; calendars have no DST. New tests cover spring-forward,
fall-back, a half-hour-offset zone, and leap days. luxon stays for
apologist-extension, which is code-split.

Finally, the CI build-size threshold drops from 300 KB to 50 KB so a future
eager import shows up in the PR comment instead of sliding through.

Claude-Session: https://claude.ai/code/session_01FvhDpx8bCWHS7JEggosnRn

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,10 @@ jobs:
 
       - name: Generate build size report
         run: |
-          ARGS=(report --head size-head.json --out size-report.md --threshold-kb 300)
+          # 50 KB rather than 300: the eager vendor chunk is now ~230 KB, so a
+          # 300 KB allowance would let it more than double without comment. A
+          # single mistakenly-eager import is the failure mode worth catching.
+          ARGS=(report --head size-head.json --out size-report.md --threshold-kb 50)
           if [ -f size-base.json ]; then
             ARGS+=(--base size-base.json)
           elif [ "${{ github.event_name }}" = "pull_request" ]; then
--- a/package.json
+++ b/package.json
@@ -134,7 +134,6 @@
     "i18next": "catalog:",
     "i18next-resources-to-backend": "catalog:",
     "imagetracerjs": "^1.2.6",
-    "luxon": "catalog:",
     "opentype.js": "^1.3.4",
     "preact-render-to-string": "catalog:",
     "prompts": "^2.4.2",
--- a/packages/seed-bible/package.json
+++ b/packages/seed-bible/package.json
@@ -30,7 +30,6 @@
     "@casual-simulation/websocket": "catalog:",
     "axios": "catalog:",
     "i18next": "catalog:",
-    "luxon": "catalog:",
     "preact": "catalog:",
     "preact-iso": "catalog:",
     "react": "catalog:",
--- a/packages/seed-bible/seed-bible/app/utils.tsx
+++ b/packages/seed-bible/seed-bible/app/utils.tsx
@@ -20,6 +20,134 @@ export const translateTitle = (
   });
 };
 
+/**
+ * How many whole calendar months separate two instants, signed and truncated
+ * toward zero. Calendar-aware rather than "every month is 30 days", so
+ * Jan 31 → Feb 28 counts as a whole month and Jan 1 → Jan 31 counts as none.
+ */
+function wholeMonthsBetween(fromMs: number, toMs: number): number {
+  // Always measure from the earlier instant forward, then re-apply the sign.
+  // Measuring backward from the later one is not symmetric: Feb 28 minus a
+  // month is Jan 28, which is earlier than Jan 31 and would drop the month,
+  // while Jan 31 forward correctly reaches Feb 28.
+  const sign = toMs >= fromMs ? 1 : -1;
+  const earlier = new Date(Math.min(fromMs, toMs));
+  const laterMs = Math.max(fromMs, toMs);
+  const later = new Date(laterMs);
+
+  let months =
+    (later.getFullYear() - earlier.getFullYear()) * 12 +
+    (later.getMonth() - earlier.getMonth());
+
+  // The calendar delta above rounds outward — Jan 31 to Feb 1 reads as one
+  // month. Step back toward zero when less than a full month has elapsed.
+  if (months > 0) {
+    const anchor = new Date(earlier.getTime());
+    const wantedMonth = (((anchor.getMonth() + months) % 12) + 12) % 12;
+    anchor.setMonth(anchor.getMonth() + months);
+    // `setMonth` overflows when the day is missing from the target month —
+    // Jan 31 plus one month lands on Mar 3. Clamp back to that month's last
+    // day so Jan 31 → Feb 28 stays the whole month it reads as.
+    if (anchor.getMonth() !== wantedMonth) {
+      anchor.setDate(0);
+    }
+    if (anchor.getTime() > laterMs) {
+      months -= 1;
+    }
+  }
+  return months * sign;
+}
+
+/**
+ * Locales the app translates but CLDR has no relative-time data for, mapped to
+ * the nearest locale it does cover. Left alone, `Intl` falls back to English,
+ * which strands an English "3 days ago" in an otherwise translated UI. Guaraní
+ * is co-official with Spanish in Paraguay, where nearly all of its speakers
+ * are, so Spanish reads far less foreign to them than English.
+ */
+const RELATIVE_TIME_LOCALE_FALLBACKS: Record<string, string> = { gn: "es" };
+
+function resolveRelativeTimeLocale(locale: string): string {
+  const fallback = RELATIVE_TIME_LOCALE_FALLBACKS[locale];
+  // The mapping only applies when the platform genuinely lacks the locale, so
+  // this corrects itself if CLDR gains data later.
+  if (
+    fallback &&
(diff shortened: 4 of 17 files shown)


Blitzy:
> [!NOTE]

## Code review — perf: cut the initial bundle download roughly in half

This is a high-quality, unusually well-documented change. The lazy-loading conversions (Yjs, Typesense, the avatar editor), the `luxon` removal, and the `$initial` / tree-shaking / `overrides` build fixes are all sound and each carries an inline comment explaining *why*. I verified the changed test suites locally (`utils.test.ts` + `ReadingPlansManager.test.ts`, **85 passed**), confirmed `luxon` is fully gone from the main app (only `apologist-extension`, which is code-split, still uses it), and confirmed each newly-lazy module has exactly one dynamic import site so nothing eagerly re-imports it and quietly undoes the split.

No blocking issues. Everything below is Low severity or a verification note. There is no linked issue on the PR, so this review is scoped to the diff and its stated goals.

---

### Low — `civilDateInZone` now throws on an invalid time zone where `luxon` degraded silently

**Where:** `packages/seed-bible/seed-bible/managers/civilDate.ts` (`getFormatter` / `civilDateInZone`)

`new Intl.DateTimeFormat("en-US", { timeZone })` throws a `RangeError` when `timeZone` is a string that isn't a recognized IANA zone. I confirmed this on the runtime here:

```
$ node -e "new Intl.DateTimeFormat('en-US',{timeZone:'Bad/Zone'})"
RangeError: Invalid time zone specified: Bad/Zone
```

The old `luxon` path (`DateTime.fromMillis(ms, { zone })`) never threw — a bad zone produced an *Invalid* `DateTime` and the arithmetic silently yielded `NaN` offsets. So this is a behavior change: `dateForSession`, `sessionsForDate`, and `getReadingCalendar` can now throw instead of degrading, and an exception from `getReadingCalendar` would break rendering of the reading-plan calendar rather than just showing wrong dates.

**Why the impact is Low today:** `progress.timeZone` is `z.string().nullable().optional()` and defaults to `null` (→ machine zone), and the field is currently marked "reserved for future local-day-boundary handling" (`ReadingPlansManager.tsx:231`). So in practice it's almost always `null` right now. The risk grows once that field starts getting populated from stored/user data, since nothing validates it as a real IANA zone.

**Reproduction:** call `getReadingCalendar(plan, progress, Date.now())` with a `progress.timeZone` of `"Bad/Zone"` — it throws `RangeError` instead of returning entries.

**Suggested fix:** make `getFormatter` fall back to the machine zone if construction throws, preserving the old lenient behavior:

```ts
function getFormatter(timeZone: string | undefined): Intl.DateTimeFormat {
  const key = timeZone ?? "";
  let formatter = formatterCache.get(key);
  if (!formatter) {
    try {
      formatter = new Intl.DateTimeFormat("en-US", { timeZone, /* … */ });
    } catch {
      // Unknown zone — match the old luxon behavior of resolving in the
      // machine zone rather than throwing.
      formatter = new Intl.DateTimeFormat("en-US", { /* … no timeZone */ });
    }
    formatterCache.set(key, formatter);
  }
  return formatter;
}
```

---

### Low — `uploadFile` axios→fetch drops axios's automatic `Content-Type` default; please confirm the signed headers carry it

**Where:** `packages/seed-bible/seed-bible/managers/OsManager.tsx` (`uploadFile`)

The switch from `axios.request({ data: encodedData })` to `fetch(url, { body: encodedData })` is a net improvement in error handling (`uploadResult.ok` is checked explicitly, and `fetch` won't silently throw before that check the way a default-`validateStatus` axios would). One difference is worth a sanity check: axios sets `Content-Type: application/octet-stream` by default when you hand it a typed-array `data` and no explicit type, whereas `fetch` sends **no** `Content-Type` for a `Uint8Array` body unless it's in `headers`.

For an S3-style presigned `PUT` this is usually *more* correct — the object's content type must match what was signed, so sending only the server-provided `uploadHeaders` is right, and axios auto-adding an unsigned `Content-Type` could actually break the signature. The thing to confirm is simply that `recordFileResult.uploadHeaders` always includes the intended `Content-Type` (it should, since `uploadFile` already computes `mimeType` and passes it to `client.recordFile`). This path has no unit coverage, so it's worth one manual smoke test of an actual upload (e.g. a profile-picture upload, which now also goes through the newly-lazy editor).

---

### Notes / verification (not defects)

- **The build-config changes are the highest-risk part of this PR and are the least verifiable by the unit suite.** The `$initial` chunk tag, the `codeSplitting.groups` change, and especially `treeshake.moduleSideEffects: false` on `aux-common/{documents,partitions}` can only fail at bundle/runtime, never in Vitest. The reasoning about `aux-common` having no `sideEffects` field and the `BlobPolyfill` side effect reaching the bundle via the package root is careful and convincing, but I'd still recommend a runtime smoke test on the built bundle of the three deferred paths before merge: (a) join/host a **shared multiplayer session** (exercises the lazy Yjs `RemoteYjsSharedDocument` chunk and confirms nothing needed was tree-shaken out of `partitions`), (b) run a **search** (lazy Typesense chunk), and (c) open **Update picture** in settings (lazy avatar editor + its `Suspense` fallback). The deployment preview link in the PR is a good place to do this.

- **`formatRelativeTime` parity with `luxon` looks intentional and correct.** `numeric: "always"` + truncate-toward-zero matches `luxon.toRelative()`'s default unit set (years/months/days/hours/minutes/seconds — no weeks/quarters), and the month/leap/DST edge cases are well covered by the new tests. No action needed — just flagging that display strings for the relative timestamps are now produced by hand-rolled `Intl` rather than the library, so this is the reference for future tweaks.

- **Nit — the Guaraní fallback table keys on the exact locale string.** `RELATIVE_TIME_LOCALE_FALLBACKS` is `{ gn: "es" }` and is looked up with the raw `language` value. If i18next ever resolves Guaraní as a regioned tag (e.g. `gn-PY`) rather than bare `gn`, the remap is bypassed and it would fall through to English again. Fine as long as `useI18n().language` yields bare codes (which matches the locale-file naming), but a base-language lookup (`locale.split("-")[0]`) would make it robust to that.

Nice work overall — the initial-download win is real and the accompanying documentation makes the risky bits reviewable.

---

PR #1561, opened 2026-07-31
Split extension list translations into per-language chunks

KallynGowdy:
## Summary

The Settings extensions list shows each extension's `title` and `description`. Those strings were inlined into the `virtual:@extensions` module for **all 77 UI languages**, and that module is a static import on the boot path — so every reader downloaded 77 languages to render one, on a screen most never open.

Measured by deleting the literal from the built chunk and re-gzipping: **138 KB raw / 72.5 KB gzipped, 34.6% of the gzipped entry chunk.**

| | before | after |
|---|---|---|
| `assets/index.js` | 630.4 KB raw / 209.4 KB gz | **500.7 KB / 138.9 KB gz** |
| Total initial download | 1123.6 KB / 312.0 KB gz | **935.7 KB / 242.2 KB gz** |

`vendor.js` is byte-identical. Nothing is deferred and no loading state is added.

> **Scope:** the vendor-chunk work this branch was cut from has since merged as #1560, so the diff here is now only the two commits below. An earlier version of this description documented #1560's changes as well — apologies to anyone who went looking for `civilDate.ts`.

## How it works

- **`script/lib/extensionsModule.ts` (new)** — source generation for the `virtual:@extensions` module family, split out of the plugin so it is unit-testable without running a build, matching how `precacheManifest.ts` is separated from its own plugin.
- **`virtual:@extensions/locale/<lang>`** — a new virtual module per language, ~2.2 KB each, holding every extension's `title`/`description` for that language. Reached through a new optional `loadListTranslations` map on `ExtensionSet`.
- **`ExtensionManager`** fetches only the active language, and re-fetches when the UI language changes. Sets that ship strings inline (uploaded extensions) are unaffected and still use `meta.translations`.

**English stays inline.** `ExtensionMeta` requires it, it is i18next's `fallbackLng` so it renders wherever a language has no string of its own, and it means the list shows real titles rather than raw ids in the moment before a locale chunk lands. Measured cost: 1.6 KB raw / 0.6 KB gz. It was all 77 languages that cost 138 KB, not one.

The existing trimming is kept — the full `extension.json` files total 1.1 MB, so inlining `title`/`description` was the right call. The defect was only shipping every language of it.

## Review fixes (`e1ccd8d`)

- **SSR listener leak.** `createExtensionManager` runs once per SSR request while `i18n` is the process-wide i18next singleton, so the `languageChanged` listener accumulated for the life of the server process. The whole list-translation path is now skipped under SSR — the list is never server-rendered and English is inline, so nothing is lost, and the initial fetch was likewise a per-request write to shared state. Two tests cover it; both fail if the guard is removed.
- **Redundant build-time reads.** `load()` re-ran `readdir` and re-parsed every manifest for each of the 78 modules it emits. Instrumenting `readExtensionMeta` across a real build: **936 reads before, 12 after.** The promise is cached so Rollup's concurrent calls share one read, a failed read clears rather than poisons the cache, and the dev watcher now also drops it on content change — `addWatchFile` re-runs `load()` on an edit, which would otherwise be served stale metas.
- **`setsWithListTranslations` growth** — noted in the code for whoever adds a set-untracking path; there is none today.

The pre-existing unguarded listener in `I18nManager.tsx` is deliberately left alone: it updates a signal the SSR render reads, so guarding it is not the same one-line change and doesn't belong in this PR.

## Testing

3,456 tests pass. New: `test/unit/script/lib/extensionsModule.test.ts` for the generator, plus `ExtensionManager` coverage for active-language-only fetching, language switching, the SSR guard, sets without the new loader, and a locale chunk that fails to load (the list degrades to ids rather than throwing). `check:ts` and `lint` clean; client and SSR builds succeed.

**Not yet verified by hand:** Settings → Extensions in a browser. The language-switch path is covered by unit tests but not by eye.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,10 @@ jobs:
 
       - name: Generate build size report
         run: |
-          ARGS=(report --head size-head.json --out size-report.md --threshold-kb 300)
+          # 50 KB rather than 300: the eager vendor chunk is now ~230 KB, so a
+          # 300 KB allowance would let it more than double without comment. A
+          # single mistakenly-eager import is the failure mode worth catching.
+          ARGS=(report --head size-head.json --out size-report.md --threshold-kb 50)
           if [ -f size-base.json ]; then
             ARGS+=(--base size-base.json)
           elif [ "${{ github.event_name }}" = "pull_request" ]; then
--- a/package.json
+++ b/package.json
@@ -134,7 +134,6 @@
     "i18next": "catalog:",
     "i18next-resources-to-backend": "catalog:",
     "imagetracerjs": "^1.2.6",
-    "luxon": "catalog:",
     "opentype.js": "^1.3.4",
     "preact-render-to-string": "catalog:",
     "prompts": "^2.4.2",
--- a/packages/seed-bible/package.json
+++ b/packages/seed-bible/package.json
@@ -30,7 +30,6 @@
     "@casual-simulation/websocket": "catalog:",
     "axios": "catalog:",
     "i18next": "catalog:",
-    "luxon": "catalog:",
     "preact": "catalog:",
     "preact-iso": "catalog:",
     "react": "catalog:",
--- a/packages/seed-bible/seed-bible/app/utils.tsx
+++ b/packages/seed-bible/seed-bible/app/utils.tsx
@@ -20,6 +20,134 @@ export const translateTitle = (
   });
 };
 
+/**
+ * How many whole calendar months separate two instants, signed and truncated
+ * toward zero. Calendar-aware rather than "every month is 30 days", so
+ * Jan 31 → Feb 28 counts as a whole month and Jan 1 → Jan 31 counts as none.
+ */
+function wholeMonthsBetween(fromMs: number, toMs: number): number {
+  // Always measure from the earlier instant forward, then re-apply the sign.
+  // Measuring backward from the later one is not symmetric: Feb 28 minus a
+  // month is Jan 28, which is earlier than Jan 31 and would drop the month,
+  // while Jan 31 forward correctly reaches Feb 28.
+  const sign = toMs >= fromMs ? 1 : -1;
+  const earlier = new Date(Math.min(fromMs, toMs));
+  const laterMs = Math.max(fromMs, toMs);
+  const later = new Date(laterMs);
+
+  let months =
+    (later.getFullYear() - earlier.getFullYear()) * 12 +
+    (later.getMonth() - earlier.getMonth());
+
+  // The calendar delta above rounds outward — Jan 31 to Feb 1 reads as one
+  // month. Step back toward zero when less than a full month has elapsed.
+  if (months > 0) {
+    const anchor = new Date(earlier.getTime());
+    const wantedMonth = (((anchor.getMonth() + months) % 12) + 12) % 12;
+    anchor.setMonth(anchor.getMonth() + months);
+    // `setMonth` overflows when the day is missing from the target month —
+    // Jan 31 plus one month lands on Mar 3. Clamp back to that month's last
+    // day so Jan 31 → Feb 28 stays the whole month it reads as.
+    if (anchor.getMonth() !== wantedMonth) {
+      anchor.setDate(0);
+    }
+    if (anchor.getTime() > laterMs) {
+      months -= 1;
+    }
+  }
+  return months * sign;
+}
+
+/**
+ * Locales the app translates but CLDR has no relative-time data for, mapped to
+ * the nearest locale it does cover. Left alone, `Intl` falls back to English,
+ * which strands an English "3 days ago" in an otherwise translated UI. Guaraní
+ * is co-official with Spanish in Paraguay, where nearly all of its speakers
+ * are, so Spanish reads far less foreign to them than English.
+ */
+const RELATIVE_TIME_LOCALE_FALLBACKS: Record<string, string> = { gn: "es" };
+
+function resolveRelativeTimeLocale(locale: string): string {
+  const fallback = RELATIVE_TIME_LOCALE_FALLBACKS[locale];
+  // The mapping only applies when the platform genuinely lacks the locale, so
+  // this corrects itself if CLDR gains data later.
+  if (
+    fallback &&
+    Intl.RelativeTimeFormat.supportedLocalesOf([locale]).length === 0
+  ) {
+    return fallback;
+  }
+  return locale;
+}
+
+/**
+ * Formatters are immutable and comparatively expensive to build — around 35x
+ * the cost of a format call — and the reader rebuilds every visible timestamp
+ * on a 15-second timer, so they are kept per locale. Keyed by the requested
+ * locale rather than the resolved one, so the fallback lookup above is paid
+ * once too. Bounded by the number of UI languages, the same as the time-zone
+ * cache in `managers/civilDate.ts`.
+ */
+const relativeTimeFormatters = new Map<string, Intl.RelativeTimeFormat>();
+
+function getRelativeTimeFormatter(locale: string): Intl.RelativeTimeFormat {
+  let formatter = relativeTimeFormatters.get(locale);
+  if (!formatter) {
+    formatter = new Intl.RelativeTimeFormat(resolveRelativeTimeLocale(locale), {
+      numeric: "always",
+    });
+    relativeTimeFormatters.set(locale, formatter);
+  }
+  return formatter;
+}
+
+/** Fixed-length units, largest first, used below the one-month threshold. */
+const RELATIVE_TIME_UNITS: [Intl.RelativeTimeFormatUnit, number][] = [
+  ["day", 86_400_000],
+  ["hour", 3_600_000],
+  ["minute", 60_000],
+  ["second", 1_000],
+];
+
+/**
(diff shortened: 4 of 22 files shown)


Blitzy:
> [!NOTE]

## Scope note

This PR's description documents a lot of ground (the `civilDate`/`formatRelativeTime` utilities, Luxon removal, `SettingsPage`/`SearchManager` lazy-loading, `vite.config.ts` tree-shaking, `pnpm-workspace.yaml` dedup). None of that is actually in this PR's diff — it already landed in `develop` via PR #1560 before this branch was opened. The commit that is actually unique to this PR (`f93e281`, "perf: fetch extension list strings per language instead of all 77") only touches:

- `packages/seed-bible/seed-bible/managers/ExtensionManager.tsx`
- `script/lib/extensionsModule.ts` (new)
- `script/lib/vite-plugin-extensions.ts`
- `test/unit/script/lib/extensionsModule.test.ts` (new)
- `test/unit/seed-bible/managers/ExtensionManager.test.ts`

The review below is scoped to that actual diff. Worth trimming the PR description so a reviewer isn't hunting for changes (e.g. `civilDate.ts`) that aren't there.

## Findings, most severe first

### 1. (High) New `languageChanged` listener leaks on every SSR request

`ExtensionManager.tsx` now does, at the top level of `createExtensionManager`:

```ts
i18n.on("languageChanged", (language: string) => {
  for (const set of setsWithListTranslations) {
    void loadListTranslationsForLanguage(set, language);
  }
});
```

`i18n` here is the process-wide `i18next` default-export singleton (`I18nManager.tsx: import i18n from "i18next"; export { i18n };` — not a per-request instance). `createExtensionManager` is not a one-time boot call: `standalone/entry-ssr.tsx`'s `render()` calls `createSeedBibleState(...)` — which calls `createExtensionManager(...)` — fresh **on every incoming HTTP request** handled by the long-running SSR server (`server/index.ts`). Each call registers a brand-new closure on the shared singleton, and there is no matching `i18n.off(...)` anywhere, so every request permanently retains a listener that closes over that request's `setsWithListTranslations` and the rest of the manager's closures. Over the life of a long-running server process, the listener count (and retained memory) only ever grows.

Note there's a pre-existing instance of the same pattern in `I18nManager.tsx` (`createI18nManager`'s own unconditional `i18n.on("languageChanged", ...)` at line 177, outside the `isInitialized` guard), so this isn't a novel anti-pattern for this codebase — but this PR adds a second, independent leaking listener rather than fixing or reusing the existing one, doubling the accumulation rate.

**Reproduction:**
1. In a REPL/test, call `createExtensionManager(login)` (or `createSeedBibleState(...)`, which the real SSR path uses) several times in a row against the real `i18next` singleton (not the mocked one the unit tests use).
2. Inspect the instance's listener count for `"languageChanged"` (i18next's emitter exposes this similarly to Node's `EventEmitter`, e.g. via its internal `observer`/`listeners` — or simpler: call `i18n.changeLanguage("es")` after N manager creations and count how many times a `console.log` placed inside the handler fires; it will fire N times).
3. In production, this happens on literally every rendered page: hit the SSR endpoint repeatedly and watch heap usage / listener count climb without bound.

**Suggested fix:** Skip registering the listener during SSR entirely — it does nothing useful there (SSR renders once and never lives long enough to react to a language change), and the codebase already has the convention for this (`import.meta.env.SSR`, used elsewhere e.g. in `BibleReadingManager.tsx`, `LoginManager.tsx`). Something like:

```ts
if (!import.meta.env.SSR) {
  i18n.on("languageChanged", (language: string) => {
    for (const set of setsWithListTranslations) {
      void loadListTranslationsForLanguage(set, language);
    }
  });
}
```

If the manager is ever expected to be torn down client-side too (e.g. in tests, or if this becomes disposable later), capturing the handler and exposing a way to call `i18n.off("languageChanged", handler)` would be the more complete fix.

### 2. (Medium) Every locale chunk re-discovers and re-parses all extensions at build time

`vite-plugin-extensions.ts`'s `load(id)` now calls `discoverExtensions()` — which re-runs `discoverExtensionFolders()` (a fresh `readdir` + per-folder `existsSync`) and re-reads + re-`JSON.parse`s every extension's `extension.json` — for **every** virtual module it's asked to build, not just the entry module. Because `generateEntryModuleSource` emits one `import("virtual:@extensions/locale/<lang>")` per language the extensions collectively support (up to the ~76 non-English languages the repo carries translations for), Rollup will statically resolve and build all of those chunks during `pnpm build`, each one re-running full extension discovery. With today's 9 extension packages, that's roughly 9 packages × ~76 languages ≈ 680+ redundant file reads/parses on top of the entry module's own discovery — pure build-time cost (this doesn't affect the runtime chunk sizes the PR is optimizing for), but it will keep growing as more languages/extensions are added.

**Reproduction:** Run `pnpm build` and trace/instrument `readExtensionMeta` (e.g. a `console.log` or a `strace`/`--trace-fs` style tool) — you'll see each `extension.json` read once per locale module generated, not once per build.

**Suggested fix:** Cache the `DiscoveredExtension[]` result for the lifetime of a build (e.g. a module-level `let cached: Promise<DiscoveredExtension[]> | null = null` in `vite-plugin-extensions.ts`, populated on first `load()` call and reused by subsequent calls), and invalidate it in the existing dev-server file-watcher callback (the same place that already invalidates the entry + locale modules on an `extension.json` change).

### 3. (Low) `setsWithListTranslations` never shrinks

`trackExtensionSet` adds a set to `setsWithListTranslations` whenever it has `loadListTranslations`, but nothing in this diff removes a set from it if/when the extension set is untracked (I didn't find an "untrack"/removal path for `ExtensionSet`s in this file at all, so this may be moot today, but flagging it in case set removal is added later — otherwise the `languageChanged` handler will keep trying to re-fetch translations for sets that are no longer relevant).

## What looks good

- `extensionsModule.ts` is a clean extraction — pulling `trimMeta`/`toJsLiteral`/module generation out of the Vite plugin so it's unit-testable without a build is a good call, and the new test suite (`extensionsModule.test.ts`) covers the interesting edge cases well: omitting extensions with no translation for a language (rather than emitting `undefined`), rejecting malformed/path-traversal-shaped language subtags in `parseLocaleModuleId`, and the empty-language-module case.
- `ExtensionManager.test.ts`'s new tests correctly cover the behaviors that matter for this feature: only the active language is fetched, a language switch fetches the new one, sets without `loadListTranslations` still fall back to inline `meta.translations`, and a rejected chunk load doesn't take the Settings page down (just degrades to showing the extension id, matching the existing `defaultValue` fallback).
- `parseLocaleModuleId`'s regex-based validation of the language subtag is a sensible guard against a malformed id causing the plugin to try to emit a module for something like `../secret`.
- The CI threshold change (300 KB → 50 KB) and its comment explaining why are a reasonable tightening given the vendor chunk's new size, and match the actual PR title/goal (catching an accidental large eager import).

---

KallynGowdy:
Thanks — all three findings were real. Fixed in `e1ccd8d`, and the description has been trimmed to this PR's actual diff.

**1. SSR listener leak (High)** — confirmed exactly as described. `createSeedBibleState` calls `createExtensionManager` at line 462, `entry-ssr.tsx` calls it per request, and `i18n` is the i18next default-export singleton.

I went slightly further than the suggested fix and skipped the whole list-translation path under SSR, not just the listener. The initial `loadListTranslationsForLanguage` call in `trackExtensionSet` had the same underlying problem in a quieter form: a per-request `addTranslations` write into state shared by every request. Nothing is lost on the server — the Settings list is never server-rendered and English is inline as the fallback.

Two tests cover it: one asserts no listener and no fetch under SSR, one asserts the client still registers exactly one. I checked both fail when the guard is removed.

**2. Redundant build-time discovery (Medium)** — confirmed, and it was worse than estimated. Instrumenting `readExtensionMeta` across a real `pnpm build`: **936 reads before, 12 after** (12 extension packages × 78 modules — the entry plus 77 locales).

Cached as you suggested. Three details worth flagging:
- The **promise** is cached rather than the result, since Rollup drives `load()` concurrently and caching the result still lets several calls race into a read.
- A failed read clears the cache rather than poisoning it for the rest of the build.
- The watcher now also drops the cache on **content** change, not just add/unlink. That one isn't in your suggestion but the cache makes it necessary: `addWatchFile` already re-runs `load()` when a manifest is edited, and without clearing it that re-run would be served stale metas. Caching would have introduced a dev-only bug otherwise.

**3. `setsWithListTranslations` never shrinks (Low)** — you're right that it's moot today; I confirmed there's no untracking path. Rather than add speculative removal logic I've left a note at the declaration so whoever adds one knows to prune there.

On the **pre-existing listener** in `I18nManager.tsx:177` — I left it. It isn't quite the same one-line fix: that handler updates the `language` signal, and `changeLanguage` is async, so during SSR the signal initialiser can capture a stale value and the listener is what corrects it. Guarding it could plausibly break server-rendered language selection. It deserves its own change with its own testing rather than a drive-by in a bundle-size PR.

**Verification:** 3,456 tests pass, `check:ts` and `lint` clean, client and SSR builds succeed, and the initial payload is unchanged at 935.7 KB raw / 242.2 KB gzipped — these were correctness and build-time fixes, not size ones.

Still outstanding on my side: I haven't opened Settings → Extensions in a browser. The language-switch path has unit coverage but no human has looked at it.

---

KallynGowdy:
I can confirm that Settings -> Extensions loads correctly in multiple locales

PR #1562, opened 2026-07-31
Add Playlist AI features

KallynGowdy:
This PR, when merged, adds support for dynamically registering AI contexts with tools that AI agents can call via the chat window. 
In particular, when editing a playlist, the user can ask Apologist to make changes for them.


Blitzy:
> [!NOTE]

## Code review — Add Playlist AI features (No. 1562)

Nice work — this is a substantial, well-structured feature. The generalization of `ChatProvider.generateResponse` into a message stream, the shared/local context merging in `ChatsManager`, and the split between the AI-facing `AIPlaylistItemSchema` and the real `PlaylistItemData` are all clean, and the added unit tests give good coverage of the timeline/tool-call rendering and the context lifecycle.

The theme of the findings below is **error handling around the new tool-calling loop**: the happy path is solid, but several failure modes (a model that omits a field, an HTTP error, an unknown tool) currently end with the typing indicator silently disappearing and an unhandled promise rejection in the console, instead of a visible error. Since AI models routinely produce imperfect tool calls, these paths will be hit in practice.

No linked issue is referenced in the PR description, and there are no prior review comments to reconcile.

---

### Process note

GitHub reports this PR's mergeable state as **`dirty`** (merge conflict with `develop`). It'll need a merge/rebase against `develop` before it can land.

---

### High severity

**H1 — Tool-call errors silently kill the whole response (no catch around `generateResponse`)**

In both `createLocalChatSession` and `createSharedChatSession`, the provider is driven inside `void (async () => { try { … } finally { … } })()` with a `finally` but **no `catch`** (`ChatsManager.tsx:2028-2050` and `:1608-1629`). Any error thrown out of the provider's `generateResponse` generator becomes an unhandled promise rejection; the only visible effect is that the "typing…" indicator vanishes with no message.

The new tool loop in `init.tsx` throws in several ordinary situations:
- `throw new Error(\`Tool not found: ${fn.name}\`)` when the model names a tool that isn't registered (`init.tsx:368-370`).
- `JSON.parse(fn.arguments)` when the model emits malformed or empty arguments (`init.tsx:372`).
- the tool `function` itself rejecting (see H2).

*Reproduction:* Open a playlist for editing, start an AI chat, and ask for something that nudges the model to call a tool with a slightly-wrong name or invalid JSON args (or force it via a mocked provider that yields a `tool_call` for an unregistered tool). The response ends with no assistant message and an `Uncaught (in promise)` in the console.

*Suggested fix:* Wrap the `processProviderResponse` call in a `try/catch` (or catch inside the tool loop) and surface a user-visible error message via `upsertMessage`, so a bad tool call degrades to an error bubble rather than a dead turn. Catching per-tool-call inside `init.tsx` (returning the error text as the tool result so the model can recover) is also worth considering.

---

**H2 — `convertToPlaylistItem` crashes on schema-valid input**

`AIPlaylistItemSchema` is a flat object where `bibleVerse` / `link` / `html` are each `nullable().default(null)` and independent of the `type` discriminant (`AIManager.tsx:60-77`). So `{ type: "bible-verse", bibleVerse: null }` passes `safeParse` cleanly. `convertToPlaylistItem` then does `item.bibleVerse!.ref.bookId` (`AIManager.tsx:96-101`) and throws `TypeError: Cannot read properties of null`. The same applies to the `link`/`html` branches, and symmetrically to `convertToAiPlaylistItem`.

Because `insertPlaylistItem` / `updatePlaylistItem` run `convertToPlaylistItem(args)` inside the tool `function` (`PlaylistManager.tsx:2415-2426` in the diff), this rejection flows straight into H1 and silently kills the response.

*Reproduction:* Have the model (or a test) call `insertPlaylistItem` with `{ index: 0, type: "bible-verse" }` and no `bibleVerse` object — an easy mistake for a model to make. The tool call throws instead of returning an error string.

*Suggested fix:* Enforce the type↔field relationship at the schema level (a discriminated union, or a `.superRefine` that requires the matching object for the given `type`), or make the converters return a descriptive `"error: …"` string when the expected variant field is missing (matching the pattern the `PlaylistManager` tools already use).

---

### Medium severity

**M1 — No `response.ok` check on the chat completions request**

In `generateResponse`, the `fetch` result is passed straight to `streamChoices(response)` with no status check (`init.tsx:270-282`). On a 4xx/5xx (bad/expired key, rate limit, upstream error) the body isn't SSE, so `parseSseJsonStream` finds no `data:` lines, the skip loop drains to `done`, and the turn loop `break`s — again leaving the user with a vanished typing indicator and no message. The previous non-streaming code would at least have thrown while reading `choices[0].message`.

*Suggested fix:* Check `response.ok` (and ideally read the error body) before streaming, and yield a `type: "text"` error message so the failure is visible.

---

### Low severity / nits

**L1 — Left-in debug logging.** `console.log("Generating response for message:", lastMessage)` and `console.log("Chat context:", chatContext)` (`init.tsx:187-188`) run on every generation; the second logs the full context including the tool list. Worth removing or gating.

**L2 — SSE tail not flushed.** `parseSseJsonStream` returns on reader `done` without processing whatever remains in `buffer` (`init.tsx:86-90`). A final `data:` line without a trailing newline would be dropped. OpenAI's terminal `data: [DONE]\n\n` makes this uncommon, but an abrupt disconnect could lose the last delta.

**L3 — Redundant tool shape.** The mapped tool object sets `name` / `description` / `parameters` / `strict` at both the top level and inside the nested `function` (`init.tsx:228-240`). The chat-completions API only reads the nested `function`; the top-level copies are dead weight.

**L4 — Doc-comment typo.** `/**d` should be `/**` in `AIProviderFunctionTool.function`'s JSDoc (`AIManager.tsx:12`).

**L5 — Vestigial fragment.** `DiscoverPaneHeader` now wraps a single `<button>` in a `<>…</>` fragment with no sibling (`DiscoverPane.tsx:734-743`) — looks like a leftover from moving the AI button onto the title row. The fragment can be dropped.

**L6 — `effect` that never reacts.** `effect(() => { enableCoreChatContext(); })` (`SeedBibleStateManager.tsx:2684-2686`) reads no signals, so it runs exactly once; a direct call would be clearer than an `effect` wrapper. (Non-blocking — just a readability note.)

---

### Things that look right

- `mergeLocalChatContexts` correctly collapses empty instructions/tools to `undefined`, and the `.peek()`-based `setContext`/`addContext`/`removeContext` writes avoid the self-retriggering-effect loop the comment calls out — good catch documenting the reasoning.
- The `PlaylistManager` editing tools consistently return `"success"` / `"error: …"` strings so the model gets actionable feedback (H2 aside), and reading `editingPlaylist` live at call time rather than snapshotting is the right call.
- Timeline grouping for tool-call events and the `ENTRY_KIND_PRIORITY` stable-sort tweak are tidy, and the new tests exercise them.

---

KallynGowdy:
@Blitzy those issues should be fixed now and could use another review

Blitzy (dismissed):
@KallynGowdy this is really cool, works very well in my testing. I only ran into a couple of odd snags but nothing I would consider blocking, could be followed up on later.

📺 Video: https://www.youtube.com/watch?v=GlXyVohplzE

### 1. Stale playlist state after discarding an edit (~1:37 to 3:50)
I asked the AI to delete the AO Lab link, then backed out without saving. Reopened the playlist and the link was correctly still there, but the AI's internal state still reported it as removed. Asking it to re-fetch the `Ryan Test 1` playlist state didn't clear the stale copy, it still said the item was gone. Its "current playlist items" listing was out of sync with what was actually in the editor.

### 2. "Make a playlist" edits the open playlist instead of creating a new one (~3:50 to 4:29)
With the playlist editor open, asking for a new playlist caused it to edit the currently open one. It makes sense given what's in context, but it isn't what I expected. Might be worth disambiguating create vs edit when an editor is already in context.

### 3. New playlist created but not opened (~4:29 to 5:20)
After closing the editor and asking for a new playlist on a different topic, it said it created one but nothing opened. I had to refresh and start a new chat before the same request worked. Possibly related to leftover context from the previous session.

### 4. Tool discoverability (~0:31 to 1:03)
There's no visible list of what tools the AI has in the current context. Low priority since you can just ask it and it answers correctly, but surfacing them in the UI might be nice?

The tool calling itself is great, and being able to say "open Genesis 5" or have it build out a whole playlist is genuinely powerful.

Blitzy:
> [!NOTE]

## Code review — Add Playlist AI features (No. 1562)

This is a re-review following @KallynGowdy's "those issues should be fixed now" comment.

**What got fixed since the last pass:** the missing `catch` around `generateResponse` is now there in both session types, so a failed turn shows an error bubble instead of vanishing silently. `convertToPlaylistItem` no longer crashes on a schema-valid-but-empty variant — it throws a readable message that the playlist tools convert into an `"error: …"` tool result, with tests covering each variant. `response.ok` is now checked before streaming. The SSE tail is flushed, the debug `console.log`s are gone, the duplicated top-level tool fields are gone, the `/**d` typo is fixed, the vestigial fragment is gone, and `enableCoreChatContext()` is a direct call instead of a one-shot `effect`. A `MAX_COMPLETION_TURNS` bound was added on top of that. Good, thorough follow-up.

**The theme this time:** the error handling was fixed at the *ChatsManager* layer (the user now sees something went wrong) but not at the *tool-loop* layer (the model can't recover from its own mistake). Three ordinary model errors still tear down the whole turn instead of being handed back as a tool result. Separately, a few paths through the new streaming loop still end with no message at all — the same "typing indicator disappears" symptom, reached a different way.

I ran the suite locally on `60568e9dc`: **296 tests pass**, `pnpm check:ts` clean, `pnpm lint` 0 errors. The new tests are genuinely behavioural — the playlist tool tests drive the real zod-validated tool functions and assert on `editingPlaylist`, and the two new error-path tests do fail on the pre-fix code. No linked issue is referenced in the PR description.

---

### Process notes

**Merge conflict.** GitHub still reports this as `CONFLICTING` against `develop`. A likely contributor: `en.json` has 242 changed lines for 10 genuinely-new keys — a large block (the `reading-plan-*` keys, `show-more-languages`, `book-not-found-*`, and others) got relocated to the end of the file. I verified no keys were lost or changed in value, so this is pure reordering churn, but it makes the diff hard to read and will keep conflicting with anything else touching `en.json`. Worth restoring the original ordering and appending only the new keys.

**Bundle size.** The CI report flags **+2.9MB total** (client +1.4MB, SSR +1.5MB) and `extension.js` going 73KB → 217KB (+31KB gzip). Some of that is real feature weight, but it's worth confirming none of it is accidental — see L4 below on the deep import path.

---

### High severity

**H1 — A single bad tool call still kills the entire turn instead of letting the model recover**

Three lines in the tool-resolution loop can throw, and none of them are caught inside the loop ([init.tsx:365-369](packages/apologist-extension/ext_Apologist/main/init.tsx#L365-L369)):

1. `throw new Error(\`Tool not found: ${fn.name}\`)` — the model named a tool that isn't registered.
2. `JSON.parse(fn.arguments)` — malformed or **empty** arguments.
3. `await tool.function(args)` — `generateFunctionTool` does `return Promise.reject(result.error)` on any zod validation failure ([AIManager.tsx:185](packages/seed-bible/seed-bible/managers/AIManager.tsx#L185)).

Each one propagates out of `generateResponse`, hits the new `catch` in `ChatsManager`, and ends the turn with `chat-ai-error`. So the user's before/after is: **before**, the typing indicator vanished silently; **now**, they get "Sorry, something went wrong generating a response (…)" and the turn is still over. The model never learns what it got wrong and never retries.

Two of these are high-likelihood, not theoretical:

- **Empty arguments on a zero-arg tool.** `getPlaylistState` takes `z.object({})`. The delta accumulator initialises `args: ""` and only appends when `delta.function?.arguments` is truthy, so if the provider sends no arguments delta at all, `JSON.parse("")` throws `SyntaxError: Unexpected end of JSON input`. OpenAI itself sends `"{}"`, so this depends on what Apologist's upstream emits — but it's a one-character fix and the cost of being wrong is the flagship "what's in my playlist right now?" flow dying.
- **A wrong book ID.** `AIBibleVerseRefSchema` deliberately rejects non-USFM book IDs, and commit `7a4698136` ("Issues with AI generated book names and IDs being wrong") says this happens in practice. Today that rejection surfaces as a rejected promise, so asking for "Song of Songs 2" and getting `SONG` back from the model ends the turn — and because `err.message` is interpolated into the error string, the user sees a raw multi-line ZodError JSON blob in the chat bubble.

*Reproduction:* register a stub provider (or point at a model) that emits a `tool_call` for `getPlaylistState` with `arguments: ""`, or a `goToReference` call with `bookId: "GENESIS"`. Either ends the turn with an error bubble rather than the model correcting itself.

*Suggested fix:* wrap each tool call so failures become the tool *result*, which is what the model is designed to react to:

```ts
for (const call of toolCalls) {
  const fn = call.function;
  let content: string;
  try {
    const tool = chatContext.tools?.find((t) => t.name === fn.name);
    if (!tool) {
      throw new Error(`Tool not found: ${fn.name}`);
    }
    const args = fn.arguments ? JSON.parse(fn.arguments) : {};
    content = JSON.stringify(await tool.function(args));
  } catch (err) {
    content = `error: ${err instanceof Error ? err.message : String(err)}`;
  }
  messages.push({ role: "tool", tool_call_id: call.id, name: fn.name, content });
  yield { type: "tool_call", name: fn.name };
}
```

That also makes the loop consistent with the `PlaylistManager` tools, which already return `"error: …"` strings on purpose. Keeping the `ChatsManager` catch as the outer net is still right — it just shouldn't be the *first* line of defence for something as routine as a mistyped tool name.

---

### Medium severity

**M1 — `strict: true` doesn't match the schema being sent, so it's either ignored or a hard 400**

Every tool is advertised with `strict: true` ([init.tsx:226](packages/apologist-extension/ext_Apologist/main/init.tsx#L226)). OpenAI-style strict function calling requires `additionalProperties: false` on every object, every property listed in `required`, and no `default` keyword. I generated the actual schema with this repo's zod:

```
required: ["type","index"]          // bibleVerse / link / html are NOT required
additionalProperties: undefined     // never emitted
"bibleVerse": { "default": null, "anyOf": [...] }
```

`.nullable().default(null)` makes those three properties optional in the input schema and adds `default`, so the schema violates strict mode three ways over. Since the feature demonstrably works in @Blitzy's video, Apologist must currently ignore the flag — which means `strict: true` is a claim the code can't back, and the day Apologist honours it (or this provider is pointed at real OpenAI) every tool call 400s.

*Suggested fix:* drop `strict: true`, since the code already handles loosely-shaped arguments (the `.default(null)` fields plus the H1 error strings exist precisely to tolerate them). If strict enforcement is actually wanted, `AIPlaylistItemSchema` needs to become a discriminated union of strict objects — which would also make H1's missing-variant case unrepresentable.

**M2 — Several paths through the turn loop end without emitting any message**

The original "typing indicator vanishes with nothing to show" symptom is still reachable, just by different routes ([init.tsx](packages/apologist-extension/ext_Apologist/main/init.tsx)):

- [L305](packages/apologist-extension/ext_Apologist/main/init.tsx#L305) `if (next.done) break;` — the stream ended with only empty deltas.
- [L384](packages/apologist-extension/ext_Apologist/main/init.tsx#L384) a tool-call turn finishing with `finish_reason: "stop"` breaks out, so the user is left with only "used 3 tools" and no answer.
- [L390](packages/apologist-extension/ext_Apologist/main/init.tsx#L390) a content-less turn that stops.
- [L257](packages/apologist-extension/ext_Apologist/main/init.tsx#L257) exhausting `MAX_COMPLETION_TURNS` falls out of the `for` loop and returns normally — 25 turns of work, no output, no error.

*Suggested fix:* track whether anything was yielded, and on any of these exits emit one `type: "text"` message ("I wasn't able to finish that — try again?" / a turn-limit message). The turn-cap case in particular should probably be an error, since silently doing 25 round-trips and returning nothing is indistinguishable from a hang.

**M3 — The SSE reader is abandoned on the success path, so `releaseLock()` never runs**

`parseSseJsonStream` releases its reader in a `finally` ([init.tsx:110](packages/apologist-extension/ext_Apologist/main/init.tsx#L110)), but that `finally` only fires if the generator is completed or `.return()`ed. On the normal path, `textDeltas()` breaks as soon as `finishReason` is set — before `data: [DONE]` — and then the outer generator `return`s. `stream` is left suspended and simply dropped, so neither generator's `finally` runs: the response body stays locked and the connection isn't cancelled until GC. Every AI turn leaks one. The `break` paths in M3 do the same.

Related: there's no cancellation path at all. `AIProviderGenerateOptions` declares a `cancelToken: AbortSignal` ([AIManager.tsx:26](packages/seed-bible/seed-bible/managers/AIManager.tsx#L26)) but that interface is never used anywhere, and `ChatContext` carries no signal — so closing the chat mid-generation can't stop the request.

*Suggested fix:* wrap the turn loop in `try { … } finally { await stream?.return?.(); }` (or restructure to `for await`, which cleans up automatically on `break`). Threading a real `AbortSignal` through `ChatContext` into `fetch` is the fuller fix, and would give `AIProviderGenerateOptions` a reason to exist.

**M4 — Playing a synthetic playlist writes a malformed `?playlist=.<id>` into the URL**

`startPlaying` now accepts `SimplePlaylist`, which has no `recordName`, and `getPlaylistLocator` papers over that with `recordName ?? ""` ([PlaylistManager.tsx:90](packages/seed-bible/seed-bible/managers/PlaylistManager.tsx#L90)). The locator is fed into the reader's query params, so the URL becomes `?playlist=.4f6a…&playlistStep=0` — leading dot, empty record name. `parsePlaylistLocator` rejects exactly that shape (`if (lastDot <= 0)`) and logs `Invalid playlist locator:`, so the link can never be restored.

Two callers hit this. The new `playPlaylist` tool ([SeedBibleStateManager.tsx:1691](packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx#L1691)) also uses a bare `uuid()` rather than the `playlist_${uuid()}` convention. And reading-plan day playback used to pass `recordName: plan.recordName`; this PR comments those fields out to satisfy the narrower type ([BibleToolsManager.tsx:754](packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx#L754)) — which also leaves the comment two lines above it stating the playlist "borrows the plan's own record name" now factually wrong.

*Reproduction:* ask the AI for a tour ("walk me through some verses about hope"), watch playback start, then look at the address bar — `?playlist=.<uuid>`. Reload: console logs `Invalid playlist locator: .<uuid>` and playback doesn't resume.

*Suggested fix:* have `transformQueryParams` emit `playlist` only when the playing playlist has a `recordName` (a transient AI/plan playlist genuinely isn't addressable, and the PR comment already acknowledges "a shared/reloaded URL won't resume it"). Emitting nothing beats emitting a locator that's guaranteed to fail. Either way, delete the commented-out fields rather than leaving them in.

**M5 — Bonfire now activates against hardcoded org/AI IDs, and a failed session start silently bricks the chat**

Dropping `bonfireApiKey` from a **URL query parameter** is a clear improvement — that was a client-side secret. Two things about what replaced it:

- `DEFAULT_ORG_ID` / `DEFAULT_AI_ID` are baked into source ([bonfire-extension/main/init.tsx](packages/bonfire-extension/ext_Bonfire/main/init.tsx)), so the extension went from "no-ops unless you pass three query params" to "always registers a Bonfire AI wired to one specific org". Requests to `bonfire.seedbible.io` now carry no credential at all, so whoever can read the shipped bundle can drive that org's AI on its quota. If `bonfire.seedbible.io` is a first-party proxy that injects the key, that's the right shape — but it needs its own auth/rate limiting, and these two IDs look like build-time env config rather than source constants. Worth confirming before this ships broadly.
- The `if (!orgId || !aiId)` guard below is now unreachable (both sides have `??` fallbacks), and `apiKey` was commented out rather than deleted, leaving its `/** The API key for the Bonfire API. */` doc orphaned above `name`.

Separately: `onJoinChat` calls `bonfireSessionStartResponseSchema.parse(await response.json())` with no `response.ok` check, and `ChatsManager` fires `onJoinChat` without awaiting or catching it. So a failed session start becomes an unhandled promise rejection, `chatSessionMap` never gets an entry, and every later message returns `null` — a permanently dead chat whose only trace is a `console.error`. Worth a `response.ok` check plus a `try/catch` that surfaces a message.

---

### Low severity / nits

**L1 — `ai-context-tool-count` renders "1 tools".** [FloatingReaderPanels.tsx:919](packages/seed-bible/seed-bible/components/FloatingReaderPanels/FloatingReaderPanels.tsx#L919) passes `count`, which puts i18next into plural mode, but `en.json` only has `ai-context-tool-count`, no `_one`. `chat-tool-used` got its `_one`/`_other` pair right — this key needs the same. Note the test can't catch it: the `useI18n` mock in `FloatingReaderPanels.test.tsx` was changed to do naive `{{var}}` substitution, so it never exercises plural resolution.

**L2 — With no AI providers, the AI button makes a new dead chat on every click.** `startAiChat(null)` looks for an existing chat via `p.providerId === providerId`, which never matches when `providerId` is `null`, so the reuse branch can't fire; then `if (providerId)` is false so no participant is added ([DiscoverPane.tsx:107](packages/seed-bible/seed-bible/components/DiscoverPane/DiscoverPane.tsx#L107)). Each click yields another chat containing just "What do you want to add/change?" and nobody to answer. Simplest fix: hide the button when `providers.length === 0`. The tests cover 0/1/many providers but always with an empty `chats` list, so the reuse branch is untested either way.

**L3 — `reorderEditingPlaylistItem` error text says "or equal" but equal is a success.** [PlaylistManager.tsx:766-768](packages/seed-bible/seed-bible/managers/PlaylistManager.tsx#L766-L768) — `from === to` now returns `"success"`, so the trailing "or equal" is left over from the merged condition and will mildly mislead the model.

**L4 — Deep `@packages/…` import from the extension.** `resolveMessageAuthors` is pulled in via `@packages/seed-bible/seed-bible/managers/ChatsManager`, bypassing the `seed-bible` package the extension actually depends on. There's precedent for the alias elsewhere, but `seed-bible/managers` is an exported subpath — `ChatsManager` just isn't in its barrel. Adding `export * from "./ChatsManager"` to `managers/index.tsx` and importing from `seed-bible/managers` would keep the boundary intact, and is worth trying to see whether it accounts for part of the `extension.js` 73KB → 217KB jump.

**L5 — Dead code.** `AIProviderGenerateOptions` ([AIManager.tsx:20](packages/seed-bible/seed-bible/managers/AIManager.tsx#L20)) has no references anywhere. `createNewPlaylist` now returns `Signal<Playlist | null> | null` ([PlaylistManager.tsx:610](packages/seed-bible/seed-bible/managers/PlaylistManager.tsx#L610)) but its only caller ignores it — looks like the remains of a create-playlist tool that didn't land. `BibleVersePlaylistItem` is exported but only used in the same file.

**L6 — `searchVerses` tool results.** `bookLabel` and `translationLabel` are set to the raw ids they duplicate ([SeedBibleStateManager.tsx:1681](packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx#L1681)), so the model gets `{"bookId":"JHN","bookLabel":"JHN"}` — two tokens' worth of nothing, and a "label" that isn't one. Either resolve real display names or drop the fields. The hit list is also uncapped; if Typesense's page size ever grows, a whole page of verse text lands in the context.

**L7 — `.sb-reading-plans-back` restyle reaches four call sites.** `DiscoverShared.css` bumps padding `4px 8px` → `0.625rem 1.25rem`, adds `font-weight: 600` and a transition, and **removes** `display: inline-flex; align-items: center; gap: 4px`. That suits the text "Cancel" button in `CreatePlaylistForm`, but the other three users are icon-only or icon+text back buttons — the two `<MaterialIcon>arrow_back</MaterialIcon>` buttons in `DiscoverPane` lose their flex centering and get noticeably chunkier. Worth a look in the browser, or splitting the text-button variant into its own class.

**L8 — Test brittleness.** The two new error-path tests (and two pre-existing ones, now amended) use three consecutive `await Promise.resolve()` to advance past the provider's microtasks. That's counting microtasks — adding one `await` in the production path silently changes what these observe, and it's the pattern CLAUDE.md's testing guidance steers away from. The other new tests already use `vi.waitFor`; these could too. Also `readingPlanPlayback.test.ts` stubs chats as `{ removeContext: vi.fn() } as any`, which will throw the day that test's playlist gets edited — `PlaylistManager.test.ts`'s `makeChats()` helper is the better model.

**L9 — `tool_call` provider messages have no manager-level test.** `ChatView.test.tsx` covers rendering the notifications well (single, collapsed run, author change, group breaking, empty state). But nothing asserts that a provider yielding `{ type: "tool_call", name }` actually becomes a `tool_call` message on the session — `grep -c tool_call test/unit/seed-bible/managers/ChatsManager.test.ts` is 0. That's the one seam between the two halves of the feature, and it's the branch at the top of `processProviderMessage`.

---

### Follow-ups from the manual-testing review

- **No. 1 (stale playlist state after discarding an edit)** — addressed as well as prompting can address it: the `getPlaylistState` tool, plus instructions saying the snapshot supersedes earlier turns and to call the tool before answering. The snapshot is rebuilt per generation, and the context is withdrawn entirely when the editor closes, so the mechanics are right. Prompt-level, so worth re-testing the exact sequence from the video.
- **No. 2 ("make a playlist" edits the open one)** — addressed by instruction: the editor tools now say never to use them for a different playlist and to tell the user to close the editor first, and `playPlaylist` says it's always a brand-new one-off. Also prompt-level.
- **No. 3 (new playlist created but not opened)** — this one still looks open. There's no create-playlist tool: the core context is `goToReference`, `searchVerses`, `playPlaylist`. With the editor closed, "make me a playlist about X" has nothing that creates or opens one, so the model's most likely moves are still to start transient playback or to claim it did something. The unused new `createNewPlaylist` return value (L5) suggests a tool was intended here. If it's deliberately out of scope, an instruction telling the model it can't create saved playlists would stop it claiming otherwise.
- **No. 4 (tool discoverability)** — done nicely. The `auto_awesome` header button lists each active context with its label and tool count, with a badge once more than one is active, and it's covered by four tests.

---

### Things that look right

- Tool execution is safe by construction in shared chats: both `generateResponse` call sites are behind `if (!participant.isAI || participant.isRemote || !authorId) return;`, so tools only ever run on the client of the user who sent the message. No risk of a remote participant driving someone else's reader or playlist editor.
- The `.peek()`-based `setContext`/`addContext`/`removeContext` writes, with the comment explaining that a tracked read would make the calling effect re-trigger itself. `PlaylistManager`'s registration effect writes to the same signals it would otherwise depend on, so this genuinely matters and the reasoning is worth having in the file.
- Reading `editingPlaylist` live inside each tool rather than closing over a snapshot, so tools always target the playlist actually open — and the tests assert exactly that, including the after-cancel case.
- The `expandCrossChapterItem` export comment ("a list of readings and the queue built from it are not the same length") is the kind of non-obvious invariant that earns a comment.
- `ChatProviderResponse` widening: `isProviderMessageStream` discriminating on the absence of `type` is a clean way to keep single-message providers working untouched, and `processProviderResponse` correctly awaits each message's own text stream to completion before pulling the next one from the provider.
- The playlist tool tests go through the real zod-wrapped `function`, so they'd catch a schema regression, and they assert on `editingPlaylist` rather than on mock calls.
- `en.json`: verified 546 → 556 keys with zero removals and zero value changes, and `chat-tool-used` correctly ships `_one`/`_other`.

---

Blitzy:
@KallynGowdy Things are improving! Here are my notes on previous issues and new findings.

## Reviewing old issues

### No. 1

It does seem better at grabbing the current state of the playlist when making edits. I could not reproduce "1. Stale playlist state after discarding an edit (~1:37 to 3:50)" from my previous QA review.

### No. 2
However "2. "Make a playlist" edits the open playlist instead of creating a new one (~3:50 to 4:29)" does still seem to be occurring but in more limited cases. If I use the explicit language "Make a new playlist..." Apologist explains the dilemma with something like: 

>I can do that — but you currently have an open playlist editor for "Ryan Test #1". I’m not allowed to create a brand-new, separate playlist while this editor is open. You have two options:
>
> - Close the current editor (so I can create a fresh new playlist), or
> - Repurpose the open playlist into the new one about "How God views sin" (I can update its title/description and replace/add items now).
>
> ...

But if i say something like "Make a playlist..." which could be inferred to be either "edit" or "make new" it generally just assumes and edits the open playlist. Possibly just a reinforcement issue in the agent prompt?

## New findings / thoughts

### No. 3

We should probably flag agents with a tool calling capability so that they do not show tool context in the chat window and do not show up in the dropdown menu for the "AI" button in the playlist editor. For example the Bonfire AI (as of this writing) does not support tool calling so asking it to do things like open a verse or make changes to a playlist it just responds that it cant do that. It would be nice to filter those agents out?

image
image


### No. 4

When asking an agent to make a new playlist, it generates the playlist and activates it for you — but you have no way of being able to save it or make edits. This makes using AI to create a baseline playlist and make your own edits to it impossible.

image

I think the natural path forward would be to open it in playlist editor instead of in play mode? This will let users immediately view the output, and make any edits either manually or with the agent since the agent would then automatically have the playlist editor context. It also is a natural place to just quickly save the AI generated playlist, which then throws the user back to their list of saved playlists where they can hit play on the new playlist.

KallynGowdy:
For 2 its probably best to just not allow creating new playlists while editing another playlist. This is an area that is ripe for confusion for the agent and I can imagine all sorts of weird things happening where the user wants it to make a change but then the agent just creates something else.

KallynGowdy:
And for number 4, this is because AI agents don't actually have access to a "create playlist" tool. Instead, it was using the "play playlist" tool that allows it to dynamically create and play a playlist that it creates. I'll change that tool to be for creating playlists and dump the user into the edit playlist form.

KallynGowdy:
@Blitzy There's now a `createPlaylist` tool that AI agents can use. I replaced the `playPlaylist` tool since it doesn't seem very useful to play a playlist but not save it.

KallynGowdy:
Now we only show AI providers that support tool calling for some actions (like clicking on the "AI" button when editing a playlist)

PR #1563, opened 2026-07-31
Feature/client branding

AmitPant9999:


Blitzy:
> [!NOTE]

## Code Review — Feature/client branding (No. 1563)

Thanks for putting this together, @AmitPant9999. The branding indirection itself is clean and lightweight. However, the PR as it stands has two blocking problems: it silently deletes the entire `obsolete/` tree, and the branding feature never actually turns on because nothing populates the config. Details below, ordered by severity.

There is no linked issue on this PR, so I reviewed against the code and the project conventions in `CLAUDE.md`.

---

### 🔴 Critical

#### C1 — The commit deletes the entire `obsolete/` directory (2,760 files, ~287k lines)

The diff for this PR reports **184 additions and 287,401 deletions across 2,822 files**. Almost all of that is the first commit (`c741186 "add client branding config"`) removing the whole `obsolete/` directory, which still exists on `develop`. This looks unintentional — a branding change shouldn't be deleting a quarter-million lines of unrelated code — and if merged it would wipe `obsolete/` from `develop` with no discussion.

**Reproduction:**
```bash
git show --stat c741186 | tail -3
#   2822 files changed, 184 insertions(+), 287401 deletions(-)
git show --numstat c741186 | grep -c '^0.*obsolete/'
#   2760
git ls-tree origin/develop --name-only | grep obsolete
#   obsolete   <-- still present on develop
```

**Suggested fix:** restore the directory and keep only the intended six code files + the en.json change:
```bash
git checkout develop -- obsolete/
git commit -m "restore obsolete/ (unintended deletion)"
```
If removing `obsolete/` is in fact desired, it should be a separate, clearly-labelled PR (ideally its own discussion), not bundled into a branding change.

#### C2 — The branding feature is never wired to a data source, so it never activates

`BrandingConfig` is added to `AppConfig` (`appConfig.tsx:12`, `appConfig.tsx:37`) and consumed in five components, but nothing ever *sets* `branding`. Both server-side render sites build the config without it, and the injected `window.__APP_CONFIG__` never carries it either:

- `server/index.ts` — the `RenderFn` config type (lines 84–89) doesn't even include a `branding` field, and both construction sites (~line 257 and ~line 569) only set `basePath` / `assetHost` / `renderedAsMobile` / `acceptedLanguages`.
- `clientConfigFromHeaders()` (server/index.ts:93) never produces branding.

The net effect: `branding` is always `undefined`, so every consumer hits its `?? "Seed Bible"` / default-icon fallback. The feature is dead code as shipped — there is currently no way for a deployment to actually supply branding.

**Reproduction:**
```bash
grep -rn "branding" server/ packages/ --include=*.ts --include=*.tsx | grep -v i18n | grep -v obsolete
# only the interface definition + the 5 consumers appear; no assignment anywhere
```

**Suggested fix:** thread branding through from a real source (env var, per-branch config file, or records lookup) into the SSR render config, add `branding` to the `RenderFn` config type in `server/index.ts`, and include it in the `__APP_CONFIG__` payload injected into the page so the client picks it up at hydration. Until that exists, the consuming code can't be exercised or tested.

---

### 🟠 Medium

#### M1 — 55 non-English translation files were modified, against project convention

`CLAUDE.md` states: *"When adding or updating any translation key, only update en.json ... don't update other translation files, leave those for the professionals."* This PR edits 55 locale files (`de`, `fr`, `th`, `sk`, …) to swap the literal "Seed Bible" for `{{appName}}`.

Beyond the convention, this is genuinely risky for gendered/article languages, where the interpolation now bakes in an article that assumed the word "Seed Bible":
- `de.json`: `"Lesen Sie ... in der {{appName}}"`
- `fr.json`: `"... les tenants et aboutissants de la {{appName}}"`

If a client's `appName` differs grammatically from "Seed Bible", these read incorrectly — which is exactly why the convention defers non-en translations to professionals.

**Suggested fix:** revert the 54 non-`en.json` changes and keep only `en.json`. Let the translation pipeline re-translate the `{{appName}}` variants.

**Reproduction:**
```bash
git diff --name-only develop HEAD -- 'packages/seed-bible/seed-bible/i18n/*.json' | grep -v /en.json | wc -l
#   55
```

#### M2 — `today-screen` `SeedBibleIcon` drops caller styling/sizing in the branded path

In `packages/today-screen/infrastructure/presentation/components/ui/SeedBibleIcon.tsx`, the default path renders `<div className="seed-bible-icon" style={style} />`, but the new branded path renders ``'&lt;img src={branding.icon} alt={branding.appName} /&gt;'`` with **no `style`, no width/height, and no class**. Callers rely on `style` for sizing:

```
Welcome.tsx:32        <SeedBibleIcon style={seedBibleIconStyle} />
SearchSection.tsx:18  <SeedBibleIcon style={seedBibleIconStyle} />
```

So a branded icon would render at intrinsic image size, unstyled, and likely break those layouts (it just won't surface today because of C2).

**Suggested fix:** apply the incoming `style` (and sensible dimensions) to the ``, e.g. ``'&lt;img src={branding.icon} alt={branding.appName} style={style} /&gt;'``, and confirm it fills the same box the `seed-bible-icon` class defines.

---

### 🟡 Low / minor

- **L1 — Cross-package context may not share an instance.** `today-screen/.../SeedBibleIcon.tsx` imports `useAppConfig` from `@packages/seed-bible/seed-bible/app/appConfig`. If `today-screen` ends up bundling its own copy of that module, its `AppConfigContext` object won't be the same one `AppConfigProvider` publishes, so it would always read `DEFAULT_APP_CONFIG` regardless of C2. Worth verifying the today-screen subtree actually renders inside the seed-bible `AppConfigProvider` and shares one module instance.

- **L2 — `main.tsx` gates `SeedBibleIcon` behind branding, but the `` doesn't respect `size`.** In `components/icons.tsx:50`, the SVG branch scales by `size`, while the branded `` (lines 58–64) sets `width={size} height={size}` but also spreads `{...props}` which may include SVG-oriented attributes. Low risk, worth a glance.

- **L3 — Doc/cosmetic.** `appConfig.tsx` drops the blank line after the imports (line 3), and the new field comment `/**the list of client config values */` (line 36) is slightly off — it's a single config object, not a list, and is missing a leading space.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| C1 | Critical | `obsolete/` directory deleted (~287k lines) — appears unintentional |
| C2 | Critical | `branding` config never populated; feature never activates |
| M1 | Medium | 55 non-`en.json` locale files edited (convention + grammar risk) |
| M2 | Medium | today-screen branded icon drops caller `style`/sizing |
| L1–L3 | Low | Cross-package context, minor icon/doc nits |

The two Critical items should be resolved before merge: restore `obsolete/`, and wire `branding` through the SSR config + injected `__APP_CONFIG__` so the feature can actually run and be tested. Happy to help with any of these.

---

AmitPant9999:
@KallynGowdy @Blitzy branding is intentionally optional in the core application. The standalone core continues to use the existing Seed Bible branding through its current defaults, so branding being undefined does not affect its behavior. BrandingConfig is an extension point for downstream client applications

PR #1565, opened 2026-07-31
Fix stale touch samples causing swipe track to jump sideways

Blitzy:
## Summary

Fixed a bug where touch events from a previous swipe gesture could arrive over a second late and be applied to a new gesture, causing the swipe track to jump sideways unexpectedly. This happened most noticeably when picking up the page to swipe again immediately after changing chapters.

## Root Cause

Touch samples can be queued and delivered out of order. When the main thread is blocked (e.g., during a chapter change), a `touchmove` event generated during the *previous* gesture could sit in the queue and be delivered mid-*next* gesture, carrying a coordinate from where the finger was back then. Acting on this stale sample threw the track 100–200px sideways for a single frame.

## Key Changes

- **Added timestamp validation in `onTouchMove`**: Each touch sample is now checked against `swipeLastStamp`, which tracks the newest sample accepted for the current gesture. Samples older than this are silently dropped as stale.

- **Extracted magic numbers to named constants**: 
  - `SWIPE_LOCK_THRESHOLD_PX = 10` — how far a touch must travel before committing to horizontal/vertical
  - `SWIPE_ANIMATION_MS = 250` — track slide duration
  - `SWIPE_SETTLE_BUDGET_MS` — how long the track waits on a neighbouring panel before recentring (half the skeleton delay)
  - `PANEL_PCT` — exported so tests can use it

- **Refactored scroll handling**: Split the monolithic `attachScroller` effect into three focused effects:
  - One for restoring scroll position on navigation
  - One for scrolling to a linked verse
  - One for observing scroll and driving mobile chrome
  
  This prevents the scroll listener from being re-attached on every render, which was yanking partly-scrolled chapters back to their saved offset.

- **Improved swipe commit cleanup**: Now explicitly clears the settle timer on unmount so a pending navigation doesn't run for a component that no longer exists.

- **Converted ref callbacks to plain refs where possible**: `swipeViewportRef` and `swipeTrackRef` are now plain `RefObject`s instead of callbacks, since they only need `.current` filled in.

## Testing

Added integration tests covering:
- Stale touch sample rejection (replays a real device capture where a sample arrived 1.2s late)
- Scroll position restoration on navigation
- Scroll listener attachment not moving the reader on ordinary re-renders

https://claude.ai/code/session_01Ez9cyKrfsnpuiaGt8cxiMD


Blitzy:
> [!NOTE]

## Code Review — PR No. 1565: Fix stale touch samples causing swipe track to jump sideways

Related issue: [#1558](https://github.com/HelloAOLab/seed-bible/issues/1558) (chapter + highlights flickering on page swipe). The core flash fix landed in the already-merged PR No. 1559; this PR adds three further mobile-swipe fixes on top: dropping stale touch samples, stopping the scroll listener from being re-attached (and yanking the reader) on every render, and removing the small jump at the start of a swipe.

### Verification performed

- ✅ `pnpm vitest run TabSlotReader` — 22/22 passing, including the three new tests
- ✅ `pnpm check:ts` — 0 visible errors
- ✅ `eslint` on all three changed files — clean
- ✅ Confirmed no dangling references to the removed symbols (`attachScroller`, `swipeParked`, `currentChapterRef`, `slotScrollerRef`/`mobileScrollerRef`, `swipeViewportRefCallback`/`swipeTrackRefCallback`)
- ✅ CHANGELOG diff vs `develop` is clean — the two new entries stack cleanly above the already-merged No. 1559 flash entry, no duplication

Overall this is a well-structured, well-tested change. The effect split (restore-position / verse-scroll / observe-scroll) is a genuine improvement over the old monolithic `attachScroller`, the memoised ref callbacks correctly stop the re-attach churn, and the new tests replay a real device capture rather than a synthetic one. I found **no high-severity or blocking issues.** Two low/informational items below.

---

### 🟡 Low — Orphaned commit timer if two swipes commit within the 250 ms slide window

`packages/seed-bible/seed-bible/components/TabsLayout.tsx` (`commitSwipe`, ~L425; cleanup, ~L501)

`swipeCommitTimer` holds only the **most recent** timer id:

```ts
swipeCommitTimer.current = window.setTimeout(() => { ... }, SWIPE_ANIMATION_MS);
```

If a second swipe commits before the first's `SWIPE_ANIMATION_MS` (250 ms) timer fires, the second `commitSwipe` overwrites `swipeCommitTimer.current`, so the first timer id is lost. The effect-cleanup only clears whatever is currently stored:

```ts
window.clearTimeout(swipeCommitTimer.current); // clears timer2 only
```

The token bump correctly prevents the *first* timer from writing a transform to a track the effect no longer owns, but the timer body still calls `navigate()` unconditionally (by design — "the reader completed the swipe that asked for it"). So the orphaned first timer's `navigate()` runs even after the effect has been torn down.

**Impact:** minor and rare. `navigate()` mutates `readingState` signals, which outlive the component (they live on the tab), so nothing crashes — but closing a pane/tab immediately after a fast double-swipe could advance the reading position one extra chapter after the reader is gone. The inner settle `budget` timer from the orphaned commit is likewise never cleared (though it is token-guarded).

**Reproduction (hard to hit by hand):** on mobile, perform two complete swipe gestures (each: touchstart → horizontal move past ~90 px → touchend) within 250 ms of each other, then unmount the slot (switch/close the tab or pane) before the first slide finishes.

**Suggested fix:** clear the previous timer before overwriting, e.g. at the top of `commitSwipe`:

```ts
window.clearTimeout(swipeCommitTimer.current);
```

(or track pending timers in a `Set` and clear the set on cleanup). Since committing a new swipe already supersedes the previous one via the token, clearing the previous slide timer first is consistent with that intent.

---

### 🔵 Informational — Navigation now needs slightly more travel to trigger

`packages/seed-bible/seed-bible/components/TabsLayout.tsx` (`onTouchMove` ~L387, `onTouchEnd` ~L458)

The new "discount the lock threshold" logic means `swipeCurrentDx.current` is stored as `travel = dx - sign(dx) * SWIPE_LOCK_THRESHOLD_PX` (i.e. `dx` minus 10 px). `onTouchEnd` then compares that discounted value against the unchanged `threshold = 80`:

```ts
const dx = swipeCurrentDx.current;   // == travel, already 10px short of raw dx
const threshold = 80;
const swipedLeft = dx < -threshold;
```

- **Before:** a chapter change committed at ~80 px of raw finger travel.
- **Now:** it commits at ~90 px of raw travel (80 + the 10 px discounted at lock).

This is almost certainly an intentional side effect of removing the start-of-swipe jump, and 10 px is small — flagging only so it's a conscious choice rather than an accident. No change needed if the slightly stiffer swipe feels right on device.

---

### Notes (no action needed)

- The stale-sample fail-safe comment ("every comparison is then false and nothing is dropped") is correct: if a browser reports a constant `timeStamp`, `t < t` is always false, so no legitimate sample is discarded. Equal timestamps are also kept (`<`, not `<=`). Good defensive reasoning.
- Wrapping the translation-change `effect()` in `useEffect` (rather than a bare `effect()` in the render body) correctly fixes the previous subscribe-on-every-render leak.
- Grouping the scroller re-attach fix under issue #1558 is reasonable — it manifests during the same swipe-to-navigate flow.

---

Blitzy:
Fixed both issues brough up by the Code review.

PR #1569, opened 2026-07-31
Remove references to ao.bot in favor of auth.seedbible.org

KallynGowdy:
We've had issues with ao.bot domains getting blocked by network filters, so we're switching as much as possible over to auth.seedbible.org

### DIFF
--- a/packages/seed-bible/seed-bible/app/initPostHog.tsx
+++ b/packages/seed-bible/seed-bible/app/initPostHog.tsx
@@ -85,7 +85,7 @@ if (
 
   const seedBibleApiKey = "phc_rEUogfrnXkdTitOTrfWK2laEINF1QwNtGNQizzuMW0";
   posthog.init(seedBibleApiKey, {
-    api_host: "https://i.ao.bot",
+    api_host: "https://i.seedbible.org",
     ui_host: "https://us.posthog.com",
     defaults: "2026-05-30",
   });
--- a/packages/seed-bible/seed-bible/managers/OsManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/OsManager.tsx
@@ -64,7 +64,9 @@ const UNSAFE_HEADERS = new Set([
   "host",
 ]);
 
-export function CasualOSManager(endpoint: string = "https://auth.ao.bot") {
+export function CasualOSManager(
+  endpoint: string = "https://auth.seedbible.org"
+) {
   const rawClient = createRecordsClient(endpoint);
   const connectionId = uuid();
   let currentWakeLock: WakeLockSentinel | null = null;
@@ -127,7 +129,7 @@ export function CasualOSManager(endpoint: string = "https://auth.ao.bot") {
 
   function getInstClient(): InstRecordsClient {
     if (!instRecordsClient) {
-      const url = new URL("wss://auth.ao.bot");
+      const url = new URL("wss://auth.seedbible.org");
       const manager = new WebsocketManager(url);
       manager.init();
       const client = new WebsocketConnectionClient(manager.socket);
--- a/packages/seed-bible/seed-bible/managers/SearchManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/SearchManager.tsx
@@ -2,7 +2,7 @@ import * as Typesense from "typesense";
 import * as z from "zod/v4";
 import type { TranslationBook } from "./FreeUseBibleAPI";
 
-const TYPESENSE_NODE_URL = new URL("https://search.ao.bot");
+const TYPESENSE_NODE_URL = new URL("https://search.seedbible.org");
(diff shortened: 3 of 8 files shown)


Blitzy:
> [!NOTE]

## Code Review — Remove references to ao.bot in favor of auth.seedbible.org

### Summary

This is a tightly scoped find-and-replace that swaps a handful of `ao.bot` hostnames for their `seedbible.org` equivalents, plus the matching test updates. The diff is clean and the runtime-facing changes are correct:

- The default endpoint in `CasualOSManager` (`OsManager.tsx`) really does take effect — the production instantiation `SeedBibleStateManager.tsx:415` calls `CasualOSManager()` with no argument, so it inherits the new `https://auth.seedbible.org` default. The websocket URL (`wss://auth.seedbible.org`) and Typesense host (`search.seedbible.org`) were updated consistently, and the corresponding tests (`BibleSelector.test.tsx`, `SearchManager.test.ts`) were kept in sync.

No blocking defects. The findings below are things to confirm are intentional given the PR's stated goal — the body says you're "switching **as much as possible**," so several of these may already be out of scope on purpose. I'm listing them so the decision is explicit rather than accidental.

---

### Medium — Origin now mismatches the endpoint in the CLI upload scripts

In three build/publish scripts the request `Origin` was changed to `https://auth.seedbible.org`, but the endpoint the client actually talks to was left as `api.ao.bot`:

- `script/lib/records.ts:16` — `Origin: "https://auth.seedbible.org"`, while `records.ts:61` still does `createRecordsClient("https://api.ao.bot")`
- `script/lib/pattern.ts:22` — `Origin` updated, while `pattern.ts:34` / `pattern.ts:140` still use `createRecordsClient("https://api.ao.bot")`
- `script/lib/extension.ts:374-375` — `--endpoint "https://api.ao.bot"` kept, `--origin "https://auth.seedbible.org"` changed

The records server validates the `Origin` header against the allow-list configured for the target record. If `api.ao.bot` has not been configured to accept `https://auth.seedbible.org` as an allowed origin, these extension/pattern uploads will start being rejected where they previously succeeded — a regression that only shows up at publish time (CI or manual `pnpm pattern pack` / extension upload), not in the test suite.

Two ways to resolve, either is fine:
- If `api.ao.bot` and `auth.seedbible.org` front the same backend and the origin is already whitelisted, this is safe — worth a one-line confirmation in the PR.
- Otherwise, move the endpoints over too (`api.seedbible.org` or whatever the equivalent is) so origin and endpoint stay on the same domain family, matching what the runtime app now does (it points both at `auth.seedbible.org`).

**Reproduction:** run an extension or pattern upload against the current branch with valid credentials and watch for a 403/origin-rejection from the records server. (I couldn't reproduce this myself — it needs a session/record key I don't have — so please verify.)

---

### Medium — Remaining active `ao.bot` domains that the same network filters would still block

The motivation here is that `ao.bot` domains get caught by network filters. These active (non-`obsolete/`) references are still on `ao.bot`, so a filtered network would still hit them:

1. `packages/seed-bible/seed-bible/app/initPostHog.tsx:88` — `api_host: "https://i.ao.bot"` (PostHog analytics ingestion). Blocked → analytics silently stops reporting for affected users.
2. `vite.config.ts:191-250` — PWA manifest icons and screenshots served from `https://favicon.ao.bot/...`. Blocked → broken install icon / store screenshots on filtered networks.
3. `packages/seed-bible/seed-bible/components/PortalComponent/PortalComponent.tsx:26` — portal iframe built from `new URL("https://ao.bot/")`. Blocked → grid/map portals fail to load. (This one may be genuinely unavoidable since the `.aux` patterns run inside the CasualOS `ao.bot` origin — but worth calling out.)
4. `server/index.ts:366` — legacy deep-link handler `302`-redirects to `https://ao.bot/${search}`.

If any of these are deliberately staying on `ao.bot` for now, No. 1 and No. 2 are the ones I'd most want an explicit decision on, since they directly affect the same end users this PR is trying to unblock.

---

### Low / Nits

- `script/lib/records.ts:38` (doc comment "If they are not logged into ao.bot…") and `records.ts:46` (`config.get("https://api.ao.bot:sessionKey")`) still reference `api.ao.bot`. The config key is internally consistent with the unchanged `api.ao.bot` endpoint, so it's correct as-is — just flagging that the comment wording will read as stale if the endpoints later move.
- `CLAUDE.md` still describes portals as loaded in "cross-origin `ao.bot` iframes." Accurate today (see No. 3 above); revisit if the portal origin ever moves.

---

### Verdict

Looks good to merge on its own terms — the changes it does make are correct and consistently tested. Before merging I'd want a quick confirmation on the **Medium — Origin/endpoint mismatch** item, since that's the one place where behavior could newly break at publish time. The remaining-domains list is informational: decide which are in scope for this pass and which are follow-ups.

---

PR #1570, opened 2026-07-31
Improvements from PageSpeed insights

KallynGowdy:


Blitzy:
> [!NOTE]

## Code review — Improvements from PageSpeed insights

Nice, focused set of changes. Removing render-blocking font requests and adding gzip are both real PageSpeed wins, and the `proxyAsset` gzip path is done carefully (it correctly strips upstream `content-encoding`/`content-length`/`transfer-encoding` before re-framing, and skips 206/304 responses).

There is one issue I'd consider blocking, plus a few smaller ones. Details below, most severe first.

---

### 🔴 High — Verse, book-title, and heading fonts are no longer loaded, but the app still asks for them

**What happened.** The single `<link>` you removed from `app/main.tsx` loaded **three** families in one stylesheet:

```
family=DM+Sans...&family=Newsreader...&family=Plus+Jakarta+Sans...&display=swap
```

The CSS/theme changes in this PR only swap **Satoshi** and **DM Sans** over to `system-ui`. But **Newsreader** and **Plus Jakarta Sans** are still the configured typefaces across the app, and after this PR nothing downloads them anymore (the only remaining Google Fonts request is Material Symbols Outlined).

Places that still reference the now-unloaded fonts:

- `app/styles/base.css` — `--sb-verse-font-family`, `--text-verse-font`, `--sb-chapter-heading-font-family`, `--sb-book-title-font-family`, `--sb-hebrew-subtitle-font-family` (lines ~50–125)
- `managers/ThemeManager.tsx` — `verseFontFamily`, `chapterHeadingFontFamily`, `bookTitleFontFamily`, `hebrewSubtitleFontFamily` in **both** `LIGHT_THEME` and `DARK_THEME` (lines ~505–520 and ~647–662)
- `managers/SettingsManager.tsx` — `TEXT_FONT_OPTIONS` still offers `Newsreader` and `Plus Jakarta Sans` as pickable fonts (lines 161–162), and the default text presets use them (lines 204–226)

The most visible consequence: **verse text** (the primary reading content) defaults to `Plus Jakarta Sans` via `ThemeManager`'s `verseFontFamily`, which — per `CLAUDE.md` — wins over `main.css`. So verses, book titles, and chapter headings will silently fall back to the generic `serif`/`sans-serif` system faces. And if a user picks "Newsreader" or "Plus Jakarta Sans" from the font settings, the selection does nothing because the font isn't loaded.

Because the PR deliberately *kept* those two families in the theme/CSS (it only replaced Satoshi/DM Sans), this looks like an accidental over-deletion of the stylesheet rather than an intended move to full system fonts.

**Reproduction:**
1. Check out `feature/page-speed`, run `pnpm dev`.
2. Open the reader with DevTools → Network → filter "Font".
3. Observe that only Material Symbols is fetched — no request for Plus Jakarta Sans or Newsreader.
4. Verse text and book titles render in the default system serif/sans, not the intended typefaces.
5. Settings → text font → pick "Plus Jakarta Sans" or "Newsreader": the on-screen font does not change.

**Suggested fix (keeps intended typography — most likely what's wanted):** re-add a Google Fonts stylesheet that loads only the families still in use, dropping DM Sans and Satoshi:

```html
<link
  href="``https://fonts.googleapis.com/css2?family=Newsreader:ital,opsz,wght@0,6..72,200..800;1,6..72,200..800&family=Plus+Jakarta+Sans:ital,wght@0,200..800;1,200..800&display=swap"``
  rel="stylesheet"
/>
```

(Or self-host + `@font-face` them for an even better PageSpeed result, since that removes the third-party origin round-trip entirely.)

**Alternative fix (if the intent really is full system fonts):** replace the remaining `Newsreader`/`Plus Jakarta Sans` references in `base.css`, `ThemeManager.tsx`, and `SettingsManager.tsx` (including removing/relabeling the two font-picker options) so the app stops advertising fonts it can't render.

---

### 🟡 Low — Streaming responses use `.pipe()` without error handling

In both `sendHtml` and `proxyAsset` the new streaming paths chain raw `.pipe()`:

```ts
Readable.from(body).pipe(createGzip()).pipe(res);
// ...
upstreamBody.pipe(createGzip()).pipe(res);
```

`.pipe()` does not forward errors or destroy upstream stages on failure. If the client disconnects mid-response, or the gzip transform errors, the source/transform streams aren't torn down (socket/memory leak), and an unhandled `'error'` on a stream can take the process down. The old SSR path was a single buffered `res.end(html)`, so this is newly introduced risk.

**Suggested fix:** use `stream.pipeline`, which destroys the whole chain and surfaces errors in one place:

```ts
import { pipeline } from "node:stream";
// ...
pipeline(Readable.from(body), createGzip(), res, (err) => {
  if (err) console.error("gzip response failed:", err);
});
```

---

### 🔵 Info — `proxyAsset` re-compression relies on fetch fully decoding every upstream encoding

Re-gzipping upstream bodies is correct **only** because Node's `fetch` transparently decompresses the response (as the docstring notes). If an upstream ever responds with a content-coding `undici` leaves encoded, the body would be gzip-wrapped a second time and the client would get corrupted bytes. This is very unlikely with the current asset host, but it's an implicit assumption worth a comment or a guard. No action needed if the asset host only ever serves plain gzip/br/identity.

---

### 🔵 Info — Minor notes

- **No CHANGELOG entry.** The repo keeps a `## TBD` section in `CHANGELOG.md` and recent PRs update it; this one doesn't. Worth a line under 🔧 Changed / 🗑️ Removed for the font and gzip changes.
- **`i.ao.bot` preconnect.** `index.html` preconnects to `https://i.ao.bot`, while PR No. 1569 is removing `ao.bot` references in favor of `auth.seedbible.org`. Worth confirming this origin is still the one serving images so the two PRs don't land contradicting each other.
- **Preconnect count.** Five preconnect origins is on the high side (each opens a socket eagerly); if `assets.seedbible.org` assets are actually served same-origin via the SSR proxy, that preconnect may go unused. Low impact, just worth a sanity check against what the reader actually requests on first paint.

---

Happy to help with the font fix once the intended direction (keep Newsreader/Plus Jakarta Sans vs. go full system) is settled — that choice drives which of the two fixes above applies.

---

PR #1573, opened 2026-08-01
fix issue-1566

### LINKED ISSUE #1566: X button on translation selection box has wrong cursor
Currently, the mouse cursor for the "X" button on the translation selection box appears like if you were typing in an input box, not pressing a button.

Image

AmitPant9999:
fixes: #1566

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.css
+++ b/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.css
@@ -1053,6 +1053,9 @@
   padding: 0.125rem;
   cursor: pointer;
 }
+.close-icon {
+  cursor: pointer;
+}
 
 .emptyCircle {
   display: inline-block;
--- a/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
@@ -1476,7 +1476,7 @@ const TranslationModal = (props: {
             </span>
             {!isMobile.value && (
               <span
-                class="material-symbols-outlined"
+                class="close-icon material-symbols-outlined "
                 onClick={() => {
                   selectingTranslation.value = false;
                   showTranslationSettings.value = false;
--- a/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
@@ -1433,7 +1433,7 @@ const TranslationModal = (props: {
           >
             {isMobile.value && (
               <span
-                class="material-symbols-outlined"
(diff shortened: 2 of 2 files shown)


Blitzy:
> [!NOTE]

## Code Review — PR #1573 (fixes #1566)

### Summary

The change does what the issue asks: the "X" (close) button on the translation selector used the default text/I‑beam cursor because it's a Material Symbols glyph (`close`) rendered as text. Adding a `.close-icon { cursor: pointer; }` rule and applying that class to the desktop close button gives it the expected pointer cursor. The diagnosis is correct, the fix is scoped (the new `.close-icon` class isn't used anywhere else, so no collateral cursor changes), and it's consistent with the neighboring `.filters-icon` button which already sets `cursor: pointer`. Nice and minimal.

No blocking problems. One low‑severity gap and one nit below.

---

### Low — The mobile close button still has the wrong cursor

`isMobile` is viewport‑width driven (`SeedBibleStateManager.tsx:587`: `computed(() => viewportWidth.value <= MOBILE_BREAKPOINT)`), so the translation modal renders **two** close buttons depending on window width — a mobile one and a desktop one. This PR only fixes the desktop branch.

The mobile branch still renders a bare glyph with an `onClick` and no pointer cursor:

`packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx:1434`
```tsx
{isMobile.value && (
  <span
    class="material-symbols-outlined"   // <-- no cursor: pointer
    onClick={() => { ... setOpen(false); }}
  >
    close
  </span>
)}
```

Because `isMobile` keys off viewport width (not a touch/pointer capability check), a desktop user with a mouse who narrows the browser below the mobile breakpoint will still see the same wrong (text) cursor on the X — the exact symptom this PR set out to fix. On true touch devices there's no hover cursor, so it's cosmetic there.

**Reproduce:**
1. Open the app on desktop with a mouse.
2. Narrow the browser window until the layout switches to its mobile breakpoint.
3. Open the translation selector and hover the "X" — the cursor is still the I‑beam.

**Suggested fix:** apply the same class to the mobile branch so both paths match:
```tsx
class="close-icon material-symbols-outlined"
```

### Nit — Trailing space (and slight redundancy) in the class string

`packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx:1479`
```tsx
class="close-icon material-symbols-outlined "
```
The trailing space is harmless but untidy. Prettier (per the repo's formatting conventions) won't touch string contents, so it'll persist — worth trimming to `"close-icon material-symbols-outlined"` while you're in there.

---

### Verdict

The fix correctly resolves the reported desktop cursor bug and is safe to merge as‑is. Folding the mobile branch into the same `.close-icon` class (No. 1 above) would make the fix complete for narrow‑viewport mouse users and avoid the same bug resurfacing; the trailing‑space nit is purely cosmetic.

---

KallynGowdy (approved):

PR #1575, opened 2026-08-01
issue_1574

### LINKED ISSUE #1574: Last verses are  obscured when Verse Toolbar is open
### what is happening
When the Verse Toolbar is open in mobile, the user cannot scroll the last verse completely into view because it is covered by the toolbar. This makes the final verse difficult or impossible to read and interact with.

Image
### what expected
The last verse should be fully visible and accessible when the Verse Toolbar is open.

AmitPant9999:
fixes: #1574

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.css
@@ -684,6 +684,9 @@
   opacity: 0.35;
   pointer-events: none;
 }
+.sb-chapter-content-margin {
+  margin-bottom: 4.5rem;
+}
 
 /* ─── Chapter loading placeholder ────────────────────────────────────────── */
 /* Stands in for the verses while the chapter the reader has navigated to is
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
@@ -1043,6 +1043,7 @@ interface ChapterContentProps {
    * the chapter they navigated to arrives.
    */
   isStale?: boolean;
+  isMobile: boolean;
 }
 
 function ChapterContent(props: ChapterContentProps) {
@@ -1058,6 +1059,7 @@ function ChapterContent(props: ChapterContentProps) {
     selectVersesFromTextSelection,
     justConvertedSelectionRef,
     scriptureElements,
+    isMobile,
   } = props;
 
   const contentRef = useRef<HTMLDivElement>(null);
@@ -1307,7 +1309,7 @@ function ChapterContent(props: ChapterContentProps) {
       ref={contentRef}
       className={`sb-chapter-content${
         props.isStale ? " sb-chapter-content-stale" : ""
-      } ${containerClasses}`}
+      } ${containerClasses} ${isMobile && selectedVerses?.value?.length > 0 ? "sb-chapter-content-margin" : ""}`}
       onPointerDown={() => {
         justConvertedSelectionRef.current = false;
       }}
@@ -1643,6 +1645,7 @@ export function BibleReader(props: BibleReaderProps) {
               selectVerse={selectVerse}
               selectFootnote={selectFootnote}
               scriptureElements={scriptureElements}
+              isMobile={isMobile}
             />
           </Suspense>
         ))}


Blitzy:
> [!NOTE]

## Code review — PR #1575 (fixes #1574: last verses obscured by the verse toolbar)

**Short version:** the approach is right — on mobile the verse toolbar is docked to the bottom, so reserving space below the chapter content is the correct way to let the last verse scroll clear of it. The concern is that the reserved space is a fixed `4.5rem` (72px), and the actual mobile toolbar is a bottom sheet whose height is both larger than that and variable. So the original bug can still reproduce. Details and suggested fixes below, ordered by severity.

---

### Medium — The fixed `4.5rem` margin is smaller than the mobile verse toolbar, so the last verse can still be obscured

The change reserves a constant `margin-bottom: 4.5rem` (`BibleReader.css:687`), but the mobile verse toolbar (`.sb-verse-toolbar-mobile`) is a fixed, bottom-docked sheet whose rendered height is bigger than 72px and grows depending on state:

- **Collapsed state** stacks a grab handle, a reference label, and a row of action **cards that are `4rem` tall each** (`.sb-verse-toolbar-cards .sb-verse-toolbar-action { height: 4rem }`), plus the sheet's own padding `0.5rem 1rem calc(0.875rem + env(safe-area-inset-bottom))`. Added up that is already roughly `6–7rem`, i.e. taller than the `4.5rem` reserved. On notched phones it is taller still, because the fix does not include `env(safe-area-inset-bottom)` while the toolbar does.
- **Expanded state** — tapping **More** switches the sheet to the multi-row card grid, which can be `12rem`+ tall. `4.5rem` leaves the last verses well behind the sheet.

**Net effect:** the exact symptom described in #1574 (last verse covered by the toolbar) still happens once the sheet is more than one short row tall — most visibly after tapping More.

**Reproduction (human):**
1. Open the site on a mobile viewport (≤ 480px, e.g. device toolbar in dev tools), or the deployed branch build linked above.
2. Navigate to the end of a chapter and select the **last** verse so the bottom verse toolbar appears.
3. Tap **More** to expand the action cards.
4. Try to scroll the last verse fully into view — it stays partly hidden behind the sheet. On a notched device it is partly hidden even in the collapsed state.

**Suggested fix — make the reserved space track the real toolbar height instead of a magic number.** A few options, roughly in order of robustness:

- Measure the toolbar element and publish its height as a CSS variable on the reader root (e.g. `--sb-verse-toolbar-height`), then use `margin-bottom: var(--sb-verse-toolbar-height)`. This automatically follows the collapsed/expanded states and the safe-area inset.
- If you want to keep it CSS-only, at minimum fold the safe area in and raise the base so the collapsed sheet is cleared, e.g. `margin-bottom: calc(7.5rem + env(safe-area-inset-bottom))` — but this still won't cover the expanded grid, so it's a partial fix.
- Alternatively (or additionally), reserve the space on the scroll side with `scroll-padding-bottom` on the scroll container (and/or `scroll-margin-bottom` on the verse elements) so any scroll-into-view already accounts for the sheet. This pairs well with the measured-height approach.

---

### Low — Redundant optional chaining

`BibleReader.tsx:1312`:

```tsx
${isMobile && selectedVerses?.value?.length > 0 ? "sb-chapter-content-margin" : ""}
```

`selectedVerses` is a required prop typed `Signal<BibleSelectedVerse[]>`, and `.value` is always an array, so neither `?.` can short-circuit. `selectedVerses.value.length > 0` says the same thing with less noise. Not a bug — type-check passes — just dead defensiveness.

### Low — `4.5rem` has no explanation, and `CLAUDE.md` calls out this exact sync hazard

The new rule hard-codes `4.5rem` with no comment tying it to the toolbar it is meant to clear. The repo already keeps toolbar dimensions in `--sb-*` variables, and `CLAUDE.md` warns that CSS values here often need to stay in sync with a second source of truth. A named variable or a one-line comment ("must clear `.sb-verse-toolbar-mobile`") would save the next maintainer from guessing. Deriving the value from the toolbar (Medium item above) removes the magic number entirely.

### Low — The margin is applied to every mounted chapter pane, not just the visible one

`ChapterContent` is rendered once per chapter in the pager's `.map` (`BibleReader.tsx:1635`+), and `selectedVerses` is a single shared signal. So when a selection exists, **all** mounted panes receive `sb-chapter-content-margin` at once, not only the on-screen chapter. Only one pane is visible so this is probably harmless, but it does change the scroll height of the off-screen panes too — worth a glance if you see any scroll-position oddities on the neighboring chapters.

### Nit — `.sb-chapter-content-margin` isn't scoped to mobile in CSS

The rule lives outside any media query and is kept mobile-only purely by the `isMobile` guard in JSX. That's fine as written, just note the class would apply on desktop if it were ever reused elsewhere. The `isMobile` guard is otherwise correct — the desktop verse toolbar floats next to the selection rather than docking to the bottom, so it needs no reserved space.

---

**Scope / correctness of what's here:** the two changed files are consistent — the `isMobile` prop is threaded from the one `ChapterContent` call site through the interface to the class toggle, the type check is clean, and the build passed on the head commit. The remaining work is entirely about the sizing of the reserved space, not the plumbing.

---

KallynGowdy:
@AmitPant9999 Can you fix the issues from the review?

AmitPant9999:
@KallynGowdy   kushagra has another pr for this task and i think has a better solution

Blitzy:
Will be taking https://github.com/HelloAOLab/seed-bible/pull/1576 instead which includes a dynamic fix for this problem.

PR #1576, opened 2026-08-02
Feature/1574

### LINKED ISSUE #1572: Mobile fullscreen pane is inconsistently showing with and without the mobile toolbar
# What is happening

This is easiest to see with the Locations extension. Select a verse with a location name and open view the location on the map. Notice that the mobile toolbar is often just missing.

## What I expected

I expected the main mobile toolbar to be visible with a full screen pane open. Other fullscreen panes like the Discover pane or the Twitch Host pane seem to do this properly.

## Note
Need to understand if this is a systemic issue or just one with the location extension?

### LINKED ISSUE #1574: Last verses are  obscured when Verse Toolbar is open

Kabi4:
Fixes #1574 

Fixes #1572


Blitzy:
> [!NOTE]

## Code Review — Dynamic bottom inset for reader content (`feature/1574`)

**Fixes:** #1574 — *"Last verses are obscured when Verse Toolbar is open"* (mobile).

### Overview

The change replaces the hard-coded `padding-bottom: 8.125rem` on the scrollable reader content with a new CSS variable, `--sb-reader-bottom-inset`, that is measured at runtime from whatever bottom chrome is currently open — the mobile verse sheet, the reader toolbar, and/or the floating chapter-nav bar. A `ResizeObserver` (height changes) plus a `MutationObserver` (chrome appearing/disappearing) keep the value in sync, with a `8.125rem` CSS fallback in `base.css`.

**Verdict: the approach is sound and the fix is correct for the reported bug.** The mobile verse toolbar (`.sb-verse-toolbar-mobile`) is measured first, its `offsetHeight` already includes the `env(safe-area-inset-bottom)` padding, and feeding that into the content's `padding-bottom` is exactly what lets the last verse scroll clear of the sheet. The observer wiring is careful — the `MutationObserver` deliberately covers the case where the floating nav appears based on signals that aren't in the effect's dependency array (audio tool / prev-next / selector becoming available), which is a real gap the dependency array alone wouldn't catch. Nice attention to detail.

No blocking issues found. The notes below are all **Low severity** — polish and things worth eyeballing on the alpha deploy, not correctness defects.

---

### Findings

#### No. 1 — Low — `MutationObserver` path measures synchronously; `ResizeObserver` path is rAF-debounced

`BibleReaderToolbar.tsx` (the `measure`/`reobserve` block, ~L840–898). The `ResizeObserver` callback debounces through `requestAnimationFrame`, but `reobserve()` — invoked directly by the `MutationObserver` — calls `measure()` synchronously. `measure()` reads `offsetHeight` and `getComputedStyle().bottom` (layout-forcing reads) and then writes a CSS custom property. On any direct-child mutation of the wrap this forces a synchronous reflow.

In practice wrap child mutations are infrequent (nav appears/disappears, toolbar swaps), so the cost is negligible — hence Low. For consistency you could route the `reobserve` measurement through the same rAF the `ResizeObserver` uses:

```ts
const scheduleMeasure = () => {
  cancelAnimationFrame(frame);
  frame = requestAnimationFrame(measure);
};
const observer = new ResizeObserver(scheduleMeasure);
// ...at the end of reobserve():
scheduleMeasure(); // instead of measure();
```

#### No. 2 — Low — Stacked bottom padding when a translation license notice is present

`BibleReader.css` L338–340 and L413. The license notice (`.sb-translation-license-notice`) is a **sibling that follows** `.sb-chapter-content` inside `.sb-reader-swipe-panel`. When it's present, both the chapter content (`padding-bottom: var(--sb-reader-bottom-inset)`) **and** the panel (`:has(.sb-translation-license-notice) { padding-bottom: var(--sb-reader-bottom-inset) }`) reserve a full inset, so the layout becomes: last verse → *[inset gap]* → license notice → *[inset gap]* → bottom.

This double-gap largely pre-existed (the old values were `8.125rem` and `7.5rem + safe-area`), so it's not a regression introduced here. But the gap **before** the notice now scales with the measured inset, which can be noticeably larger than 8.125rem when a tall verse sheet (e.g. the expanded "More" state) is open. Worth a quick visual check on a translation that shows a license line (e.g. read to the end of a chapter with the verse sheet expanded) to confirm the whitespace doesn't look broken. If it does, the cleaner model is to let only the trailing element reserve the inset rather than both.

#### No. 3 — Low — Silent no-op when `ResizeObserver` is unavailable

`BibleReaderToolbar.tsx` — `if (typeof ResizeObserver === "undefined") return;`. On SSR this is correct (and desired). On a client with no `ResizeObserver`, the effect bails entirely and the static `8.125rem` fallback stands. Since 8.125rem (130px) can be shorter than an expanded mobile verse sheet, the original bug would technically remain there. This is acceptable — every browser that renders the mobile reader supports `ResizeObserver` — so I'd just flag it as a known, tolerable degradation rather than something to fix.

#### No. 4 — Low — Cross-component dependency on a runtime-set variable

`BelowReaderToolbar.css` L10 now consumes `var(--sb-reader-bottom-inset)`, but that variable is only written to `document.documentElement` while `BibleReaderToolbar` is mounted (and is removed on its unmount). In `TabsLayout` the two render together, so this is fine and the `8.125rem` fallback covers any standalone case. Calling it out only so it's a conscious coupling: `--sb-reader-bottom-inset` is now effectively "owned" by `BibleReaderToolbar`'s lifecycle, and any future surface that reuses `.sb-below-reader-toolbar` without the reader toolbar mounted will get the static fallback, not a measured value.

---

### Suggested verification (manual — this is measurement-driven DOM behavior that's hard to unit-test)

Using the alpha deploy on this PR (**https://alpha.seedbible.org/b/feature_1574**), in mobile viewport / device emulation:

1. **Primary repro (the bug):** open a chapter, scroll to the last verse, tap it to open the verse toolbar. The last verse should now be fully readable above the sheet. ✅ expected fixed.
2. **Expanded sheet:** with a verse selected, tap **More** to expand the sheet — the last verse should stay clear as the sheet grows (this is the `ResizeObserver` + `isVerseSheetExpanded` dependency at work).
3. **Floating nav appearing:** open a chapter that has audio (audio-reader extension) so the floating nav pill appears above the bottom bar — confirm the end-of-chapter content still clears both bars (this exercises the `MutationObserver` path).
4. **License notice (finding No. 2):** read to the end of a chapter whose translation shows a license/attribution line, with the verse sheet expanded, and confirm the spacing around that line looks reasonable.
5. **iOS safe area:** on an iPhone-class device (home indicator), confirm nothing regresses at the very bottom in the normal (no selection) reading state.

Consider adding a lightweight note in the PR description about which of these you've eyeballed, since there's no automated coverage for it.

---

Blitzy (approved):
This is a pretty elegant fix for this problem, and is a dynamic solution for if/when the sizes of these toolbars change in the future. I added a couple of minor optimizations.

PR #1577, opened 2026-08-02
Feature/1549

### LINKED ISSUE #1549: `parseVerseReferences()` doesn't detect 1 or 2 Corinthians references
For some reason, the `parseVerseReferences()` and `parseVerseReference()` functions in BibleDataManager don't detect references to 1 or 2 Corinthians.

We should fix that issue and also ensure that we have a comprehensive test suite checking all 66 books, both by book ID and also by english book name.

Kabi4:
Fixes #1549


Blitzy:
> [!NOTE]

## Code review — PR No. 1577 (`Feature/1549`)

Reviewed against issue No. 1549, which asks to (a) fix `parseVerseReferences()`/`parseVerseReference()` not detecting **1/2 Corinthians**, and (b) add a comprehensive test suite covering all 66 books by ID and by English name.

**What this PR does well:** The root cause of the Corinthians miss is correctly diagnosed. With the global (`/g`) regex, a false candidate like `See 1` advanced `lastIndex` **past** the leading `1`, so `1 Corinthians` was never re-examined. Resetting `lastIndex` to `match.index + 1` on every rejected candidate is the right fix, and the `songofsongs` alias plus the 66-book `it.each` matrix are good additions. All 460 tests pass locally (`pnpm vitest run test/unit/seed-bible/managers/BibleDataManager.test.ts`).

Below are the issues I found, most severe first.

---

### 🟠 Medium — New `of …` branch creates false verse references for real "Song of …" phrases

The added book pattern `(?:\s+of\s+[A-Za-z][A-Za-z0-9]*)?` captures `Song of <word>` as a single book string. That string is then handed to `getBookId()`, whose **`startsWith` fallback** (BibleDataManager.tsx:557-561) matches any input beginning with `song` → `SNG`. So phrases that are *not* Song of Solomon get silently turned into `SNG` references.

This is a **regression**: before this PR the regex stopped `Song` at the space, these phrases matched nothing. Now they resolve to a wrong, clickable reference — and this is a Bible-study app where "the Song of Moses" (Deut 32 / Exod 15) and "the Song of Mary" (the Magnificat, Luke 1:46) are common phrases.

**Reproduction** (verified locally):

```ts
parseVerseReferences("the Song of Moses 2:1")
// => [{ ref: { book: "SNG", chapter: 2, verse: 1 }, start: 4, end: 21 }]   ❌ should be []

parseVerseReferences("See Song of Mary 1:46")
// => [{ ref: { book: "SNG", chapter: 1, verse: 46 }, start: 4, end: 21 }]   ❌ should be []
```

Mechanism confirmed at the `getBookId` layer:

```ts
getBookId("song of moses")  // => "SNG"   (via the startsWith("song") fallback)
```

**Suggested fix.** The safest change is to stop the loose `startsWith` fallback from firing on multi-word ("… of …") inputs, while keeping it for the single-token names that legitimately rely on it (e.g. `Leviticus` → `lev`). All canonical multi-word names (`Song of Solomon`, `1 Corinthians`, `1 Kings`, …) already have **exact** entries in `BOOK_ID_MAP`, so restricting the fallback to space-free inputs keeps every current positive working:

```ts
export function getBookId(book: string): BookId | null {
  const hadSpaces = /\s/.test(book.trim());
  const bookLower = book.toLowerCase().replaceAll(/\s+/g, "");

  const id = BOOK_ID_MAP.get(bookLower);
  if (id) return id;

  // Only use the loose prefix fallback for single-token inputs. A multi-word
  // phrase like "Song of Moses" must match a book name exactly, or not at all.
  if (!hadSpaces) {
    for (const [key, mappedId] of BOOK_ID_MAP) {
      if (bookLower.startsWith(key)) return mappedId;
    }
  }
  return null;
}
```

Please re-run the full suite after this change (it should stay green) and add regression cases for `Song of Moses 2:1` and `Song of Mary 1:46` asserting `[]`.

---

### 🟡 Low — `parseVerseReference()` (singular) is named in the issue but has no 66-book coverage

Issue No. 1549 asks to fix and comprehensively test **both** functions. The new matrix exercises `parseVerseReferences()` (plural) and `getBookId()`, but the singular `parseVerseReference()` isn't in the 66-book coverage.

Good news: it already works today — I verified `parseVerseReference("1 Corinthians 13:4")`, `"2 Corinthians 5:17"`, `"Song of Solomon 1:1"` etc. all resolve correctly (its greedy `[0-9A-Za-z\s]+` book group backtracks into the number, so it never had the plural's `lastIndex` bug). But to satisfy the issue's "comprehensive test suite … both by book ID and by English name" for both functions, consider extending the `it.each(PROTESTANT_CANON)` blocks to also assert on `parseVerseReference()`.

---

### 🟡 Low — The `of` connector is case-sensitive, so title-cased input is missed

The literal `of` in the regex has no `i` flag, so title-case input slips through:

```ts
parseVerseReferences("Song Of Solomon 2:1")   // => []   (capital "O" not matched)
parseVerseReferences("Song of Solomon 2:1")   // => SNG 2:1  (lowercase works)
```

Minor, but "Song Of Solomon" is a natural way to write it. Using `\s+[Oo][Ff]\s+` (or `[Ff]?…` — keep it scoped, don't add a blanket `i` flag) would cover it. Worth deciding alongside the fix above.

---

### ⚪ Nit — `retryFromNextChar()` makes worst-case parsing O(n²)

Resetting `lastIndex` to `match.index + 1` on every rejection means a string full of near-miss candidates (e.g. many `"See 1 "` fragments) is re-scanned character-by-character. It's correct and fast enough for normal UI text — I confirmed a 2 KB pathological input parses instantly — so this is informational only. A short comment noting the trade-off, or capping the input length parsed, would be enough if you want to be defensive.

---

**Summary:** The core fix is correct and the test matrix is a solid addition. The one thing I'd want addressed before merge is the Medium false-positive on "Song of …" phrases, since it turns valid prose into wrong references. The Low items are polish and can be folded into the same change.

---

KallynGowdy (dismissed):

KallynGowdy (approved):

PR #1578, opened 2026-08-02
Feature/1528

### LINKED ISSUE #1528: Auto-switching translations when the user changes language causes them to be reset to Genesis 1
It should keep them at the same book, chapter and verse if possible.

Kabi4:
Fixes: #1528


Blitzy:
> [!NOTE]

## Code Review — PR #1578 (Fixes #1528)

### Summary

The change does what issue #1528 asks for: when a UI language switch triggers an automatic Bible-translation switch, the reader now stays on the **same book and chapter** instead of being reset to Genesis 1 — as long as the new translation contains that book. If the book is missing, it correctly falls back to the translation's first book via the existing `selectTranslation` path.

The implementation is small and reads cleanly, and the test coverage is genuinely good — the three added cases exercise the happy path (same book kept), the missing-book fallback, and the nearest-translation confirmation flow (Gujarati → Hindi). All 52 tests in the file pass locally. The out-of-range-chapter case is also safe because `selectTranslationAndChapter` runs the chapter through `resolveChapterInBook`, which falls back to the book's first chapter rather than requesting a chapter the new translation doesn't have.

Findings below, most to least important.

---

### Medium — Failed book-catalog fetch now breaks the language switch instead of degrading gracefully

**File:** `packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx:1619`

The newly added `const books = await data.getTranslationBooks(translation.id);` sits in the applicator body **outside any try/catch**, and **before** the final `await selector.selectTranslation(translation.id);` on line 1633.

`getTranslationBooks` returns a cached catalog when it has one, but on a cache miss it awaits a network request (`api.getTranslationBooks`) that can reject. The target translation's book list is often *not* cached yet at this point (the user is switching to a translation they may not have opened), so this is a live failure path, not a theoretical one.

Before this PR, that same fetch happened **inside** `selectTranslation`, whose `try/catch` swallowed the error, set `error.value`, and let the flow continue — and `selector.selectTranslation(...)` still ran. After this PR, a fetch failure throws straight out of the applicator, which means:

- The rejection escapes as an unhandled promise rejection. The settings UI calls this via `void setLanguage(languageCode)` (`SettingsPage.tsx:2261`), so there is no `.catch` to absorb it.
- `selector.selectTranslation(translation.id)` on line 1633 never runs, leaving the Bible selector out of sync with the UI language that `changeLanguage` already applied.
- No error is surfaced in the reading pane, so the user sees a half-applied language change with no feedback.

So the failure path is strictly worse than before this change.

**Reproduction (conceptual — needs a fetch fault):** switch the UI language to one whose Bible translation's book catalog isn't cached, while that catalog request fails (offline / 5xx). Previously: reading pane shows an error and the selector updates. Now: unhandled rejection, selector unchanged, no visible error.

**Suggested fix:** guard the added fetch and degrade to the old behavior on failure, e.g.:

```ts
const tab = selectedTab.value;
if (tab) {
  const currentBookId = tab.readingState.bookId.peek();
  const currentChapterNumber = tab.readingState.chapterNumber.peek();
  let matchingBook: TranslationBook | false | undefined = false;
  try {
    const books = await data.getTranslationBooks(translation.id);
    matchingBook =
      currentBookId && books.books.find((book) => book.id === currentBookId);
  } catch {
    // fall through to selectTranslation, which handles its own errors
  }
  if (matchingBook && currentChapterNumber != null) {
    await tab.readingState.selectTranslationAndChapter(
      translation.id,
      matchingBook.id,
      currentChapterNumber
    );
  } else {
    await tab.readingState.selectTranslation(translation.id);
  }
}
```

---

### Low — Verse position isn't carried over (partial vs. the issue's acceptance criteria)

**File:** `packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx:1624`

Issue #1528 asks to keep the reader "at the same book, chapter **and verse** if possible." This PR preserves book and chapter only. `selectTranslationAndChapter` already accepts a `scrollToVerse` option (see its `options?.scrollToVerse` handling in `BibleReadingManager.tsx`), so the plumbing to carry a verse anchor exists but isn't used here.

This is a genuine gap against the stated requirement, though a lower-severity one: chapter-level restoration already eliminates the "reset to Genesis 1" complaint that motivated the issue, and exact verse mapping across translations is imperfect because versification differs between them. Worth either wiring through the current verse anchor or explicitly noting on the issue that verse-level restoration is deferred, so it isn't silently dropped.

---

### Nit — Redundant catalog lookup

**File:** `packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx:1619-1627`

The applicator calls `data.getTranslationBooks(translation.id)` to find the matching book, and then `selectTranslationAndChapter` immediately calls `getTranslationBooks` again for the same translation. The second call is a cache hit, so the cost is negligible — noting only for completeness. Also, `matchingBook.id` on line 1626 is by construction equal to `currentBookId`; passing either is fine.

---

### Verdict

The core fix is correct, minimally scoped, and well tested. I'd like to see the Medium error-handling item addressed before merge (it's a small change), and a decision made on whether verse-level restoration is in scope for this issue or explicitly deferred. The nit is optional.

---

Kabi4:
Current Logiccally is it impossible to have verese select and auto switch langauge but covered it too.

Blitzy (approved):

PR #1583, opened 2026-08-03
fix issue_1582

### LINKED ISSUE #1582: Consider making chapter numbers inline-end aligned when their book is in the right column
Currently, all chapter numbers in the Bible Selector are default aligned. (left for LTR language and right for RTL language). This can make books with only a few chapters look a little weird when in the 5 column or 3 column layouts. For example:

Image

Image

Image

In the images above, opening the book doesn't show the chapter numbers directly beneath the book, but instead quite a bit away. This can make it visually challenging to follow and makes finding the right chapter a lot more difficult.

In these cases, I think setting `justify-content` to `flex-end` can greatly improve the visual appearance:

Image

Image

AmitPant9999:
fixes : #1582

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
@@ -455,6 +455,7 @@ const SideBarBooks = (props: {
       index: number,
       chapterPos: number,
       separator: number,
+      chapterAtEnd: boolean,
       chapterHint?: number,
       itemStyle?: JSX.CSSProperties,
       narrowChapterStyle?: boolean,
@@ -511,10 +512,13 @@ const SideBarBooks = (props: {
                 style={{
                   ...chapterGridStyle,
                   justifyContent:
-                    ws <= MOBILE_BREAKPOINT ||
-                    bd.numberOfChapters < 4 * separator
-                      ? "flex-end"
-                      : "space-between",
+                    ws <= MOBILE_BREAKPOINT
+                      ? "flex-start"
+                      : bd.numberOfChapters < 4 * separator
+                        ? chapterAtEnd
+                          ? "flex-end"
+                          : "flex-start"
+                        : "space-between",
                 }}
               >
                 {narrowChapterStyle && ntColumns === 2 && (
@@ -547,6 +551,8 @@ const SideBarBooks = (props: {
         openVisualRow !== null
           ? { gridColumn: "1 / -1", gridRow: openVisualRow + 2 }
           : undefined;
+      const openVisualCol = columns > 1 ? Math.floor(lbc / rows) : 0;
+      const chapterAtEnd = openVisualCol === columns - 1;
 
       return (
         <div
@@ -559,6 +565,7 @@ const SideBarBooks = (props: {
               index,
               chapterPos,
               columns,
+              chapterAtEnd,
               chapterHint,
               itemStyle,
               narrowChapterStyle,
--- a/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
@@ -513,7 +513,7 @@ const SideBarBooks = (props: {
                   justifyContent:
                     ws <= MOBILE_BREAKPOINT ||
                     bd.numberOfChapters < 4 * separator
-                      ? "flex-start"
+                      ? "flex-end"
                       : "space-between",
                 }}
               >


Blitzy:
> [!NOTE]

## Code review — PR No. 1583 (fixes issue #1582)

**What the PR does:** In `SideBarBooks`, the chapter panel's `justify-content` for short books (and mobile) is flipped from `flex-start` to `flex-end`:

```diff
  justifyContent:
    ws <= MOBILE_BREAKPOINT ||
    bd.numberOfChapters < 4 * separator
-      ? "flex-start"
+      ? "flex-end"
      : "space-between",
```

**One thing done right:** using the logical `flex-end` value (rather than a physical `right`) is the correct choice — the chapter row lives inside a `dir`-aware container, so `flex-end` resolves to the inline-end in both LTR and RTL. That matches the issue's "inline-end aligned" wording and keeps RTL working.

However, the change is applied unconditionally, and the layout it lives in is **column-major with a full-width chapter row**, so it produces two problems.

---

### Findings by severity

#### 🟠 Medium — Short books in the left/middle column now misalign (the fix is not column-aware)

The chapter panel is a single full-width grid row (`chapterGridStyle = { gridColumn: "1 / -1", ... }`, line 548). The expanded book, however, can sit in **any** column — its column is `Math.floor(lbc / rows)`. Issue #1582 explicitly scopes the request to *"when their book is in the right column."* This PR ignores the column and right-aligns **every** short book, so it fixes right-column books but breaks left/middle-column ones — the mirror image of the original bug.

Before this PR: left-column short books were correct (chapters under the book), right-column ones were wrong.
After this PR: right-column short books are correct, left-column ones are now wrong. It's a lateral trade, not a general fix.

**Reproduction (desktop, wide viewport so OT is 3 columns / NT is 2 columns):**
1. Open the Bible Selector on a wide screen.
2. Expand **Ruth** (OT, left column, 4 chapters) — or **Galatians** (NT, left column, 6 chapters).
3. Observe: the chapter numbers now render bunched at the far **inline-end** (right in LTR), several columns away from the book you just expanded — the exact "hard to follow" problem #1582 set out to remove, only now for left-column books.

**Suggested fix — make the alignment depend on the expanded book's column.** The needed data (`rows`, `columns`, `lbc`) is already computed in `renderBooksGrid`, so compute the alignment there and thread it into `renderBook` instead of deciding it inline:

```ts
// in renderBooksGrid, where rows/columns/lbc are in scope:
const openVisualCol = columns > 1 ? Math.floor(lbc / rows) : 0;
const chapterAtEnd = openVisualCol === columns - 1; // book sits in the inline-end column
```

Then in `renderBook`, keep `space-between` for many-chapter books but choose the short-book alignment by column:

```ts
justifyContent:
  ws <= MOBILE_BREAKPOINT
    ? "flex-start"
    : bd.numberOfChapters < 4 * separator
      ? (chapterAtEnd ? "flex-end" : "flex-start")
      : "space-between",
```

Note: this cleanly handles the two extreme columns (leftmost → `flex-start`, rightmost → `flex-end`). A short book in a **middle** column (only possible in the 3-column OT grid) still can't be placed directly under its book with `justify-content` alone — that would need a column offset on the chapter row. That refinement is out of scope here; the important thing is to stop regressing left-column books.

#### 🟠 Medium — Mobile layout changed, which #1582 did not ask for

On mobile, `otColumns`/`ntColumns` are both `1` (lines 385–387), so there is no "right column" — the book already spans the full width. The `ws <= MOBILE_BREAKPOINT` term in the condition means this PR also flips **every** mobile chapter panel from inline-start to inline-end, pushing the chapter numbers to the right edge, away from the left-aligned book title. Issue #1582 is explicitly about the desktop *"5 column or 3 column layouts,"* so this is an unintended, out-of-scope visual regression.

**Reproduction:**
1. Open the app at a narrow (mobile) width.
2. Expand any book.
3. Observe: chapter numbers sit at the inline-end (right) instead of starting under the book title as before.

**Suggested fix:** keep `flex-start` on mobile — the `ws <= MOBILE_BREAKPOINT ? "flex-start"` branch in the snippet above already does this.

#### 🟢 Low / Nits
- **No test coverage.** The alignment logic now has three meaningful cases (mobile, short-book-by-column, many-chapter). Even a small unit test asserting the chosen `justify-content` per case would lock in the intended behavior and prevent the next person from reintroducing the same regression.
- **Readability.** The `bd.numberOfChapters < 4 * separator` heuristic (`separator` is the column count, so this is "fewer than 4 rows' worth of chapters") is non-obvious. A one-line comment on why `4 * columns` is the threshold would help future readers.

---

### Summary
The direction-aware `flex-end` value is the right primitive, but applying it unconditionally means the PR doesn't yet implement the actual request in #1582 — *align to the inline-end only when the book is in the inline-end column* — and it regresses both left/middle-column desktop books and the single-column mobile layout. Making the alignment depend on the expanded book's column (data already available in `renderBooksGrid`) and leaving mobile at `flex-start` would resolve both Medium findings.

---

PR #1587, opened 2026-08-03
Fix stale highlights after switching accounts

### LINKED ISSUE #1564: Highlights show the previous user's highlights after switching accounts (stale cache?)
## What is happening

Highlights don't load properly after signing out of one account and into a different one. It looks like a caching issue: the current page's highlights refresh correctly, but nearby pages still show the previous user's highlights until a full page refresh.

Repro:

1. User A loads Genesis 10-13, viewing highlights along the way (so those chapters get cached).
2. User A navigates back to Genesis 10.
3. User A signs out, and User B signs in. User B also has highlights on Genesis 10-13.
4. Genesis 10 shows User B's highlights correctly.
5. Navigating to Genesis 11 or Genesis 12 shows User A's highlights instead of User B's.
6. This doesn't correct itself until the page is fully refreshed.

#### User A Genesis 10-13 highlights 

These are User A's true highlights

Image
Image
Image

#### User B Genesis 10-13 highlights (user's true highlights)

These are User B's true highlights

Image
Image
Image

#### User B Genesis 10-13 highlights after following reproduction steps

Notice that the first shot is User B's true highlights for Genesis 10 but Genesis 12-13 are User A's highlights.
Image
Image
Image

## What is expected

Highlights should always reflect the currently signed-in user, on every chapter, without needing a full page refresh. Signing out and into a different account should invalidate any cached highlight data from the previous user.

Blitzy:
## What was happening

After signing out of one account and into another, highlights on chapters you'd already visited kept showing the *previous* account's highlights, not the new one's. It only fixed itself on a full page refresh.

Closes #1564

## Why

Highlights were cached in a signal per chapter address (`translationId/bookId/chapter`), with no account in the key. On sign-in, the manager tried to detect the account change and clear stale data — but there was a window where a slow-arriving response for account A could land after account B had signed in, overwriting B's entry with A's data.

## The fix

Highlights are now cached per **account + chapter**, not just per chapter. A response for account A can only ever be written into account A's own cache entry, so there's no window where it can land on account B's data — the old detect-and-clear logic is gone because there's nothing left for it to detect.

The chapter view returned to callers is unaffected by this internally: it's a computed value that follows whichever account is currently signed in, updating in place when the account changes so nothing holding onto the view has to re-request it.

## Two related bugs caught during review, fixed here too

**Highlighting/unhighlighting a verse could write to the wrong account.** Adding or removing a highlight reads the chapter's existing highlights, merges the change in, then saves. The read and the save each independently asked "who's signed in?" — normally the same answer, but if the account changed in the gap between them (e.g. an auto sign-out mid-request), the merged result got saved under the *new* account, overwriting their highlights with the old account's data plus the edit. The save now always writes to the account the data was actually read from.

**A save could get silently reverted by a slow load.** If a chapter's highlights were still loading when a save completed, the load could land afterward and overwrite the save with older, pre-save data. Loads now check whether something newer has already settled before applying their result.

## Testing

- 38 unit tests in `HighlightsManager.test.ts` (14 new), including a dedicated `account switching (regression for #1564)` suite that reproduces the original bug plus the two additional cases above.
- Coverage on the touched file: 98.5% statements / 94% branches / 100% functions.
- `pnpm check:ts`, `pnpm lint`, and `pnpm vitest run HighlightsManager.test.ts` all pass.
- Manually verified in browser following original repro steps:: sign in as A, view a few chapters, sign out, sign in as B, revisit those chapters without refreshing.


Blitzy:
> [!NOTE]

## Code review — PR #1587 "Fix stale highlights after switching accounts"

### Summary

This PR re-keys the highlights cache from **chapter address** to **account + chapter address**, and hands callers an identity-stable `computed` "view" per chapter that follows whichever account is signed in. That structural change makes the original bug (#1564) impossible by construction: a response for account A can only ever be written into account A's own entry, so it can never land on account B's data. Two related races the PR describes — a mutation writing to whoever is signed in by the time the write starts, and a slow load reverting a completed save — are also closed (write pinned to the account the data was read from; loads bail if the entry already `settled`).

Overall this is a solid, well-reasoned fix with strong test coverage. I could not find a correctness defect. The items below are all low-severity / nits.

### Verification performed

- `pnpm vitest run HighlightsManager.test.ts` → **38 passed** (14 new).
- Confirmed the `Signal` → `ReadonlySignal` return-type change is safe: the only non-test callers are `BibleReadingManager.tsx:1818` (assigns the view into an outer signal, reads `.value.value`) and `packages/twitchPub-extension/ext_twitchPub/host/init.tsx:80` (reads `.value` only). Neither writes to the returned signal. `SessionsManager` does not call `getChapterHighlights` directly.
- Traced the account-switch, sign-out/in, late-response, and mid-load-mutation paths against the eviction `effect` + per-account keying; the stale-load and wrong-account-write windows are genuinely closed.

---

### Findings by severity

#### Low

**L1 — `views` map is never pruned, unlike `entries`.** `HighlightsManager.tsx:389`, eviction effect at `:484-495`.
The eviction effect drops stale `entries` on account change but leaves `views` to accumulate one `computed` per distinct chapter address for the lifetime of the manager. In practice this is bounded by (translations × chapters), so it is not a practical leak for a singleton app-lifetime manager — but the asymmetry (one map pruned, the sibling map not) is a foot-gun for the next reader, who may assume both are cleaned up.
*Repro:* not user-observable; visible only by inspecting `views.size` growing monotonically as a reader browses chapters.
*Suggested fix:* either add a short comment explaining that `views` is intentionally retained (views are cheap and identity-stability across account switches is the whole point), or drop the view alongside its entries when no entry for that address remains.

**L2 — The manager-level `effect()` is created but never disposed.** `HighlightsManager.tsx:484`.
`effect()` returns a dispose function that is discarded, so the subscription to `login.userId` lives forever. Fine for a manager instantiated once per app, but it means the manager can't be cleanly torn down (relevant if a test or a future multi-instance scenario ever creates and discards managers — leaked effects would keep firing).
*Suggested fix:* if managers are ever expected to be disposable, capture the returned disposer and expose it; otherwise a one-line comment noting the intentional app-lifetime subscription would prevent a future "why isn't this cleaned up?" question.

#### Nitpick

**N1 — Side effects run inside the `computed` view.** `HighlightsManager.tsx:466-474`.
`getOrCreateView`'s computed mutates the `entries` map (`getOrCreateEntry`) and kicks async work (`ensureLoaded`) during evaluation. This is safe today because the operations are idempotent and no signal is written *synchronously* — and the inline comment on `ensureLoaded` (`:442-444`) already calls this out. The nit is only that a Preact `computed` is conventionally expected to be pure; anyone extending this later needs to preserve the "idempotent, no synchronous signal writes" invariant. A one-line note on the computed itself (not just on `ensureLoaded`) would make that contract local to where it's easy to break.

**N2 — Stale interface JSDoc.** `HighlightsManager.tsx:264-269`.
The `HighlightsManager` interface doc still reads *"Highlights are keyed by `translationId/bookId/chapterNumber` ... persisted per user."* After this PR the cache is keyed by **account + address**, which is the crux of the fix. The detailed factory-function comment (`:366-381`) is correctly updated, but this interface-level comment now understates the design. Worth a one-line update so the two don't disagree.

**N3 — CHANGELOG omits the third fix.** `CHANGELOG.md`.
The two added bullets cover the stale-highlights bug and the wrong-account write. The PR description's third fix — "a save could get silently reverted by a slow load" (guarded by the `entry.settled` check at `:418`) — isn't listed. It's arguably too internal to be user-facing, so this is a judgment call, but if the first two warranted an entry the third may too.

---

### Out of scope (pre-existing, not introduced here)

Noting these only so they aren't mistaken for regressions — both predate this PR and behave the same as before:

- **Optimistic local write has no rollback.** `writeChapterHighlights` (`:536-539`) sets `entry.data.value` and `settled = true` *before* `os.recordData` resolves. If the persist throws, the local cache keeps the unpersisted value (and `settled = true` means it won't be re-fetched until an account switch or refresh). The old `saveChapterHighlights` had the identical optimistic pattern, so this isn't new — just now shared by `highlightVerses`/`unhighlightVerses`.
- **Concurrent read-modify-write on the same chapter can lose an update.** Two mutations awaiting the same in-flight load both read the same pre-mutation `entry.data`, so last-write-wins. Also pre-existing (the previous `getChapterHighlights` + `awaitChapterLoad` flow had the same shape).

Neither blocks this PR; flagging in case a follow-up wants to harden the mutation path.

---

### Verdict

No blocking issues. The fix is correct, the design choice (per-account keying + account-tracking views) is the right one, and the regression suite for #1564 is genuinely convincing. The items above are polish, not blockers.

---

KallynGowdy (approved):

PR #1588, opened 2026-08-03
fix: correct broken dynamic import path in policy modals

KallynGowdy:
TermsOfServiceModal, PrivacyPolicyModal, and CodeOfConductModal each
lazily import their policy HTML from an i18n JSON bundle, but the
import path was left at one directory level up after a refactor moved
these components into their own subfolders. The path resolved to a
nonexistent location, so the import silently rejected and the modals
always rendered with an empty body.

Fix the path to go up two levels, matching the correct static
useI18n import already in each file, and log a console error if the
import ever fails again instead of failing silently.

Add regression tests for all three modals that render with a real
i18n instance and assert the actual policy text loads.

### DIFF
--- a/packages/seed-bible/seed-bible/components/CodeOfConductModal/CodeOfConductModal.tsx
+++ b/packages/seed-bible/seed-bible/components/CodeOfConductModal/CodeOfConductModal.tsx
@@ -25,12 +25,14 @@ function loadCodeOfConduct(
 ): Promise<void> {
   let promise = loadingPromises.get(language);
   if (!promise) {
-    promise = import(`../i18n/policies/code-of-conduct/${language}.json`).then(
-      (resources) => {
+    promise = import(`../../i18n/policies/code-of-conduct/${language}.json`)
+      .then((resources) => {
         i18n.addResourceBundle(language, "code-of-conduct", resources.default);
         loadedLanguages.value = new Set(loadedLanguages.value).add(language);
-      }
-    );
+      })
+      .catch((error) => {
+        console.error("Failed to load code of conduct policy bundle", error);
+      });
     loadingPromises.set(language, promise);
   }
   return promise;
--- a/packages/seed-bible/seed-bible/components/CodeOfConductModal/CodeOfConductModal.tsx
+++ b/packages/seed-bible/seed-bible/components/CodeOfConductModal/CodeOfConductModal.tsx
@@ -19,6 +19,11 @@ const loadingPromises = new Map<string, Promise<void>>();
  * the bundle is available — callers can await it (during SSR) to block rendering
  * until the policy text is present.
  */
+// Resolves — never rejects, even on failure (logged below). A rejected
+// promise thrown during `renderToStringAsync` surfaces as a render exception
+// and takes down the whole SSR document (see BibleReadingManager's
+// `chapterDataPromise`); resolving lets the modal render with an empty body
+// instead.
 function loadCodeOfConduct(
   i18n: I18nInstance,
   language: string
@@ -32,6 +37,8 @@ function loadCodeOfConduct(
       })
       .catch((error) => {
         console.error("Failed to load code of conduct policy bundle", error);
+        // Don't cache the failure — let the next modal open retry the import.
+        loadingPromises.delete(language);
       });
     loadingPromises.set(language, promise);
   }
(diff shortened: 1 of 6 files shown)


Blitzy:
> [!NOTE]

## Code Review — fix: correct broken dynamic import path in policy modals

**Summary:** This is a tight, well-scoped fix. The root cause is correctly identified and the correction is right. All three policy modals (`TermsOfServiceModal`, `PrivacyPolicyModal`, `CodeOfConductModal`) now resolve their lazily-imported policy bundle two directory levels up, which matches where `i18n/policies/` actually lives relative to each component's subfolder. I verified the fix and ran the new tests — all 6 pass.

### Verification performed

- **Path is correct.** Each component sits in `components/<Name>Modal/<Name>Modal.tsx`. From there, `../../i18n/policies/...` resolves to `packages/seed-bible/seed-bible/i18n/policies/...`, which exists (`code-of-conduct/`, `privacy-policy/`, `terms-of-service/`, each with `en.json`). The old `../i18n/policies/...` pointed at `components/i18n/policies/...`, which does not exist — confirming the reported "silent empty body" bug was real.
- **Consistency.** Grepped the whole `packages/` tree; all three imports are now `../../` and no single-level `../i18n/policies` references remain. The corrected path also matches the already-correct static `useI18n` import (`../../i18n/I18nManager`) in each file.
- **Tests pass.** `pnpm vitest run` on the three new test files → 3 files / 6 tests passing. The assertions check real bundle content (e.g. `ao-lab-web-services-acceptable-use-policy` / `AO Lab Web Services Acceptable Use Policy`), which matches the actual `en.json`, so they genuinely guard the load path rather than trivially passing.

### Issues by severity

**None blocking.** The change is correct and safe to merge. The notes below are low-severity / optional and can be deferred.

#### Low — SSR failure now degrades to a silent-but-logged empty body

The new `.catch` makes the load promise always *resolve*, even on failure. On the client this is an improvement (it prevents an unhandled rejection). But on the server path — `if (isOpen && import.meta.env.SSR && !isLoaded) throw loadCodeOfConduct(...)` — the suspense promise now resolves instead of rejecting when the import genuinely fails, so the modal re-renders with an empty policy body. That is arguably intentional graceful degradation, but it re-creates the exact "empty content area" this PR set out to eliminate, just with a `console.error` instead of a hard error. Worth a conscious decision on whether an SSR bundle-load failure should surface anything (error boundary / user-facing fallback) rather than render blank.

#### Low — a failed load is cached and never retried within the session

On failure, `.catch` swallows the error, the resolved promise stays in `loadingPromises`, and `loadedLanguages` never records the language (so `isLoaded` stays `false`). Re-opening the modal calls `loadCodeOfConduct` again, which returns the cached already-resolved promise — no new import attempt. A transient failure therefore leaves that language's modal permanently empty until a full page reload. This caching-of-failures behavior is largely pre-existing, but since the PR's stated goal is robustness, consider deleting the map entry inside `.catch` so a later open can retry:

```ts
.catch((error) => {
  console.error("Failed to load code of conduct policy bundle", error);
  loadingPromises.delete(language); // allow a retry on next open
});
```

#### Nit — three near-identical `loadX` helpers

`loadCodeOfConduct` / `loadPrivacyPolicy` / `loadTermsOfService` (and their module-level `loadedLanguages` / `loadingPromises` signals) are copy-paste identical apart from the folder segment and namespace string. Not worth changing in this PR, but a small shared `createPolicyLoader(namespace, folder)` factory would remove the drift risk that caused this very bug — the three paths fell out of sync during the refactor precisely because they're maintained independently.

### Reproduction (original bug, for reference)

1. On `develop` before this change, open the app and trigger any of the three policy modals (e.g. from the footer/login policy links).
2. The dialog opens with a title but an empty body — the dynamic `import()` rejects because `components/i18n/policies/...` does not exist.
3. Before this PR the rejection was silent (only a `.then`, no `.catch`); after this PR the same failure would log `Failed to load … policy bundle` to the console.

### CI note

At review time the `build (24.x)` check was still in progress and `mergeable_state` was `blocked` (pending the requested review from @Blitzy), not a failing check. The three added unit tests pass locally.

---

KallynGowdy:
Thanks for the review — addressed both low-severity items in 1e9c31c:

- **Cached-failure retry**: `.catch` now deletes the `loadingPromises` entry for that language on failure, so a transient import failure no longer permanently empties the modal for the rest of the session — the next open retries the import. Applied to all three modals.
- **SSR silent-empty-body**: researched how `preact-render-to-string`'s suspense-style thrown-promise handling actually behaves here — a rejected promise thrown during `renderToStringAsync` fails the *entire* SSR document (server falls back to the un-rendered shell for that request), not just this component. The codebase already has this exact tradeoff documented deliberately in `BibleReadingManager`'s `chapterDataPromise` ("resolves — never rejects... a rejected promise... takes down the whole document"). So resolving-with-empty-body here is the correct, consistent choice, not an oversight — added the same doc comment to make that explicit rather than changing the behavior.

Left the "three near-identical `loadX` helpers" nit as-is, per your own note that it's not worth doing in this PR.

Full suite (202 files / 3703 tests), `check:ts`, and `lint` all pass.

---

PR #1593, opened 2026-08-03
Add the ability to save custom annotations

### LINKED ISSUE #1516: Design and add support for user annotations
We need to let users write their own annotations attached to verses.

Requirements:
- This requires some data structure design.
- Saves to user record like bookmarks.
- Works anonymously and/or offline to device and syncs to record when signed in.

Design questions:

- Do we show these inline with verses like we do with annotations that come with translations?
- The annotation viewer could support showing multiple annotations for a single verse.

KallynGowdy:
Fixes #1516


Blitzy:
> [!NOTE]

Reviewed against the linked issue (No. 1516, "Design and add support for user annotations") and the current diff of `feature/user-annotations` vs `develop` (34 files, ~6k lines). Overall the feature is well-built: the manager-level tests (`AnnotationsManager.test.ts`, `DiscoverPane.test.tsx`, `CreateAnnotationForm.test.tsx`) exercise real state transitions (signal updates, save/cancel flows, live-selection syncing) rather than tautological mock-echoes — one of them (`keeps a new draft's verse targeting live-synced...`) even demonstrates the exact mechanism behind issue No. 1 below, just not across a tab switch. Found 5 issues, none blocking the core save/list/edit/delete flow, but two are user-facing correctness bugs worth fixing before merge.

## 1. Tapping an already-selected verse's annotation number hides the note instead of opening it (High)

**File:** `packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx` (`handleAnnotationVerseClick`, ~line 1660)

`handleAnnotationVerseClick` unconditionally calls `selectVerse(verse, ...)`, but `selectVerse` (`BibleReadingManager.tsx`) is a pure toggle — if the verse is already selected, it *removes* it from `selectedVerses` instead of adding it.

**Repro:**
1. On mobile, tap a verse's text to select it (opens the mobile verse toolbar).
2. That verse has a note, so its verse number renders boxed/clickable. Tap the boxed number to view the note.
3. `selectVerse` sees the verse is already selected and deselects it. `isVerseToolbarVisible` is gated on `hasVerseSelection`, so the whole verse toolbar disappears, and the note never opens — the tap effectively looks like "hide toolbar" instead of "show note."

**Fix:** Don't route the annotation-number click through the toggle. Either add the verse to the selection unconditionally (only remove-toggle from the plain verse-text click path), or check `isSameSelectedVerse` first and skip the `selectVerse` call when already selected, going straight to the scroll/open-Discover logic.

## 2. Clicking an annotation list row highlights verses the annotation doesn't actually cover (Medium)

**File:** `packages/seed-bible/seed-bible/components/DiscoverPane/DiscoverPane.tsx` (~line 746) and `emphasizeVerses` in `BibleReadingManager.tsx` (~line 1141)

Annotations can target non-contiguous verses (`verseNumbers`, e.g. `[3, 5, 7]` when a user selects verses 3 and 5 while skipping 4 — confirmed possible via `deriveVerseTargeting`, and covered by the `preserves gaps in a non-contiguous selection` test). But navigating to an annotation from its Discover-pane row calls `emphasizeVerses({ verse: annotation.verseNumber, endVerse: annotation.endVerseNumber })`, and `emphasizeVerses` decorates `range(verse, endVerse + 1)` — i.e. *every* verse between the min and max, not the annotation's actual `verseNumbers`.

**Repro:** Create an annotation over verses 3 and 5 only (skip 4). Open it from the Discover pane list. Verses 3, 4, and 5 all flash as "covered," even though verse 4 was never part of the annotation.

**Fix:** Have `emphasizeVerses` accept an explicit verse list (or thread `annotation.verseNumbers` through when present) instead of always expanding `[verse, endVerse]` into a contiguous range.

## 3. Switching reading tabs mid-draft silently changes (or wipes) the annotation's verse targeting (Medium)

**File:** `packages/seed-bible/seed-bible/managers/AnnotationsManager.tsx` (live-sync `effect`, ~line 628)

While `isDraftingNewAnnotation` is true, the sync effect reads `activeTab.value` — whichever tab is currently selected in `TabsManager` — not the tab the draft was started on. On desktop the Discover pane/composer is a docked side panel, not a modal, so switching to a different open reading tab while a draft is open is a normal, unblocked action.

**Repro:**
1. In tab A (Genesis 1), select verse 5 and start a new annotation (`createNewAnnotation`). The draft targets GEN 1:5.
2. With the composer still open, switch to tab B (e.g. Exodus).
3. The effect re-runs against tab B's selection. If tab B has no matching selection for GEN/1 (the draft's own book/chapter), `deriveVerseTargeting` returns `{ verseNumber: null, endVerseNumber: null, verseNumbers: null }`, and `editingAnnotation.value` is overwritten — the verse-5 targeting is silently lost and the draft becomes a whole-chapter annotation with no warning.

**Fix:** Capture the tab the draft was started on (e.g. store its id alongside `isDraftingNewAnnotation`) and sync against that specific tab rather than whatever `activeTab` currently resolves to, or pause syncing while the active tab differs from the draft's tab.

## 4. `AnnotationAuthor` calls a hook after a conditional early return (Medium, latent)

**File:** `packages/seed-bible/seed-bible/components/DiscoverPane/DiscoverPane.tsx` (~line 590-596)

```tsx
useEffect(() => { ... }, [userId]);

if (!userId) {
  return null;
}

const { t } = useI18n();
```

`useSignal`/`useSignal`/`useEffect` run unconditionally, but `useI18n()` only runs when `userId` is truthy. This breaks the rule that a component must call the same hooks in the same order on every render. It won't misbehave as long as a given `AnnotationAuthor` instance's `userId` prop never flips between falsy and truthy across re-renders of the *same* instance (same React/Preact key) — which may hold today — but it's fragile: any future change that reuses the same list-item instance while `userId` gets backfilled (e.g. a legacy annotation missing `userId` normalized in place) would desync Preact's per-instance hook slot order.

**Fix:** Move `const { t } = useI18n();` above the `if (!userId) return null;` guard.

## 5. `groupAnnotationsByVerseRange` can merge annotations with different actual verse sets (Low)

**File:** `packages/seed-bible/seed-bible/managers/AnnotationsManager.tsx` (`annotationVerseRangeKey`, ~line 194)

The grouping key is `${min(verseNumbers)}-${max(verseNumbers)}`, so an annotation over `[3, 5]` (gap at 4) and a separate annotation over `[3, 4, 5]` both key to `"3-5"` and get merged into one displayed group/header in the Discover pane and mobile verse sheet, even though they don't target the same verses. This looks intentional per the function's doc comment ("same start *and* end verse"), so it's more a design edge case than a bug, but it can mislead a reader into thinking both comments are about verse 4 when one explicitly skips it.

**Suggested fix (optional):** Key on the full sorted `verseNumbers` array (or its `formatAnnotationVerseNumbers` string) instead of just min/max, so only annotations covering the exact same verses group together.

---

PR #1594, opened 2026-08-03
Repair highlighting in shared sessions.

### LINKED ISSUE #1590: Highlights are broken in shared sessions: built-in colors don't work, clear button does nothing
## What is happening

Highlighting is broken specifically inside shared sessions, in two ways:

- **Built-in highlight colors don't work.** Only custom colors seem to apply/sync correctly in a shared session. The 3 built-in colors do not.
- **The clear button does nothing.** Clearing a highlight has no effect while in a shared session.

Both of these are scoped to shared sessions. Neither issue happens in regular, non-shared reading.

_Evidence:_ Loom walkthrough (timestamp 0:55): https://www.loom.com/share/b24a81faa78e44ad943ccc58585a23ac

## What is expected

In a shared session, highlighting should work the same as it does outside of a session:

- All highlight colors, built-in and custom, should apply and sync correctly to other participants.
- Clearing a highlight should remove it.

Blitzy:
Closes #1590.

## What was wrong

Highlighting inside a shared session never created a highlight. It created a **decoration** — the generic per-verse styling channel — and hand-wrote CSS to imitate what a highlight looks like. That imitation broke in July.

`a89016678` ("highlights are now a separate svg layer rendered behind text content") moved highlight backgrounds out of CSS and into an SVG ribbon layer, reducing `.sb-highlight-<id>` to a font colour only. Anything still painting a highlight by applying that class silently stopped painting anything at all.

That accounts for both reported symptoms:

- **Built-in colours didn't work.** The decoration sent `className: "sb-highlight-yellow"` and no background. Custom colours survived only because they carried an inline `backgroundColor`. The person highlighting still saw their own colour, because the same action also saved a real highlight — but that lives in their own record and never reached anyone else.
- **Clear did nothing.** Clear removed the broadcast decoration but never the saved highlight, so the verse stayed coloured.

## The fix

Rather than patch the imitation, decorations can now carry a real highlight:

```ts
rs.decorateVerses(bookId, chapter, verse, { highlight: { colorId: "yellow" } }, id)
```

The reader folds those into its own highlight lookup, so they render through the existing ribbon layer — same rounded shape, same run merging, same fade in and out. Because the `colorId` travels rather than a resolved colour, a preset now renders in **each viewer's own theme**: a participant in dark mode sees their dark-mode yellow, not the sender's.

Three separate call sites were hand-rolling that CSS. This removes the pattern from the codebase.

## Rules for session highlights

Settled with the team over the course of this branch:

| Context | Saves personally | Broadcasts |
| --- | --- | --- |
| Not in a session | yes | — |
| In a session, not permitted to broadcast | yes | no |
| Broadcasting, duration = ∞ | yes | yes |
| Broadcasting, finite duration | no | yes |
| Broadcasting, signed out | no | yes |

A saved copy is always the author's alone — participants receive the broadcast, never a highlight of their own.

A broadcast **covers** a personal highlight rather than replacing it, so when a timed highlight expires the reader's own colour reappears underneath. An earlier design deleted the personal highlight instead; that was shipped in `ef4f1b261`, removed as a bug in `f11f39e55`, briefly restored from a stale comment during this work, and removed again.

## Visual treatment

A broadcast highlight draws as an **outline** — same colour, stroked, reduced fill — while the reader's own stays solid. Without this, a covering highlight and a personal one are indistinguishable, and expiry looks like losing your work.

Under ∞ the highlight is saved *and* broadcast, so it shows as an outline for everyone while the session is live and returns to solid when the user leaves. That needs no special handling: the decoration only exists inside the session tab.

image

## Also fixed

- **A restricted participant couldn't highlight at all.** `8e3e79997` is titled *"Save user highlights when they aren't allowed to decorate sessions"* but wrote the inverted condition, so those users matched no branch: no saved highlight, and a local-only decoration that never synced and vanished after 16 seconds.
- **Signed-out users were interrupted by a login modal**, with the highlight applied behind it and then quietly not saved. Broadcasting needs only a connection id, so it no longer reaches for the save path. Where signing in *is* the point, the prompt now settles before anything is applied.
- **Clearing prompted for login even when nothing was saved.** Unhighlighting now returns early when no saved highlight covers the selection, which also drops a wasted record write for signed-in users.
- **Twitch chat highlights** had the same broken `sb-highlight-<id>` pattern and are migrated to the new field.
- **An unrecognised colour id painted a solid black bar.** `var(--sb-highlight-<id>-color)` with no fallback is invalid at computed-value time, and `fill` inherits down to its initial value of black. Now falls back to `transparent`. Reachable once Twitch colour ids — which come from a chat config, not our picker — started going through this layer.

## Known limitations

- Colour ids aren't validated against the theme's presets. An unsupported colour from Twitch chat (their test fixture uses `red`) now renders invisibly rather than as a black bar, but it still won't show.
- When two highlight-carrying decorations land on the same verse — a session broadcast and a Twitch highlight, say — the winner is decided by decoration array order, which can differ between clients. Session highlights use a deterministic per-verse id, so two of those can't collide.
- The broadcast stroke is centred on the ribbon path, so it extends ~1px past the measured bounds. Nothing clips it; adjacent differently-coloured ribbons may touch.

## Not included

Two session bugs found during two-user testing are **diagnosed but deliberately not in this branch**, to keep it reviewable:

- Re-highlighting a verse doesn't reset the expiry timer — the broadcast path arms a `setTimeout` per publish and never cancels the previous one, so a re-highlight inherits the original deadline.
- A second user highlighting a verse the first already highlighted is dropped — `decorationOwners` makes decoration ids first-writer-wins, so the second user's write is filtered out before it reaches the CRDT.

## Testing

199 test files / 3720 tests passing, `pnpm check:ts` clean, `pnpm lint` 0 errors. Roughly 950 lines of new test coverage across the reader, toolbar, sessions, and the Twitch extension.

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.css
@@ -455,6 +455,22 @@
   }
 }
 
+/* A highlight arriving from a shared session (or an extension) is outlined
+   rather than solid, so it reads as somebody else's mark laid over the page
+   instead of one of the reader's own. The reader's highlight is still there
+   underneath and returns to solid when the broadcast goes away. Same colour on
+   the stroke — only the amount of fill differs — so the colour still says who
+   picked what. One opacity serves both themes; the presets are light enough in
+   light mode and dark enough in dark mode to read at the same strength. */
+.sb-highlight-ribbon-broadcast {
+  fill-opacity: 0.5;
+  stroke-width: 2;
+  transition:
+    opacity 0.25s ease,
+    fill 0.25s ease,
+    stroke 0.25s ease;
+}
+
 @media (prefers-reduced-motion: reduce) {
   .sb-highlight-ribbon {
     transition: none;
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
@@ -181,6 +181,16 @@ function getVersePlainText(content: ChapterVerse["content"]): string {
   return content.map((part) => getInlineText(part)).join("");
 }
 
+/**
+ * A highlight resolved for one verse, plus where it came from. `broadcast` is
+ * true for a highlight carried by a decoration — a session peer's, or an
+ * extension's — as opposed to one the reader saved themselves.
+ */
+interface ResolvedHighlight {
+  highlight: ChapterHighlight;
+  broadcast: boolean;
+}
+
 function hasContentTargeting(decoration: VerseDecoration): boolean {
   const hasTargetContent =
     typeof decoration.targetContent === "string" &&
@@ -566,10 +576,15 @@ function renderChapterContent(
   // `showHighlights` hides the reader's *saved* highlights. Decoration
   // highlights are a live signal from a session peer or an extension, so they
   // stay visible either way — as they did when they were plain CSS.
-  const getVerseHighlight = (verseNumber: number): ChapterHighlight | null => {
+  //
+  // `broadcast` distinguishes the two for rendering: a decoration highlight is
+  // drawn as an outline, a saved one as a solid ribbon. A broadcast covers the
+  // reader's own highlight rather than replacing it, so the outline is what
+  // says "this isn't yours, and yours is still underneath".
+  const getVerseHighlight = (verseNumber: number): ResolvedHighlight | null => {
     const decorated = decorationHighlights.get(verseNumber);
     if (decorated) {
-      return decorated;
+      return { highlight: decorated, broadcast: true };
     }
 
     if (!scriptureElements.showHighlights) {
@@ -579,14 +594,14 @@ function renderChapterContent(
     for (const highlight of highlights) {
       if (typeof highlight.verse === "number") {
         if (highlight.verse === verseNumber) {
-          return highlight;
+          return { highlight, broadcast: false };
         }
         continue;
       }
 
       const [start, end] = highlight.verse;
       if (verseNumber >= start && verseNumber <= end) {
-        return highlight;
+        return { highlight, broadcast: false };
       }
     }
 
@@ -597,20 +612,24 @@ function renderChapterContent(
   // ChapterContent), so a highlighted run's wrapper paints no background itself.
   // It only carries the readable font color and a `fill` (a CSS-var reference for
   // preset colors, or the custom hex) that the layer reads back off the DOM.
-  const getHighlightPresentation = (highlight: ChapterHighlight | null) => {
-    if (!highlight) {
+  const getHighlightPresentation = (resolved: ResolvedHighlight | null) => {
+    if (!resolved) {
       return {
         className: "",
         style: undefined as JSX.CSSProperties | undefined,
         fill: null as string | null,
+        broadcast: false,
       };
     }
+
+    const { highlight, broadcast } = resolved;
 
     if (highlight.customColor && highlight.customFontColor) {
       return {
         className: "sb-highlight",
         style: { color: highlight.customFontColor } as JSX.CSSProperties,
         fill: highlight.customColor,
+        broadcast,
       };
     }
 
@@ -618,6 +637,7 @@ function renderChapterContent(
       className: `sb-highlight sb-highlight-${highlight.colorId}`,
       style: undefined as JSX.CSSProperties | undefined,
       fill: `var(--sb-highlight-${highlight.colorId}-color)`,
+      broadcast,
     };
   };
 
@@ -645,14 +665,18 @@ function renderChapterContent(
     );
   };
 
-  const getHighlightColorKey = (highlight: ChapterHighlight | null) => {
-    if (!highlight) {
+  // Also keys on `broadcast`, so a broadcast highlight never merges into one run
+  // with a saved highlight of the same colour — they draw differently.
+  const getHighlightColorKey = (resolved: ResolvedHighlight | null) => {
+    if (!resolved) {
       return null;
     }
+    const { highlight, broadcast } = resolved;
+    const prefix = broadcast ? "broadcast:" : "";
     if (highlight.customColor && highlight.customFontColor) {
-      return `custom:${highlight.customColor}:${highlight.customFontColor}`;
+      return `${prefix}custom:${highlight.customColor}:${highlight.customFontColor}`;
     }
-    return highlight.colorId;
+    return `${prefix}${highlight.colorId}`;
   };
 
   // Renders a single verse's `<span class="sb-verse">`. The highlight background
@@ -939,6 +963,7 @@ function renderChapterContent(
           style={presentation.style}
           data-highlight-fill={presentation.fill ?? undefined}
           data-highlight-key={runKey}
+          data-highlight-broadcast={presentation.broadcast ? "true" : undefined}
         >
           {runIndices.map((idx) =>
             renderVerseNode(entries[idx] as ChapterVerse, idx)
@@ -1007,6 +1032,8 @@ interface Ribbon {
   key: string;
   d: string;
   fill: string;
+  /** Drawn as an outline rather than a solid fill — see `ResolvedHighlight`. */
+  broadcast: boolean;
   first: number;
   last: number;
   enter: boolean;
@@ -1161,6 +1188,7 @@ function ChapterContent(props: ChapterContentProps) {
         el,
         key: el.getAttribute("data-highlight-key") || `i${index}`,
         fill: el.getAttribute("data-highlight-fill") ?? "",
+        broadcast: el.getAttribute("data-highlight-broadcast") === "true",
         lines: collectLineRects(el, box.left, box.top),
         leadPad: padX,
         trailPad: padX,
@@ -1192,6 +1220,7 @@ function ChapterContent(props: ChapterContentProps) {
       key: string;
       d: string;
       fill: string;
+      broadcast: boolean;
       first: number;
       last: number;
     }> = [];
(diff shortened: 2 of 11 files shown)


Blitzy:
> [!NOTE]

## Code review — Repair highlighting in shared sessions

**Overall:** This is a strong, well-documented change. It correctly diagnoses the root cause (the July move of highlight backgrounds from CSS into the SVG ribbon layer in `a89016678`, which silently neutered the `sb-highlight-` imitation), and the fix — letting a decoration carry a real `highlight` that the reader draws through its own ribbon layer — is the right shape. The save-vs-broadcast truth table is implemented faithfully, the inverted-permission bug in `applyHighlightWithSession` is genuinely fixed, and the test coverage is thorough.

I verified locally on the branch head (`d14b91e`):

- **Affected suites pass:** `BibleReader` + `BibleReaderToolbar` (79 tests), `SessionsManager` + `HighlightsManager` + `twitchSubManager` (68 tests) — all green.
- **CI is green** on the PR (build + deploy both succeeded).
- The `applyHighlightWithSession` branch logic matches the documented table for all five contexts, and the `unhighlightVerses` early-return sits before `saveChapterHighlights` (where the login modal lives), so the "clear no longer prompts login" fix holds in isolation.

There is **one blocking issue** and a few lower-severity notes below.

---

### 🔴 High (blocking): Semantic merge conflict with `develop` in `HighlightsManager.tsx`

GitHub reports this PR as `mergeable_state: dirty`, and it is **not just a textual conflict** — it is a genuine design collision that needs careful re-application.

While this branch was open, `develop` merged #1587 (`fix/1564-stale-highlights`), which refactored highlight persistence into a **per-account entry model**: `saveChapterHighlights` now delegates to a new `writeChapterHighlights`, and both save and unhighlight resolve the account first via a new `resolveEntryToMutate(address)`. This PR edits the *old* structure in two places that `develop` rewrote:

1. **`saveChapterHighlights`** — this PR moves the optimistic `chapterHighlightsSignal.value = …` update to *below* the login gate (so a highlight no longer sits on screen looking saved behind the login modal). `develop` extracted that whole body into `writeChapterHighlights`.
2. **`unhighlightVerses`** — this PR adds a `coversAnyVerse` early-return so clearing an unsaved verse makes no write and prompts no login.

The subtle part is No. 2. This PR's early-return reads `current.value.highlights` **before any login happens**. In `develop`, reading is coupled to login: `unhighlightVerses` gets its data from `resolveEntryToMutate(address)`, which calls `login.login()` when the user is signed out. So the fix **cannot be re-applied verbatim** — dropping the early-return after `resolveEntryToMutate` would defeat its own purpose (the login prompt would already have fired). The coverage check has to be restructured to test verse coverage *without* resolving/logging-in the account.

**Reproduction:**
```bash
git checkout fix/1590-highlights-sessions
git merge origin/develop
# → CONFLICT (content): Merge conflict in
#   packages/seed-bible/seed-bible/managers/HighlightsManager.tsx
```

**Suggested fix:**
- Rebase (or merge `develop`) and re-apply both intents into the new entry-based structure.
- For the empty-clear fix, read current highlights from the already-loaded chapter signal (`getChapterHighlights(...).value`) *before* calling `resolveEntryToMutate`, and early-return when no saved highlight covers the selection — so the login path is never reached for a verse that was never saved.
- Keep a regression test asserting that clearing a verse with no saved highlight triggers **no** `login.login()` call and **no** write, exercised against the `resolveEntryToMutate` path (the current test may pass trivially against the old structure and give false confidence post-merge).

---

### 🟡 Low: Two diagnosed session bugs are deferred — make sure they get tracked

The description candidly lists two bugs found in two-user testing that are intentionally out of scope to keep the branch reviewable:

- Re-highlighting a verse doesn't reset the expiry timer (`setTimeout` per publish, previous one never cancelled).
- A second user highlighting a verse the first already highlighted is dropped (`decorationOwners` is first-writer-wins).

No objection to deferring them — that's a reasonable scoping call. The only ask is that they land as their own tracking issues (linked back here) so they don't get lost once this PR merges and closes #1590. Neither is a regression introduced by this PR.

### 🟡 Low: Unrecognised `colorId` now fails invisible (Twitch)

The `transparent` fallback (`var(--sb-highlight-${colorId}-color, transparent)`) correctly kills the solid-black-bar bug, and that trade-off is well reasoned. But as the description notes, an unsupported colour from a Twitch chat config (the extension's own fixture uses `red`) now renders **invisibly** rather than showing at all. Failing invisible is better than a black bar, but the highlight still doesn't appear.

Worth a follow-up: map/validate incoming `colorId`s from external sources (Twitch) against the theme's known presets, and fall back to a supported default (e.g. `yellow`) rather than to `transparent`, so those highlights actually render. Not blocking.

### 🟢 Nit: Decoration-highlight precedence is client-order-dependent (already documented)

When two highlight-carrying decorations land on the same verse, the winner is decided by decoration array order, which can differ between clients. Session broadcasts use a deterministic per-verse id so session-vs-session can't collide; the residual risk is session-vs-extension. Already called out under "Known limitations" — flagging only so it's on the reviewer's radar. Fine to defer.

---

**Recommendation:** Resolve the `develop` conflict (with the semantic care described above) and re-run the suite; once that's done and the empty-clear regression test is confirmed against the new persistence structure, this is in good shape to merge. The in-isolation behaviour, test coverage, and documentation are all solid.

---

Blitzy:
Thanks — the blocker was real and the warning about semantic re-application was exactly right. Merged `develop` and re-applied both intents in `8488e76ba`.

## `saveChapterHighlights` — nothing to re-apply

`develop` already does what this branch's reorder was for: it resolves the account and returns early *before* `writeChapterHighlights` applies the highlight optimistically, so a signed-out user can't end up with a highlight sitting on screen behind the login modal. I took `develop`'s version wholesale and left a comment noting the ordering is load-bearing, since it's now structural rather than visible in one function.

## `unhighlightVerses` — the trap fired as described

Git auto-merged the `coversAnyVerse` block into place *below* `resolveEntryToMutate`, where the login prompt has already fired, and still referencing `current` from the old structure. Confirmed it would have defeated its own purpose.

One adjustment to the suggested shape, though. Reading `getChapterHighlights(...).value` before `resolveEntryToMutate` doesn't await the load, so a read that lands mid-flight reports empty and the early-return skips a clear that should have happened. Awaiting the load first fixes that but introduces a second account resolution across the await — and **that broke `develop`'s own #1564 regression test**, `unhighlightVerses() writes to the account it merged from when the account changes mid-load`. If the account switches during the load, the second resolve picks up the new one and writes the first account's highlights into the second account's record: precisely the bug #1587 exists to prevent.

What works is to resolve once, without prompting, and reuse that entry for the write:

```ts
const userId = login.userId.peek();
if (!userId) {
  return; // signed out: nothing saved, so nothing to clear and nothing to prompt for
}

const entry = getOrCreateEntry(userId, address);
await ensureLoaded(userId, address, entry);

const coversAnyVerse = /* … */;
if (!coversAnyVerse) {
  return;
}
// `entry` is then used for the write — same account it merged from
```

This is `resolveEntryToMutate` minus the login attempt, which is safe to drop here because a signed-out user has no saved highlights to remove.

## Regression test

Good call on the false-confidence risk. The existing `unhighlightVerses() does nothing when the user is not logged in` asserted only that no write happened, which passes against either structure. It now also asserts `login.login` and `getData` are never called, and there's a second test covering the signed-in case where no saved highlight covers the selection.

I checked both fail when the guard is removed rather than assuming they bite.

Blitzy:
> [!NOTE]

## Code review — Repair highlighting in shared sessions (follow-up on `8488e76b`)

This is a re-review after the previous round's blocking issue (a semantic merge conflict with `develop`'s #1587 entry-based highlight persistence) was addressed. Summary of what I independently checked before writing this:

- Fetched the PR branch fresh and ran the affected test files directly against head `8488e76b` (not just trusting CI): `HighlightsManager`, `BibleReader`, `BibleReaderToolbar`, `SessionsManager`, `twitchSubManager` — **165/165 passing**.
- Confirmed there is **no remaining merge conflict** between this branch and `develop` (`git merge-tree` on the current merge-base is clean). GitHub's `mergeable_state: blocked` looks like a required-review gate, not a real conflict.
- Read the rewritten `unhighlightVerses` in `HighlightsManager.tsx` line by line against the account-switch race it has to avoid (#1564): it resolves the account once via `getOrCreateEntry`/`ensureLoaded`, checks verse coverage against that same entry, and reuses that entry for the write — so an account switch mid-load can't cross-write into the wrong account's record. This is what the prior review's blocking comment asked for, and it holds up.
- CI is green on `8488e76b` (build, which runs lint + `check:ts` + `pnpm test`).

**No blocking issues found in this round.**

---

### 🟡 Medium: The two deferred session bugs still have no tracking issues

The PR description candidly lists two bugs found in two-user testing that are intentionally out of scope here ("Not included" section): the expiry timer not resetting on re-highlight, and a second user's highlight being dropped via `decorationOwners`' first-writer-wins behavior. The previous review round asked that these get their own linked tracking issues so they aren't lost once this PR merges and closes #1590 — I searched open issues in the repo for either one and found nothing tracking them yet.

**Suggested fix:** Open two follow-up issues (or one, if preferred) before merging, and link them from the PR description in place of the current prose, so they survive after #1590 closes.

### 🟢 Low: Unrecognized external colour ids still render invisibly (Twitch)

Already an explicitly-documented, accepted trade-off from this branch (transparent fallback beats the old solid-black-bar bug), not a regression — flagging only as a reminder that a Twitch highlight using a colour outside the theme's presets (their own test fixture uses `red`) currently shows nothing to viewers rather than something. Optional follow-up: map unrecognized external `colorId`s to a default preset (e.g. `yellow`) instead of `transparent`.

### 🟢 Nit: `.value` vs `.peek()` inconsistency reading `login.userId`

`saveChapterHighlights` reads `login.userId.value`, while the new guard in `unhighlightVerses` reads `login.userId.peek()`. Functionally equivalent here (neither call site is inside a reactive computed), but worth picking one convention for readability. Not blocking.

---

**Recommendation:** Good to merge from a correctness standpoint — the previous blocker is genuinely fixed and I verified that independently rather than taking the description's word for it. The only open item is process (tracking issues for the two deferred bugs), not code.


---

PR #1595, opened 2026-08-04
Right and center align chapter buttons in the Bible Selector

### LINKED ISSUE #1582: Consider making chapter numbers inline-end aligned when their book is in the right column

AmitPant9999:
fixes: #1582

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
@@ -466,6 +466,7 @@ const SideBarBooks = (props: {
       index: number,
       chapterPos: number,
       separator: number,
+      chapterAtEnd: boolean,
       chapterHint?: number,
       itemStyle?: JSX.CSSProperties,
       narrowChapterStyle?: boolean,
@@ -522,10 +523,13 @@ const SideBarBooks = (props: {
                 style={{
                   ...chapterGridStyle,
                   justifyContent:
-                    ws <= MOBILE_BREAKPOINT ||
-                    bd.numberOfChapters < 4 * separator
+                    ws <= MOBILE_BREAKPOINT
                       ? "flex-start"
-                      : "space-between",
+                      : bd.numberOfChapters < 4 * separator
+                        ? chapterAtEnd
+                          ? "flex-end"
+                          : "flex-start"
+                        : "space-between",
                 }}
               >
                 {narrowChapterStyle && ntColumns === 2 && (
@@ -558,6 +562,8 @@ const SideBarBooks = (props: {
         openVisualRow !== null
           ? { gridColumn: "1 / -1", gridRow: openVisualRow + 2 }
           : undefined;
+      const openVisualCol = columns > 1 ? Math.floor(lbc / rows) : 0;
+      const chapterAtEnd = openVisualCol === columns - 1;
 
       return (
         <div
@@ -570,6 +576,7 @@ const SideBarBooks = (props: {
               index,
               chapterPos,
               columns,
+              chapterAtEnd,
               chapterHint,
               itemStyle,
               narrowChapterStyle,
--- a/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
@@ -563,7 +563,7 @@ const SideBarBooks = (props: {
           ? { gridColumn: "1 / -1", gridRow: openVisualRow + 2 }
           : undefined;
       const openVisualCol = columns > 1 ? Math.floor(lbc / rows) : 0;
-      const chapterAtEnd = openVisualCol === columns - 1;
+      const chapterAtEnd = columns > 1 && openVisualCol === columns - 1;
 
       return (
         <div
--- a/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx
@@ -466,7 +466,7 @@ const SideBarBooks = (props: {
       index: number,
       chapterPos: number,
       separator: number,
-      chapterAtEnd: boolean,
+      chapterColumnPosition: "start" | "middle" | "end",
       chapterHint?: number,
       itemStyle?: JSX.CSSProperties,
       narrowChapterStyle?: boolean,
@@ -526,9 +526,11 @@ const SideBarBooks = (props: {
                     ws <= MOBILE_BREAKPOINT
                       ? "flex-start"
                       : bd.numberOfChapters < 4 * separator
-                        ? chapterAtEnd
+                        ? chapterColumnPosition === "end"
                           ? "flex-end"
-                          : "flex-start"
+                          : chapterColumnPosition === "middle"
+                            ? "center"
+                            : "flex-start"
                         : "space-between",
                 }}
               >
@@ -563,7 +565,12 @@ const SideBarBooks = (props: {
           ? { gridColumn: "1 / -1", gridRow: openVisualRow + 2 }
           : undefined;
       const openVisualCol = columns > 1 ? Math.floor(lbc / rows) : 0;
-      const chapterAtEnd = columns > 1 && openVisualCol === columns - 1;
+      const chapterColumnPosition: "start" | "middle" | "end" =
+        columns <= 1 || openVisualCol === 0
+          ? "start"
+          : openVisualCol === columns - 1
+            ? "end"
+            : "middle";
 
       return (
         <div
@@ -576,7 +583,7 @@ const SideBarBooks = (props: {
               index,
               chapterPos,
               columns,
-              chapterAtEnd,
+              chapterColumnPosition,
               chapterHint,
               itemStyle,
               narrowChapterStyle,


Blitzy:
> [!NOTE]

## Code review — fix 1582

**What this PR does:** Issue #1582 asks that when a book is expanded in the **right/last column** of a multi-column Bible Selector layout, its short chapter list be aligned toward the book (inline-end) instead of the far edge. The PR adds a `chapterAtEnd` flag computed from the expanded book's visual column (`Math.floor(lbc / rows) === columns - 1`) and switches the chapter panel's `justify-content` to `flex-end` in that case, keeping `flex-start` otherwise.

The core idea is sound, and using `flex-end` (a direction-aware flex value) rather than a hard-coded `right` is the correct instinct. There's one behavioral regression worth fixing before merge, plus a couple of smaller notes.

---

### Findings by severity

#### 🟠 Medium — No. 1: Regression for the single-column New Testament layout (tablet widths)

`MOBILE_BREAKPOINT` is `480`, but the NT grid uses `ntColumns = 1` for **every** width below `1200` (see `BibleSelector.tsx:396–408` and the NT render at `BibleSelector.tsx:661`). Only the `ws <= 480` case is caught by the new `ws <= MOBILE_BREAKPOINT ? "flex-start"` guard — the **481–1199px tablet range renders NT as a single full-width column that is not guarded**.

For a single column, the new code computes:

```ts
const openVisualCol = columns > 1 ? Math.floor(lbc / rows) : 0; // → 0
const chapterAtEnd = openVisualCol === columns - 1;             // → 0 === 0 → true
```

So `chapterAtEnd` is `true` for a one-column grid, and any short book (`numberOfChapters < 4 * separator`, i.e. fewer than 4 chapters when `columns === 1`) now renders its chapters **right-aligned (`flex-end`) across the full-width column**. Before this PR those books were `flex-start`. This is the opposite of what the issue wants — it pushes the numbers *away* from the left-aligned book title — and it affects many common books (Titus, 2 Thessalonians, Haggai, Philemon, 2 John, 3 John, Jude, Obadiah, …).

**Reproduction:**
1. Open the branch deploy: https://alpha.seedbible.org/b/feature_issue_1582
2. Resize the browser to a tablet width between **481px and 1199px** (e.g. 800px), so the dual Old/New Testament view renders the NT as one full-width column.
3. Open a short NT book such as **Philemon (1 chapter)** or **Titus (3 chapters)**.
4. Observe the chapter buttons hugging the **right** edge of the column, far from the left-aligned book title. On `develop` they sit at the left, directly under the book.

**Suggested fix** — a single column has no "end column", so exclude `columns <= 1` (`BibleSelector.tsx:565–566`):

```ts
const openVisualCol = columns > 1 ? Math.floor(lbc / rows) : 0;
const chapterAtEnd = columns > 1 && openVisualCol === columns - 1;
```

This restores `flex-start` for the single-column NT while preserving the intended `flex-end` behavior for genuine multi-column layouts.

#### 🟡 Low — No. 2: Books in a middle column remain disconnected

Because the chapter panel spans the whole width (`gridColumn: "1 / -1"`), `justify-content` can only pin it left, right, or spread it. For layouts with 3+ columns (desktop OT = 3, single-testament desktop = 5), a short book in a **middle** column still falls to `flex-start` and renders its chapters under the left edge — the same visual disconnect the issue describes, just for interior columns rather than the last one. This is outside the strict scope of #1582 (which only mentions the right column), so it's fine to defer, but worth calling out: fully solving "chapters appear under the book" for middle columns would need a different mechanism (e.g. an explicit `grid-column`/offset for the panel) rather than `justify-content` alone.

#### 🔵 Info — No. 3: RTL looks correct, but please verify on the deploy

The logic is pleasingly direction-agnostic: in RTL the last logical column (`columns - 1`) is rendered on the visual left, and `flex-end` in an RTL flex row also resolves to the left — so chapters should still land under the book. Since #1582 explicitly calls out RTL alignment, it's worth a quick manual check with an RTL translation (e.g. Hebrew/Arabic) at a multi-column width to confirm the last-column book aligns as expected.

#### 🔵 Info — No. 4: No test coverage for alignment

`test/unit/seed-bible/components/BibleSelector.test.tsx` doesn't assert on `justify-content`, so neither the intended behavior nor the No. 1 regression is caught by the suite. A small render test that expands a short book in a known column and asserts the panel's `justify-content` (covering single-column, first-column, and last-column cases) would lock this in and guard against future breakage.

---

### Recommendation

Address **No. 1** (the one-liner `columns > 1 &&` guard) before merging — it's a real, easily-triggered regression at common tablet widths. Nos. 2–4 are non-blocking; No. 2 in particular is reasonable to leave for a follow-up since it's beyond the issue's stated scope. Nice, minimal implementation otherwise.

---

Blitzy:
@KallynGowdy this fixes what the issue specifically mentioned but the Code Review brought up a good point — Low No. 2 (which i dont actually think is Low 😆) layouts with middle columns get left in the dust, and even the 5 column layouts you can get when you are filtering by Old Testament and New Testament all the columns in the middle are not accounted for.

So like here for instance, Im filtering by New Testament and 3 John looks good, the first chapter is easily reachable.
image

However if I open 2 Thessalonians which is decidedly in the middle, the chapter buttons are all the way on the left, and even if they were `flex-end` aligned it would be the same UX problem.
image

I'm not actually sure what the best UX here would be. Maybe @amir2iqbal447 would have some ideas?

I played around with keeping the chapter containers the width of the column track but i felts that it was doing a disservice to the amount of screen real estate you have in a desktop view and many of the larger books were still making you scroll.

KallynGowdy:
I think we could just center align the buttons?

KallynGowdy:
@AmitPant9999 can you try center aligning chapters for middle columns?

KallynGowdy:
Pushed 6a12aaaf: books opened in a middle column now get `justify-content: center` for their chapter list, instead of falling to `flex-start`. `chapterAtEnd` (boolean) became `chapterColumnPosition` (`"start" | "middle" | "end"`), computed from the same `openVisualCol`/`columns` check as before — first column → `flex-start`, last column → `flex-end`, any column in between → `center`. Single-column layouts still resolve to `"start"`, so the earlier tablet-width fix is unaffected.

`pnpm check:ts` and the existing `BibleSelector.test.tsx` suite (33 tests) pass; there's still no alignment-specific test, per Info No. 4 above.

---

KallynGowdy:
This looks good now. Books in the center column now have their chapter buttons centered when they don't cover a full line.

Testing video: https://youtu.be/UfXwUTH2bKs

Blitzy:
There are some minor cases where it can look off. Mostly when viewing Old testament or New testament only and on a book that is very small. Its worth noting that i consider this to be **minor** and not worth blocking on something that is overall a good UX improvement. Also this only really occurs when filtering for Old/New testament **and** in the full 5 column width layout of the book selector.

image

image

image

Blitzy (approved):

Blitzy:
> [!NOTE]

## Review of head commit `6a12aaaf`

This picks up after the in-thread discussion: the original `chapterAtEnd` boolean regressed the tablet-width single-column NT layout (fixed in `7b9f2bd3`), and the middle-column disconnect Ellery/Blitzy raised is now handled by replacing the boolean with `chapterColumnPosition: "start" | "middle" | "end"`, which center-aligns short books opened in a middle column. I re-verified this from scratch against `BibleSelector.tsx` and `BibleSelectorManager.tsx` on the current head:

- `openVisualCol = columns > 1 ? Math.floor(lbc / rows) : 0` matches the existing `visualCol` formula in `getItemGridStyle`, so the column math stays consistent with how books are actually placed.
- `ghostArray` pads each testament's book list to a multiple of the column count before `rows` is derived, so every column has a uniform `rows` height — no ragged-column edge case that could make `openVisualCol` land in the wrong bucket.
- The `columns <= 1 || openVisualCol === 0 ? "start" : ...` guard correctly keeps single-column (tablet NT, mobile) layouts at `"start"`, so the earlier regression stays fixed.
- `pnpm check:ts` and `pnpm vitest run BibleSelector.test.tsx` both pass clean (33/33) on this commit.

No new correctness bug found in the alignment logic itself. Two things worth flagging:

#### 🟡 Low — Still no automated coverage for the alignment behavior

This is the core behavior the PR (and issue #1582) is about, and it has no test asserting on it — not `justify-content`, not `chapterColumnPosition`. This was already called out as Info No. 4 in the first review pass and again acknowledged in the latest commit message ("there's still no alignment-specific test"), but it hasn't been added across three pushes.

It's cheap to add with the patterns already in this test file: other tests already read inline styles off elements returned by `container.querySelector` (e.g. `.style.display` at `BibleSelector.test.tsx:707`), so a test can expand a short book known to land in the first/middle/last visual column at a multi-column width and assert `justify-content` on the chapter panel's style, without needing real layout/rendering. As-is, a future refactor could silently flip `"start"`/`"middle"`/`"end"` (or reintroduce the tablet regression) and the suite would still report green — which is exactly the "tests that don't actually catch anything" risk worth avoiding here.

#### 🔵 Info — The 5-column "middle" bucket isn't geometrically centered

Small, and Blitzy already flagged + explicitly deemed this non-blocking in the thread, but noting it for the record since it follows directly from the code: the chapter panel always spans the full grid width (`gridColumn: "1 / -1"`), and `chapterColumnPosition` only has three buckets. In the 5-column single-testament layout, columns 1, 2, and 3 (0-indexed) all map to `"middle"` and get centered across the *entire* row — which only visually lines up with the book for column 2 (the true center); a book in column 1 or 3 gets chapters centered further from its own position than a plain `flex-start`/`flex-end` would be. This matches the "looks a little off" screenshots in the thread. No action needed given the team's call, just confirming the root cause for anyone revisiting it later.

---

PR #1596, opened 2026-08-04
Feature/1507

### LINKED ISSUE #1507: Chat on mobile should not automatically open the keyboard
## What is happening

Currently when you open a chat, it automatically focuses the input field and opens the keyboard. This is a somewhat jarring experience compared to how most chat apps work (ie Telegram, WhatsApp).

## What is expected

I expect the chat to open but the keyboard to not open until you tap the input field, following the UX decision from major chat apps like Telegram and WhatsApp. There should still be a blinking cursor in the input field though as a subtle hint to users to type into the input field.

Kabi4:
Fixes #1507


Blitzy:
> [!NOTE]

## Code Review — PR #1596 (Fixes #1507)

**Summary of the change:** In `ChatView.tsx`, the mount-time `useEffect` that auto-focuses the chat input now returns early when `state.app.isMobile.value` is `true`, so opening a chat on mobile no longer pops the soft keyboard. Desktop behavior is unchanged. The test mock gains an `isMobile: signal(false)` field.

This is a clean, minimal, correctly-scoped change and it uses the established `state.app.isMobile` signal the way the rest of the codebase does. The core problem from the issue — the jarring auto-keyboard on mobile — is solved. Two things are worth addressing before merge, plus one nit.

---

### Medium — The "blinking cursor" half of the requirement is not implemented

The issue asks for two things on mobile:

1. The keyboard should **not** open until the user taps the field. ✅ Done.
2. "There should **still be a blinking cursor** in the input field though as a subtle hint to users to type into the input field." ❌ Not done.

By returning early and skipping `inputRef.current?.focus()` entirely, the input is left unfocused — and an unfocused `<input>` renders **no caret at all**. So the subtle "type here" hint the issue explicitly requested is missing on mobile.

This is a genuinely hard trade-off, and it's worth calling out *why*: on mobile browsers, focusing a text input is what summons the soft keyboard, so you generally can't get a caret without the keyboard through a plain `.focus()`. Showing a caret while keeping the keyboard down usually requires a workaround (e.g. temporarily marking the input `readonly`, focusing it, then clearing `readonly` on the first real tap — the "readonly focus" trick), and those workarounds are browser-flaky.

**Recommendation:** Either (a) implement the caret-without-keyboard hint if the team wants to honor the issue as written, or (b) if this simplified "no focus on mobile" approach is the agreed direction, note that on the PR/issue and get sign-off from @Blitzy (issue author) so the dropped requirement is a documented decision rather than an oversight. I'd lean toward (b) being acceptable for now given the fragility of the workarounds, but that's a product call, not a code call.

**Reproduction:**
1. Open the app on a mobile device (or a viewport ≤ 480px wide — see the nit below).
2. Open a chat.
3. _Before:_ keyboard auto-opened. _Now:_ keyboard stays down (good), **but** the input shows no blinking cursor, so there's no visual hint to type.

---

### Low — No test locks in the new behavior

The only test change is adding `isMobile: signal(false)` to `createMockState()`. That keeps the existing suite from throwing on the new `state.app.isMobile.value` read, but it does **not** assert the new behavior — a future refactor could delete the guard and every test would still pass.

**Recommendation:** Add two focused cases against the real branch, e.g.:
- With `isMobile` `false` (desktop), assert the input receives focus on mount (`document.activeElement === input`, or spy on `HTMLInputElement.prototype.focus`).
- With `isMobile` `true` (mobile), assert `focus` is **not** called / the input is not the active element.

The existing helpers already render `ChatView` with a mock state, so this should be a small addition, and it's the piece that guarantees the desktop path never silently regresses.

---

### Nit — `isMobile` is viewport-width based, not device/keyboard based

`isMobile` is `computed(() => viewportWidth.value <= MOBILE_BREAKPOINT)` with `MOBILE_BREAKPOINT = 480`. So a desktop user who narrows their window to ≤ 480px will also lose the auto-focus, even though they have a hardware keyboard and no soft-keyboard problem to avoid. This is consistent with how `isMobile` is used everywhere else in the app, so it's a reasonable choice and I would **not** block on it — just flagging that the guard keys off layout width, not the presence of a soft keyboard, in case a more precise touch/`(pointer: coarse)` check is ever wanted.

---

### Verdict

The change does what it primarily set out to do and is safe to ship in that respect. Before merging I'd want a decision recorded on the missing blinking-cursor requirement (finding No. 1) and, ideally, a small test asserting the desktop-focus path (finding No. 2). The nit is optional.

---

Blitzy (approved):

Issue #1520, opened 2026-07-26
Add more default highlight colors


KallynGowdy:
Currently, we only have 3 default highlight colors. We should have a lot more default colors. At least:

- yellow
- green
- cyan
- red/crimson
- magenta
- pink
- white/cream
- black/gray
- tan/light brown

Kabi4:
Fixedvia PR #1597

PR #1597, opened 2026-08-04
Feature/1520

Kabi4:


Blitzy:
> [!NOTE]

## Code Review — Feature/1520 (Add more default highlight colors)

### Summary

This PR closes [issue No. 1520](https://github.com/HelloAOLab/seed-bible/issues/1520) (add more default highlight colors) and bundles in a fix for the desktop verse toolbar clipping off-screen. Overall it's a **clean, well-structured change** — I found no functional bugs. The notes below are all low-severity polish/accessibility suggestions.

**What changed:**
- **Colors:** Adds `cyan`, `red`, `magenta`, `cream`, `gray`, `tan` to the highlight palette (light + dark themes in `ThemeManager.tsx`, plus matching CSS vars in `base.css`). Combined with the existing `yellow`/`green`/`blue`/`pink`/`purple`/`orange`, the palette now covers every color the issue asked for and more.
- **Single source of truth:** The verse toolbar picker now iterates the exported `DEFAULT_HIGHLIGHT_IDS` instead of a hardcoded `["yellow","green","blue"]`. Because `SettingsPage.tsx` already iterates the same constant, all six new colors automatically appear in Settings too — nice.
- **Layout:** The picker was reworked from a single horizontal row into a 6-column, 2-row swatch grid, with the Back chevron moved into the reference header and a dedicated Add/Clear action row. Keyboard navigation was upgraded from linear (`handleHorizontalListKeyNav`) to 2-D (`handleGridKeyNav`), which is the right call for a grid.
- **Off-screen fix:** New height-measuring (`ResizeObserver`) + clamping logic keeps the floating desktop toolbar (which grows taller with the 12-color grid open) inside the viewport top/bottom.

**Verification performed:** confirmed `handleGridKeyNav` exists and is sound; `viewportHeight` is a real signal (never 0, safe SSR fallbacks); all i18n keys used (`highlight-colors`, `back`, `add`, `clear`, etc.) already exist in `en.json`; no leftover references to the removed `DEFAULT_HIGHLIGHT_COLOR_IDS`; the hidden `color` input is `position:absolute; width:0` so it does **not** consume a cell in the 2-column action grid; and contrast of every new color against its font color looks readable in both themes.

---

### Findings

#### Low — Wider picker can still clip off the left/right screen edge

The vertical clamp is thorough, but the horizontal clamp (`clampedToolbarLeft`) still uses `inset = 84 * scale` as the half-width margin:

```ts
const inset = 84 * uiScale.value;
const x = floatingX.value + verseToolbarOffset.value.dx;
return Math.min(Math.max(x, inset), Math.max(inset, viewportWidth.value - inset));
```

The toolbar is centered via `transform: translate(-50%, -100%)`, so `left` is its center. The color picker is up to `13.5rem` (~216px) wide → ~108px half-width, which exceeds the 84px inset. Near a screen edge the picker can therefore extend ~24px past the viewport.

- **Repro:** On desktop, narrow the browser window, select a verse near the far-right edge, open the color picker → its right edge can clip.
- **Suggested fix:** Use the measured toolbar width (analogous to `verseToolbarHeight`) for the horizontal inset, or bump the inset to cover the picker's half-width when it's open. Low priority — desktop-only edge case, and the vertical case (the reported bug) is fully handled.

#### Low (a11y) — Arrow keys don't reach the Add/Clear buttons

The picker container carries `role="toolbar"`, but arrow-key navigation (`handleGridKeyNav`) is wired only onto the inner `sb-verse-toolbar-swatches` group. The Add/Clear pills live in a separate `sb-verse-toolbar-picker-actions` div with no key handler, so they're reachable by <kbd>Tab</kbd> but not by arrow keys. The WAI-ARIA toolbar pattern expects arrow navigation across all controls. Functional (Tab works), just slightly inconsistent with the role. Consider either extending grid nav to include the action buttons or dropping `role="toolbar"` in favor of the group semantics.

#### Low (a11y/i18n) — Swatch labels expose raw internal IDs, untranslated

Each swatch uses `aria-label={\`Highlight ${colorId}\`}` and `title={colorId}`, so screen readers/tooltips announce raw IDs like "magenta", "tan", "gray" in English only. This pattern pre-dates the PR (it was there for yellow/green/blue), but it's more noticeable now with 12 colors, some non-obvious. A follow-up could map IDs to localized color names. Not blocking.

#### Info — `--sb-highlight-blue-color` value changed in `base.css`

`base.css` changed `--sb-highlight-blue-color` from `#fff59d` (a pale yellow) to `#90caf9` (actual blue), which now matches `ThemeManager.tsx` (already `#90caf9`). Per `CLAUDE.md`, `ThemeManager`'s body-scoped values win over `base.css`'s `:root`, so rendered highlights were almost certainly already blue and this just removes a stale/misleading value — a good consistency fix. Flagging only so a human can confirm nothing reads that specific `:root` var directly and that existing "blue" highlights look as expected.

---

### Non-blocking notes

- **Scope:** The PR does more than "add colors" — it also refactors the picker layout, changes keyboard nav, and fixes the off-screen toolbar. That's reasonable (12 colors don't fit the old single-row layout), but the PR body is empty and the title (`Feature/1520`) isn't descriptive. A short description + a `Closes #1520` line would help reviewers and auto-close the issue.
- **Tests:** These are UI/CSS changes with no test coverage, which is consistent with the surrounding code; no action needed, just noting.

### Verdict

Meets the issue's requirements cleanly, with good reuse of `DEFAULT_HIGHLIGHT_IDS` as a single source of truth. No blocking issues — the items above are polish. Nice work. 👍

---

KallynGowdy:
Looks pretty good!

### Dark Theme

image

### Light Theme

image

KallynGowdy:
@Kabi4  However, on mobile the new options are pretty small and there's a lot of extra padding that doesn't need to be there.

image

We should try to make it more like YouVersion, where there's a single scrollable list of colors and each of the colors are larger and easier to select.

image

Kabi4:
image


Once Swiped

image

Blitzy:
@Kabi4 remember to not add translations for anything but `en.json`. @KallynGowdy or myself will run a Google Cloud translation pass on the localization to update them.

Blitzy (approved):
This looks good. I like the new mobile color picker for highlights, swiping left and right feels nice and responsive.

I did find a number of issues with the general UX of color picking itself lackluster but these all pre-date the changes of this PR and so I'm not going let those block it. I do think we need to in general go with a unified color picking component for Seed Bible but I've logged that issue here #1602

Issue #1522, opened 2026-07-26
The Bible Selector should scroll the currently chapter to the center of the screen


KallynGowdy:
Currently, the Bible Selector only scrolls the chapter into the screen. This means that the current chapter will be either at the bottom of the screen or at the top of the screen. This can make it more difficult for users to select a different chapter in the same book. Instead, we should scroll the chapter to the center of the screen so that surrounding chapters are always visible.

### what it looks like
![Screenshot_20260726-103700.png](https://github.com/user-attachments/assets/e95c59b5-444e-481c-8f49-7450137c21e1)

### what it should look like
![Screenshot_20260726-103745.png](https://github.com/user-attachments/assets/39160e67-eb24-4ca9-8b21-e1cb01711ec4)

Kabi4:
This is done via PR #1598

PR #1598, opened 2026-08-04
Feature/1522

Kabi4:


Blitzy:
> [!NOTE]

## Code Review — Feature/1522 (center the current chapter in the Bible Selector)

**Scope:** one file, `BibleSelector.tsx` — the `scrollTargetInto` helper inside the selector's scroll-into-view effect (+11 / −7). This implements issue #1522: scroll the current chapter to the **center** of the panel instead of just barely into view, so neighboring chapters stay visible.

**Verdict:** the core change is correct and solves the issue well. The centering math is sound and the reliance on `scrollTo` clamping to handle the first/last chapters is the right approach. I have no blocking issues — one low-severity behavior caveat around the fallback target, plus a couple of nits.

For context, here is the new logic:

```js
const targetFits = targetRect.height <= scrollerRect.height;
let delta = 0;
if (targetFits) {
  const targetCenter = targetRect.top + targetRect.height / 2;
  const scrollerCenter = scrollerRect.top + scrollerRect.height / 2;
  delta = targetCenter - scrollerCenter;
} else if (targetRect.top < scrollerRect.top) {
  delta = -(scrollerRect.top - targetRect.top + 8);
}
if (Math.abs(delta) > 1) { scroller.scrollTo({ top: scroller.scrollTop + delta, behavior: "auto" }); }
```

I verified the sign of `delta` (target below center → positive delta → `scrollTop` increases → content scrolls up → target rises toward center: correct) and confirmed that `scrollTo` clamps `top` to `[0, scrollHeight − clientHeight]`, so Genesis 1 and Psalm 150 land "as close to center as possible" without opening a gap — matching the issue's requested behavior. The non-fitting branches are unchanged from before, so there is no regression in the tall-grid path.

---

### Low — fallback target (whole chapter grid) can now hide the book title when it *fits*

**Where:** `BibleSelector.tsx`, the `if (targetFits)` branch.

**What:** The code comment states the intent is *"For a tall fallback (e.g. full chapter grid) only pin the top — centering would hide the book title above the grid."* But that pin-top only lives in the `else if` branch, which runs when the target **doesn't** fit. When the target is the fallback `chapterPanel` (used when there is no visible active-chapter button — i.e. the open book is not the current reading book) and the grid **does** fit inside the scroller while `title + grid` overflows, `targetFits` is `true`, so the grid gets **centered**, pushing the book title off the top. Before this PR, on initial open the grid's top was at/below the scroller top, `delta` stayed `0`, and the title remained visible.

Net effect: opening a *non-current* book whose chapter grid is shorter than the panel but whose title+grid overflows will now scroll its title out of view. This only affects the fallback path, not the primary "center the current chapter" feature, so severity is low — but it contradicts the code's own stated intent.

**Reproduction (best-effort — depends on viewport/book proportions):**
1. Open the Bible Selector while reading, say, Genesis.
2. Browse to a different book that is *not* your current reading book and has a moderate chapter count, on a viewport short enough that the book title + chapter grid overflow the panel but the grid alone fits.
3. Observe the grid centering with the book's title scrolled out of view at the top, instead of the title staying pinned.

**Suggested fix:** only center when the target is an actual chapter button; keep pin-top for the panel/book-tab fallback. For example, gate centering on whether a real chapter button was found:

```js
const isChapterTarget = target === visibleChapterButton;
if (targetFits && isChapterTarget) {
  // center
} else if (targetRect.top < scrollerRect.top) {
  // pin top (+8)
}
```

(`visibleChapterButton` is already computed just above the target selection, so this is a cheap check.) If browsing non-current books is considered out of scope, a one-line note in the comment clarifying that the fitting-grid case intentionally centers would also resolve the mismatch.

---

### Nits (non-blocking)

- **Commit message typo:** `Fixes Scroll to nearest possible Egdes or center` → "Edges". Cosmetic, but worth a quick amend before merge if you like a clean history.
- **No PR description / issue link:** the PR body is empty. Adding `Closes #1522` would auto-close the issue on merge and give reviewers the before/after screenshots for free.
- **Testing:** there is no automated coverage for this scroll math, which is understandable — jsdom does no layout, so `getBoundingClientRect()` returns zeros and the centering can't be exercised in the existing `BibleSelector.test.tsx`. Please confirm manual verification on the alpha deploy (https://alpha.seedbible.org/b/feature_1522) across both the desktop `.books-item` scroller and the mobile nested `.books-container` / `.sidebar-results` scrollers, and for a large book like Psalms (the auto-expanding group path).

---

### CI

At review time the `build (24.x)` check was still **in progress** and the PR shows `mergeable_state: blocked` — that appears to be the pending build plus the requested review, not a merge conflict (the branch is a clean fast-forward over `develop`). Worth a glance once the build settles.

Nice, tightly-scoped change overall — the centering behavior is exactly what the issue asked for.

---

KallynGowdy:
@Kabi4 Currently, this runs every time that the user expands a book in the book selector, which can be pretty jarring. Instead, it should only try to center the book when first opening the book selector, not when expanding a book while the book selector is already open. We want the scrolling behavior to be exactly like `develop` when the user is clicking on books to expand them and see chapters (where it scrolls the chapters into view if they aren't in view, but otherwise it doesn't do anything to the scroll position), but when they open the book selector we want their current book (and chapter) to be in the center of the screen.

Blitzy (approved):

PR #1599, opened 2026-08-04
feat: Let users follow each other and see their reading data

KallynGowdy:
Adds a follow graph so a user can subscribe to other people's highlights,
playlists, and reading history.

All three streams were already stored world-readable under the author's own
record (`recordName = <userId>`, `publicRead*` markers), so the read side
needed almost nothing — what was missing was any notion of *whose* data to
read besides your own. That gap had four features stubbed out waiting on it,
all of which now work:

- `today-screen`'s Community section, whose `SubscribedUsersProvider`
  returned an empty list into fully-built UI
- `scripture-map`'s reading-history timeline, which could only ever see
  users who were online in a session
- `seed-bible-utils`' `UserDatabase.getSubscribedUsers()`
- `InvitationsManager`, commented out in full with a TODO asking for exactly
  this ("re-enable when we can only listen for entries from users the current
  one follows")

Design:

- Following is asymmetric and needs no approval, matching how the data
  already behaves — nothing is written to the followee's record.
- The follow list is a single blob at address `following` in the follower's
  own record, under the `private` marker. It is the one artifact here with no
  reader but its owner.
- Each entry carries a denormalized name/picture snapshot so the list renders
  on first paint; `refreshProfiles` re-reads them.
- Discovery is a `?follow=<userId>` share link (confirmed via a modal, never
  automatic) and a Follow button on shared-session participants.
- `HighlightsManager.getUserChapterHighlights` reads a named account's
  highlights. Its entry cache was already keyed by account; explicit entries
  are exempted from the account-switch eviction sweep so a sign-in doesn't
  discard followed users' data.
- `PlaylistManager.listPlaylistsForUsers` takes ids rather than the manager,
  so playlists stay unaware that following exists.

Gated behind two PostHog flags: `following`, and `followed-sessions` for the
shared-session registry, which opens a shared document on every client and so
needs to be switchable off on its own.

Followers see everything; there are no per-stream sharing toggles in this
change.

Claude-Session: https://claude.ai/code/session_01BsA8PBufWnPkRAKhCMYW8Z


Blitzy:
> [!NOTE]

## Code review — Let users follow each other and see their reading data

Nice, well-documented change. The design is coherent: following is a purely local statement over data that was already world-readable, the stale-response guards in `FollowsManager` (`loadFollows`/`persist`/`refreshProfiles`) are careful about account switches, and the four previously-stubbed consumers are all wired through thin adapters that don't have to know following exists. Test coverage for the new managers and adapters is solid.

There is one **blocking** problem (CI is red on Node 24) and two functional gaps around the feature flags that I'd want fixed before merge. Details below, most severe first.

No linked issue was referenced in the PR body, and there are no prior review comments to reconcile — the only existing comment is the deployment bot.

---

### 🔴 High — CI `build (24.x)` fails: 78 unhandled `Event` errors introduced by this PR

**What happens:** The `build (24.x)` job fails (both attempts, so not flaky). Every one of the 3758 assertions **passes**, but Vitest reports `78 errors` and exits 1:

```
TypeError: The "event" argument must be an instance of Event. Received an instance of Event
Serialized Error: { code: 'ERR_INVALID_ARG_TYPE' }
This error originated in "test/unit/seed-bible/managers/SeedBibleStateManager.test.ts"
... and "test/unit/seed-bible/components/BibleReaderToolbar.test.tsx"
```

This is the Node 24 + jsdom cross-realm `Event` incompatibility: an async callback dispatches an event **after** the test's jsdom window has been torn down, and Node 24's stricter `Event` brand-check rejects it.

**Why it's this PR's fault (attribution):**
- The PR's exact base commit `43a29f37` (Merge No. 1587) **passes** the same Node 24 `ci.yml` run.
- PR No. 1599 on top of it **fails** with these 78 errors.
- The errors are attributed to `SeedBibleStateManager.test.ts` and `BibleReaderToolbar.test.tsx` — the tests that build the full app state, whose **source** this PR changed. `createSeedBibleState` now kicks off new construction-time async work that these tests trigger and never await:
  - `createFollowsManager`'s `effect()` → `loadFollows(userId)` → `os.getData(...)`
  - `createInvitationsManager` → `void openRegistry()` → `os.getSharedDocument(...)` (in DEV/test the `followed-sessions` flag reads as enabled, so this actually runs)
  - `void setupInitialFollow()`

  One or more of those promises settles after the test's environment is gone, producing the late event dispatch.

**Why it looked green locally:** the project requires Node `>=24.15.0` (`package.json` `engines`), but on Node 22 the full suite passes with **zero** errors — I verified this. The failure only reproduces on the Node version CI actually uses. (I could not run Node 24 in this environment, so I could not isolate the exact throwing line; the evidence above is from the CI logs and the base-vs-head CI comparison.)

**Reproduction:**
1. `nvm use 24` (or any Node ≥ 24.15.0) — this is the required engine.
2. `pnpm install && pnpm test` (the same coverage run CI uses).
3. Observe `Tests 3758 passed` alongside `Errors 78`, exit code 1.
4. `git checkout 43a29f37 && pnpm test` → clean, for contrast.

**Suggested fix:** Reproduce on Node 24, then make the new construction-time async paths tear down cleanly — e.g. give `FollowsManager`/`InvitationsManager` disposers that abort in-flight `loadFollows`/`openRegistry` work (and have the tests dispose the state + flush microtasks), or guard the callbacks that fire after teardown (`if (disposed) return;`) so nothing dispatches into a dead jsdom window. `InvitationsManager` already has a `disposed` guard for its document open path; the follow-list effect and `setupInitialFollow` don't have an equivalent.

---

### 🟠 Medium — followed-sessions registry never opens when the PostHog flag loads after construction

**Where:** `InvitationsManager.tsx` (`openRegistry` / the auth effect) + `FeaturesManager.tsx`.

`openRegistry()` bails early on `!isEnabled()`, and it's only invoked once at construction (`if (typeof window !== "undefined") void openRegistry();`) plus inside `publishSession`. But PostHog flags load **asynchronously** — `createFeaturesManager` starts with `flags = signal<string[]>([])` and only fills them in the `onFeatureFlags` callback. At the moment `createInvitationsManager` runs, `followed-sessions` reads as **false** in production, so `openRegistry` returns immediately. When the flag later flips true, nothing re-invokes `openRegistry`: the `stopAuthEffect` effect watches `login.userId` and `follows.following` but not the feature flag, and only calls `applyEntries` when `registryMap` already exists (it never will).

**Effect (before/after):** The headline "when someone you follow creates a shared tab, it auto-appears here" won't work on a normal load in production. The registry only ever opens if the user themselves hosts a session (which calls `openRegistry` again). This passes in DEV only because `isFeatureEnabled` is hardcoded to `true` when `import.meta.env.DEV`.

**Reproduction:** In a production build with `followed-sessions` enabled in PostHog (not bootstrapped into the initial payload), load the app while someone you follow is already hosting a shared session. Their session never appears in `availableSessions` until you publish one of your own.

**Suggested fix:** Make the enabled-flag reactive and open on transition. In the existing effect, read the flag so the effect re-runs when it changes, and open the registry when it becomes enabled:

```ts
const stopAuthEffect = effect(() => {
  void login.userId.value;
  void follows.following.value;
  const enabled = isEnabled(); // reads the feature signal -> effect re-runs when it flips
  if (enabled && !registryDoc && typeof window !== "undefined") {
    void openRegistry();
  }
  if (registryMap) {
    applyEntries(readStoredEntries());
  }
});
```

(Have `enabled` read the signal directly rather than through a `() => ....value` thunk that the effect can't track.)

---

### 🟠 Medium — the `following` flag only hides one entry point; the feature stays reachable when it's off

**Where:** `FEATURE_KEY_FOLLOWING` is checked in exactly one place — the "Following" toolbar tool in `BibleToolsManager.tsx:731`.

Everything else that reads or writes the follow graph is **ungated**:
- `setupInitialFollow` (the `?follow=<userId>` share link) opens the confirm prompt and follows regardless of the flag (`SeedBibleStateManager.tsx:1430`).
- `ParticipantFollowButton` in `SessionParticipants.tsx` always renders.
- The read surfaces — Today's Community section (`SubscribedUsersProvider`), the scripture-map timeline (`useReadingHistoryProvider`), and `UserDatabase.getSubscribedUsers` — all read `follows.following` unconditionally.

So with `following` **off**, a user can still follow via a share link or a session, and their followed data still surfaces on the Today screen and the map. The flag hides the management pane but does not actually turn the feature off, which is probably not the intent given the PR describes it as the feature's gate.

**Suggested fix:** Gate the write/entry paths (and ideally the read surfaces) on `FEATURE_KEY_FOLLOWING` too — short-circuit `setupInitialFollow` when the flag is off, hide `ParticipantFollowButton`, and have the adapters return empty when the flag is off. A single shared `canFollow` computed passed into these call sites would keep it consistent.

---

### 🟡 Low — miscellaneous

1. **Follow-from-session is mobile-only.** `ParticipantFollowButton` is added only to `MobileSessionParticipants`. The desktop participant lists in `Tabs.tsx` (~lines 433 and 568) render the same participants with `getUserDisplayName(user)` but get no Follow button. If "follow from a shared session" is meant to be a general discovery path, desktop users miss it. Confirm whether that's intentional.

2. **Unbounded cache growth for followed users' highlights.** In `HighlightsManager`, `explicit` entries and `userViews` are deliberately exempt from the account-switch sweep and are never pruned. That mirrors the existing `views` behavior, but it's now multiplied by (followed users × chapters browsed). Over a long session this only grows. Consider a bound / LRU for the explicit path.

3. **Un-internationalized fallback name.** `SubscribedUsersProvider.getUserProfile` hardcodes `` `User ${id.slice(0, 8)}` `` instead of using a translation key. The app already has `follow-unnamed-user` for exactly this. Minor, and it's in the `today-screen` package where i18n wiring may differ — worth a look.

4. **Possible lost update on concurrent follows.** In `FollowsManager.follow`, the duplicate/`MAX_FOLLOWS` check and the final `persist([...follows.peek(), entry])` straddle an `await login.getUserProfile(userId)`. Two follows of different accounts fired in the same tick could each capture the pre-persist list, and the later `persist` would drop the earlier entry. Very unlikely from the UI (one button, disabled while busy), but a serialized mutation queue would remove the race entirely.

---

### ✅ Things that look good
- The account-switch stale-response guards in `loadFollows`, `persist` (identity-checked rollback), and `refreshProfiles` (version + identity re-check before writing) are careful and correct.
- Storing the follow list under the `private` marker with a denormalized name/picture snapshot is a sensible call, and the privacy model is clearly documented.
- Good adapter boundaries — `PlaylistManager.listPlaylistsForUsers` taking ids, `UserDatabase`/`SubscribedUsersProvider` as thin passthroughs — and genuinely good test coverage for the new code.

---

Blitzy:
> [!NOTE]

## Code review — Let users follow each other and see their reading data (re-review)

This is a follow-up after three commits landed since the Aug 4 review: two CI fixes and one flag removal. All three are narrowly scoped and each resolves a specific concern from that review. No linked issue is referenced in the PR body, so I'm reviewing against the earlier pass and the PR description.

I read the actual diffs for `fd11d2675` (registry open-gating fix), `8f65d6bf` (test mock), and `e0a4fe046` (flag removal) against the previously-reviewed commit, plus the current state of `InvitationsManager.tsx`, `FeaturesManager.tsx`, and `BibleToolsManager.tsx`.

### 🔴 → ✅ Resolved: CI was red on Node 24 (78 unhandled `Event` errors)

The root cause turned out to be different from what I guessed at in the first pass, but the fix is correct: `InvitationsManager` was unconditionally opening a live WebSocket (`getSharedDocument`) to the shared-sessions registry any time `window` existed — true in jsdom too — so every test constructing a `SeedBibleState` kicked off a real connection attempt to `wss://auth.seedbible.org` that failed asynchronously under Node 24's stricter `undici`, surfacing as uncaught exceptions attributed to whatever test happened to be running.

The fix gates opening behind `signed in AND following ≥1 account AND the `followed-sessions` flag`, which also happens to mean the overwhelming majority of tests (signed out, no follows) never touch the network. `publishSession()` still force-opens the registry regardless of follows (correctly — a host doesn't need followers to publish), and the one test file that exercises that path now mocks `InvitationsManager` the same way its sibling managers already are in that file. Both `build (24.x)` check runs are green.

### 🟠 → ✅ Resolved: followed-sessions registry not reactive to the async-loading flag

The original finding was that `openRegistry()` only ran once at construction, so a PostHog flag that resolved *after* construction would never retrigger it. The new code moves the open call inside the existing `effect()` and adds the enabled-check there too:

```ts
const stopAuthEffect = effect(() => {
  const userId = login.userId.value;
  const hasFollows = follows.following.value.length > 0;
  if (registryMap) {
    applyEntries(readStoredEntries());
  } else if (typeof window !== "undefined" && userId && hasFollows && isEnabled()) {
    void openRegistry();
  }
});
```

I traced `isEnabled` back to its call site in `SeedBibleStateManager.tsx`: `enabled: () => features.isFeatureEnabled(FEATURE_KEY_FOLLOWED_SESSIONS).value`. Because that `.value` read happens synchronously inside the effect body (even though it's one function call away), Preact Signals tracks it as a dependency of this effect just like `userId` and `hasFollows` — so the effect genuinely re-runs and opens the registry the moment the flag flips true, not just on account/follow changes. This is the right fix.

### 🟠 → ✅ Resolved (by removal): the `following` flag only gated the toolbar entry point

Rather than gate the remaining ungated paths (share-link follow, the follow button on session participants), `e0a4fe046` removes `FEATURE_KEY_FOLLOWING` entirely, so the toolbar tool is now visible unconditionally (`isVisible: (context) => !!context.follows && !!context.login`). That's a reasonable call if the intent is a full launch — it removes the inconsistency I flagged rather than papering over it, and I confirmed there are no orphaned `FEATURE_KEY_FOLLOWING` references left anywhere in `packages/seed-bible` (i18n key, storage key, and UI copy all still legitimately use the plain string `"following"` for unrelated purposes). Worth a quick gut-check with product/Ryan that shipping to everyone (vs. a staged flag rollout) is actually intended, since that's a product decision, not a code defect.

### Still open from the last review (unchanged by these three commits, all non-blocking)

None of these files were touched since Aug 4, so they carry over as-is:
- Follow-from-session button is mobile-only (`ParticipantFollowButton` isn't wired into the desktop participant lists in `Tabs.tsx`).
- `HighlightsManager`'s `explicit`/`userViews` cache entries for followed users are exempt from the account-switch sweep and never pruned — unbounded over a long session.
- `SubscribedUsersProvider.getUserProfile`'s fallback name is hardcoded English instead of using the existing `follow-unnamed-user` i18n key.
- A narrow window in `FollowsManager.follow` where two concurrent follows could race past the `await login.getUserProfile(userId)` and the later `persist` could drop the earlier one.

None of these are new, and none block this PR — flagging just so they don't get lost.

### Test quality

The new `InvitationsManager.test.ts` (480 lines, previously untested code) mocks only at the `OsManager` boundary (`os.getSharedDocument` is spied/mocked; nothing deeper), matching this repo's testing convention. More importantly, it's genuine regression coverage: "does not open while signed out, even with follows" and "does not open while signed in but following nobody" would both fail against the pre-fix code that opened unconditionally on any `window`-defined environment — these aren't tests that pass regardless of the fix. The `SeedBibleStateManager.test.ts` addition mirrors the existing sibling-manager mocking pattern already used in that file rather than introducing a new approach.

### Process note

`mergeable_state` is still reported as `blocked` (not `dirty`/conflicting) — that generally means a required check or review approval is outstanding rather than a merge conflict, so no branch work should be needed here.

---

KallynGowdy:
@Blitzy This one is a larger change with a hodgepodge of things, but it's a good start for a lot of the today screen improvements that we want to make (like inviting people to your circle)

KallynGowdy:
## Recent changes

Two commits landed since the last review pass:

### `feat: Add the ability to display reading plans from people you follow` (cd0cdf55b)

Adds a "Reading plans from people you follow" section to the Reading Plans pane, mirroring the pattern already shipped for playlists/annotations. For each followed account, it shows the reading plan(s) they're currently progressing through, with their own day/streak progress, and lets you open one and start it yourself.

The interesting part: it resolves a followed user's progress to its plan **regardless of who authored it** (not just plans they wrote themselves), via a new `parseReadingPlanId` — the inverse of the existing `formatReadingPlanId`. A `ReadingPlanProgress.planId` is always `` rp_${recordName}_${address} ``; since plan addresses are always `plan_${uuid()}` (so they always contain their own underscore) while `recordName` never does, splitting on the *first* underscore after the `rp_` prefix losslessly recovers both parts. No schema changes needed — everything was already stored world-readable.

New manager APIs: `getUserReadingPlanProgresses(userId)` and `getReadingPlanByLocator(recordName, address)`, both cached per-key (mirroring `PlaylistManager.getUserPlaylists`). Starting a followed user's plan for yourself, and opening its detail view, needed zero changes — `startReadingPlan`/`selectReadingPlan`/`openPlanDetail` were already author-agnostic.

### `feat: Get rid of the FollowingPane and display it all in the DiscoverPane instead` (3f46d80cb)

Folds the standalone "Following" pane (the followed-accounts list, share-my-link CTA, unfollow) into a new bottom section of `DiscoverPane`, and removes it as its own toolbar destination. Now that playlists, annotations, and reading plans of people you follow all live in Discover, the follow list itself joins them there instead of being a separate menu entry.

- `FollowingPane.tsx` is deleted; its logic (`FollowingSection`/`FollowRow`/`getFollowUrl`) is ported into `DiscoverPane.tsx`.
- `FollowingPane.css` is split: the `.sb-follow-prompt*` rules move to a new `FollowPrompt.css` (still needed by the unrelated `?follow=<userId>` deep-link dialog), the rest moves into `DiscoverPane.css`.
- The `open-following` toolbar registration and its import are removed from `BibleToolsManager.tsx`.
- `DiscoverPane.test.tsx`'s mock `SeedBibleState.follows` fixture gained `unfollow`/`refreshProfiles`/`isLoading` (needed since the new section now renders in every test that mounts `DiscoverPane`), plus new tests for the signed-out/empty/list/unfollow/share-link behavior.

Both commits pass `check:ts`, `lint`, and their respective test suites (`ReadingPlansManager.test.ts`, `DiscoverPane.test.tsx`).

PR #1601, opened 2026-08-04
fix: Start a shared session at the chapter you're reading (#1589)

### LINKED ISSUE #1589: Starting a shared session resets reading position instead of keeping current location
## What is happening

While reading Luke 21, starting a shared session (Share > Start shared session) jumped the reading position all the way back to Genesis 1, instead of starting the session from the chapter that was currently open. The user had to manually navigate back to Luke 21 after starting the session.

_Evidence:_ Loom walkthrough (timestamp 0:21): https://www.loom.com/share/b24a81faa78e44ad943ccc58585a23ac

## What is expected

Starting a shared session should keep the current reading position. If you start a session while reading Luke 21, the session should begin at Luke 21, not reset to Genesis 1.

Blitzy:
Closes #1589

## What was happening

Starting a shared session from Luke 21 dropped you at Genesis 1. You had to navigate all the way back, and anyone who joined in the meantime landed on Genesis 1 with you.

A shared session builds its own reader, separate from the tab you were reading, and it was being built with no starting position at all. With no book to open, the reader falls back to the first book of the translation, which is Genesis. Nothing corrected it afterwards: the session's shared position starts empty for a brand new session, so there was nothing to pull the reader back to where you were. A moment later that Genesis 1 position was published as the session's official position for everyone else.

## What happens now

The session opens at the active tab's translation, book and chapter. Start a session while reading Luke 21 and the session tab opens on Luke 21.

The position is handed to the session's reader when it is built, rather than navigated to a moment later. That distinction matters: navigating afterwards would still load Genesis first and then jump, which flickers and wastes a chapter download. This mirrors what the Bible selector already does when it opens a chapter in a new tab.

The starting position is also written into the session's shared state immediately, instead of waiting for the normal publish delay. Sessions are auto-published to other signed-in users the instant they are created, so there was a window where a joiner arriving early found no position, settled on Genesis 1 themselves, and published that back over the host.

## Not included: the selected verse

The original plan was to also open the session at the verse you had selected. It is left out, because a verse seeded into a new session does not survive to the rendered tab. The session still opens at the top of the chapter regardless, so shipping it would have meant carrying a setting that quietly does nothing.

I found and fixed one cause along the way (the empty shared position was wiping the reader's scroll target on the first sync) and confirmed it with a failing test, but it was not the only cause, so that fix was reverted rather than left in as an unmotivated change to session sync. Happy to raise it as a separate issue.

## Testing

Four new unit tests:

- `SessionsManager`: the session's reader is built at the given position; that position reaches the shared state without waiting out the 150ms publish debounce; and creating a session with no position writes nothing.
- `SeedBibleStateManager`: the regression itself, with two tabs on different chapters so that reading the position off the wrong tab cannot pass by accident.

Full suite passes (3725 tests). Verified by hand in the browser: reading a chapter, starting a session, and confirming the new tab opens on the same chapter and translation, with a second browser joining via the copied link landing there too.

## Notes for the reviewer

The eager write to the shared state sits deliberately *after* the initial sync in `createBibleReadingSession`. Written any earlier, that sync reads our own position back out and re-navigates the reader to the chapter it is already on, which costs a browser history entry. It is also guarded on the shared state having no book yet, matching how `hostUserId` is claimed at creation.

`SessionStartPosition` is chapter-level by design and documents why, so the verse cannot be reintroduced by accident while it still has no effect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### 🐛 Fixed
 
+- Fix starting a shared session dropping you back at Genesis 1 instead of opening at the chapter you were already reading, in your translation.
 - Fix highlights from a previous account staying visible on already-visited chapters after signing out and into a different one, instead of updating to the signed-in account immediately. ([#1587](https://github.com/HelloAOLab/seed-bible/pull/1587))
 - Fix a highlight added as your session was ending being saved to whichever account signed in next, overwriting that account's highlights for the chapter. ([#1587](https://github.com/HelloAOLab/seed-bible/pull/1587))
 
--- a/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
@@ -9,6 +9,7 @@ import {
 import {
   bibleLanguageToUiLocale,
   uiLocaleForDefaultTranslation,
+  type BibleReadingState,
 } from "../managers/BibleReadingManager";
 import { buildReadingPath } from "../managers/ReadingUrlPath";
 import type { OfflineTranslationStore } from "../managers/OfflineTranslationStore";
@@ -71,6 +72,7 @@ import {
   isSessionHost,
   type BibleReadingSession,
   type SessionsManager,
+  type SessionStartPosition,
 } from "../managers/SessionsManager";
 import {
   createAnnotationsManager,
@@ -211,7 +213,10 @@ export interface AppState {
   selectPane: (paneId: string) => void;
   /** Closes any pane filling the reader. */
   closeFullscreenPanes: () => void;
-  /** Creates a shared reading session and opens it in a new tab. */
+  /**
+   * Creates a shared reading session and opens it in a new tab, starting from
+   * the active tab's translation, book and chapter.
+   */
   createSharedSession: () => Promise<BibleReadingSession>;
   /** Joins an existing shared session and opens it in a new tab. */
   joinSharedSession: (id: string) => Promise<BibleReadingSession>;
@@ -397,6 +402,17 @@ export interface CreateSeedBibleStateOptions {
    * IndexedDB; tests pass an in-memory store, and null disables the feature.
    */
   offlineStore?: OfflineTranslationStore | null;
+}
+
+/** Where a shared session started from this reading surface should open. */
+function getSessionStartPosition(
+  readingState: BibleReadingState
+): SessionStartPosition {
+  return {
+    initialTranslationId: readingState.translationId.value,
+    initialBookId: readingState.bookId.value,
+    initialChapterNumber: readingState.chapterNumber.value,
+  };
 }
 
 export function createSeedBibleState(
@@ -1301,7 +1317,13 @@ export function createSeedBibleState(
 
   const handleCreateSharedSession = async () => {
     closeSidebarAndSettings();
-    const session = await sessions.createSession();
+    // Start the session where the user is reading, not at the default book.
+    const activeReadingState = selectedTab.value?.readingState ?? null;
+    const session = await sessions.createSession(
+      activeReadingState
+        ? getSessionStartPosition(activeReadingState)
+        : undefined
+    );
     if (typeof posthog !== "undefined" && posthog) {
       posthog.capture("create_session", {
         sessionId: session.id,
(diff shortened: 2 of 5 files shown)


Blitzy:
> [!NOTE]

## Code review — PR #1601 (Closes #1589)

I read the linked issue, the PR description, the diff, and the surrounding session-sync code, then ran the linter, the type-check, and the two affected test files locally against this branch.

**Short version:** the fix itself is correct and the reasoning in the description holds up. There is one blocker, and it is not in the fix — it is a stray leftover in a test file that will fail CI lint.

---

### 🔴 High — will fail CI (blocks merge)

**No. 1 — Unused `selectedVerse` helper in the state-manager test fails `pnpm lint`.**

`test/unit/seed-bible/managers/SeedBibleStateManager.test.ts` defines a `selectedVerse(...)` helper (lines 124–135) that nothing calls. It is left over from the selected-verse test removed in the last commit ("removed createSharedSession selected verse test"). The repo's ESLint config sets `@typescript-eslint/no-unused-vars` to `error`, and CI runs `pnpm lint` **before** the tests (`.github/workflows/ci.yml`), so the `build (24.x)` check will stop at the Lint step before it ever gets to the suite.

Reproduction:

```bash
pnpm install
pnpm exec eslint test/unit/seed-bible/managers/SeedBibleStateManager.test.ts
# test/unit/.../SeedBibleStateManager.test.ts
#   124:10  error  'selectedVerse' is defined but never used  @typescript-eslint/no-unused-vars
# ✖ 1 problem (1 error, 0 warnings)
```

Suggested fix — delete the dead helper (lines 124–135):

```diff
-function selectedVerse(
-  bookId: string,
-  chapterNumber: number,
-  verseNumber: number
-) {
-  return {
-    bookId,
-    chapterNumber,
-    translationId: "AAB",
-    verse: { type: "verse", number: verseNumber, content: [] },
-  } as any;
-}
-
```

With that one function removed, the branch is green locally for me: `pnpm check:ts` reports 0 visible errors, and the two touched test files pass (91 tests).

---

### 🟢 Low / optional

**No. 2 — A partially-loaded active tab produces a partial start position.**

`getSessionStartPosition` (`SeedBibleStateManager.tsx`) reads `translationId.value`, `bookId.value`, and `chapterNumber.value` straight off the active tab. If a session is started before the tab has fully settled — book chosen but chapter still `null`, say — the start position is written with a `null` chapter (or `null` book). The code already degrades gracefully in that case: the eager write in `createBibleReadingSession` is guarded on `startPosition?.initialBookId`, a joiner's `canLoadSessionData` rejects a position with a `null` field and falls back, and the seeded reader defaults the missing piece. So there is no bug here — worst case is the pre-existing "starts at the default" behavior for that one edge. Flagging only so it is a known, deliberate trade-off rather than an accident. No change required.

---

### What I checked and found solid

- **Seeding at construction vs. navigating after.** Handing the position to `createBibleReadingState` (rather than calling `selectTranslationAndChapter` later) is the right call and matches `addTab`'s `initialReadingOptions` — it avoids loading Genesis first and then jumping.
- **Ordering of the eager write.** Writing the position into the shared map *after* the initial `queueRemoteSync` is correct: the initial sync reads the still-empty map, and `applySessionDataToReadingState` preserves the seeded position via its `?? peek()` fallbacks (and skips a `null` chapter), so the reader is not knocked off its start. Placing the write earlier would, as the description says, re-navigate to the chapter it is already on and cost a history entry. The write also lands before `stateMap.changes.subscribe`, so it does not echo back as a local re-sync.
- **The `!toStringOrNull(stateMap.get("bookId"))` guard** mirrors how `hostUserId` is claimed only when unset — a good defensive match for the create-vs-join split (join passes no `startPosition`, so joiners never overwrite).
- **`{ ...startPosition, isShared: true }`** with `isShared` last correctly prevents a caller's position from flipping a session's reading state to unshared.
- **Tests.** The mock cleanly separates the `reading_state` map from `options`/`decorations`, so the new "publishes without waiting for the debounce" and "no start position leaves the shared position empty" assertions are checking the right map and not passing by accident. The `SeedBibleStateManager` regression test using two tabs on different chapters is a good guard against reading the position off the wrong tab.
- **Scope.** Leaving the selected verse out is well justified given `scrollToVerse` does not survive to the rendered session tab; documenting `SessionStartPosition` as chapter-level keeps it from being reintroduced silently.

Nice, careful change overall — just clear out that one dead test helper so CI can go green.

---

PR #1603, opened 2026-08-05
fix: Today screen no longer auto-opens over path-based deep links (#1600)

### LINKED ISSUE #1600: Fix Today screen opening up always after path-based URL change
The Today screen inspected the initial URL that the Seed bible was loaded with in order to determine whether to automatically open the today screen. After the path-based URL changes were merged, the query parameters that the Today screen used to parse to determine whether to open on load are no longer there so now it is always opening on load even if the initial URL is a deep link to a book.

The implementation needs to be updated to check the path for the translation, book, chapter, and verse instead of the URL query parameters.

Blitzy:
## Summary

Fixes #1600. Today's "auto-open on cold load" check only looked at the legacy `?book=`/`?chapter=`/`?verse=` query params to detect a deep link. PR #1547 moved the reading position into the URL path, so those params are never present anymore — Today opened over every deep link instead of just bare loads.

Also found and fixed a related regression from the same PR: the effect that closes a fullscreen pane on chapter navigation read the same dead query params, so it silently stopped firing.

## Changes

- `ReadingUrlPath.ts`: new `hasReadingUrlPosition(url, basePath)` helper, backed by `parseReadingPath()`.
- `SeedBibleStateManager.tsx` / `today-screen/bootstrap.tsx`: both now call the shared helper instead of their own copy of the check.

## Test plan

- [x] Unit tests for `hasReadingUrlPosition()`.
- [x] Regression test: fullscreen pane closes on chapter navigation.
- [x] Full suite passing, typecheck clean.

### DIFF
--- a/packages/seed-bible/seed-bible/managers/ReadingUrlPath.ts
+++ b/packages/seed-bible/seed-bible/managers/ReadingUrlPath.ts
@@ -222,3 +222,13 @@ export function buildReadingUrl(params: {
 
   return url;
 }
+
+/**
+ * Whether `url` points at a specific reading position (the canonical
+ * `/{lang}/{translationId}/{book}/{chapter}` path). No query-param fallback
+ * needed — `entry-ssr.tsx` already redirects legacy `?book=`/`?chapter=`
+ * URLs onto the canonical path before the app ever sees them.
+ */
+export function hasReadingUrlPosition(url: URL, basePath: string): boolean {
+  return parseReadingPath(url.pathname, basePath) !== null;
+}
--- a/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
@@ -10,7 +10,11 @@ import {
   bibleLanguageToUiLocale,
   uiLocaleForDefaultTranslation,
 } from "../managers/BibleReadingManager";
-import { buildReadingPath } from "../managers/ReadingUrlPath";
+import {
+  buildReadingPath,
+  hasReadingUrlPosition,
+  parseReadingPath,
+} from "../managers/ReadingUrlPath";
 import type { OfflineTranslationStore } from "../managers/OfflineTranslationStore";
 import { createBibleToolsManager } from "../managers/BibleToolsManager";
 import type { ToolsManager } from "../managers/BibleToolsManager";
@@ -609,8 +613,9 @@ export function createSeedBibleState(
     i18n,
     readingExtensions
   );
-  // Close any fullscreen pane when the book/chapter params change, so
-  // navigating reveals the reader (every navigation path writes these params).
+  // Close any fullscreen pane when the book/chapter in the URL path changes,
+  // so navigating reveals the reader (every navigation path writes this
+  // position into the path — see `commitSelectedTabToUrl` in TabsManager).
   // The first location only sets a baseline, so load-time init doesn't close a
   // pane auto-opened for the same load (e.g. Today via `?today=open`).
   //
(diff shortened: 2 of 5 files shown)


Blitzy:
> [!NOTE]

## Code Review — PR #1603

**Verdict: looks solid, no blocking issues.** The fix is correctly targeted at issue #1600, the approach (a shared `hasReadingUrlPosition()` helper backed by `parseReadingPath()`) is consistent with how the rest of the codebase already reads the path first and falls back to legacy query params (`getInitialFirstTabBookId`, `getInitialTranslationId`, etc. in `TabsManager.tsx`), and the diff is well-commented. The central correctness claim — that `entry-ssr.tsx` promotes legacy `?book=`/`?chapter=` requests onto the canonical path before the app sees them — holds for the SSR cold-load path, which is exactly what `initialUrl` reflects (both `legacyReadingUrlRedirect` and `acceptLanguageRedirect` cover the bare-root `?book=` + `?lang=`/Accept-Language cases).

The findings below are all **Low / informational** — worth a look, none block merge.

> _Note on verification: I reviewed by code inspection. `node_modules` isn't installed in this environment, so I did not execute the suite or typecheck myself; I'm taking the PR's "full suite passing, typecheck clean" claim at face value._

---

### Low — No. 1: The dropped `verse` check is a real (though harmless) behavior change

`hasCompetingDeepLink` (and `todayWillAutoOpen` in `SeedBibleStateManager.tsx`) previously suppressed Today on `initialUrlParams.has("verse")`. That check is now gone, replaced entirely by `hasReadingUrlPosition(...) || has("sessionId")`.

In practice this is safe: a verse deep link is always `/{lang}/{translationId}/{book}/{chapter}?verse=N`, so the path position is present and `hasReadingUrlPosition` returns `true`. But there are two cold-load URLs where a legacy param survives to the app *without* a path, and Today now auto-opens where it previously stayed closed:

1. A bare `/?verse=5` (no book/chapter). `legacyReadingUrlRedirect` returns null (0 segments, no `?book=`), so it reaches the app as-is. Previously suppressed by `has("verse")`; now Today opens over the default chapter.
2. A legacy `/?book=<unresolvable>&lang=en`. When the book doesn't resolve even via the fuzzy fallback, `legacyReadingUrlRedirect` returns null and no path is ever built, so `hasReadingUrlPosition` is `false`. Previously suppressed by `has("book")`; now Today opens over the reader's not-found state.

Both are arguably *more* correct (a lone `?verse=` with no book is effectively a bare load), and both are vanishingly rare. **No change requested** — just confirm this is intended, and consider a one-line note in the comment block that legacy params which never get promoted to a path are deliberately no longer treated as deep links.

**Repro (case 1), on the branch deployment:**
1. Open `https://alpha.seedbible.org/b/fix_1600-path-url-parse-fixes/a47acebe8b19331877c4ed9356374499c2246cf3/?verse=5`
2. Observe Today auto-opens. On `develop` (pre-path-URL migration semantics) it would not have.

---

### Low — No. 2: No test locks in the actual issue #1600 behavior

The PR adds good tests for the **helper** (`hasReadingUrlPosition`) and for the **pane-close regression** (fullscreen pane closes on chapter nav). But the behavior that actually regressed — *Today must not auto-open when `initialUrl` is a path-based deep link, and must auto-open on a bare cold load* — has no direct test at the `todayWillAutoOpen` / `bootstrapExtension` predicate level. That predicate is what silently broke after PR #1547, and it's what would silently break again.

Suggestion: add a small test around `createSeedBibleState` (or the today-screen bootstrap) asserting, via `jsdom.reconfigure({ url })`:
- `url: "https://example.com/"` → Today auto-opens.
- `url: "https://example.com/en/BSB/john/3"` → Today does **not** auto-open.
- `url: "https://example.com/?today=open"` on top of a deep-link path → Today opens (explicit param wins).

This is the highest-value addition here — it's cheap and it directly guards the reported bug.

---

### Low — No. 3: `hasReadingUrlPosition` returns `true` for an unresolved book

Because it only checks `parseReadingPath(...) !== null`, a path like `/en/BSB/notabook/3` (4 well-formed segments, valid chapter, `bookMatch: "unresolved"`) counts as "has a reading position." For the Today-suppression use this is exactly right — an unresolved-book deep link still shouldn't be buried under Today. Flagging only because the helper's name reads like "points at a *valid* position," which it doesn't guarantee. Consider a one-word doc note ("path-shaped", not "resolvable") so a future caller that needs a *resolvable* position checks `bookMatch` rather than trusting this helper. **No code change needed.**

---

### Nice-to-have — pane-close effect now parses on every URL change

The reworked effect in `SeedBibleStateManager.tsx` runs `parseReadingPath()` on every `currentUrl` change (verse selection mirrors into the query, so this fires fairly often), where the old code just read two query params. This is negligible — canonical in-app paths hit `getBookId`'s exact-match fast path and never reach the `findClosestBookId` fuzzy fallback — and keying the location on the resolved `bookId` is actually an *improvement*, since self-heal/language rewrites of the path no longer read as a navigation and won't spuriously close a pane. Noting only that it was considered; no action needed.

---

PR #1604, opened 2026-08-07
Fold `test/` into `pnpm lint`, fix real errors it surfaced

Blitzy:
## Summary

`pnpm lint` only ever ran `eslint packages` — `test/` was never linted, no matter what the config allowed. This surfaced recently: a dead function in a test file only got caught by hand, in commit `58a2e34`, because lint had no way to see it.

This PR adds `test` to the `lint`/`lint:fix` scripts and fixes everything that turned up:

- **Real dead code removed**: an unused import, an unused local, two stale `eslint-disable` comments for rules that were never even enabled, and a couple of `let`→`const` corrections.
- **Config change**: added an underscore-prefix convention (`_arg`, `_omit`, etc.) to `no-unused-vars` — several tests already used it to mark an intentionally-unused mock parameter or destructured field, it just wasn't wired into the rule. This lives in the same shared rule block as `packages/` and `script/`, not a separate one, so tests aren't linted more strictly than the code they cover.
- **Also added**: Node + vitest globals for the `test/**` block, so a future `.js`/`.mjs` test file won't fail lint on `describe`/`vi`/`process` being "undefined."

## Why this is safe

One of my first passes at this accidentally turned a deliberate forward-reference pattern in `PanesManager.test.ts` into a `let`→`const` change that would have thrown a `ReferenceError` on a specific test path. A review round caught it before it landed — it's not in this diff. Every other file here is either dead code deleted or a genuine warning-severity match to the existing `packages/` baseline.

## Test plan

- [x] `pnpm lint` — 0 errors (549 pre-existing CSS baseline warnings unrelated to this change)
- [x] `pnpm vitest run` on all 13 touched test files — 335 tests pass
- [x] Manually verified the `PanesManager.test.ts` "calls onClose exactly once" test still guards the forward-reference case it's designed for


Blitzy:
> [!NOTE]

## Code Review — Fold `test/` into `pnpm lint`

Nice, well-scoped cleanup, and the PR description is unusually clear about *why* each change is safe. I verified the 13 file changes individually: every removed import, local, and `eslint-disable` directive is genuinely unused elsewhere in its file, the `void trigger.value` fix in `PanesManager.test.ts` correctly preserves the signal subscription, and `globals.vitest` does exist in `globals@16.5.0` so the new test-block globals resolve. Config-wise the flat-config ordering is correct: the shared block turns `no-explicit-any` on, and the later `test/**` block turns it back off for tests.

There is one blocker, though — the branch is currently red, and it's caused by this PR.

---

### 🔴 High — No. 1: type regression breaks `pnpm check:ts` (fails the `build` CI job)

**Where:** `test/unit/scripture-map/hooks/useReadingHistoryTimeline.test.tsx:312` and `:371`

**What happened.** This PR simplified two mocks by dropping their unused parameter:

```diff
-      // eslint-disable-next-line
-      const getColorByReadingTime = vi.fn((args: any) => "#computed");
+      const getColorByReadingTime = vi.fn(() => "#computed");
```

But a few lines below each one, the test reads back the *first argument* the mock was called with:

```ts
const callArg = getColorByReadingTime.mock.calls[0]?.[0] as any;
expect(callArg.userColor).toBe("#ffffff");
```

When the callback is declared as `vi.fn((args: any) => …)`, TypeScript infers the mock's parameters as `[args: any]`, so `mock.calls` is an array of one-element tuples and `.calls[0]?.[0]` type-checks. After removing the parameter, `vi.fn(() => …)` infers the parameter list as the **empty tuple `[]`**, so indexing `[0]` on it is now a compile error:

```
test/unit/scripture-map/hooks/useReadingHistoryTimeline.test.tsx:312:61 - error TS2493:
  Tuple type '[]' of length '0' has no element at index '0'.
test/unit/scripture-map/hooks/useReadingHistoryTimeline.test.tsx:371:61 - error TS2493:
  Tuple type '[]' of length '0' has no element at index '0'.
```

**Why the test plan missed it.** `pnpm vitest run` passes because Vitest transpiles with esbuild and does **no type checking** — at runtime the production code still calls the mock with a real argument, so `mock.calls[0][0]` returns the expected object. Only `tsc` sees the problem. The PR's test plan ran `pnpm lint` and `pnpm vitest run` but not `pnpm check:ts`, which is the job that's now failing (`build (24.x)` → `pnpm check:ts` → exit code 2). This is also why `mergeable_state` is `blocked`.

**Reproduction:**
1. `git checkout fix/lint-tests`
2. `pnpm check:ts`  (or just `pnpm check:ts:client`)
3. Observe the two `TS2493` errors above → exit code 2.

**Suggested fix.** Restore a parameter, but prefixed with `_` so it stays lint-clean under the very `argsIgnorePattern: "^_"` convention this PR adds. Apply to both mocks (lines 288 and 339):

```ts
const getColorByReadingTime = vi.fn((_args: any) => "#computed");
```

That gives `mock.calls` a `[_args: any]` element type again (so `.calls[0]?.[0]` type-checks), keeps the runtime behavior identical, and — because `_args` matches the new ignore pattern and `no-explicit-any` is off for `test/**` — trips neither `no-unused-vars` nor `no-explicit-any`.

---

### 🟡 Low — No. 2: a newly-surfaced `prefer-const` warning left unaddressed

**Where:** `test/unit/seed-bible/managers/PanesManager.test.ts:592`

Now that `test/` is linted, a pre-existing `let paneId` surfaces as:

```
592:11  warning  'paneId' is never reassigned. Use 'const' instead  prefer-const
```

It's only a warning (`prefer-const` is `warn`, and `lint` has no `--max-warnings`), so it does **not** fail CI — but since this PR's whole spirit is "lint now sees `test/`, so fix what it surfaces," and it already did the same `let`→`const` cleanup in `FeaturesManager.test.ts`, folding this one in would be consistent. Non-blocking.

---

### ⚪ Informational — No. 3: the underscore relaxation also loosens production linting

**Where:** `eslint.config.mts:104-111`

The new `no-unused-vars` options (`argsIgnorePattern`/`varsIgnorePattern`/`caughtErrorsIgnorePattern: "^_"`) live in the shared rule block, so they apply to `packages/**` and `script/**` as well as `test/**`. That's called out in the PR description as deliberate ("same shared rule block … so tests aren't linted more strictly than the code they cover"), and it's a common, reasonable convention — flagging only so it's a conscious call that `_`-prefixed unused variables are now permitted in production code too, not just tests. No change requested.

---

**Summary:** No. 1 must be fixed to get the build green (two-line change, suggested above). No. 2 and No. 3 are optional. Once `check:ts` passes, this is a clean, low-risk change that does exactly what it says.

---

Blitzy:
Fixed No 1 and No 2. As for No 3 the plan is to keep the rules unified between production code and tests for now.

Blitzy:
> [!NOTE]

## Follow-up Code Review — Fold `test/` into `pnpm lint`

Re-reviewed against the current head (`fa4f690`), since the last two commits address the previous review's findings.

**No. 1 (High, TS2493 type regression) — resolved.** Commit `6d4ed68` restores the `_args: any` parameter on both `getColorByReadingTime` mocks in `useReadingHistoryTimeline.test.tsx` (lines 288 and 339), exactly as suggested. This gives `vi.fn()`'s inferred call-tuple type an element at index 0 again, so `mock.calls[0]?.[0]` type-checks. Confirmed via CI: the `build (24.x)` check run is green, and its steps run `pnpm lint` → `pnpm check:ts` → `pnpm test:coverage` → `pnpm build` in sequence, so a `check:ts` failure would have failed the whole job.

**No. 2 (Low, `prefer-const` warning) — resolved, via a different route than suggested.** Rather than changing `let paneId` to `const` in `PanesManager.test.ts:592` (which the PR description explains would break — `paneId` is intentionally read by the `onClose` closure before it's assigned a few lines later, a forward-reference pattern this PR has already had to work around once), commit `fa4f690` instead adds `ignoreReadBeforeAssign: true` to the shared `prefer-const` config in `eslint.config.mts`. That's a real, documented ESLint option for exactly this situation (a `let` that's never reassigned but is read inside a closure defined before its one assignment), and it's arguably the more correct fix — it stops the warning at its actual source instead of leaving the underlying rule mismatched with a legitimate pattern the codebase already relies on. Since this lives in the shared rule block, it also applies to `packages/**` and `script/**`, not just this test file; that's consistent with the PR's existing stance ("tests aren't linted more strictly than the code they cover") and I didn't find any case where it would suppress a genuine unintentional-`let`.

**No. 3 (Informational, underscore convention loosens production linting)** — the author's follow-up comment confirms this is intentional: "the plan is to keep the rules unified between production code and tests for now." No action needed.

**New commits introduce no new issues.** Both are minimal, targeted diffs scoped to exactly what they claim to fix — no incidental changes.

**Verdict:** The blocker is cleared, CI is green end-to-end (lint, typecheck, tests, build), and the two follow-up fixes are both correct. No outstanding issues from this review.

---

Issue #1401, opened 2026-07-12
Swiping to change chapter on mobile does not track in history.


Blitzy:
## What is happening 

When you use swipe to change the chapter on mobile, the back and forward browser navigation do not navigate through the history of those pages.

However, if you use the mobile navigation toolbar button history is tracked. 

## What is expected

Swiping to change chapters on mobile should track chapter history like the mobile navigation toolbar buttons do.

PR #1605, opened 2026-08-08
fix: swipe chapter navigation history (#1401)

Bored-Wizard:
Addresses  #1401.

## What was wrong

Swiping between chapters on mobile wrote history entries that the browser's Back/Forward buttons would sometimes skip — pressing Back after a swipe could jump over the previous chapter or leave the app entirely, even though the entry was visible in the long-press history list. The toolbar chevrons never had this problem.

## Root cause (browser behavior, not app code)

Chrome grants one user-credited history entry per real interaction (tap, click, keypress), consumed by each `pushState`. A swipe is a drag, which Chrome does not count as an interaction — so a swipe's history entry is often uncredited, and Chrome's [history-manipulation intervention](https://chromestatus.com/feature/5585378122907648) flags the app's entries as "skip on back/forward". The chevrons only ever worked because each click re-arms the credit. We confirmed this model empirically by logging `navigation.entries()` around every history write: the swipe and chevron paths produce byte-identical history stacks, and the flags clear the moment the user taps the page again.

Nothing written from the page can synthesize that credit — no timing, ordering, or push/replace shape changes it. An earlier revision of this PR pushed the current URL at touchend and then replaced it with the destination; that produces an identical stack to a plain push and Chrome treats it identically, so it has been removed.

## What this PR actually fixes

- `touch-action: pan-y`, `overscroll-behavior: none`, and a non-passive `preventDefault` on the swipe viewport stop the browser's own gesture handling from fighting the chapter swipe. This is the substantive fix: without it, a horizontal drag could be claimed by the browser as a scroll or its own back/forward gesture.
- The swipe now navigates through the exact same call as the chevrons (`loadNextChapter()` / `loadPreviousChapter()`, no options), so both paths share one history behavior, including the coalescing that keeps a fast skim down to one Back press.
- Added a `touchcancel` handler: if the browser takes a gesture over mid-swipe, the track previously stayed stuck wherever the finger left it.

## Known limitation, by browser design

Swipe → immediately press Back can still skip the previous chapter, because Chrome never credited the swipe's entry. Any tap on the page clears the flags and Back works normally from then on — so in practice this only surfaces when Back is the very next thing the user touches after a swipe. There is no page-side fix for this; the full mechanism is documented in the `commitSwipe` comment in `TabsLayout.tsx`.

Note: the Loom above predates this rework and shows the earlier push-then-replace approach.


Blitzy:
> [!NOTE]

## Code review — PR No. 1605 (swipe navigation history, issue #1401)

### The short version

The core fix here is right: making the `touchmove` listener non‑passive, calling `event.preventDefault()` once the swipe locks horizontal, and adding `touch-action: pan-y` stops the browser's native horizontal swipe‑back/forward gesture from fighting the in‑app navigation. That is almost certainly why history wasn't tracking before, and this addresses it well.

The one thing I'd push back on is the extra history plumbing that came along for the ride — the manual `navigation.push(currentUrl)` in `commitSwipe` plus threading `replace: true` down through `loadNextChapter`/`loadPreviousChapter`. It reproduces the correct result for a single swipe, but it introduces a regression on rapid swipes and quietly diverges from the toolbar behaviour the issue asked swipe to match.

---

### Medium — Rapid swipes leave dead/duplicate history entries

**Files:** `components/TabsLayout.tsx` (L445‑447, L505‑512), `managers/NavigationManager.tsx` (`push`)

`commitSwipe` now runs `navigation.push(navigation.currentUrl.peek())` **unconditionally**, every time a swipe passes the threshold. But the actual chapter navigation is deferred into a `setTimeout(..., SWIPE_ANIMATION_MS)` (250 ms), and the next `commitSwipe` calls `window.clearTimeout(swipeCommitTimer.current)` — cancelling the previous swipe's pending `navigate()`. So a second swipe within 250 ms cancels the first swipe's navigation, **but its `push` has already fired.**

`NavigationManager.push` calls `window.history.pushState` directly, which always creates a new entry regardless of whether the URL changed — so pushing the *current* URL adds a duplicate entry.

Concrete trace, starting at history `[P, A]` (A = current chapter):

| Step | Action | History | Content |
| --- | --- | --- | --- |
| Swipe 1 commits | `push(A)`, timer1 set | `[P, A, A]` | A |
| Swipe 2 commits (<250 ms) | `clearTimeout(timer1)` cancels swipe 1's navigate; `push(A)` | `[P, A, A, A]` | A |
| timer2 fires | `loadNextChapter({replace:true})` → replaces top entry with B | `[P, A, A, B]` | B |

Result: content advances **one** chapter (A→B, the cancelled‑navigate behaviour is pre‑existing) but **three** entries are added, two of which are dead duplicates of A. Pressing Back goes B→A (visible), then A→A (no visible change — a "dead" back press), then A→P.

For comparison, the base `develop` code (which just called `loadNextChapter()`) ends at `[P, A, B]` for the same two rapid swipes — one entry, no dead presses. So this is a fresh regression on the exact feature the PR fixes.

**Repro (needs a mobile viewport / device):**
1. Open a chapter, e.g. John 3.
2. Swipe to the next chapter twice in quick succession (both swipes inside the ~250 ms slide animation).
3. Note the reader advances a single chapter, then press the browser Back button repeatedly.
4. **Observed:** one or more Back presses produce no visible change before you actually leave the chapter. **Expected:** the number of Back presses to unwind equals the number of chapters actually navigated (as the toolbar chevrons do).

**Suggested fix (simplest):** drop the manual push and the `replace` plumbing on the swipe path, and just call the same thing the toolbar and keyboard nav already call:

```ts
if (shouldLoadNext && hasNext) {
  commitSwipe(track, `translateX(${sign * PANEL_PCT * 2}%)`, () =>
    readingState.loadNextChapter()
  );
} else if (shouldLoadPrev && hasPrev) {
  commitSwipe(track, "translateX(0%)", () =>
    readingState.loadPreviousChapter()
  );
}
```

With the `preventDefault`/`touch-action` fix in place, the existing `emitPositionNavigate` machinery already pushes exactly one entry per navigation (and coalesces a fast skim into one entry — see the next point). That is the behaviour issue #1401 asks for. If you keep this route, `ReadingNavigationOptions` and the `replace` threading in `BibleReadingManager` are no longer needed.

If there's a specific reason the entry must be created synchronously at gesture time (e.g. to cover a Back press *during* the 250 ms animation), then the push should be moved next to the `navigate()` call inside the timer and guarded by the commit token, so a superseded swipe never contributes a push.

---

### Low — Swipe now bypasses history coalescing, diverging from the toolbar

**File:** `managers/BibleReadingManager.tsx` (`emitPositionNavigate`, L1436‑1442)

`emitPositionNavigate` deliberately collapses a burst of navigations within `NAVIGATION_COALESCE_MS` (400 ms) into a single history entry — its own comment explains this is so "skimming ten chapters" costs one Back press, not ten. The toolbar chevrons and keyboard nav go through this path (they call `loadNextChapter()` with no options).

By passing `replace: true` explicitly, the swipe path opts out of that heuristic entirely — every committed swipe is forced to its own entry. So a fast swipe‑skim gives one Back press per chapter, while the same skim on the toolbar gives one Back press total. Issue #1401 explicitly wants swipe to behave "like the mobile navigation toolbar buttons," so this is a behavioural divergence worth a conscious decision. The fix in the Medium item above resolves this too (it routes swipe through the same coalescing path).

---

### Low — `replace: !!navigation` couples "should replace" to "does a navigation manager exist"

**File:** `components/TabsLayout.tsx` (L427)

`const readingNavigationOptions = { replace: !!navigation };` conflates two unrelated things: whether the app has a `NavigationManager`, and whether this navigation should replace vs. push the history entry. It happens to work because `navigation` is effectively always present, but it reads as a coincidence rather than intent, and it's the mechanism that produces the divergence above. If the manual‑push approach is kept, a named boolean with a comment explaining *why* swipe replaces would be clearer.

---

### Nits / observations (non‑blocking)

- **`managers/NavigationManager.tsx` L161‑166 (the `onNavigate` guard):** skipping the `currentUrl` write when `href` is unchanged mirrors the existing guard in `syncCurrentUrl` and is a sound, defensive change — no issue. Worth noting it's the thing that keeps the duplicate‑`push` from also spamming `currentUrl` subscribers, so it's a little load‑bearing for the current approach.
- **`emitNavigate({ replace: options?.replace ?? false })` (L2335, the hook‑"handled" branch):** if an extension navigation hook *handles* a swipe (`replace: true`) after `commitSwipe` has already pushed the current URL, the position may not change yet the URL is replaced — leaving a stray duplicate entry. Niche (requires an extension intercepting swipe nav), and it also goes away with the simplified approach.
- **No test coverage** was added for the new history behaviour. Given how subtle the push/replace ordering is here, a small unit test around `loadNextChapter`/`emitPositionNavigate` history intent (push vs. replace, and the coalescing window) would guard against exactly the regression above. `BibleReadingManager` sits at 66% function coverage, so there's likely a harness to extend.

---

I verified the trace above against the base `develop` version of `commitSwipe` (which lacked the `push`) and the `emitPositionNavigate`/`applyPosition` paths, so the "before = `[P, A, B]`, after = `[P, A, A, B]`" comparison reflects the actual code, not just the diff. Happy to be corrected if the deferred‑push timing was intentional for a case I haven't reproduced.

---

Blitzy:
Follow-up on the root cause, for the record before merge.

The remaining Back-button behaviour is a Chrome policy, not something this PR can fix. Chrome credits one history entry per real interaction (tap, click, keypress); a swipe is a drag and earns no credit, so the entry it pushes gets flagged "skip on back/forward" by the history-manipulation intervention. We confirmed the model empirically by logging `navigation.entries()` around every history write, and by verifying that a single tap anywhere on the page makes Chrome clear the flags. Both the failure and the tap-clears-it behaviour reproduce on real Android hardware, not just in DevTools emulation. The toolbar chevrons were never doing anything smarter — their click just re-arms the credit each time.

Nothing page-side changes that: we measured both timings (sync vs deferred) and both shapes (plain push vs push-then-replace) and they produce identical outcomes. Activation cannot be synthesised.

So the residual case is: swipe, then press Back with no tap in between. Any interaction heals it. Merging this is still a clear improvement — it fixes the browser fighting the gesture, fixes a stuck track on `touchcancel`, and removes a history hack that was a no-op — and it doesn't make the Chrome behaviour any worse than it is on `develop` today.

One thing worth checking before we consider #1401 fully closed: whether a single uncredited swipe flags only its own entry or the whole same-document stack. Our repro never had more than one app entry to observe. If it's the latter, one swipe could cost a user several chevron-created Back steps, which would be worth a design rethink rather than a documented caveat.

Removing swipe navigation was considered and rejected — it's the primary mobile reading gesture, and the chevrons already give an unaffected path back.

Blitzy:
> [!NOTE]

## Code review — PR No. 1605 (swipe navigation history, issue #1401)

### Context

This is a re-review after the branch was marked ready for review again. The previous review from Blitzy (comment above) flagged a Medium issue: the earlier revision manually pushed a history entry in `commitSwipe` and threaded a `replace` option through `loadNextChapter`/`loadPreviousChapter`, which produced duplicate/dead history entries on rapid swipes. **That issue is fixed** — the current head commit (`fix: navigate swipes through the same path as the toolbar chevrons`, 9e28e61) removes the manual push and the `replace` plumbing entirely; swipe now calls `readingState.loadNextChapter()` / `loadPreviousChapter()` with no arguments, exactly like the toolbar chevrons, so it goes through the same `emitPositionNavigate` coalescing path. I diffed the current `TabsLayout.tsx`/`NavigationManager.tsx` against develop and confirmed no trace of the old manual-push/`ReadingNavigationOptions` approach remains.

I re-reviewed the current diff independently rather than just trusting that the prior fix was complete, and ran the tests (including reverting the `touchcancel` fix locally to confirm the new regression test actually catches it — details below).

### The short version

The remaining changes are sound:
- `touch-action: pan-y` on `.sb-reader-swipe-viewport` + `overscroll-behavior: none` on `.sb-reader-swipe-panel` + the guarded `event.preventDefault()` in `onTouchMove` (only once direction locks horizontal, and only when `event.cancelable`) stop the browser's own horizontal gesture handling from fighting the in-app swipe, without breaking native vertical scroll (the viewport keeps `pan-y`, and `preventDefault` never fires before the direction lock decides "vertical").
- The new `touchcancel` handler recentres the track and resets swipe state when the browser claims the gesture mid-drag (system edge gesture, second finger) — previously the track would stay stuck wherever the finger left it.
- `NavigationManager`'s `onNavigate` now skips writing `currentUrl` when the destination href is unchanged, mirroring the existing guard in `syncCurrentUrl` (line 82) — a real existing pattern, not a one-off.

No new correctness issues found in this pass. Two small, non-blocking notes below.

---

### Low — No direct unit test for the new `onNavigate` href guard

**File:** `managers/NavigationManager.tsx` L164-166

The new guard (`if (currentUrl.peek().href !== destination.href) { currentUrl.value = destination; }`) isn't itself covered by `NavigationManager.test.ts` — the existing tests there exercise `dispose()` behavior and `syncCurrentUrl`, not this specific branch in `onNavigate`. It's a low-risk, defensive no-op mirroring a tested pattern elsewhere, so this isn't blocking, but a quick test asserting that a `navigate` event to the same URL doesn't trigger a new `currentUrl` signal write (e.g. via a `computed`/effect spy) would close the gap and guard against a future regression here specifically.

### Nit — `touchcancel` doesn't bump `swipeCommitToken`

**File:** `components/TabsLayout.tsx` (`onTouchCancel`)

`onTouchStart` bumps `swipeCommitToken.current` to take ownership of the track from any in-flight committed swipe, but `onTouchCancel` doesn't. In practice this isn't a bug — `onTouchCancel` only ever fires for a gesture that never reached `commitSwipe` (that only happens in `onTouchEnd`), so there's no in-flight commit for it to invalidate, and the next `onTouchStart` still takes ownership correctly. Just flagging it so the asymmetry is a documented choice rather than an oversight if this code is touched again.

---

### Test coverage

I verified the tests are checking real behavior, not just passing trivially:
- Ran the full `TabSlotReader.test.tsx` suite on the PR branch — all 25 tests pass (3 new: swipe-matches-chevron-call, waits-for-animation, touchcancel-recentres).
- Temporarily reverted the `onTouchCancel` handler to a no-op locally and reran the suite: the new "recentres the track and navigates nowhere when the browser cancels the gesture" test correctly **failed** (`translateX(calc(-33.33% + -110px))` instead of the centred transform), confirming it's a real regression test and not a tautology. Restored the fix afterward.
- The other two new tests (matching the chevron call signature, and waiting for `SWIPE_ANIMATION_MS` before navigating) assert on observable `loadNextChapter`/`loadPreviousChapter` call behavior via fake timers, not implementation internals — consistent with the repo's test conventions.

CI (`build (24.x)`) is green on the current head commit.

### On the residual Chrome history-credit caveat

Blitzy's follow-up comment (2026-08-11) explains the remaining "swipe → immediate Back" edge case is a browser-side interaction-credit policy that no page-side code can work around, and that this was verified empirically on real hardware. That's consistent with what's achievable here and isn't something this PR can or should try to fix further — the documented caveat in the PR description is the right way to close this out.

---

Blitzy (approved):

PR #1606, opened 2026-08-08
Feature/1550

### LINKED ISSUE #1550: Improve `parseVerseReferences()` and `parseVerseReference()` to support non-english languages
Currently, `parseVerseReferences()` and `parseVerseReference()` only search for book names by book ID or by English book name. We should improve them to take an optional list of books that are searched through first before checking by book ID or english book name. For example, if given the list of books is from the `spa_onbv` translation, then "Esdras 3" would match to book ID `EZR` referencing chapter 3 (Ezra 3).

The function signature should look something like this:

```typescript
function parseVerseReferences(text: string, books?: TranslationBooks)
```

Finally, we should go through the usages of `parseVerseReferences()` (and `parseVerseReference()`) and update them to include the list of books from the current reading state

Kabi4:
Fixes: #1550


Blitzy:
> [!NOTE]

## Code review — Feature/1550 (localized verse-reference parsing)

Reviewed against issue #1550. The PR delivers the core ask: both `parseVerseReference()` / `parseVerseReferences()` now accept an optional `books` list, localized exact names resolve (e.g. `spa_onbv` "Esdras" → `EZR`), and the callers that scan prose — footnote bodies (`VerseReferenceText`) and chat messages (`ChatsManager`) — are wired to the current reading tab's books via a new `selectedTabTranslationBooks` signal. Type check is clean (`pnpm check:ts`, 0 visible errors) and all 654 tests in the two affected suites pass. Test coverage of the new behavior is genuinely good.

Two behavioral issues are worth addressing before merge, plus a couple of smaller cleanups. Severity is my estimate of user impact, not a merge veto.

---

### Medium — Prefix matching in the prose scanner creates false-positive verse links

`resolveBookId()` in `BibleDataManager.tsx` (the `else` branch of the exact match, lines ~264–271) does a **unique-prefix** match on book names. That is the right behavior for a user deliberately typing a reference into a box, but `parseVerseReferences()` here is the *scanner* that runs over arbitrary prose (chat messages, footnote text). Applying unique-prefix matching there means ordinary short English words followed by a number now render as scripture links.

Before this PR the scanner used `getBookId()` only, whose prefix fallback is conservative (it checks whether the *input* starts with a known key, not whether a book name starts with the input), so these did not match. Now they do.

**Reproduction** (verified locally against an English books list):

| Text in a chat/footnote | Before | After |
| --- | --- | --- |
| `Is 3 enough for everyone?` | (no link) | links "Is 3" → Isaiah 3 |
| `So 3 people showed up` | (no link) | links "So 3" → Song of Solomon 3 |
| `Am 3 sure about this` | (no link) | links "Am 3" → Amos 3 |

Any book with a unique 2-letter prefix is exposed: `Is`, `Am`, `So`, `Ho`, `Na`, `Ne`, `Ob`, `Ru`, etc. — all common words in normal conversation.

**Suggested fix:** don't apply prefix matching in the scanning path. Issue #1550 only requires *exact* localized names ("Esdras 3"), which the exact branch already covers. Options:
- Drop the prefix branch from `resolveBookId()` and rely on exact localized match + `getBookId()`'s own (conservative) rules; or
- Give the deliberate single-reference parse and the prose scanner different resolvers, so prefix matching stays available where a user is intentionally typing an abbreviation but is not used when scanning free text.

---

### Medium — Scanner can't detect localized book names that contain non-ASCII characters

The scanner's book-name pattern in `parseVerseReferences()` (`BibleDataManager.tsx` line ~369) matches book tokens with `[A-Za-z][A-Za-z0-9]*`. Most non-English book names contain accented or non-Latin characters, so the scanner never even forms the right token for them — which undercuts the feature's stated purpose (non-English support) in the chat/footnote path.

**Reproduction** (verified locally with a Spanish books list whose names carry accents):

```
parseVerseReferences("Génesis 3", spaBooks)  // → []  (misses it)
parseVerseReferences("Éxodo 2",  spaBooks)   // → []  (misses it)
```

The exact-name data is present in `books`; the ASCII-only regex just can't reach it. Note the tests exercise only accent-free Spanish names in this path ("Esdras", "Filipenses", "1 Corintios"), so this gap isn't caught. (The single-reference parser in `parseVerseReference.ts` splits on `(.+?)` and *does* handle accented input — the gap is specific to the scanner.)

**Suggested fix:** broaden the book-name character class to be Unicode-aware — e.g. use `\p{L}` with the `u` flag instead of `[A-Za-z]`. The `\b` word-boundary anchor and the `retryFromNextChar()` logic will need a re-check under the `u` flag, and a test with a genuinely accented name (e.g. "Génesis 3", "Nehemías 1") would lock the behavior in.

---

### Low — Duplicated book-matching logic across the two modules

`normalizeBookName()` + `resolveBookId()` in `BibleDataManager.tsx` re-implement `normalize()` + `exactBook()` + `prefixBooks()` from `parseVerseReference.ts`. They have already drifted: `parseVerseReference.ts` narrows prefix matches by `bookHasChapter()`, whereas `resolveBookId()` does not. Two copies of "match a book name" will keep diverging. Consider extracting the shared matching into one helper both modules import. (Having two separate `parseVerseReference`/`parseVerseReferences` families with different return shapes — `book` vs `bookId` — is pre-existing and out of scope, but this PR is a reasonable moment to at least share the name-matching primitive.)

### Low / nit — Prefix branch ignores chapter containment

`resolveBookId()`'s prefix match returns the first uniquely-prefixed book without checking that it actually contains the requested chapter, unlike the `parseVerseReference.ts` path which filters by `bookHasChapter()`. Minor here since the scanner has no better signal, and it becomes moot if the prefix branch is removed per the first finding.

---

### Notes / non-blocking

- Wiring looks complete: every prose caller of the `BibleDataManager` scanner (`VerseReferenceText` for footnotes, `ChatsManager` for chat) now receives books, and shared vs. local chat sessions each source books correctly (session reading state vs. selected tab). No un-updated callers of the singular `parseVerseReference` were found.
- Preserving the raw token as `book` when a reference is unresolved (`resolveBookId(book, books) ?? book`) is unchanged prior behavior, and the new tests document it — fine.
- CI: at the time of review the `build (24.x)` check was still in progress; worth confirming it goes green.

---

Blitzy:
> [!NOTE]

## Code review — Feature/1550 (localized verse-reference parsing)

This is a re-review. Commits landed after the previous pass (which flagged two Medium and two Low findings) — this review checks what changed and looks for anything new.

**What got fixed since the last review:** both Medium findings are resolved, with tests added for each.
- The prefix-matching branch was removed from `resolveBookId()` in `BibleDataManager.tsx`, so the scanner (`parseVerseReferences`, used for chat/footnote prose) no longer expands short tokens like "Is"/"So"/"Am" into false-positive book links. New tests ("does not unique-prefix expand short tokens when scanning prose", "does not treat ordinary short English words as book abbreviations") cover this directly.
- The scanner's book-name regex and word-boundary check were made Unicode-aware (`\p{L}`/`\p{N}` with a negative lookbehind instead of `[A-Za-z]`/`\b`), so accented localized names now match. New tests cover "Génesis 3", "Éxodo 2", "Nehemías 1:5".
- The Low finding about duplicated book-matching logic was also addressed: `normalizeBookName`/`exactTranslationBook`/`prefixTranslationBooks` were extracted into a new shared `managers/bookNameMatch.ts`, imported by both `BibleDataManager.tsx` and `parseVerseReference.ts`.

I re-ran the affected suites on the current head (`f4f2e871`): all 748 tests in `BibleDataManager.test.ts`, `parseVerseReference.test.ts`, and `ChatsManager.test.ts` pass. `pnpm check:ts` is clean and `pnpm lint` reports 0 errors (only pre-existing, unrelated CSS baseline warnings). CI (`build (24.x)`, `deploy (24.x)`) is green on the current head.

I didn't find any new correctness bugs in the fix commits. One coverage gap is worth calling out before merge.

---

### Medium — The manager-level wiring for localized chat resolution has no test that actually exercises it

The pure parsing functions (`parseVerseReferences`, `resolveBookId`, `bookNameMatch.ts`) are thoroughly tested with real Spanish book lists. But the two places that plumb a translation's books *into* chat — which is the actual feature this issue asked for ("go through the usages of `parseVerseReferences()` and update them to include the list of books from the current reading state") — have no test that passes a non-trivial books list through them:

- **Local chat:** `createChatsManager(loginManager, i18nManager, translationBooks)` takes the books signal, but every call site in `ChatsManager.test.ts` (`createChatsManager(loginManager, mockI18nManager)`) omits the third argument, so `translationBooks` is `undefined` in every local-session test. The `verse_reference` tests around lines 3460–3540 all use plain English names (GEN, JHN, ROM), which resolve via the pre-existing `getBookId()` fallback regardless of the new `books` plumbing — they'd pass identically if the third parameter were deleted entirely.
- **Shared chat:** `createSharedSessionMock()` sets `readingState.translationBooks: signal(null)` in every test (added in the latest commit purely to stop a mock crash), so `session.readingState?.translationBooks?.value?.books` is always `undefined` there too.
- **The tab → chat wiring in `SeedBibleStateManager.tsx`** — the new `effect()` that keeps `selectedTabTranslationBooks` in sync with `selectedTab.value?.readingState.translationBooks.value?.books` — has no test at all (`SeedBibleStateManager.test.ts` has no reference to `selectedTabTranslationBooks` or to the books passed into `createChatsManager`).

None of this is wrong today — I traced the wiring by hand and it looks correct — but nothing would fail if it broke. A regression here (e.g. forgetting to pass `selectedTabTranslationBooks` into `createChatsManager`, or reading the wrong signal in the effect) would silently fall back to English-only resolution with all existing tests still green.

*Suggested fix:* add at least one test per integration point using a Spanish (or other localized) books fixture — e.g. in `ChatsManager.test.ts`, create a local session via `createChatsManager(loginManager, mockI18nManager, signal(spaBooks))` and assert a message containing "Esdras 3" parses to a `verse_reference` part with `ref: { book: "EZR", ... }`; do the same for a shared session with `readingState.translationBooks` populated. For `SeedBibleStateManager.tsx`, a test that sets a tab's `translationBooks` signal and asserts `chats`'s injected signal follows it would catch a wiring regression directly.

---

### Things that look right

- `resolveBookId()` in `BibleDataManager.tsx` is now exact-only against the provided books, falling back to the already-conservative `getBookId()` — I confirmed `getBookId()`'s prefix fallback only matches when the *input* starts with a known key (not the reverse), so ordinary short words still don't resolve even without a books list.
- The Unicode regex changes (`\p{L}[\p{L}\p{N}]*` with `(?<[\p{L}\p{N}])` in place of `\b`) preserve the "Song of Solomon"-style multi-word matching and don't introduce any obvious backtracking risk.
- `readingState` is a required (non-optional) field on `BibleReadingSession` in `SessionsManager.tsx`, so the `?.` added around `session.readingState?.translationBooks?.value?.books` in the latest commit is purely defensive for test doubles, not masking a real production nullability gap.
- Book matching correctly stays split by purpose: the scanner (`BibleDataManager.tsx`) is exact-only to avoid false positives in free text, while the deliberate single-reference parser (`parseVerseReference.ts`, used by the scripture-reference input) still does prefix matching, which is appropriate there and unchanged by this PR.

---

Blitzy (approved):

PR #1609, opened 2026-08-08
test: unmount the reader between tests to stop a post-teardown render

Blitzy:
The suite intermittently failed with an unhandled
"ReferenceError: document is not defined" raised inside preact's DOM
diff. Vitest attributed it to BibleReader.test.tsx, and every assertion
passed — only the stray rejection failed the run.

BibleReader.test.tsx never unmounted what it rendered: the afterEach
detached the container but left the commented-out `render(null,
container)` in place, so all 61 trees stayed mounted for the whole file
with their timers, listeners and signal subscriptions live. A late async
update — the reader's CHAPTER_SKELETON_DELAY_MS skeleton timer, or the
un-awaited retry handler's `setRetrying(false)` — could therefore still
reach preact's diff after Vitest had torn the file's jsdom environment
down, at which point `document` no longer existed.

Unmounting inside `act` runs the effect cleanups and clears each
component's parent DOM, which is what makes preact drop a late update
instead of diffing it. Measured before and after on this file: two
un-cleared 500ms timers were pending when it finished, now zero.

Claude-Session: https://claude.ai/code/session_01NmPLJDywRiYx7LjTYcf994

### DIFF
--- a/test/unit/seed-bible/components/BibleReader.test.tsx
+++ b/test/unit/seed-bible/components/BibleReader.test.tsx
@@ -311,7 +311,15 @@ describe("BibleReader", () => {
   });
 
   afterEach(() => {
-    // render(null, container);
+    // Detaching the container is not enough: the tree stays mounted, so its
+    // effect cleanups never run and a late async update (the reader's skeleton
+    // timer, an un-awaited retry) still diffs into a document that Vitest has
+    // since torn down — an unhandled "document is not defined" rejection.
+    // Unmounting clears the component's parent DOM, which makes preact drop
+    // those updates instead.
+    act(() => {
+      render(null, container);
+    });
     container.remove();
   });
 


Blitzy:
> [!NOTE]

## Code review — test hygiene fix for `BibleReader.test.tsx`

**Summary:** The PR replaces a commented-out `render(null, container)` in the `describe`-scoped `afterEach` with a real, `act()`-wrapped unmount so each rendered tree is torn down between tests. This is a correct, well-scoped fix for the intermittent unhandled `"document is not defined"` rejection. I verified the mechanism against the component under test and ran the suite. **No blocking issues found.**

### Verification performed
- **Mechanism confirmed against the source.** The two leak sources named in the description are real and are both closed by unmounting:
  - The skeleton timer (`BibleReader.tsx:1545-1549`) is created inside a `useEffect` whose cleanup calls `window.clearTimeout(timer)`. Unmounting runs effect cleanups, so the pending 500ms timer is cleared instead of firing after teardown.
  - The retry handler (`BibleReader.tsx:1684-1692`) calls `setRetrying(false)` in a `finally` after an un-awaited `await readingState.retryLoad()`. Once the component is unmounted, Preact drops that queued state update rather than diffing it into a torn-down document.
- **Hook ordering is correct.** The new unmount lives in the `describe`-scoped `afterEach`; the module-level `afterEach(teardown)` is outer. Preact runs `afterEach` innermost-first, so the unmount runs *before* `teardown()` — the required order.
- **All render targets share one container.** Every `render(...)` call and the `renderMobileReader` helper render into the single `describe`-scoped `container`, so `render(null, container)` unmounts whatever the test last mounted. Nothing bypasses it.
- **Suite passes:** `pnpm vitest run test/unit/seed-bible/components/BibleReader.test.tsx` → 61 passed (61). ESLint on the file → clean. `render` and `act` were already imported, so no new imports.

### Does it "just always succeed"?
This PR adds no new test cases or assertions — it is pure teardown hygiene, so it introduces no tautological/always-green tests. The existing 61 tests it protects assert real DOM outcomes (selector opening on title click, not-found/error panels, retry flow, stale-content dimming, skeleton timing) and use `vi.fn` spies, so they remain meaningful. The change makes the run *more* honest by surfacing/clearing async work that was previously leaking across the file rather than being cleaned up.

### Observations (non-blocking)

**Low — informational only, no change requested.** The root cause of one leak lives in the component, not the test: `retryChapterLoad` performs `setRetrying(false)` after an `await` with no unmount guard. In production this is a benign no-op (Preact discards updates to unmounted components), which is why this is not a bug and does not need fixing here. Flagging only so it is a known pattern; the test-layer fix in this PR is the right fix for the flake.

**Low — latent maintenance note.** The teardown only unmounts the shared `container`. If a future test renders into its own locally-created container without unmounting it, this `afterEach` will not catch it and the same class of flake could return. No action needed for this PR; worth keeping in mind for future additions to the file.

### Reproduction note
The original failure was an intermittent unhandled rejection (assertions all passed; only the stray rejection failed the run), so it is inherently hard to reproduce on demand. Rather than chase the race, I confirmed the fix at the source: the timer effect cleanup and the unmount-drops-update behavior together account for both leak paths described in the PR, and the "two un-cleared 500ms timers → zero" measurement is consistent with the `useEffect` cleanup clearing the skeleton timer on unmount.

### Verdict
Correct, minimal, and well-documented. Safe to merge.

---

PR #1610, opened 2026-08-09
added config for disabling multiple tools

AmitPant9999:


Blitzy:
> [!NOTE]

## Code Review — added config for disabling multiple tools

### What this PR does

Adds an optional `disabledToolbarTools?: string[]` field to `BrandingConfig`. When present, `getDefaultToolbarTools()` filters out any default toolbar tool whose `id` appears in that list. `createBibleToolsManager()` now takes a `branding` argument and threads it into the default-tools builder, and the three reader/tools test files are updated to pass a `testBranding` object through the new parameter.

The mechanism itself is small and reasonable. My main concern is that the feature ships with **no test that actually exercises it**, plus a type-safety shortcut. Findings below are ordered by severity.

---

### 🔴 High — the new behavior has no real test coverage

Every test touched by this PR passes `disabledToolbarTools: []` (an empty list). Across all three files:

- `test/unit/seed-bible/managers/BibleToolsManager.test.tsx`
- `test/unit/seed-bible/components/BibleReader.test.tsx`
- `test/integration/seed-bible/components/TabSlotReader.test.tsx`

...the `testBranding` fixture is only used to satisfy the new required parameter so the existing tests still compile. **Nothing asserts that a tool named in `disabledToolbarTools` is actually removed, or that a tool _not_ named is kept.**

This is exactly the "tests that always succeed" case: the filter is the entire point of the PR, yet it is unverified. You could delete the filter, or invert it (`includes` → `!includes`), or misspell the field, and the whole suite would stay green.

1. In `packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx`, change the end of `getDefaultToolbarTools` from
   ```ts
   return tools.filter(
     (tool) => !branding?.disabledToolbarTools?.includes(tool.id)
   );
   ```
   to just `return tools;`
2. Run `pnpm vitest run BibleToolsManager.test.tsx` (and the other two files).
3. All tests still pass — confirming the feature is not covered.

**Suggested fix** — add a test that drives the public API with a non-empty list. The valid main-toolbar IDs include `stop-playing`, `previous-chapter`, `previous-item`, `open-sidebar`, `open-selector`, `open-search`, `open-chat`, `open-plans`, `open-discover`, `share`, `next-chapter`, `next-item`:

```ts
it("omits tools listed in disabledToolbarTools", () => {
  const manager = createBibleToolsManager({
    ...testBranding,
    disabledToolbarTools: ["open-search", "share"],
  });
  const ids = manager.listToolbarTools().map((t) => t.id);
  expect(ids).not.toContain("open-search");
  expect(ids).not.toContain("share");
  // a non-disabled default still registers
  expect(ids).toContain("previous-chapter");
});

it("keeps all default tools when disabledToolbarTools is empty", () => {
  const withNone = createBibleToolsManager(testBranding);
  const withSome = createBibleToolsManager({
    ...testBranding,
    disabledToolbarTools: ["open-search"],
  });
  expect(withNone.listToolbarTools().length).toBe(
    withSome.listToolbarTools().length + 1
  );
});
```

With that mutation from step 1 in place, this test fails as it should.

---

### 🟡 Medium — `branding!` non-null assertion hides a genuinely-optional value

`packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx:480`

```ts
const branding = options.config?.branding; // BrandingConfig | undefined
...
const tools = createBibleToolsManager(branding!);
```

`branding` really can be `undefined`: `options.config` is optional, `config.branding` is optional, and `DEFAULT_APP_CONFIG` has no `branding` key at all (so the common/default path has none). Everywhere else in this same file branding is handled defensively — e.g. line 966 uses `branding?.appName ?? "Seed Bible"`. The `!` here is the odd one out and asserts something that is often false.

It does not crash today only because `getDefaultToolbarTools(branding?: BrandingConfig)` uses optional chaining (`branding?.disabledToolbarTools?.includes(...)`). So the runtime is saved by a coincidence in a different function, while the declared contract of `createBibleToolsManager(branding: BrandingConfig)` (non-optional) is violated. That is a fragile setup — a future reader who trusts the non-optional type and drops the optional chaining would introduce a real crash on the default deployment.

**Reproduction:** call `createSeedBibleState({})` (no config). `branding` resolves to `undefined` and is passed as `branding!`; `strictNullChecks` is silenced rather than satisfied.

**Suggested fix** — make the parameter honestly optional and drop the assertion:
```ts
// BibleToolsManager.tsx
export function createBibleToolsManager(
  branding?: BrandingConfig
): ToolsManager { ... }

// SeedBibleStateManager.tsx
const tools = createBibleToolsManager(branding);
```
(`getDefaultToolbarTools` already accepts `branding?`, so no other change is needed.)

---

### 🟢 Low — filtering scope is narrower than the field name suggests

`disabledToolbarTools` only filters the **main** toolbar (`getDefaultToolbarTools`). The verse toolbar, quick toolbar, empty-slot, and below-reader tool sets are untouched. If main-toolbar-only is the intent, that's fine — but the generic name reads broader than the behavior. Consider a short doc comment on the field in `appConfig.tsx` clarifying that it targets main toolbar tool IDs and that unknown IDs are silently ignored.

---

### 🟢 Low — disabling is only applied to defaults, not enforced on registration

The filter runs once, when the default tool set is built. An extension (or any later `registerToolbarTool` call) using a disabled ID would re-add that tool, since registration does not consult `disabledToolbarTools`. If the goal is only "don't show these defaults," current behavior is correct. If the goal is "this tool must never appear," `registerToolbarTool` would also need to honor the list. Worth a one-line decision either way; no change strictly required.

---

### Notes

- No linked issue or PR description was found, so this review is based on the diff and surrounding code. If there's a tracking issue with acceptance criteria, linking it would help confirm the intended scope (e.g. main-toolbar-only vs. all tool sets).
- CI `build (24.x)` was still in progress at review time; the findings above are independent of the build result.

The functional change is fine and low-risk. The **High** item (real test coverage for the filter) is the one I'd treat as blocking before merge; the **Medium** item is a quick, worthwhile cleanup while you're in there.

---

PR #1611, opened 2026-08-09
Add Compare extension

Blitzy:
Select verses, press **Compare** on the verse toolbar, and a side pane shows those verses in every translation you've chosen — abbreviation left, full name right.

Before, comparing across translations meant switching translation in the reader and losing your place, or opening a second tab.

Ships as an auto-installed extension (`packages/compare-extension/`), not core app code.

## Behaviour

- **The selection is frozen at press time**, not followed live. On mobile the pane renders fullscreen and hides the verse toolbar, so a pane tracking the live selection would blank itself as it opened.
- **The translation you're reading is always first and is never saved.** If it's also in your saved list it's hoisted to the top and shown once. Hoisting is display-only, so switching translations never reshuffles the order you set.
- **First use seeds a default set** so the pane isn't a single lonely block: a curated shortlist for English (AAB, BSB, KJAV, NASB95), otherwise the other translations in your reading language. A list you've explicitly emptied is left alone.
- **Clicking a block header moves the reader to that translation**, landing on the verses being compared and briefly fading the rest of the chapter — the same highlight search results and `?verse=` links use.
- **RTL translations mirror their header** via each translation's own `textDirection`. DOM order is unchanged; `dir` does the reversal.
- Each translation loads independently behind skeleton placeholders. Verses missing from a translation (versification differs) and failed fetches each get their own message rather than an endless skeleton.

## Screenshots

Select verses and use the compare tool from the verse toolbar:
image

Comparing selected verses:
image

Compare settings, you can reorder the translations and also easily remove them:
image

Adding translations to Compare:
image

## Persistence

The saved set is one profile-config key (`compareTranslations`) written through the existing `saveProfileConfigValue` — device-local when logged out, profile when logged in, re-derived when the profile resolves, so a set saved on another device just appears. The list is a `computed` over config rather than a second copy of state.

Writes are debounced into one request per burst, with an optimistic pending value so ticks and chips update instantly.

Worth knowing: `LoginManager` deliberately discards device-local config when an existing account logs in on a device, so an anonymous Compare set doesn't follow you into an established account. Same as every other setting.

## The translation picker is now shared with the reader

Rather than Compare having its own list, the reader's picker was extracted into two reusable pieces:

- **`managers/translationGrouping.ts`** — grouping, search rules, view modes, sort and limit. `BibleSelectorManager` now delegates to it, so the reader's results are unchanged.
- **`components/TranslationList/`** — the list itself. The reader passes its own offline/share row controls through `renderActions`; its modal chrome, filter popover and custom-translation form stay where they were.

Compare's use differs in two ways: it's multi-select, and it drops the per-row offline/share actions.

Net effect on the reader's own files: **716 lines deleted, 131 added**. The styles moved into `TranslationList.css` so they travel with the component, retiring the very generic global `.item` and `.language-list` for namespaced names.

## Core changes

- `components/index.tsx` / `managers/index.tsx` — export `TranslationList`, `Skeleton`/`SkeletonContainer`, `useDragReorder` and `translationGrouping` so extensions can use them.
- `BibleToolsManager.tsx` — the private `extractVerseText` is generalised into an exported `extractVerseContentText`, with the old one kept as a one-line wrapper, instead of duplicating its normalisation in the extension.
- `i18n/en.json` — three new keys for the shared list. English only, per CLAUDE.md.

## Testing

`pnpm check:ts` clean · `pnpm lint` 0 errors · `pnpm test` 4432 passing across 209 files · `pnpm build` succeeds, with the extension code-split into its own chunk (6.4 kB gzipped).

## Notes

- `app/main.tsx` and `LoginManager.tsx` pick up small `import.meta.env.DEV`-guarded debug hooks that are unrelated to Compare — happy to split them out. But will be very useful to have for poking at Seed Bible app state while while running in development mode so I think worth keeping.
- Verse-toolbar priority is 250, putting Compare between Copy and Share.


Blitzy:
> [!NOTE]

Reviewed the diff between the merge-base and head (`826d06ee`) locally, read through `compareState.ts`, `ComparePane.tsx`, `CompareVerses.tsx`, `CompareSettings.tsx`, `TranslationPicker.tsx`, the `translationGrouping.ts` extraction, and the `BibleSelector`/`TranslationList` refactor, and ran the affected test suites (all passing). No linked issue is referenced from this PR body. There are no earlier review comments on this PR to reconcile with.

One note up front on tests: `test/unit/compare-extension/ComparePane.test.tsx` and `compareState.test.ts` both hold up well — they assert on real DOM output, mocked-fetch call counts, debounce timing, and concrete state transitions rather than tautological `not.toThrow()`/`toBeDefined()` checks. `translationGrouping.test.ts` and the updated `BibleSelector.test.tsx` are similarly substantive (concrete fixtures, real expected outputs) — the refactor didn't weaken existing reader test coverage.

### Medium

**1. A failed chapter fetch is silently dropped instead of shown as an error, when only *some* of a multi-chapter selection fails.**
`packages/compare-extension/ext_Compare/CompareVerses.tsx` (around line 127–170, in `TranslationBlock`)

Repro:
- Select verses spanning two chapters (e.g. John 1:51 and John 2:1) and open Compare with a translation added.
- Have the fetch for one of the two chapters fail (network blip) while the other succeeds.

`chapterStates` is then `[loaded, error]`. Neither the "some loading" skeleton branch nor the "every error" failure branch fires, so the code falls through to the normal render. `verses` is built with `flatMap`, which just omits the errored chapter — the user sees only the John 1 verse for that translation, with no error message or retry control, indistinguishable from "this translation simply doesn't have this verse" (the `verse-not-available` case).

Suggested fix: track whether *any* chapter in `chapterStates` has `status === "error"` (not just "every"), and when so, render a partial-failure notice (and ideally a retry limited to the failed chapter) alongside whatever verses did load, instead of silently dropping them.

### Low

**2. Unbounded growth of the `chapters` cache over a session.**
`packages/compare-extension/ext_Compare/compareState.ts` (`chapters` signal, `setChapterState`, ~line 500; `reset()` ~line 682; `dispose()` ~line 687)

Every `translationId|bookId|chapterNumber` ever fetched is added to the `chapters` map and never removed. `reset()` (called on every pane close, via `onClose: () => state.reset()` in `init.tsx`) only clears `snapshot`/`sourceReadingState`; `dispose()` only runs when the extension itself is uninstalled, not per pane-close, and doesn't touch `chapters` either. Over a long session — opening Compare on many different verses, or adding/removing translations repeatedly — this grows without bound.

Suggested fix: clear (or prune to just the current `order`/`snapshot`) the `chapters` map in `reset()`. Since a fresh `snapshot` on the next open re-fetches anyway, clearing on close trades a little re-fetching for bounded memory.

**3. Vacuous-truth edge case in the "every chapter errored" check.**
`packages/compare-extension/ext_Compare/CompareVerses.tsx` (~line 127): `chapterStates.every(chapter => chapter.status === "error")`

If `chapterStates` is ever empty (e.g. a snapshot with `groups: []`, which `test/unit/compare-extension/ComparePane.test.tsx`'s "keeps an Add Translation button anchored..." test constructs directly via `snapshotSelection([])`), `[].every(...)` is vacuously `true`, so the block would render "Could not load this translation." with a Retry button despite nothing ever having been fetched. Not currently reachable through the real UI flow — `openComparePane` in `init.tsx` guards `if (selected.length === 0) return;` before ever calling `snapshotSelection` — so this is latent rather than user-facing today, but worth a defensive `chapterStates.length > 0 && chapterStates.every(...)` guard in case a future caller sets `snapshot` directly (as the test file already does).

### Nitpick — PR description accuracy

**4. The reader's own translation-picker behavior did change slightly, despite the PR description's "the reader's results are unchanged."**
`packages/seed-bible/seed-bible/managers/BibleSelectorManager.tsx` (new `filterTranslationGroups`/`matchingTranslationGroupCount`) and `packages/seed-bible/seed-bible/components/TranslationList/TranslationList.tsx` (~line 726–754)

Two literal, user-visible changes to the existing reader (not the new Compare extension) came along with the extraction:
- The "Load more" control's visibility threshold changed from comparing the paging limit against the size of the *entire unfiltered* catalog (old behavior: in "complete" view or a narrow search, the control stayed visible/clickable long after every match was already shown, and clicking it did nothing until the limit exceeded the ~1000-language catalog) to comparing against the *actual* number of matching results. This is a genuine improvement, not a regression, but it is a behavior change to existing reader UI.
- The reader's load-more control used to render as an unlabeled chevron; it now shows "Show more languages" and a "Showing X of Y" count.

Neither is a bug — the new behavior is more correct and the new copy is a usability improvement — but since the PR frames this refactor as behavior-preserving for the reader, it's worth calling out explicitly in the description so reviewers approve with full knowledge of the (good) side effect. Also worth noting: neither the old nor new load-more threshold logic has direct test coverage, so this wasn't caught by CI in either state — a small test asserting the new visibility threshold would lock in the fix.

---

Everything else — grouping, search/filter matching, sort order and tie-breaks, view-mode gating, per-row selection, auto-expand, and the reader's offline/share `renderActions` (correctly omitted only for Compare's multi-select mode) — checked out as unchanged between the old `BibleSelector.tsx` and the new `translationGrouping.ts`/`TranslationList` split.

The `LoginManager.tsx`/`app/main.tsx` dev-only debug hooks are already called out and justified in the PR description, so no finding there.


---

PR #1612, opened 2026-08-10
Feature/1581

### LINKED ISSUE #1581: Allow bookmarks to belong to multiple categories/folders
## What is happening

Bookmarks are currently forced to belong to a single category (folder) at a time.

## What is expected

Bookmarks should be allowed to belong to any number of categories.

### Schema change

The `category` property of `bookmarkSchema` needs to support both:

- a `string` (for backwards compatibility with existing bookmarks), and
- an `Array<string>` (for bookmarks in multiple categories).

It's important that the schema support **both** forms, not just replace one with the other.

### UI changes

- The bookmark modal should use **checkboxes** instead of radio buttons, so a bookmark can be assigned to multiple categories at once.
- The bookmark modal's title should change to **"Edit bookmark."**
- The **"Move to folder"** option in the bookmark context menu should change to **"Edit bookmark,"** and should simply open the bookmark modal so users can change which categories the bookmark belongs to.

_Mockup of changes to bookmark modal:_
Image

_Mockup of "Move to folder" -> "Edit bookmark":_
Image

Kabi4:
Fixes: #1581


Blitzy:
> [!NOTE]

## Code review — PR No. 1612 (Feature/1581: bookmarks in multiple categories)

Reviewed against issue [#1581](https://github.com/HelloAOLab/seed-bible/issues/1581). Overall this is a solid, careful implementation that meets the issue's requirements:

- ✅ `category` schema now accepts **both** a `string` (legacy/single) and a `string[]` (multi), rather than replacing one with the other — backwards compatibility preserved.
- ✅ Modal switched from radio buttons to **checkboxes** (multi-select).
- ✅ Modal title and the context-menu "Move to folder" item both became **"Edit bookmark"**.
- ✅ `deleteCategory` correctly drops bookmarks that lived *only* in the deleted folder while keeping multi-folder bookmarks in their other folders; `renameCategory` updates every membership; single-membership collapses back to a plain string on write.

I ran the suite and the type checker locally: **`BookmarksManager.test.ts` — 17/17 pass**, and **`pnpm check:ts` — 0 visible errors**. No dangling references to the removed `moveBookmark` / `excludeCategory` remain in source.

### On the tests (real-world behavior?)
The manager tests are genuine — they assert on actual membership arrays, the resulting category list, `recordData` call counts, and order-insensitive no-op behavior. None are trivially-always-true. Good examples: the no-op assertions (`setBookmarkCategories` with the same set does **not** persist a third time) and the delete test that verifies a sole-membership bookmark is removed while a multi-folder one survives in `Favorites`. These would fail if the logic regressed.

The one real gap is that **all** new coverage is at the manager level — there are **no tests for the modal component** (`BookmarkCategoryPickerContent`), which is where the riskiest new behavior lives (see below). See Low No. 4.

---

## Findings by severity

There are no correctness bugs that break the feature. The items below are one behavior change worth a product decision plus some low-severity cleanups.

### Medium

**No. 1 — "Create folder" persists an empty folder immediately, so cancelling the modal leaves an orphan folder behind.**

`handleCreateNewCategory` (`Tabs.tsx:1405`) calls `bookmarks.createCategory(...)`, which persists right away (`BookmarksManager.tsx:706`). Before this PR, a new folder was only written when the bookmark was actually saved (the old `savingToNewCategory` path). Now the folder is committed the moment "Create folder" is clicked, independent of whether the user ever saves a bookmark.

Reproduction:
1. Open the bookmark modal on any verse/chapter.
2. Click **Add to new**, type `Temp`, click **Create folder**.
3. Close the modal with the X or Escape (do **not** click Save).
4. Open the bookmarks sidebar → the empty `Temp` folder is still there and persisted across reload.

This is clearly deliberate (there's a dedicated test, "creates an empty category without assigning any bookmarks", and code comments describing it), so I'm flagging it as a **decision to confirm** rather than a defect. Mitigation: users can delete the stray folder via the folder kebab menu, so it's recoverable. Suggested options, in order of preference:
- Revert to deferred creation — keep the typed name pending in local state and only call `createCategory` inside `handleSave`, so abandoning the modal creates nothing; **or**
- Keep the immediate-create behavior but make the trade-off intentional and documented (it already partly is).

### Low

**No. 2 — Modal title is always "Edit bookmark", even for a brand-new bookmark.**

`openBookmarkCategoryModal` hard-codes the title to `edit-bookmark` for both `add` and `edit` modes (`Tabs.tsx:1614`). Issue #1581 does literally ask for the title "Edit bookmark", so this matches the written request — but titling the first-time *add* flow (from the verse toolbar / chapter button) as "Edit bookmark" reads slightly off. Worth a quick confirmation with @Blitzy on whether add vs. edit should show different titles.

**No. 3 — Once you click "Add to new", you can't Save your existing checkbox selection without backing out.**

While `isAddingNew` is true, the action button renders as **Create folder**, not **Save** (`Tabs.tsx:1553`). A user who clicks "Add to new", changes their mind, and just wants to save the folders they already ticked has to press Escape (or clear the input) to get the Save button back. Minor friction; consider showing Save alongside Create, or auto-exiting add mode when the input is empty.

**No. 4 — No component-level tests for the modal flow.**

The multi-select toggling, the create-then-keep-selecting flow, the `canSave`/`canCreateNew` gating, and the empty-folder-on-cancel behavior in No. 1 are all untested. Since the modal is the most behavior-heavy new surface, a couple of `BookmarkCategoryPickerContent` render tests (toggle a checkbox, create a folder, verify Save calls `addBookmark`/`setBookmarkCategories` with the expected set) would lock this down.

**No. 5 — Stale i18n keys.**

`move-bookmark` ("Move to folder") and `move-to-new` are no longer referenced anywhere in source now that the context menu uses `edit-bookmark`, but they remain in `en.json` (and every locale). Not harmful, but consider removing `move-bookmark` / `move-to-new` from `en.json` to keep the key set tidy (per CLAUDE.md, leave the other locale files to the translators).

**No. 6 — Accessibility nit.**

The checkbox container uses `role="group"` (`Tabs.tsx:1463`) without an accessible name. Adding `aria-label` (e.g. the "Folders" label) would give screen-reader users context for the group independent of the modal title.

---

### Verdict
Functionally correct, backwards-compatible, and well-tested at the manager layer. The only item that warrants a real decision before merge is **No. 1** (empty-folder-on-cancel). The rest are low-severity polish.

---

KallynGowdy:
@Kabi4 There's one other thing I found: 

- Removing a bookmark removes it from all folders. Steps to reproduce:
    1. Add a bookmark to multiple folders
    2. Display the bookmarks in the sidebar
    3. Click the three dots menu on the bookmark you just added in one of the folders
    4. Click "Remove bookmark"
    5. See that it is removed from all folders even though you clicked on only one of the bookmarks
    6. See [this video](https://youtu.be/QXrNKZ0ka9g) for an example

### What I expected
That the bookmark is only removed from the folder that it was in. If after removing it from the folder it isn't in any folders, then it should be removed completely.

KallynGowdy (approved):

PR #1614, opened 2026-08-10
Feature/1501

### LINKED ISSUE #1501: Option to install app in settings menu does not reappear
## What is happening

After installing the app as a PWA, if you the  uninstall the PWA and then go to seedbible.org in the browser, the option to install as app does not reappear.

## What I expected

I expect the Install App option to appear when we are no longer have the app installed.

## Notes

I'm not sure this is even possible, will need to investigate how install state is tracked and if browsers are even able to tell. This is very likely some kind of local storage key that is hiding the button if I had to guess so there likely isn't any kind of "is the app installed" check occuring if it's at all possible?

Kabi4:
Fixes: #1501 

We only treat the app as installed when:

you’re actually using the installed app (standalone mode), or
you just finished installing it in this same visit
We no longer save “installed” forever in local storage or the profile for this purpose. “Maybe later” still works the same.

Trade-off
If someone still has the app installed but opens the website in a normal browser tab, they might see Install app again. That’s okay — better than the option never returning after uninstall.


Blitzy:
> [!NOTE]

## Code Review — PR No. 1614 (Fixes No. 1501)

### Summary

This is a clean, well-scoped fix that correctly addresses the issue. Before, "app installed" was cached in `localStorage` (`sb-app-installed`) and mirrored to the user's profile (`appInstalled`), and a standalone session even wrote that flag to the backend. Because browsers have no reliable "the PWA was uninstalled" signal, those sticky flags hid the Settings **Install App** entry forever after a single install — which is exactly what issue No. 1501 reported.

Now `installed` is derived only from the **current session**: `true` when the page is running standalone (installed-PWA display mode) or right after `markInstalled()` in this page load, and never persisted. Since the Settings entry is gated on `!state.onboarding.installed.value` (`SettingsPage.tsx:2088`), uninstalling the PWA and reopening in a normal browser tab makes `isStandalone()` false → the entry reappears. This matches the approach suggested in the issue thread (detect via the `display-mode` media query rather than sticky storage).

I confirmed the cleanup is complete: no remaining references to `sb-app-installed` / `APP_INSTALLED_KEY` / `appInstalled` / `PROFILE_APP_INSTALLED` anywhere in `packages/` or `test/` (aside from the new regression test), the now-unused `effect` import was removed, and the documented trade-off (a browser tab may re-offer install to someone who still has it installed) is accurately described. The full onboarding suite passes locally (6/6).

**No correctness bugs found.** All findings below are about test coverage — the tests, as written, would stay green even if the core mechanism of this fix regressed.

---

### Findings by severity

#### Medium — the fix's core mechanism (`isStandalone()`) is never actually exercised by a test

`installed` now hinges entirely on `isStandalone()` reading `matchMedia("(display-mode: standalone)")` / `navigator.standalone`. But under jsdom `window.matchMedia` is `undefined`, so `isStandalone()` **always returns `false`** in the test environment. Every current test therefore runs the not-standalone branch only. The suite has no test that asserts the standalone case (`installed === true`, Settings entry hidden), which is the exact behavior the whole PR pivots to.

Consequence: if someone later typo'd the media-query string, broke the `navigator.standalone` check, or short-circuited `isStandalone()`, **all onboarding tests would still pass** — the suite gives false confidence about the primary behavior.

Reproduction (why the gap is real): temporarily change `isStandalone()` to `return false` unconditionally and run `pnpm vitest run OnboardingManager` — the suite stays green, even though a standalone PWA session would then incorrectly show install affordances.

Suggested fix — add a test that mocks standalone and asserts the installed path:

```ts
it("treats a standalone (installed-PWA) session as installed", () => {
  const mql = { matches: true, addEventListener() {}, removeEventListener() {} };
  vi.stubGlobal("matchMedia", vi.fn().mockReturnValue(mql));

  const onboarding = createOnboardingManager(createLogin());

  expect(onboarding.installed.value).toBe(true);
  expect(onboarding.installAvailable.value).toBe(false);

  vi.unstubAllGlobals();
});
```

#### Low — the profile-persistence half of the fix has no regression test

The new test (`does not treat a leftover localStorage install flag as still installed`) is a genuine regression guard — I verified it fails against the pre-fix code (the old `signal(standalone || readFlag(APP_INSTALLED_KEY))` would make `installed` true). Good.

But the fix removed **two** sticky sources, and only the `localStorage` one is covered. For a logged-in user the more impactful half was the profile: the old code read `appInstalled` from `profile.config` and an `effect` wrote it to the backend on any standalone session. Nothing now asserts that a profile carrying `appInstalled: true` is ignored, so a future refactor could quietly reintroduce profile reading and no test would catch it.

Suggested fix — mirror the localStorage test for the profile path:

```ts
it("ignores a legacy appInstalled flag persisted on the profile", () => {
  const login = createLogin();
  (login.profile as any).value = { config: { appInstalled: true } };

  const onboarding = createOnboardingManager(login);

  expect(onboarding.installed.value).toBe(false);
  expect(onboarding.installAvailable.value).toBe(true);
});
```

(`getProfileConfigValue` reads `profile.config[key]`, so seeding `profile.value = { config: { appInstalled: true } }` reproduces the legacy state.)

#### Low / informational — documented trade-off worth confirming with the team

Because `installed` is session-only, a logged-in user who installed the app on a previous visit and now opens the site in a **normal browser tab** will be re-offered install (both the Settings entry and, if not previously dismissed, the auto-prompt). The PR body calls this out and it's the accepted cost of fixing No. 1501 — flagging only so a reviewer signs off on the UX consciously.

One pre-existing detail (not changed by this PR, so not a blocker): in `Onboarding.tsx` `handleInstall()`, the `finally { onboarding.dismissInstall() }` runs on every outcome, and `dismissInstall()` persists `installPromptDismissed: true` to localStorage + profile permanently. So after an accepted install → uninstall → browser reopen, the **Settings entry returns** (issue fixed) but the **auto-prompt won't** re-trigger, because `installAvailable = !installed && !dismissed` and `dismissed` stays sticky. That's consistent with "Maybe later works the same," just worth being aware the two affordances diverge by design.

---

### Verdict

Solid fix that resolves the reported bug with minimal, readable changes and accurate comments. Nothing blocks correctness. I'd recommend adding the standalone-path test (Medium) before merge so the suite actually protects the behavior this PR introduces; the profile-path test (Low) is a nice-to-have that closes the other half of the regression surface.

---

PR #1615, opened 2026-08-10
fix: enhance RTL support and localization for various components

### LINKED ISSUE #1613: Arabic / RTL localization fixes
The Arabic experience is mostly good, but a few localization and RTL issues still need to be fixed.

### Required fixes

* In Arabic, the navigation arrow shown in the tutorial should point **right**, not left.
* Replace the Arabic translation for **Tutorial** with:

  * `شرحاً إرشادياً`
* Localize the two mode labels shown in the mode selector.
* Replace the Arabic translation for **Footnotes** with:

  * `اظهر الهوامش`
* Replace the Arabic translation for **Highlights** with:

  * `عرض الآيات المميّزة`
* Localize all tutorial step titles/text shown during the walkthrough.
* Fix the tabs layout/alignment issue in RTL mode.

### Acceptance criteria

* All mentioned UI strings are properly localized in Arabic.
* Tutorial navigation direction is correct for RTL.
* Tutorial steps contain no untranslated English text.
* Tabs render and align correctly in RTL.
* No regressions are introduced in the LTR/English experience.

mazenemad:
fixes #1613


Blitzy:
> [!NOTE]

## Code Review — PR No. 1615: "fix: enhance RTL support and localization for various components"

Reviewed against linked issue No. 1613 (Arabic / RTL localization fixes) and its parent QA issue No. 1544. Overall this is a clean, well-scoped change that meets the acceptance criteria, and the inline comments explaining each RTL decision are genuinely helpful. The findings below are mostly about the safety net and test coverage rather than functional defects — I did not find a runtime bug in the RTL or i18n behavior.

> Note: I reviewed statically (the sandbox had no `node_modules`, so I could not run `pnpm lint` / `check:ts` / `test` here). Please rely on CI for the final lint/type/build gate.

### What I verified as correct

- **Acceptance criteria coverage.** Arrow direction (Tutorial), the three exact Arabic strings, the two theme/mode labels, all tutorial step titles/bodies, and the RTL tabs layout are all addressed.
- **Arabic strings match the issue verbatim** — `شرحاً إرشادياً` (launch-tutorial), `اظهر الهوامش` (show-footnotes), `عرض الآيات المميّزة` (show-highlights).
- **All 24 tutorial keys referenced in `TutorialManager.tsx` exist in both `en.json` and `ar.json`**, plus `theme-light` / `theme-dark`. No duplicate keys; both JSON files parse.
- **The `[dir="rtl"]` selectors actually take effect.** `.sb-app-root` sets `dir` from `isRtl` (`app/main.tsx:135`) and the `BibleSelector` portal sets its own `dir` (`BibleSelector.tsx:102`), and `<Tutorial>` renders inside `.sb-app-root`, so both `.rtl-mirror` and `[dir="rtl"] .sb-tour-btn-next` resolve.
- **No LTR regression from the logical-property swaps.** In `Tabs.css`, `padding-inline: 0.75rem 0.375rem` and `margin-inline-start`/`text-align: start` are exact LTR equivalents of the old physical values, and mirror correctly in RTL.
- **`.rtl-mirror` chevrons are safe.** `MaterialIcon` forwards `className` (`icons.tsx:1364`), and none of the five chevron/separator classes carry their own `transform`, so there's nothing for the RTL flip to clobber.

---

### Findings by severity

#### 🟠 Low–Medium — The error-level lint rule does not validate the new "key-as-data" tutorial keys

`getTranslationUsageStats.ts` now teaches the **usage** extractor about `titleKey`/`bodyKey`, which feeds `translation-unused-keys` (a *warning*). But the `translation-missing-keys` rule — the one configured as **`error`** (`eslint.config.mts:137`) — only inspects `t(...)` `CallExpression`s and tool-title objects (`i18nMissingKeysRule.ts:312`, `:370`). It never looks at tutorial-step `titleKey`/`bodyKey`.

Consequence: if a developer adds a new tutorial step (or mistypes a key) whose `titleKey` is **absent from `en.json`**, lint stays green, and at runtime `t(step.titleKey, { defaultValue: step.titleDefault })` silently falls back to the English default. That is exactly the failure mode this PR set out to eliminate ("Tutorial steps contain no untranslated English text") — only now it can regress with no automated signal.

- Reproduction (for a human dev, after this PR lands):
  1. In `TutorialManager.tsx`, change one step's `titleKey` to a key that is not in `en.json` (e.g. `"tutorial.doesNotExist"`).
  2. Run `pnpm lint`.
  3. Observe: no `missing_key` error is raised, even though the key is unresolved.
- Suggested fix: extend `i18nMissingKeysRule` to also read tutorial-step `titleKey`/`bodyKey` (reuse the same `getObjectPropertyStaticString` helper the usage extractor uses) so a missing key is an error, symmetric with the usage side. At minimum, add the test described in the next finding so the gap is documented.

#### 🟠 Low–Medium — No tests for the new logic

The PR adds two pure, easily-testable pieces and no tests for either:

- `getTutorialStepTranslationKeys` in `script/getTranslationUsageStats.ts` — the i18n lint tooling depends on this being correct, yet nothing guards it.
- `localizedThemeName` in `SettingsPage.tsx`.

Given the routine's emphasis on tests reflecting real behavior: a regression test here would be high-value and cheap. A good test would parse a small in-memory source snippet containing a tutorial-step object literal and assert that `getTranslationUsageStats` reports those keys as used — plus a negative case (an object literal without `titleKey`/`bodyKey`) to prove it isn't matching everything. That negative case is what stops the test from "always succeeding."

#### 🟡 Low — Extractor matches `titleKey`/`bodyKey` on *every* object literal project-wide

`getTutorialStepTranslationKeys` is applied to every `ObjectLiteralExpression` in `packages/**` and keys off two generic property names. Today only tutorial steps use them (confirmed — the only other `titleKey`/`bodyKey` references are the `t(...)` calls in `BibleSelector.tsx`), so it's correct right now. But it silently couples the i18n tooling to those names: any future unrelated object with a static `titleKey`/`bodyKey` string would be counted as a "used" translation key and could mask a real unused-key warning.

There's also a sharp edge inherited from the shared helper: `getObjectPropertyStaticString` returns the property *name* for a `ShorthandPropertyAssignment` (`getTranslationUsageStats.ts:168-174`). So a shorthand `{ titleKey }` would record the literal string `"titleKey"` as a used key — garbage, but harmless-until-it-isn't.

- Suggested fix: scope the match to the tutorial-step shape rather than the bare property names — e.g. only treat an object as a tutorial step when it also has the sibling `titleDefault`/`bodyDefault` (or `target`/`group`) properties. This keeps the intent explicit and removes the accidental global coupling.

#### ⚪ Informational — Direct edits to `ar.json`

`CLAUDE.md` asks that translation-key changes touch only `en.json` and leave other locale files "for the professionals." This PR edits `ar.json` directly. I'm flagging it only for awareness, not as a defect: the linked issue provided the exact Arabic strings, the PR author is the team member who filed the issue in Arabic, and the strings match the issue verbatim — so this reads as authoritative rather than machine-translated. No action needed unless the team wants to keep the convention strict.

---

### Minor notes (non-blocking)

- The `dir="ltr"` on the tutorial "Next" arrow span is belt-and-suspenders: `→` (U+2192) is not bidi-mirrored by the algorithm, so the real fix is the `flex-direction: row-reverse` in `Tutorial.css`. The extra attribute is harmless and defensible as future-proofing, matching the comment's intent.
- Relocating `dir` from `.sb-tab-row` to `.sb-tab-main-content` is an intentional, sound design choice (card chrome follows the UI direction; only the passage label follows the translation's direction). Worth a QA glance in the LTR-UI + RTL-translation combination to confirm the label still right-aligns as expected.

### Verdict

Functionally solid and meets issue No. 1613. Nothing here needs to block merge, but I'd encourage (1) closing the missing-key lint gap so the new key-as-data pattern is protected the same way `t(...)` calls are, and (2) a small unit test for the extractor so this tooling doesn't drift silently.

---

KallynGowdy:
@mazenemad Can you address the findings from the code review?

mazenemad:
On it

KallynGowdy:
@mazenemad Are you still working on this?

mazenemad:
already worked on it will push the updates
i forgot to add it yasterday

Blitzy:
> [!NOTE]

## Follow-up review — commit `e59efd9` ("fix: add tutorial step translation key extraction and tests")

This is a follow-up to the review posted on 2026-08-10 (comment above), after Kallyn asked the author to address its findings and commit `e59efd9` was pushed in response. Re-reviewed against linked issue No. 1613 and re-checked the two findings that were rated 🟠 Low–Medium.

> Note: reviewed statically again — no `node_modules` in the sandbox, so `pnpm lint` / `check:ts` / `test` could not be run here. Please rely on CI for the final gate.

### Previously-flagged findings — status

- **"Lint rule doesn't validate tutorial-step keys" — Resolved.** `script/eslint/i18nMissingKeysRule.ts` now has `getTutorialStepTranslationInfo`, wired into the `ObjectExpression` visitor, which reports the `missing_key` error (the rule configured as `error` in `eslint.config.mts`) when a tutorial step's `titleKey`/`bodyKey` isn't in `en.json`. This closes the exact gap called out before: a mistyped or new tutorial step key is now a lint error, not a silent runtime fallback.
- **"No tests for the new logic" — Mostly resolved.** `test/unit/script/getTranslationUsageStats.test.ts` and `test/unit/seed-bible/components/localizedThemeName.test.ts` are both new. I read through both: each includes a real negative case (an object literal that has *some* but not all of the required shape, and one that has unrelated `titleKey`/`bodyKey` names) rather than only positive/always-pass assertions, so they do exercise real behavior, not just happy-path text.

### New finding

#### 🟡 Low — The actual bug-catching code path (the lint rule fix) still has no test of its own

The two new test files cover `getTutorialStepTranslationKeys` (in `getTranslationUsageStats.ts`, which only feeds the *unused-key warning*) and `localizedThemeName`. Neither test file touches `getTutorialStepTranslationInfo` in `i18nMissingKeysRule.ts` — the function that is the actual fix for the previous finding, i.e. the one that turns a missing tutorial-step key into a build-breaking `error`.

- Why it matters: `getTutorialStepTranslationInfo` reimplements the same "all four properties or bail" matching logic independently (different AST library — `TSESTree` vs. `ts-morph` — so it can't just call the other function), sharing only the `TUTORIAL_STEP_TEXT_PROPERTIES` name/property-pairing constant. If a future edit to one copy (e.g., changing the bail-out semantics, or the property order) isn't mirrored in the other, the missing-key *error* and the unused-key *warning* could silently disagree about what counts as a tutorial step, and nothing would catch that.
- Reproduction (for a human dev): temporarily swap the loop body in `getTutorialStepTranslationInfo` so it reports on the *first* property match instead of requiring all four (i.e., reintroduce the exact false-positive bug the shape-matching was designed to prevent) — no test fails.
- Suggested fix: none of the other i18n ESLint rules in `script/eslint/` have direct unit tests either (this repo tests them by running lint against the real project, not via `RuleTester`), so this isn't a new gap introduced by this PR so much as an existing convention this PR could have closed while it was already touching the file. A small `RuleTester`-based test (or exporting `getTutorialStepTranslationInfo` for a direct unit test, mirroring what was done for `getTutorialStepTranslationKeys`) would remove the drift risk cheaply.

### Still open from the original review (informational, non-blocking)

- The shape match is still keyed off the bare property names `titleKey`/`bodyKey` rather than a more specific tutorial-step marker (e.g. requiring the sibling `target`/`group` fields too). Still correct today (no other object literals in the project use these names), just an accepted, low-risk coupling that wasn't tightened.
- The direct `ar.json` edits (vs. `CLAUDE.md`'s "only touch `en.json`" convention) — unchanged, still not a defect given the author is the Arabic-speaking issue filer and the strings match the issue verbatim.

### Verdict

The two 🟠 Low–Medium findings from the first pass are functionally addressed and the fix is sound. The one new note above is low severity and non-blocking — nothing here should hold up merge.

---

PR #1617, opened 2026-08-10
Cache API responses server-side and seed client to avoid refetches

KallynGowdy:
## Summary

This change eliminates redundant API requests when the client hydrates after server-side rendering. The server now captures all API responses it fetches during SSR, embeds them in the page as JSON, and the client uses them to pre-populate its own API cache instead of re-fetching the same data.

## Key Changes

- **API response caching in `FreeUseBibleAPI`**: Added `snapshotResponseCache()` and `seedResponseCache()` methods to capture successfully resolved responses and pre-populate a fresh instance's cache. Only successful responses are snapshotted; failed requests are excluded.

- **Server-side WebKit detection**: Moved the `isWebKit()` check from client-only to run on the server during request handling. The result (`renderedAsWebKit`) is passed through `AppConfig` so the client can use it as a fallback when `navigator` is unavailable (e.g., during SSR hydration).

- **WebKit detection refactored**: Extracted `isWebKitUserAgent()` into a reusable function in `ssrEnv.ts` (client) and duplicated in `server/index.ts` (server can't import from packages). The client's `isWebKit()` now prefers the live `navigator.userAgent` but falls back to the SSR-computed value.

- **Component structure refactored**: Split `Main` into three components (`Main`, `MainWithState`, `MainCreatingState`, `MainBody`) to avoid conditionally calling `useMemo` across re-renders, which would violate Preact's hook invariant. Each component now has a fixed, internally-consistent hook sequence.

- **API response injection**: The server embeds the snapshotted API responses as JSON in a `<script id="app-seed-data">` tag. The client reads this via `readInjectedApiResponseSnapshot()` and passes it to `createSeedBibleState()`, which seeds the `FreeUseBibleAPI` instance before any requests are made.

- **Tests added**: New test file `ssrEnv.test.ts` validates WebKit detection across Safari, desktop Chrome, and Chrome on iOS. Extended `FreeUseBibleAPI.test.ts` with tests for snapshot/seed behavior. Updated server tests to verify `renderedAsWebKit` is computed correctly.

## Implementation Details

- The `_resolvedResponses` map in `FreeUseBibleAPI` is kept separate from the promise cache (`_responseCache`) because promises can't be read back synchronously — the snapshot needs actual values for embedding in HTML.

- `seedResponseCache()` never overwrites an already-cached URL, so if the client makes a request before hydration completes, the seeded value won't override the in-flight result.

- The WebKit check is duplicated between client and server because the server bundle must remain dependency-free of the app's browser-oriented code.

- The refactored component structure ensures that whether `initialState` is provided (SSR hydration) or not (client-side creation) doesn't change which hooks are called on each render.

https://claude.ai/code/session_01T2HcadY9uu84Lh5VSVLZ57


Blitzy:
> [!NOTE]

Reviewed the diff (14 files, +352/-28), the SSR/server code paths it touches, and the test suite. This PR has no linked issue. Overall the API-cache-snapshotting mechanism (`snapshotResponseCache`/`seedResponseCache`) is well-scoped and its tests genuinely exercise real behavior (success vs. failure, no-refetch-after-seed, no-override-of-cached-value). The `Main`/`MainWithState`/`MainCreatingState` split is also a correct fix for the pre-existing conditional-`useMemo` hook-order hazard. One part of the change, however, doesn't actually do what it says on the box in production — see High below.

### High — server-side WebKit detection never actually uses the per-request value, because Bun's runtime polyfills a global `navigator`

**Where:** `packages/seed-bible/seed-bible/app/ssrEnv.ts`, the new `isWebKit()`:

```ts
export function isWebKit(ssrRenderedAsWebKit: boolean): boolean {
  return typeof navigator === "undefined"
    ? ssrRenderedAsWebKit
    : isWebKitUserAgent(navigator.userAgent);
}
```

The whole point of this PR's WebKit refactor is to fix the *old* bug (webkit-class was computed once, at module-load time, from whatever `navigator` happened to be when the SSR bundle was first imported — a value that then stuck for every request served by that process). The fix routes a correctly-computed-per-request `renderedAsWebKit` through `AppConfig`. But `isWebKit()`'s guard for "are we actually in a browser" is `typeof navigator === "undefined"`, and this project's server (`pnpm dev` and the built server bundle) runs under **Bun**, not Node — and Bun ships a global `navigator` object (`{ userAgent: "Bun/x.y.z", platform, hardwareConcurrency }`) on every runtime, including the SSR render path invoked from `server/index.ts`. I confirmed this directly:

```
$ bun -e 'console.log(typeof navigator, navigator.userAgent)'
object Bun/1.3.11
```

So on the server, `typeof navigator === "undefined"` is **always false**, meaning the function always falls into the `else` branch and evaluates `isWebKitUserAgent("Bun/1.3.11")` — which never matches the Safari/iOS regexes — instead of using the `ssrRenderedAsWebKit` argument that was correctly derived from the real request's `User-Agent` header. Net effect: every server-rendered page, for every visitor (including genuine Safari/iOS users), gets `webkitClass = ""` from SSR. It's the same *category* of bug this PR sets out to fix (a stale/fake `navigator` deciding the class instead of the real client), just reintroduced through a different code path. The visible symptom is a flash of un-Safari-patched layout on first paint that self-corrects only once the client-side `render()` call re-renders with the real browser's `navigator.userAgent` (there the check works correctly, since a real browser has no `userAgent` collision with Bun's fake one).

**Repro steps:**
1. `bun server/index.ts` (i.e. `pnpm dev`), or run the built server bundle.
2. Request any page with a Safari/iOS `User-Agent` header (e.g. `curl -H "User-Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 17_0 like Mac OS X) AppleWebKit/605.1.15 ..." http://localhost:<port>/en/AAB/genesis/1`).
3. Inspect the returned HTML — the element that should carry `is-webkit` (via `MainContent`'s `webkitClass`) never does, regardless of the UA sent.

**Suggested fix:** don't use `typeof navigator === "undefined"` as a proxy for "are we in a real browser" — it no longer holds under Bun (and is increasingly unreliable under Node too, which has been adding spec globals). Check something that's actually browser-exclusive, e.g. `typeof document === "undefined"` / `typeof window === "undefined"`, or just always prefer the SSR-computed value when running inside the server bundle rather than re-deriving it from `navigator`.

**Test gap:** the new `ssrEnv.test.ts` case "falls back to the SSR-derived value when there is no live navigator" only passes because it does `vi.stubGlobal("navigator", undefined)` to force that branch. That scenario doesn't occur in any real environment this code runs in — not the Bun server (navigator is always an object), and not the Vitest suite either, since `vite.config.ts` sets `environment: "jsdom"`, and jsdom also provides a real `navigator` object by default. So the test can't and doesn't catch this regression; it only proves the ternary's true-branch returns its argument, which was never in question.

### Low — the `.replace()` placeholder-substitution chain is vulnerable to `$`-pattern corruption, and this PR adds a second occurrence of it

**Where:** `standalone/entry-ssr.tsx`, the final `return` in `render()`:

```ts
html: options.html
  .replace("<!-- META -->", metaHtml)
  .replace("<!-- CONFIG_JSON -->", configJson)
  .replace("<!-- SEED_JSON -->", seedJson)
  .replace("<!-- APP_HTML -->", appHtml),
```

`String.prototype.replace(search, replacement)` treats `$&`, `` $` ``, `$'`, `$$`, and `$1`–`$99` specially in the *replacement* string even when `search` is a plain string rather than a `RegExp` — this applies to all four calls here, not just the new one. `configJson`/`appHtml`/`seedJson` are all built from live Bible translation content this project doesn't control the source of (modern-English paraphrases in particular are prone to things like dollar amounts in footnotes). If a translation's JSON ever contains a literal `$1`, `$&`, etc., that specific `.replace()` call silently corrupts the output (re-inserting the matched placeholder text, or dropping a capture group that doesn't exist) instead of throwing — the resulting page would ship broken HTML/JSON with no error logged.

This pattern already existed for `META`/`CONFIG_JSON`/`APP_HTML` before this PR, so it's not new, but this change adds a second copy of the same raw chapter/translation text (`seedJson`, unescaped-for-HTML JSON rather than rendered HTML) to the exposed surface, which — being rawer text — is if anything more likely to contain a bare `$` next to a digit than the already-HTML-escaped `appHtml` is.

**Suggested fix:** swap the chained `.replace(literal, value)` calls for `.split(literal).join(value)` (immune to `$`-substitution), or a function-based replacer (`.replace(literal, () => value)`), while touching this code.

### Test coverage — otherwise solid, one shallow spot

The `entry-ssr.test.ts` addition ("injects the fetched API responses into the #app-seed-data JSON script tag") only asserts that a URL substring shows up in the seed data — it doesn't round-trip a chapter payload containing a "risky" character (like a bare `$` before a digit) through the actual `render()` → HTML → `JSON.parse` path, so it wouldn't have caught the Low finding above either. Worth adding once the `.replace` fix lands, as a regression guard.


---

KallynGowdy:
Fixed both findings in bfabaa82:

- **High** — `isWebKit()` now guards on `typeof document === "undefined"` instead of `typeof navigator === "undefined"`. Confirmed the repro logic: Bun's global `navigator` (`{ userAgent: "Bun/x.y.z" }`) made the old guard always false server-side, so it always evaluated `isWebKitUserAgent("Bun/...")` (never a match) instead of the correct per-request `renderedAsWebKit`. `document` isn't defined by Bun or Node, so it correctly distinguishes "real browser" from "server." Added a regression test that stubs a Bun-shaped global `navigator` alongside `document: undefined` to catch this specifically, plus fixed the now-incorrect "no live navigator" test (jsdom always provides a real `document`, so the old assertion no longer exercised the code path it claimed to).
- **Low** — swapped the chained `.replace(literal, value)` calls in `entry-ssr.tsx` for a `split`/`join`-based substitution, immune to `$1`/`$&`-style corruption from live translation content. Added a round-trip regression test that fetches a chapter engineered to contain a literal `$1` and verifies it survives `render()` → HTML → `JSON.parse` unchanged.

`pnpm check:ts`, `pnpm lint`, and `pnpm test` (204 files / 4327 tests) all pass.

---

Blitzy (approved):

PR #1618, opened 2026-08-10
Make chapter navigation links crawlable and middle-clickable

KallynGowdy:
## Summary

Chapter navigation (previous/next) now renders as real `<a href>` links instead of buttons when a destination can be named with certainty. This makes them followable by web crawlers and gives readers the standard link affordances — middle-click to open in a new tab, right-click to copy the address, etc. A plain click still navigates in-app without a page reload.

## Key Changes

- **New `ToolActionElement` component** — A toolbar action that renders as a link when it has an `href`, and as a button when it doesn't. A plain click is intercepted and handled in-app; modifier keys (cmd/ctrl/shift/alt) and middle-click are left to the browser.

- **Chapter position tracking** — Added `nextChapterPosition` and `previousChapterPosition` signals to `BibleReadingState`. These return the adjacent chapter's book, chapter number, and translation ID — but only when it can be named with certainty. They return null at the end of the canon, before the book catalog has loaded, or when a reading extension owns that navigation direction (and may send it anywhere).

- **Chapter tool hrefs** — The previous/next chapter tools now have a `getHref` that builds a canonical URL from the adjacent position. The URL follows the translation's language (not the UI language), includes the deployment path prefix, and has no query string — matching the same rules as `canonicalUrl` in `SeedBibleStateManager`.

- **Server-side suspend** — The toolbar now suspends on the initial chapter load and book catalog arrival during SSR. Without this, the server would render the chapter page before the catalog arrived, leaving the prev/next arrows as disabled buttons with no links out — exactly what a crawler would see. The suspend is client-side no-op (browsers fill in the links as soon as the catalog lands, with no render to block).

- **Updated `BibleReaderToolbar`** — Now passes `navigation` to the tools context and uses `ToolActionElement` for the floating nav arrows.

## Implementation Details

- The chapter position resolver (`resolveAdjacentPosition`) is stricter than the availability check: it needs an actual book and chapter number, not just "can we move?" It returns null if an extension owns the direction, because that extension may navigate anywhere.

- The `chapterToolHref` function derives the language the same way `canonicalUrl` does: from the translation's language tag, falling back to the default for that translation, then to the UI default. This ensures a Spanish translation read through an English interface still points to the Spanish page, avoiding unnecessary redirects.

- The SSR suspend uses `initialCatalogSettled` (a new latch that fires when the book catalog arrives or the timeout expires) alongside the existing `initialChapterLoadSettled`. Both must be true before the toolbar renders, ensuring the chapter links are ready.

- Tests cover the new component, the position tracking across book boundaries and at the canon's end, the href generation with language and path prefix handling, and the server-render behavior.

https://claude.ai/code/session_013Ve9ZrEKRg2nsE5UJ1fwLp


Blitzy:
> [!NOTE]

I read through the diff, the two new signals (`initialChapterLoadSettled` / `initialCatalogSettled`) in `BibleReadingManager.tsx`, the new `ToolActionElement` component, and the added tests. No linked issue is referenced in the PR body, so this review is based on the diff and the PR description alone. Overall the design is solid and well-explained, but there's one race condition that undercuts the core promise of the PR (crawlable links land in the SSR HTML), plus a test gap and a smaller edge case worth a look before merge.

## High severity

**The SSR suspend check only waits on one of the two latches it needs, so the exact bug this PR fixes can still happen.**

`BibleReaderToolbar.tsx` decides whether to suspend during SSR like this:

```tsx
if (
  import.meta.env.SSR &&
  !readingState.value.initialChapterLoadSettled.value
) {
  throw readingState.value.chapterDataPromise;
}
```

But `chapterDataPromise` only resolves once **both** `initialChapterLoadSettled` **and** `initialCatalogSettled` are true (`BibleReadingManager.tsx`, the effect right after the catalog-latch effect). The guard above never reads `initialCatalogSettled` — and it can't, because that signal isn't exposed on `BibleReadingState` at all (it's a local closure variable in `createBibleReadingState`).

The catalog fetch (`dataManager.getTranslationBooks`, fired from `requestContent`) and the chapter fetch are two independent, un-awaited network requests. The PR's own comment acknowledges this is only usually true: "that request is issued before the chapter's own and returns the smaller payload, so it is **normally** already home." Whenever it isn't — a slower catalog response, a cold cache for that translation, request contention — `applyChapterContent` sets `initialChapterLoadSettled.value = true` (line ~2006) while the catalog is still in flight. At that instant the guard above sees `initialChapterLoadSettled === true`, does **not** throw, and the toolbar renders immediately with `translationBooks.value` still null — which is exactly the "disabled buttons, no links for crawlers" bug described in the PR's own rationale.

**Repro (conceptual, since it's timing-dependent):** in a test/harness, make the chapter response resolve before the catalog response (e.g. delay whatever mock backs `dataManager.getTranslationBooks`), then call `renderToStringAsync` on `<BibleReaderToolbar>` with `import.meta.env.SSR = true`. Expect: the rendered HTML has no `<a>` for next/previous chapter (or has a stale/incorrect one), even though the chapter itself renders.

**Suggested fix:** expose `initialCatalogSettled` on `BibleReadingState` (or better, expose a single derived "ready" signal that ANDs both latches) and check it alongside `initialChapterLoadSettled` in the guard:

```tsx
if (
  import.meta.env.SSR &&
  (!readingState.value.initialChapterLoadSettled.value ||
    !readingState.value.initialCatalogSettled.value)
) {
  throw readingState.value.chapterDataPromise;
}
```

## Medium severity

**The new SSR test can't catch the race above, so it gives false confidence.**

`BibleReaderToolbar.test.tsx`'s "puts the chapter links in the server-rendered HTML" test builds a `warm` state via `createTestSeedBibleState` first, then a second `state` via `createSeedBibleState`, and renders that. If `dataManager`'s translation-books cache is shared/warmed by the first call (or the mocked responses simply resolve in the same tick regardless of order), the catalog will already be available by the time the chapter settles — the test can pass identically whether or not the guard correctly waits on `initialCatalogSettled`. Given the task's request to make sure tests reflect real-world timing rather than always succeeding, this one doesn't currently exercise the failure mode it claims to guard against.

**Suggested fix:** add a test where the catalog response is deliberately delayed relative to the chapter response (e.g. a controllable/deferred promise behind the `getTranslationBooks` mock), and assert the rendered HTML still contains the chapter link once both resolve — this would fail against the current guard and pass once the fix above is applied.

**`chapterToolHref` doesn't route the translation id through `data.buildTranslationId()`, unlike `canonicalUrl`.**

`SeedBibleStateManager.tsx`'s `canonicalUrl` (which this PR's own comments say `chapterToolHref` matches "the same rules as") does:

```ts
const translationId = data.buildTranslationId(readingState.translationId.value);
```

`buildTranslationId` is a no-op for default/official translations, but for a **custom translation** added via a non-default API endpoint it returns a full absolute URL rather than the short id. `chapterToolHref` in `BibleToolsManager.tsx`, by contrast, uses `position.translationId` directly — the raw short id — because `BibleToolContext` has no reference to the data manager to call `buildTranslationId` at all.

Practical effect: for a custom translation, the generated `href` on the prev/next chapter link would encode a different (and likely non-resolvable) translation-id segment than the page's own `<link rel="canonical">` for that same destination. The everyday in-app click path is unaffected (it calls `onSelect` → `loadNextChapter()`, not the href), so this only shows up via middle-click/ctrl-click-to-new-tab, right-click-copy-link, or a crawler — but it does mean the "real link" behavior this PR is built around is inconsistent for anyone reading a custom translation.

**Repro:** add a custom translation via a custom API endpoint, open a chapter in it, inspect the "Next Chapter" link's `href` attribute, and compare it against the `<link rel="canonical">` tag rendered for that same target chapter — they diverge for a custom translation, and would match for an official one.

**Suggested fix:** either thread a way to call `buildTranslationId` into `chapterToolHref` (e.g. add a minimal `data` reference to `BibleToolContext`, mirroring how `navigation` was added in this PR), or explicitly document/test that chapter links are only guaranteed canonical for default-endpoint translations.

---

Nice-to-have, not blocking: `resolveToolHref` allocates a fresh `computed(() => null)` for every tool without a `getHref` on every `getToolbarTools()` recomputation. It's consistent with the existing pattern for `disabled`/`visible`, so not a regression, just noting it in case it's ever worth memoizing a shared "always null" signal.

---

Blitzy:
> [!NOTE]

I checked out the PR head (`fcd2455`), installed dependencies, and ran the changed tests, `pnpm lint`, `pnpm check:ts`, and Prettier locally. All green: 242 unit tests + 25 integration tests pass, lint reports 0 errors, typecheck reports 0 visible errors, formatting is clean. There is no linked issue on the PR, so this review is based on the diff, the PR description, and the earlier review thread.

**First, the good news on the earlier review.** The high-severity race it flagged — the SSR guard watching only `initialChapterLoadSettled` while `chapterDataPromise` waited on both latches — is genuinely fixed by `208a197e`. I verified the new test is a real regression test: I reverted the guard back to `initialChapterLoadSettled` locally and `still server-renders the links when the catalog arrives after the chapter` failed with `expected [] to include '/en/AAB/genesis/2'`, then passed again once restored. I also instrumented the test's fetch mock to confirm it reproduces the intended ordering rather than passing by luck: the chapter response lands and the latch flips while `books.json` is genuinely still in flight (`catalogPresent = false` at the moment the render starts). That is exactly the failure mode it claims to guard. The `NO_HREF` allocation nit was addressed too.

The design is sound and the comments are unusually good. What follows is one real problem introduced by the fix, two behaviour changes worth a decision, and some smaller items.

## High severity

### A slow book catalog now stalls the whole server response for 5 seconds — and still ships a page with no links

Making `chapterDataPromise` wait on the catalog didn't just delay the toolbar. `BibleReader` throws that same promise, so the entire document render now waits for the catalog, not just the chapter.

**Before:** the chapter arrives, the server responds. If the catalog was late, the arrows served as disabled buttons — degraded, but fast.
**Now:** the chapter arrives, and the server sits on the response until the catalog lands or the 5-second `SSR_INITIAL_CHAPTER_TIMEOUT_MS` expires.

The part that makes this a bug rather than a trade-off: **when the timeout is what ends the wait, the page has no chapter links anyway.** The 5 seconds buy nothing. Measured locally with the chapter response home at 2 ms and `books.json` hanging:

| Gating | `renderToStringAsync` | Chapter link in HTML |
| --- | --- | --- |
| Pre-PR (`initialChapterLoadSettled`) | **17 ms** | no |
| This PR (`initialLoadSettled`) | **5006 ms** | no |

Identical output, 5 seconds slower. Any upstream hiccup on `books.json` now becomes a five-second TTFB on the live host.

**Reproduction.** The quickest deterministic version is a scratch test alongside the existing SSR ones — same `renderToolbarOnServer` scaffolding, but never release the catalog:

```tsx
globalThis.fetch = (async (url: string) => {
  const response = responses[url];
  if (!response) throw new Error(`No mocked response for ${url}`);
  if (url === booksUrl) await waitFor(() => released, 20000); // never released
  return response;
}) as typeof globalThis.fetch;

await waitFor(chapterHasSettled, 8000);   // flips at ~2ms
const t = Date.now();
await renderToStringAsync(/* …toolbar… */);
console.log(Date.now() - t);              // 5006 on this branch, 17 on develop
```

In a browser: run `pnpm dev`, add a temporary `if (import.meta.env.SSR) await new Promise(r => setTimeout(r, 30000));` at the top of `BibleDataManager.getTranslationBooks`, then `curl -s -o /dev/null -w '%{time_total}\n' http://localhost:<port>/en/BSB/john/3`. It goes from roughly 0.2s to 5.0s, and the served HTML has no chapter `<a>` either way.

**Suggested fix.** Give the catalog its own, much shorter deadline instead of sharing the chapter's. A catalog that hasn't answered in a second is not going to save the links, so cut the loss early:

```ts
const SSR_INITIAL_CATALOG_TIMEOUT_MS = 1000;
const initialCatalogTimer = import.meta.env.SSR
  ? setTimeout(() => {
      initialCatalogSettled.value = true;
    }, SSR_INITIAL_CATALOG_TIMEOUT_MS)
  : null;
const clearInitialCatalogTimer = () => {
  if (initialCatalogTimer !== null) clearTimeout(initialCatalogTimer);
};
effectDisposers.push(clearInitialCatalogTimer);
```

…cleared in the same effect that sets `initialCatalogSettled` when `translationBooks` arrives. Worth a regression test asserting the render finishes well under the chapter deadline when the catalog never answers.

## Medium severity

### Space no longer activates the chapter arrows

`<button>` activates on both Enter and Space. `<a href>` activates on Enter only — Space scrolls the page. So a keyboard-only reader who tabs to "Next Chapter" and presses Space used to change chapter and now scrolls instead. `ToolActionElement`'s own JSDoc says the opposite: *"Runs on a plain activation — click, Enter, or Space."*

This affects both the desktop toolbar chapter tools and the mobile floating nav arrows, and it's the primary navigation control in the app.

**Reproduction.** Desktop, keyboard only: Tab to the "Next Chapter" toolbar button, press Space. On `develop` the chapter advances; on this branch the page scrolls.

**Suggested fix** — restore Space in the anchor branch:

```tsx
onKeyDown={(event: JSX.TargetedKeyboardEvent<HTMLAnchorElement>) => {
  if (event.key !== " " && event.key !== "Spacebar") return;
  event.preventDefault();
  onActivate();
}}
```

jsdom won't synthesise the native difference, but a unit test asserting `onActivate` fires and `defaultPrevented` is true on a `keydown` with `key: " "` would fail today and pass with the handler. Worth adding either way — `ToolActionElement.test.tsx` currently has no keyboard coverage at all.

### Middle-clicking a chapter arrow drops the reader out of a live shared session

`chapterToolHref` builds a bare path with no query string. That's the right call for crawlers, but `?sessionId=` is how a reader stays in a shared session (`SeedBibleStateManager.tsx:543`, `:726`, `:1660`), and `?pattern=` carries the portal.

So the new link affordances the PR is selling cut against session readers: middle-click or ctrl-click "Next Chapter" while in a session and the new tab opens **outside** the session, silently. Right-click → "Copy link address" hands out a URL that won't put the recipient in the session either — which is a plausible thing to do, since that's how you'd expect to share where you are.

**Reproduction.** Start or join a shared session (URL shows `?sessionId=…`), then middle-click the "Next Chapter" arrow. The new tab has no `sessionId` and isn't in the session. Compare with the Share modal's link, which keeps it.

**Suggested fix.** Crawlers are never in a session, so suppressing the href when one is active costs nothing for SEO and preserves today's behaviour for session readers:

```ts
if (context.navigation?.currentUrl.value.searchParams.has("sessionId")) {
  return null; // falls back to a button, as before
}
```

If you'd rather accept the drop, it's worth saying so explicitly in `chapterToolHref`'s doc comment — right now the comment only discusses dropping `?verse=`, which reads as though that's the only param in play.

## Low severity

### `chapterToolHref` and `canonicalUrl` disagree on the last language fallback

The comment says the href is *"Built the same way as `canonicalUrl` … same language derivation"*. The first two steps match; the third doesn't:

```ts
// chapterToolHref
… ?? uiLocaleForDefaultTranslation(translationId) ?? DEFAULT_UI_LANGUAGE;   // "en"
// canonicalUrl
… ?? uiLocaleForDefaultTranslation(translationId) ?? i18n.language.value;   // the UI language
```

That third step is reached whenever the translation's language has no UI locale **and** it isn't one of the ~24 hardcoded defaults — which is most of the API's catalog. Practical effect: a reader on `/fr/SWH_XYZ/john/3` sees a "Next Chapter" href of `/en/SWH_XYZ/john/4`, so middle-clicking flips their interface to English.

Worth noting this is arguably the *better* choice for crawling — `script/lib/sitemap.ts` uses the same `uiLocale || DEFAULT_UI_LANGUAGE` fallback, so the links agree with what the sitemap publishes. If that's deliberate, say so in the comment rather than claiming the derivations match; if not, use `i18n.language.value` and the two agree everywhere.

### Third copy of the language-derivation chain

That `bibleLanguageToUiLocale(…) ?? uiLocaleForDefaultTranslation(…) ?? fallback` chain now lives in `SeedBibleStateManager.tsx:1191`, `BibleToolsManager.tsx:572` (new here), and `BibleSelector.tsx:1550` (with the first two swapped), plus the sitemap's variant. CLAUDE.md's rule is to extract a shared helper at the third copy, because copies drift — and this one drifted on arrival, which is the item above. A small `canonicalLanguageFor({ translation, translationId, fallback })` in `ReadingUrlPath.ts` would collapse all of them and make the fallback difference an explicit argument rather than an accident.

## Nits

- `ToolActionElement` accepts `ariaCurrent` but no call site passes it. Dead prop unless something's planned for it.
- The `event.button !== 0` check in the anchor's click handler is effectively unreachable: middle-click fires `auxclick`, not `click`, in every current browser. Middle-click works because the browser handles it natively, not because of this branch. Harmless, but the comment implies the check is doing the work.
- Anchors are draggable by default, so the chevrons now start a link-drag ghost if you press and move on desktop. `draggable={false}` on the anchor if that bothers anyone.
- `ToolActionElement.test.tsx` covers meta/ctrl/shift/alt clicks but nothing for keyboard, which is where the Space gap lives.

## Test assessment

The tests hold up well against the "must actually fail on the pre-fix code" bar. I verified this by reverting fixes locally rather than reading the assertions:

- ✅ `still server-renders the links when the catalog arrives after the chapter` — reproduces the real ordering (instrumented and confirmed the catalog is in flight at render time) and fails on the pre-fix guard.
- ✅ `returns null for the adjacent chapter at the end of the canon` — asserts `nextChapterApiLink` is truthy *and* the position is null, so it pins the exact distinction the feature rests on rather than just checking a null.
- ✅ `stops naming the adjacent chapter once an extension owns that direction` — checks the untouched direction still resolves, so it can't pass by nulling everything.
- ⚠️ Gap: no keyboard-activation test on `ToolActionElement`, which is where the Space regression hides.
- ⚠️ Gap: nothing asserts an SSR time bound, which is what let the High-severity stall through.

---

Nothing here blocks the concept — the feature works and the SSR fix is real. The item I'd want resolved before merge is the 5-second stall, since it costs real TTFB on the live host for no benefit. The Space key and the shared-session drop are quick calls once someone decides which way they want them.

---

Blitzy:
> [!NOTE]

I reviewed the current head (`bfcbf880`) against `develop`, re-read the two prior review threads on this PR, and traced whether the fix commit (`333cf3bd`) genuinely addresses the three issues raised previously (the 5-second catalog stall, the Space-key regression, and the shared-session middle-click drop). No linked issue is referenced in the PR body, so this review is based on the diff and description alone. I ran the affected test suite (358 tests, all passing), `pnpm check:ts` (clean), and `pnpm lint` (clean).

**On the previously-flagged issues:** all three look genuinely fixed and covered by regression tests that fail on the pre-fix code — the SSR catalog timeout now has its own short deadline, `ToolActionElement` handles Space again, and the session-href suppression is in place. Nice work closing those out.

Two smaller things turned up in this pass. Neither is a regression from the fix commit and neither blocks merge on its own, but both are worth a look:

## Medium severity

**A third copy of the language-fallback chain still has the steps in the wrong order.**

This PR extracted `resolveTranslationUiLanguage` (in `BibleToolsManager.tsx`) specifically because an earlier review flagged that `BibleSelector.tsx` had a copy of this chain with the first two steps swapped (`uiLocaleForDefaultTranslation` before `bibleLanguageToUiLocale`, instead of after). `chapterToolHref` and `canonicalUrl` were updated to share the new helper, but `BibleSelector.tsx`'s "Copy translation share link" fallback (around line 1743) still calls the two functions in the swapped order directly, instead of using the shared helper.

**Effect:** for a translation where both `bibleLanguageToUiLocale` and `uiLocaleForDefaultTranslation` resolve to different non-null values, "Copy translation share link" would name a different language segment in the URL than the page's own canonical link and chapter-nav links for that same translation.

**Suggested fix:** replace the inline chain at that call site with `resolveTranslationUiLanguage(...)`, the same helper `chapterToolHref`/`canonicalUrl` now use. That's also the third occurrence of this exact chain (`SeedBibleStateManager.tsx`, `BibleToolsManager.tsx`, and this one in `BibleSelector.tsx`), which is the threshold CLAUDE.md's duplication rule uses to call for consolidation — worth doing now rather than letting a fourth copy show up later.

## Low severity

**`resolveAdjacentPosition`'s extension-ownership check is all-or-nothing, but the actual navigation runner isn't.**

`resolveAdjacentPosition` (`BibleReadingManager.tsx`, around line 3213) returns `null` for a direction as soon as an extension merely *defines* `navigateNext`/`hasNext`, treating that as full, permanent ownership of that direction. But `runNavigationHooks` (the code that actually runs the hook at click time) treats a hook returning `{type: "default"}` as declining to intervene, and falls through to normal catalog-based stepping.

**Effect:** an extension that only conditionally redirects (e.g. an audio/playlist extension that redirects only at the end of a track, and otherwise returns `{type: "default"}`) would suppress the crawlable `<a href>` / next-chapter link for the *entire* time it's enabled, not just on the chapters where it actually intervenes — undercutting this PR's own goal on the common case where the real click-time behavior is identical to plain stepping.

No shipping extension in this repo currently defines `navigateNext`/`hasNext`, so this isn't exercised today, and there's no test covering a hook that sometimes returns `default`. Worth a comment noting the trade-off (or a follow-up) if the intent is genuinely "any ownership disables the link," since that's a reasonable conservative choice — just flagging it since it's a bit surprising given the rest of the PR's care about naming destinations "with certainty."

---

Blitzy (approved):

PR #1619, opened 2026-08-10
Route meta tag image URLs through Vite's asset pipeline

KallynGowdy:
## Summary

Social preview images referenced in `<meta content="...">` tags (like `og:image`) were not being processed by Vite's asset pipeline during builds. This meant the image files were never emitted to the build output, and the URLs in production pointed to paths that only existed in the repository, resulting in 404s.

This change makes Vite treat meta tag image URLs the same way it treats other HTML asset references: emitting them to `assets/` with content hashes and rewriting them to their final absolute URLs.

## How it works

Rather than re-implementing Vite's asset emission and hashing logic, the solution borrows Vite's existing machinery:

1. **Before Vite processes HTML** (`pre` hook): For each targeted meta tag (`og:image`, `og:image:secure_url`, `twitter:image`), inject a temporary carrier `<link>` tag with the image path in its `href` attribute.

2. **Vite's normal processing**: Vite rewrites the carrier link's `href` exactly as it does for favicons and other assets — emitting the file, hashing it, and rewriting the URL.

3. **After Vite processes HTML** (`post` hook): Copy the rewritten URL from the carrier link back into the meta tag's `content` attribute and remove the carrier link.

The carrier uses a custom `rel` value that browsers don't recognize, so they ignore it and don't fetch the social image on every page load. The tag is removed before the HTML ships anyway.

## Key changes

- **`script/lib/htmlMetaAssets.ts`** — Pure string transformation functions:
  - `isLocalAssetPath()` — Identifies image URLs that should be built (repo-relative paths with image extensions), excluding remote URLs and data URIs
  - `injectMetaAssetCarriers()` — Adds temporary carrier `<link>` tags before Vite processes the HTML
  - `resolveMetaAssetCarriers()` — Copies rewritten URLs back into meta tags and removes carriers after Vite processes the HTML

- **`script/lib/vite-plugin-html-meta-assets.ts`** — Vite plugin that wires the transforms to the `transformIndexHtml` hooks that bracket Vite's own HTML processing

- **`test/unit/script/lib/htmlMetaAssets.test.ts`** — Comprehensive unit tests for the transform functions, covering local vs. remote URLs, placeholder resolution, and edge cases

- **`vite.config.ts`** — Registers the new plugin in the build pipeline

- **`index.html`** — Updated `og:image` meta tag with a comment explaining that its `content` is rewritten at build time

## Implementation notes

- The transforms are pure string functions so they can be unit tested without running a full build
- Vite asset placeholders (`__VITE_ASSET__<hash>__`) are passed through verbatim and resolved by Vite's global replacement pass
- Root-relative URLs are absolutized against a configurable site origin (defaults to `https://seedbible.org`, can be overridden via `SITE_ORIGIN` env var)
- The plugin only runs during builds; in dev, the paths in `index.html` already point at files Vite's static middleware serves

https://claude.ai/code/session_01ExTFFfwvQaagCQG9yrdv4J

### DIFF
--- a/index.html
+++ b/index.html
@@ -25,6 +25,8 @@
     <link rel="dns-prefetch" href="https://i.seedbible.org" />
     <link rel="dns-prefetch" href="https://auth.seedbible.org" />
     <meta property="og:type" content="website" />
+    <!-- `content` is rewritten at build time to the emitted asset's absolute
+         URL — see script/lib/vite-plugin-html-meta-assets.ts. -->
     <meta
       property="og:image"
       content="/standalone/img/SeedBibleLogoBlackOnWhiteBackground.jpg"
--- a/script/lib/htmlMetaAssets.ts
+++ b/script/lib/htmlMetaAssets.ts
@@ -1,0 +1,157 @@
+/**
+ * Routing image URLs written in `<meta content="...">` through Vite's asset
+ * pipeline.
+ *
+ * Vite rewrites asset URLs in the HTML entry for a fixed set of tag/attribute
+ * pairs — `link[href]`, `img[src|srcset]`, `source`, `video`, and two SVG ones.
+ * `meta[content]` is not among them, so `index.html`'s `og:image` was left
+ * verbatim and its file was never emitted: the social preview pointed at
+ * `/standalone/img/...`, a path that exists only in the repo (`publicDir` is
+ * off, so nothing copies it into the build) and 404s in production.
+ *
+ * Rather than re-implement emitting, hashing and base-prefixing, the transform
+ * borrows the machinery that already works. Before Vite reads the HTML, each
+ * targeted meta's path is copied onto a carrier `<link>`; Vite rewrites that
+ * `href` exactly as it does the favicon's; afterwards the rewritten value is
+ * copied back into the meta and the carrier is deleted.
+ *
+ * Pure string functions so `test/unit/script/lib/htmlMetaAssets.test.ts` can pin
+ * them down without running a build.
+ */
+
+/** Meta tags whose `content` names an image that should be built like an asset. */
+export const META_ASSET_KEYS = [
+  "og:image",
+  "og:image:secure_url",
+  "twitter:image",
+];
+
+/**
+ * The carrier's `rel`. Deliberately not a real one: `preload` would work just as
+ * well for the build, but it also makes browsers download the 1200x630 social
+ * image on every page load. A rel no browser knows is ignored, and the tag is
+ * removed before the HTML ships anyway.
+ */
+const CARRIER_REL = "seed-bible-meta-asset";
+
+const META_TAG_RE = /<meta\b[^>]*>/gi;
+const LINK_TAG_RE = /<link\b[^>]*>/gi;
+const IMAGE_EXT_RE = /\.(png|jpe?g|gif|svg|webp|avif|ico)(\?[^#]*)?(#.*)?$/i;
+
+function readAttr(tag: string, name: string): string | null {
+  const match = new RegExp(
+    `\\b${name}\\s*=\\s*("([^"]*)"|'([^']*)')`,
+    "i"
+  ).exec(tag);
+  if (!match) return null;
+  return match[2] ?? match[3] ?? null;
+}
+
+function setAttr(tag: string, name: string, value: string): string {
+  const escaped = value.replace(/&/g, "&amp;").replace(/"/g, "&quot;");
+  return tag.replace(
+    new RegExp(`(\\b${name}\\s*=\\s*)("[^"]*"|'[^']*')`, "i"),
+    (_match, prefix: string) => `${prefix}"${escaped}"`
+  );
+}
+
+/** The `property` or `name` a meta tag is identified by, lowercased. */
+function metaKey(tag: string): string | null {
+  const key = readAttr(tag, "property") ?? readAttr(tag, "name");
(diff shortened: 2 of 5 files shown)


Blitzy:
> [!NOTE]

I built this branch locally to verify the fix end-to-end (not just the unit tests), and found that the core scenario the PR documents as supported is actually broken.

## High severity

### `og:image` still ends up as a relative URL in the exact "plain build" case this code claims to handle

**What the code claims:** `ResolveMetaAssetOptions.siteOrigin`'s doc comment (`script/lib/htmlMetaAssets.ts:107-113`) says a root-relative URL left by Vite gets absolutized against `siteOrigin`, and that this "is only reached when `base` is `/` — a plain local build." The unit test `"absolutizes a root-relative URL against the site origin"` asserts exactly this.

**What actually happens:** I ran a real build with no deploy env vars set (`pnpm build:client`, nothing in the environment) — the same conditions CI's own `Build` step uses (see `.github/workflows/ci.yml` No. 60-62 and No. 81-84, which explicitly runs `pnpm build` with no `ASSET_BASE_URL`/`DEPLOY_BRANCH`/`DEPLOY_BUILD_ID` on every PR). The resulting `standalone/dist/client/index.html` contains:

```html
<meta property="og:image" content="/assets/SeedBibleLogoBlackOnWhiteBackground-D_M_K5v5.jpg" />
```

That's a root-relative path, not the absolute `https://seedbible.org/assets/...` URL the code and its tests say should be produced. Open Graph/Twitter Card consumers require an absolute URL for `og:image` — a relative path is invalid per spec and social crawlers (Facebook, X, Slack, Discord, iMessage, etc.) won't render a preview from it. This is the same failure class the PR sets out to fix, just reappearing in a different shape.

**Repro steps:**
1. `git checkout claude/html-image-assets-ag70ot` (clean env, no `ASSET_BASE_URL`/`DEPLOY_BRANCH`/`DEPLOY_BUILD_ID`)
2. `pnpm build:client`
3. `grep -A2 'og:image"' standalone/dist/client/index.html`
4. Observe `content="/assets/....jpg"` instead of an absolute URL.

For comparison, I re-ran with `ASSET_BASE_URL=https://assets.seedbible.com/ DEPLOY_BRANCH=testbranch DEPLOY_BUILD_ID=abc123 pnpm build:client` (matching what `cd.yml` actually sets for real deployments) and got the correct absolute URL: `content="https://assets.seedbible.com/branches/testbranch/abc123/assets/....jpg"`. So today's alpha/production deploys via `cd.yml` are not affected — but any build that doesn't go through that exact pipeline (a local `pnpm build`, a self-hosted/Docker deploy, or any future deploy path) will silently ship an invalid `og:image`.

**Likely root cause:** `resolveMetaAssetCarriers` runs in `transformIndexHtml`'s `"post"` phase, which executes before Vite's own internal asset-URL placeholder (`__VITE_ASSET__<hash>__`) is substituted with its final `base`-relative value (that substitution happens later, during bundle generation, as a raw text replace across the emitted HTML). So at the point this plugin reads the carrier's `href`, it's still the placeholder text, which doesn't start with `/` — the `href.startsWith("/")` check never fires, the origin-prefix step is skipped, and whatever `base` ultimately resolves the placeholder to (absolute when `ASSET_BASE_URL` is set, root-relative when it isn't) is copied into the meta tag completely unprefixed.

**Suggested fix:** Do the origin-prefixing after Vite has fully resolved the placeholder, not before — e.g. move this step into a `writeBundle`/`generateBundle` hook that post-processes the actually-emitted `index.html` asset (at which point `__VITE_ASSET__` placeholders are gone), or use Vite's `experimental.renderBuiltUrl` config option, which is designed to customize how asset URLs are rendered with full knowledge of the final resolved path, instead of round-tripping through a carrier `<link>` and a second text-based rewrite pass.

## Test coverage gap

The unit test `resolveMetaAssetCarriers()` → `"absolutizes a root-relative URL against the site origin"` (`test/unit/script/lib/htmlMetaAssets.test.ts`) uses a hand-rolled `afterVite()` helper that manually substitutes an already-resolved absolute-looking URL into the carrier's `href` *before* calling `resolveMetaAssetCarriers`. That doesn't match how Vite actually processes the HTML (the value is still an unresolved placeholder at that point, as shown above), so the test passes while the real build output is wrong — it's exercising a code path that a real `vite build` with `base: "/"` never reaches. Since this is exactly the condition CI's own build step runs under, an integration-style test that runs an actual (even minimal) `vite build` and asserts on the emitted `index.html`'s `og:image` content — for both a `base: "/"` config and an absolute-`base` config — would have caught this before merge, and would guard against regressions here going forward. The current suite passing has been giving false confidence for the specific case that's broken.

## Nit

`isLocalAssetPath()` only recognizes paths that literally start with `/` or `./`; if `og:image`'s `content` in `index.html` is ever changed to a bare relative path like `img/logo.jpg` (no prefix), it will silently be treated as "not a local asset" and skipped — no warning, no build failure, just the original 404-prone path shipping again. Not a bug today (the current `index.html` always uses `/`-prefixed paths), but worth a comment near `META_TAG_RE`/`isLocalAssetPath` calling out that this is deliberately narrow, so a future edit to `index.html` doesn't quietly regress past this fix without anyone noticing at build time.


---

KallynGowdy:
For the high severity issue, it isn't an issue because we always set `ASSET_BASE_URL`. It would be a bug to not set `ASSET_BASE_URL` during build.

PR #1620, opened 2026-08-10
Quote chapter text in meta description tags

KallynGowdy:
## Summary

This PR adds chapter text excerpts to the `<meta name="description">` and `<meta property="og:description">` tags, so search engines and social media platforms display a preview of the actual scripture rather than a generic app description.

## Key Changes

- **New module `ChapterText.ts`**: Pure text-processing helpers for converting chapter content into search-result snippets
  - `extractContentText()` — flattens a verse's content union (strings, formatted text, headings, footnotes) into plain text, dropping non-quotable elements like verse numbers and inline headings
  - `buildChapterExcerpt()` — concatenates verses into prose, stopping early once a grapheme budget is met (avoids reading all 5,000 characters of Psalm 119 when only 155 are needed)
  - `truncateForMeta()` — shortens text to fit a grapheme budget while breaking on word boundaries where possible, with special handling for scripts without spaces (Chinese, Thai, etc.), combining marks (pointed Hebrew), and surrogate pairs (emoji)

- **Updated `SeedBibleStateManager.tsx`**: The `app.description` signal now quotes the chapter when one is loaded
  - Falls back to a generic app description when no chapter is available (site root, failed load, missing book)
  - Composes a template like `"Genesis 1 (BSB): In the beginning God created…"` and truncates the whole string to stay within the 155-grapheme budget
  - Uses `chapter.translation.shortName` (e.g., "BSB") instead of the full translation name to save space

- **Updated `entry-ssr.tsx`**: Fixed Open Graph meta tags to use `property=` instead of `name=` (parsers only read the `property` attribute)

- **Updated `BibleToolsManager.tsx`**: Reuses `extractContentText()` to avoid duplicating the verse-flattening logic

- **Test coverage**: 
  - 247 lines of unit tests for the text helpers, covering grapheme boundaries, word segmentation, combining marks, surrogate pairs, and 77 locales' worth of edge cases
  - Integration tests in `SeedBibleStateManager.test.ts` and `entry-ssr.test.ts` verify the composed description stays within budget and escapes correctly

- **i18n**: Added two new translation keys (`chapter-meta-description` and `app-meta-description`) to `en.json`

- **Grapheme counting, not code units**: Pointed Hebrew runs ~1.7 code units per visible glyph, so a budget measured in code units would give Hebrew readers ~60% of the snippet Latin readers get. The budget is 155 grapheme clusters (what a reader sees as one character).

- **Early stopping in `buildChapterExcerpt()`**: Psalm 119 is ~5,000 characters; reading the whole thing to throw it away is a real performance regression. The function stops as soon as the budget is met.

- **Word boundary fallback**: For scripts with dictionary-based segmentation (Chinese, Thai), the function tries to break on a word boundary. If the only boundary is too early (one enormous unbroken token, like a URL or a dictionary miss), it cuts mid-token rather than returning almost nothing.

- **Sentence-end detection**: If the cut already ends a sentence (`.`, `!`, `?`, etc.), no ellipsis is added — `"…the deep."` reads as complete, `"…the deep.…"` reads as a bug.

- **Trailing separator cleanup**: Commas, semicolons, and other separators are trimmed before the ellipsis so `"over the surface of the deep, and the Spirit"` truncates to `"over the surface of the deep…"` not `"over the surface of the deep,…"`.

https://claude.ai/code/session_01RgK2L2SLr1LotdNiZHtp3b


Blitzy:
> [!NOTE]

I read through `ChapterText.ts`, the `description` signal in `SeedBibleStateManager.tsx`, and the new test suites, and ran the tests, `tsc`, and `lint` against the branch. I also wrote small standalone scripts against the real `extractContentText`/`buildChapterExcerpt` functions to confirm behavior rather than just reading the code. No linked issue or prior review comments exist on this PR yet, so this is a first pass.

Overall the approach is solid — grapheme-based budgeting, word-boundary-aware truncation, and the space-less-script handling in `buildChapterExcerpt` are all well thought through and well tested at that layer. The issues below are about a gap in *where* that space-less-script handling was applied, and two places where the "always fits in 155 graphemes" guarantee this PR is built around has a hole.

### High: `extractContentText` still inserts a wrong space for space-less scripts, inside a single verse

**File:** `packages/seed-bible/seed-bible/managers/ChapterText.ts`, line 99

`extractContentText` joins a verse's content parts with `.join(" ")` unconditionally. The PR added `NO_LEADING_SPACE` specifically so CJK/Thai text doesn't get a bogus space inserted — but that fix was only wired into `buildChapterExcerpt`'s *inter-verse* join (line 142). The identical problem inside a single verse's own content array — the exact array `extractContentText` flattens — was never fixed.

**Repro:** call the function directly (or via `vitest`):
```ts
extractContentText(["起初神创造天地", { noteId: 1 }, "地是空虚混沌"])
// actual:   "起初神创造天地 地是空虚混沌"   (wrong: space inserted)
// expected: "起初神创造天地地是空虚混沌"
```
This happens for any verse whose content array has more than one text-bearing part — a mid-verse footnote reference, inline heading, line break, or a `wordsOfJesus`-marked span (common in red-letter Gospel text) all split a verse's content this way. `extractContentText` is also reused by `BibleToolsManager.tsx` for the existing verse-copy feature, so this isn't only a new-SEO-excerpt bug.

None of the new tests catch it: `extractContentText`'s "drops inline headings, line breaks, and footnote references" test (test file line 46) and the CJK/Thai test (test file line 97, under `buildChapterExcerpt`) never combine the two — a space-less-script verse with an embedded non-text part.

**Suggested fix:** give `extractContentText` the same incremental-join logic `buildChapterExcerpt` already has, checking `NO_LEADING_SPACE` against each part before deciding whether to insert a space, instead of an unconditional `.join(" ")`.

### Medium: the "chapter has no quotable text" fallback isn't run through `truncateForMeta`

**File:** `packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx`, lines 1036–1041

Every other branch of the `description` computed signal is bounded to `META_DESCRIPTION_MAX_GRAPHEMES` (155). This one isn't:
```ts
if (!excerpt) {
  return t("seed-bible-description", {
    bookName: chapter.book.name,
    chapterNumber: chapter.chapter.number,
    defaultValue: "Read {{bookName}} {{chapterNumber}} in the Seed Bible",
  });
}
```
`chapter.book.name` comes from the translation's own book catalog, not a fixed short label, so a translation with a long or descriptive book name produces a `<meta name="description">` with no length ceiling — the exact thing this PR exists to prevent. It's triggered whenever `buildChapterExcerpt` returns `""` (e.g., a chapter payload with only headings/line breaks, or a malformed/partial fetch).

The two tests that exercise this branch (test file lines 1374 and 1387) both use the book name `"Genesis"`, so they'd pass even with no truncation at all — they don't actually probe the budget on this path.

**Suggested fix:** wrap the same return in `truncateForMeta(..., META_DESCRIPTION_MAX_GRAPHEMES)`, same as line 1055 does for the excerpt branch.

### Low: `truncateForMeta` always trims from the end, but the code's own comment anticipates templates that don't put the excerpt last

**File:** `packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx`, lines 1053–1055

The comment reads: "The whole composed string is what has to fit, not just the excerpt: book names vary in length and a translated template may reorder its parts." That's the right reason to truncate the composed string rather than just the excerpt — but `truncateForMeta` unconditionally cuts from the end (`ChapterText.ts` line 172/189, via `.slice(0, end)`). If a locale's `chapter-meta-description` template puts the citation *after* the excerpt (e.g. `"「{{excerpt}}」— {{bookName}} {{chapterNumber}} ({{translationName}})"`), a long excerpt causes the trailing book/chapter/translation citation to get chopped off instead of the scripture text — defeating the point of citing the reference.

This is currently dormant since only `en.json` has `chapter-meta-description`/`app-meta-description` (per this repo's translation convention, other locales fall back to English), so it won't surface until a translator adds a non-English template with a different part order. Since the comment already flags the reordering case as a known possibility, it'd be worth resolving now rather than leaving it for whoever adds that translation to discover.

**Suggested fix (either works):** reserve grapheme budget for the citation portion and truncate only the excerpt before composing, or leave a translator-facing note (e.g. in the i18n key's comment/description, if this repo has a mechanism for that) that the template must keep `{{excerpt}}` last so truncation lands on scripture text, not the citation.

---

KallynGowdy:
All three findings were real and are fixed in 33abf03f. I confirmed each new test fails against the previous code before keeping it, so they're genuine regression tests rather than restatements of current behavior.

**High — space inserted inside a verse for space-less scripts.** Confirmed by calling the function directly, exactly as described:

```
before: 起初神创造天地 地是空虚混沌
after:  起初神创造天地地是空虚混沌
```

Worth adding to your analysis: the `wordsOfJesus` case makes this more common than it first looks — red-letter Gospel text splits a verse's content array on nearly every verse, so for Chinese and Thai this was closer to the norm than an edge case. And as you noted it reaches verse copy/share through `extractContentText`; that path predates this branch, so this fixes a pre-existing bug too.

Rather than duplicating the `NO_LEADING_SPACE` check in both places, both joins now go through one `appendText` helper, so the intra-verse and inter-verse rules can't drift apart again.

**Medium — reference-only fallback unbounded.** Fixed. One thing you missed while looking at this: the *no-chapter* branch immediately above has the same hole. `app-meta-description` is a translatable key, so a translation is free to exceed 155 graphemes even though the English default is 131. Both branches are now truncated, so every path through `description` is bounded.

**Low — truncation always trims from the end.** Fixed rather than documented. The excerpt is now trimmed to whatever the citation leaves, so the cut lands on scripture regardless of the template's part order, with the composed-string truncate kept as a backstop for a template whose own literal text overruns the budget. Tested with a reordered template (`「{{excerpt}}」— {{bookName}} {{chapterNumber}} ({{translationName}})`), which loses the citation entirely without the fix.

You were right that this was worth resolving now — the two-pass version is also just simpler to reason about than "truncate the whole thing and hope the citation is first."

English output is byte-identical after all three changes; I re-rendered Genesis 1, Psalm 51, Psalm 119, pointed Hebrew and Chinese through SSR to confirm. 4390 tests pass, `tsc` and `lint` clean.

One caveat on my end-to-end verification, since it affects how much weight to give it: the Bible API isn't reachable from my sandbox, so those renders go through a local mock serving realistic payloads rather than the live API.

---

PR #1625, opened 2026-08-11
fix: restore missing branding declaration in createSeedBibleState

Blitzy:
## Summary
- `develop` currently fails `pnpm check:ts` (TS2304: Cannot find name 'branding') at `SeedBibleStateManager.tsx:497`, which also fails CI on every other open PR since Actions checks out the PR/develop merge ref.
- Traced the regression to the merge of #1610 ("added config for disabling multiple tools"): that PR changed `createBibleToolsManager()` to `createBibleToolsManager(branding)` and declared `const branding = options.config?.branding;` on its own branch (where CI was green), but the merge into `develop` kept the call site and dropped the declaration — likely a conflict-resolution mistake with the concurrent OpenTelemetry PR (#1557) touching the same function.
- Restores the one-line declaration so `branding` is back in scope before it's used (both for `createBibleToolsManager(branding)` and the existing `appName: branding?.appName` usage further down).

## Test plan
- [x] `pnpm check:ts` — clean (was failing with TS2304 before this change)
- [x] `pnpm test` — 4433/4433 passing

### DIFF
--- a/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
@@ -448,6 +448,7 @@ export function createSeedBibleState(
     initialHref: options.initialHref,
     basePath: options.config?.basePath,
   });
+  const branding = options.config?.branding;
   const api = new FreeUseBibleAPI(
     getDefaultAPIEndpoint(navigation.currentUrl.value)
   );


Blitzy:
> [!NOTE]

## Code Review — PR #1625

**Verdict: Approve.** This is a correct, minimal, one-line hotfix. It restores `const branding = options.config?.branding;` in `createSeedBibleState` so `branding` is in scope before it's passed to `createBibleToolsManager(branding)`. I verified the fix resolves a real, active break on `develop`, and I found no correctness issues. The notes below are informational — none are blocking.

### What I verified

- **Type-check is clean with the fix.** Ran `pnpm check:ts` on this branch → exit 0, 0 visible errors (the 6 "suppressed errors" are pre-existing, controlled by `tsc-silent.config.cjs`, and unrelated to this change).
- **The regression was real on `develop`.** The base commit (`7a87d26`) has the call `createBibleToolsManager(branding)` but no declaration of `branding`, which is the reported `TS2304: Cannot find name 'branding'`.
- **The type is correct.** `options.config?.branding` is `BrandingConfig | undefined` (`app/appConfig.tsx:39`), and `createBibleToolsManager(branding?: BrandingConfig)` (`BibleToolsManager.tsx:1118`) accepts exactly that. No `any`, no narrowing issues.

### Severity: none blocking

**Informational — the impact was worse than the PR describes (runtime crash, not just a type error).** The PR summary frames this as a `pnpm check:ts` failure. It was actually a runtime break too: Vitest transpiles without type-checking, so on `develop` the undeclared reference throws `ReferenceError: branding is not defined` the moment `createSeedBibleState` runs — i.e. core app init crashes. I reproduced it by deleting the line on this branch and running one existing test:

```
× created with default values
ReferenceError: branding is not defined
 ❯ createSeedBibleState packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx:497:41
 ❯ createTestSeedBibleState test/unit/seed-bible/testUtils/createTestSeedBibleState.ts:177:17
```

This strengthens the case for the fix — it's not merely a type-checker nicety. No code change needed; just flagging that the merge left `develop` runtime-broken.

**Low — the PR description references a usage that doesn't exist.** The summary says the declaration is needed "both for `createBibleToolsManager(branding)` and the existing `appName: branding?.appName` usage further down." There is no `appName: branding?.appName` in `SeedBibleStateManager.tsx` — `grep` finds only two `branding` occurrences: the restored declaration (line 451) and the single consumer at line 498. The code is correct; only the description overstates the scope. Worth tidying so future readers aren't sent looking for a second consumer that isn't there.

### On tests

This PR adds no test, and that is the right call here. The dropped line is already covered by the existing `createSeedBibleState` suite (`test/unit/seed-bible/managers/SeedBibleStateManager.test.ts`) — those tests fail without the line (shown above) and pass with it, so the guardrail exists. A new test would duplicate that coverage. Nothing here "always succeeds"; the existing tests genuinely exercise the crashing path.

### One process observation (not for this PR to fix)

The existing test suite *would* have caught this — a test that calls `createSeedBibleState` throws without the line. That it still landed on `develop` suggests merges into `develop` aren't gated on a green `pnpm test` / `pnpm check:ts` run of the merged result (the conflict resolution against #1557 dropped the line and wasn't re-checked). Ensuring `develop` pushes/merges re-run CI would prevent a conflict-resolution slip like this from shipping a runtime-broken base again. Out of scope for this hotfix — merge it as-is.

### Reproduction (for a human, to confirm the fix)

1. Check out `develop` at `7a87d26` (or delete line 451 on this branch).
2. `pnpm check:ts` → `TS2304: Cannot find name 'branding'` at `SeedBibleStateManager.tsx`.
3. `pnpm vitest run test/unit/seed-bible/managers/SeedBibleStateManager.test.ts -t "created with default values"` → `ReferenceError: branding is not defined`.
4. Restore the line (this PR) → both are clean/green.

---

PR #1631, opened 2026-08-11
Bonfire improvements

KallynGowdy:
This PR fixes issues with calling the Bonfire API and adds support for streaming messages

### DIFF
--- a/packages/bonfire-extension/ext_Bonfire/main/bonfire.tsx
+++ b/packages/bonfire-extension/ext_Bonfire/main/bonfire.tsx
@@ -1,4 +1,37 @@
 import { type SeedBibleState } from "seed-bible";
+import { z } from "zod";
+
+const bonfireSessionStartResponseSchema = z.object({
+  session: z.object({
+    session_id: z.string(),
+  }),
+});
+
+const bonfireChatResponseSchema = z.object({
+  session_id: z.string(),
+  message: z.object({
+    id: z.string(),
+    role: z.string(),
+    content: z.string(),
+  }),
+  sources: z.array(z.unknown()).optional(),
+  usage: z
+    .object({
+      input_tokens: z.number(),
+      output_tokens: z.number(),
+      cost_usd: z.number(),
+    })
+    .optional(),
+  quota: z
+    .object({
+      org_limit: z.number(),
+      org_used: z.number(),
+      org_remaining: z.number(),
+      org_used_pct: z.number(),
+      request_interaction_cost: z.number(),
+    })
+    .optional(),
+});
 
 export interface BonfireOptions {
   /** The organization ID for the Bonfire API. */
@@ -58,7 +91,9 @@ export function* registerBonfireChatProvider(
           headers,
         }
       );
-      const data = await response.json();
+      const data = bonfireSessionStartResponseSchema.parse(
+        await response.json()
+      );
       console.log("[Bonfire] Session created", data);
       chatSessionMap.set(chatContext.chatId, data.session.session_id);
     },
@@ -113,7 +148,7 @@ export function* registerBonfireChatProvider(
             org_id: orgId,
             ai_id: aiId,
             session_id: sessionId,
-            stream: false,
+            stream: true,
             input: {
               content: lastMessage?.type === "text" ? lastMessage?.text : "",
             },
@@ -123,7 +158,7 @@ export function* registerBonfireChatProvider(
         }
       );
 
-      const data = await response.json();
+      const data = bonfireChatResponseSchema.parse(await response.json());
 
       const message = data.message;
 
--- a/packages/bonfire-extension/ext_Bonfire/main/bonfire.tsx
+++ b/packages/bonfire-extension/ext_Bonfire/main/bonfire.tsx
@@ -219,7 +219,7 @@ export function* registerBonfireChatProvider(
             input: {
               content: lastMessage?.type === "text" ? lastMessage?.text : "",
             },
-            custom_instructions: `You are chatting with a user who is reading the Bible. They are currently reading: ${readingState?.bookId} ${readingState?.chapterNumber}.  Your responses should be in the same language as the user's messages.`,
+            custom_instructions: `You are chatting with a user who is reading the Bible. They are currently reading: ${readingState?.bookId} ${readingState?.chapterNumber}`,
           }),
           headers,
         }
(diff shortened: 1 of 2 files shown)


Blitzy:
> [!NOTE]

## Code review — PR No. 1631 (Bonfire improvements)

No linked issue was found for this PR, so this review is based on the diff, the four commits, and the CI results (build/deploy both green) alone.

### The short version

The core rework is sound: moving from a single non-streaming `chat/completions` call to a session-based `session/start` → `session/chat` → `session/end` flow against a new host (`bonfire.seedbible.io`), with a hand-rolled SSE reader and zod validation on the responses. The two biggest things I'd want addressed before merging are (1) that all client-side authentication was removed with nothing put in its place, and (2) this PR adds real, non-trivial parsing/streaming logic with zero test coverage.

---

### High — Client-side API key removed with no replacement auth, and default org/AI IDs are now hardcoded into shipped code

**Files:** `packages/bonfire-extension/ext_Bonfire/main/bonfire.tsx` (headers), `packages/bonfire-extension/ext_Bonfire/main/init.tsx` (L4-5, `DEFAULT_ORG_ID` / `DEFAULT_AI_ID`)

Before this PR, calling Bonfire required `bonfireOrgId`, `bonfireAiId`, **and** `bonfireApiKey` to all be supplied via the URL, and the key was sent as `X-API-Key`. After this PR, `apiKey` is gone entirely (the field is commented out, not removed — see the Low item below), the only header sent is `Content-Type: application/json`, and `orgId`/`aiId` now fall back to hardcoded constants (`DEFAULT_ORG_ID`, `DEFAULT_AI_ID`) baked into the extension bundle when the URL doesn't supply them.

Net effect: the Bonfire chat provider now activates with **no credentials and no URL parameters at all**, calling `https://bonfire.seedbible.io/...` with an org/AI id that anyone can read straight out of the shipped JS. If `bonfire.seedbible.io` is a proxy that holds the real Bonfire API key server-side and does its own authn/rate-limiting, that's a reasonable design — but nothing in this diff (or the PR description) confirms that, and there's no origin check, token, or rate limit visible from the client side. If the proxy doesn't defend itself, this is an open door for anyone to run (and pay for, on your Bonfire account) unlimited chat completions just by hitting the endpoint directly with the IDs they can read out of the bundle.

**Suggested fix:** confirm (and ideally note in a comment) that `bonfire.seedbible.io` enforces its own auth/rate-limits independently of the client. If it doesn't yet, this needs some form of server-side gate before merging — an origin/referrer check, a short-lived token issued by your own backend, or similar.

---

### Medium — No test coverage for the new parsing/streaming logic

`git log`/the file list show only `bonfire.tsx` and `init.tsx` changed — no test file was added or touched, and there were none before (`packages/bonfire-extension` has no test directory at all).

This PR adds a hand-rolled SSE line parser (`parseBonfireEventStream`), a delta-extraction generator (`streamBonfireMessageDeltas`), and two zod schemas that `.parse()` (throw) on mismatch — all logic that is easy to get subtly wrong (buffering across chunk boundaries, the end-of-stream flush, event names that don't match `message.delta`, malformed/empty `data:` lines) and easy to unit test without any network mocking beyond a fake `ReadableStream`. Per this repo's own testing conventions (CLAUDE.md: "cover error paths and edge cases... a failed record call, an empty result... are where the real bugs hide"), this is exactly the kind of code that should ship with tests, e.g.:

- A multi-chunk SSE response where an `event:`/`data:` pair is split across two `reader.read()` calls.
- A stream that ends without a trailing newline after the last `data:` line (exercises the `done` branch's single `consumeLine(buffer.trim())` call).
- A non-`message.delta` event (e.g. `message.done`) being correctly skipped.
- A `session/chat` response with `response.ok === false` — confirm the thrown error surfaces as the chat error message rather than an unhandled rejection.
- A `session/start` response that doesn't match `bonfireSessionStartResponseSchema` — confirm the failure mode is acceptable (right now it throws inside an un-awaited `onJoinChat` call in `ChatsManager`, which would be an unhandled rejection with no user-visible error and a chat participant stuck in "no session" state; that particular call site is outside this PR's diff, but this PR is what makes hitting it more likely since it now validates strictly instead of just indexing into the parsed JSON).

---

### Low — Silent behavior change: dropped length/language instructions from `custom_instructions`

**File:** `bonfire.tsx` (`generateResponse`)

The old prompt included `"Keep responses tweet-length. Your responses should be in the same language as the user's messages."` — the new one drops both sentences. That may be intentional (perhaps the new `ai_id`/session config on the Bonfire side already encodes tone/length/language), but if not, this is a user-visible regression: longer replies, and replies that may not match the reader's language. Worth confirming this was deliberate.

---

### Low — Dead code: `apiKey` field commented out instead of removed

**File:** `bonfire.tsx`, `BonfireOptions` interface

```ts
/** The API key for the Bonfire API. */
// apiKey: string;
```

The field itself is unused everywhere now, so this should just be deleted (along with the now-orphaned doc comment above it) rather than commented out.

---

### Low — `zod` is imported but not declared as a dependency of this package

**Files:** `packages/bonfire-extension/ext_Bonfire/main/bonfire.tsx` (new `import { z } from "zod"`), `packages/bonfire-extension/package.json`

`packages/bonfire-extension/package.json` only lists `seed-bible` under `dependencies`; `zod` isn't there. This currently builds fine (confirmed: CI's `build (24.x)` / `deploy (24.x)` are green) only because the workspace root `package.json` also depends on `zod: catalog:`, so module resolution finds it by walking up to the repo root. That's a fragile, implicit dependency — other packages in this repo (e.g. `packages/seed-bible/package.json`) declare `"zod": "catalog:"` explicitly. Worth adding it to `bonfire-extension`'s own `package.json` so it doesn't silently break if the root's dependency graph ever changes.

---

### Nit — `Idempotency-Key` doesn't do anything yet

**File:** `bonfire.tsx`, `onLeaveChat`

```ts
headers: {
  ...headers,
  "Idempotency-Key": crypto.randomUUID(),
},
```

Idempotency keys only help if the *same* key is reused across retries of the *same* logical request, so the server can recognize a duplicate. Here a fresh random UUID is generated on every call, and there's no retry logic around this `fetch` at all, so the header currently has no effect either way — harmless, but not doing what its name implies. Not blocking, just flagging in case the intent was to guard against duplicate session-end calls (e.g. from a fast unmount + cleanup double-fire), which would need the key to be stable per logical "end this session" attempt, not per call.

---

PR #1633, opened 2026-08-11
feat: Record annotations offline and sync them when online

KallynGowdy:
Annotations were the least protected data in the app despite being the only user-written prose. Every write went straight to the server, so with no connection saving threw and whatever had been typed was lost. Reading was worse: a failed list request was cached as "no annotations" *and marked settled*, so an offline visit reported an empty chapter and kept reporting one for the life of the page.

Notes are now recorded on the device first and pushed when a connection is available. Creating, editing and deleting all work offline, and a count of what is still waiting is shown in the notes pane.

Conflict handling had to come with it rather than after it. A chapter's list was fetched once per page load and never revalidated, and `recordData` is a blind overwrite with no server version or etag — so two devices editing one note already clobbered each other silently. Deferring writes makes that strictly worse, since a change queued for hours is far more likely to be stale. Every write now reads the server's copy first and checks it is still the one the local edit was based on; when it isn't, the user is asked which version to keep before anything is overwritten.

Notable decisions:

- One IndexedDB store carrying its own sync metadata, not a mirror plus a separate outbox. IndexedDB skips null index values, so the index over `pendingOp` *is* the queue. Repeated offline edits then coalesce for free, and a create-then-delete offline costs no requests at all.
- Conflicts compare against a remembered base version, not "whose timestamp is larger" — timestamps are written by whichever device made the edit, so clock skew between devices makes "newer" unreliable. A content fingerprint covers records that carry no timestamp.
- A separate `seed-bible-annotations` database rather than a version bump on `seed-bible-offline`, whose open path rejects on `onblocked` and would break offline translations in a tab holding the old connection.
- Notes written while signed out go to a device-local bucket and are adopted into the account on first sign-in, cleared in the same transaction so a second account on a shared device cannot inherit them — the leak `LoginManager` already guards against for anonymous config.
- Sign-out drops synced rows but keeps unsent writing: synced notes are re-fetchable, and leaving readable prose on a shared device is what clearing the profile cache exists to prevent.
- A sync pass captures the account once and re-checks it before every write, so a queued change can never be pushed into whichever account happens to be signed in when connectivity returns (#1587).

Claude-Session: https://claude.ai/code/session_0125RRhUgDbdQFxpTAtB9V3S


packages/seed-bible/seed-bible/managers/AnnotationSyncManager.tsx
```diff
@@ -0,0 +1,721 @@
...
+        ...syncedRow(owner, copy),
+        annotation: copy,
+        baseUpdatedAtMs: null,
+        baseFingerprint: null,
+        pendingOp: "upsert",
+        updatedAtMs: row.updatedAtMs,
+      });
+      await store.put(syncedRow(owner, conflict.server));
+      onSynced?.(conflict.server, owner);
+      await store.delete(owner, row.annotationId);
```
Blitzy: **Bug (confirmed): "keep both" deletes the server's own note instead of preserving it.**

`conflict.server.id` is always the same id as `row.annotationId` (it's the same annotation, just the server's version of it). So the `store.put(syncedRow(owner, conflict.server))` on line 616 writes a row keyed `${owner}/${row.annotationId}` — and then line 618's `store.delete(owner, row.annotationId)` deletes that *exact same row* right afterward. `onRemoved` then strips it from the UI cache too. Net effect: after "keep both", the server's version disappears from the local mirror and the UI, directly contradicting the PR's own description ("yours is saved as a second note... leaves theirs untouched").

I verified this empirically with a standalone test exercising `resolveConflict(id, "keep_both")`: `store.get(owner, "ann-1")` returns `null` afterward. The row does reappear on the *next* full server-list reconcile (since nothing was erased server-side), but until then it's gone from the open chapter's view.

Note the existing test for this ("keep_both writes ours under a new id and leaves theirs untouched" in `AnnotationSyncManager.test.ts`) doesn't catch this because its assertion is `expect(original?.annotation?.data.html ?? theirs.data.html).toBe(theirs.data.html)` — the `??` fallback makes the assertion trivially pass even when `original` is `null`.

Fix: drop the redundant `store.delete`/`onRemoved` call for the original row — the `store.put(syncedRow(owner, conflict.server))` two lines above already leaves it correctly synced.

---

packages/seed-bible/seed-bible/managers/AnnotationSyncManager.tsx
```diff
@@ -0,0 +1,721 @@
...
+      // Left in `awaitingUser` on purpose: the question is still open, and
+      // dropping it would silently abandon the user's decision.
+      return;
+    }
+
+    await refreshPendingCount();
+    // Awaited so the choice has actually been carried out by the time this
+    // resolves — the modal keeps its buttons disabled until then, and a caller
+    // that checks the result isn't racing the push.
+    await sync();
```
Blitzy: **Likely bug: resolving a conflict while its own sync pass is still running can leave the resolution un-pushed, contradicting the comment above.**

`runPass`'s `for` loop doesn't stop when it records a conflict (`conflicts.value = [...]; continue;`) — it keeps pushing the *other* pending rows, each with its own `await` (network round trip). Since `conflicts.value` updates as soon as the first conflict is found, the modal can open — and the user can call `resolveConflict` — while that same pass is still mid-loop on later rows.

At that point `running` is still set, so this `sync()` call just joins the *already in-flight* pass instead of starting a fresh one. That in-flight pass captured its `rows` snapshot before the resolution happened and won't revisit the just-resolved row, and `resolveConflict` never sets `dirty = true` (unlike `notifyLocalChange`), so the do-while loop won't repeat either. For `keep_mine` (where `pendingOp` deliberately stays set so the next pass pushes the new base), the row is left pending with nothing scheduled to retry it — it sits until some unrelated future trigger (another edit, an online/offline toggle, or a reload) happens to run `sync()` again.

This directly contradicts the comment here ("Awaited so the choice has actually been carried out by the time this resolves"). Not covered by the existing tests, which always `await sync.sync()` to completion *before* calling `resolveConflict`, so the race never occurs in them.

A fix could have `resolveConflict` set `dirty = true` before its trailing `await sync()`, the same way `notifyLocalChange` does.

---

packages/seed-bible/seed-bible/managers/AnnotationsManager.tsx
```diff
@@ -651,13 +961,19 @@ export function createAnnotationsManager(
 
   const createNewAnnotation = async (): Promise<void> => {
     let userId = login.userId.value;
-    if (!userId) {
+
+    // Offer a sign-in, but don't insist on one. Without a local store there is
+    // nowhere else to put the note, so an account is still required; with one,
+    // the note is kept on the device and adopted when the user does sign in.
+    // The prompt is skipped entirely while offline, where it cannot succeed.
+    if (!userId && (!store || sync.isOnline.value)) {
```
Blitzy: Minor: this condition doesn't match its own comment. `(!store || sync.isOnline.value)` prompts for login whenever there's *no store*, regardless of online state — so on a browser with no IndexedDB (e.g. Safari private browsing) while offline, this still calls `login.login()` and waits on a login flow that cannot succeed, instead of the documented "skipped entirely while offline, where it cannot succeed." It falls through gracefully afterward (the next check warns and returns), but the user sees an attempted/failing sign-in prompt rather than an immediate "can't create notes offline" message. If the intent really is "skip while offline" in all cases, this should be `(!store && sync.isOnline.value) || (store && sync.isOnline.value)` — i.e. just gate on `sync.isOnline.value` outright — or the comment should be corrected to describe the current (store-dependent) behavior.

---

Blitzy (commented):
Single-pass review (Agent tool unavailable in this context, so no multi-agent fan-out or subagent verify pass ran — this is one careful read-through, not the full pipeline).

3 findings posted inline, most severe first. #1 is empirically confirmed (reproduced with a standalone test against the real store); #2 and #3 are reasoned through carefully but not reproduced against a live race/browser.

---

Blitzy:
> [!NOTE]

Full review posted as [inline comments and a review summary](https://github.com/HelloAOLab/seed-bible/pull/1633#pullrequestreview-4910407296) on this PR. Highlights, most severe first:

1. **`AnnotationSyncManager.tsx` (No. 618)** — "keep both" conflict resolution deletes the server's own note right after writing it, instead of preserving it alongside the new copy. The delete key is identical to the write key, so the server's note vanishes from the local mirror and the UI until the next full reconcile — the opposite of the PR's stated "leaves theirs untouched." Confirmed empirically against the real store.
2. **`AnnotationSyncManager.tsx` (No. 575)** — `resolveConflict`'s trailing `sync()` can silently join an already-running pass without setting `dirty = true`, so a just-resolved row (especially a "keep_mine" choice that deliberately leaves `pendingOp` set) may not actually get pushed or retried, despite the code's comment claiming the await guarantees the choice "has actually been carried out."
3. **`AnnotationsManager.tsx` (No. 969)** — `createNewAnnotation`'s login-prompt condition (`!userId && (!store || sync.isOnline.value)`) attempts to sign in even while offline when there's no local store, contradicting its own comment that the prompt "is skipped entirely while offline."

Tests: the existing conflict-resolution test for finding No. 1 doesn't catch the bug because its assertion uses `original?.annotation?.data.html ?? theirs.data.html` — the `??` makes it pass even when `original` is null.

---

PR #1635, opened 2026-08-11
Refactor/bible stack

### LINKED ISSUE #776: Fix bible stack typechecking issues
This task is to fix typescript issues in the bible stack extension.

### LINKED ISSUE #1118: [refactor] Port the Bible Stack to the refactor
This task is to port the Bible Stack extension to the refactor.

soukigabriel:
Fixes #776, Fixes #1118.


Blitzy:
> [!NOTE]

## Summary

This PR ports the **Bible Stack** extension to the refactored architecture and tightens a couple of arrangement types (fixes No. 776 and No. 1118). It's large — 734 files, ~51k added lines — but the bulk is the ported pattern: 383 new `.tsx` source files plus generated `.aux` bots and `legacy_*.txt` reference dumps (committing those is the repo's existing convention, so not flagged here). The hand-written integration changes outside the pattern are small and mostly safe.

I focused on real runtime correctness in the new source. Findings: **3 High**, **4 Medium**, **4 Low**, plus a test-coverage gap. The type tightening is safe against the static arrangement config (all 25 `subset` entries already provide the now-required fields).

**On tests:** the port ships **no new automated tests** — the only test change is a 2-line factory fix (`startIndex`/`endIndex`) so the pre-existing `ScriptureService` tests still compile. Those existing tests are genuine — real assertions on computed output, error paths, and caching, so they'd fail on a regression rather than always passing. But several bugs below (H1, H2) would have been caught by even a trivial unit test on the ported services.

---

### 🔴 High

**H1 — `InfoLabelData.addActivityIndicator` never stores new indicators (inverted guard)**
`patterns/bible-stack/bible-stack/domain/entities/InfoLabelData.tsx:94`

The guard is inverted — it stores an indicator only when the map *already* contains its id, so a genuinely new indicator is never added:
```ts
addActivityIndicator(indicator: ActivityIndicator) {
  if (this.#activityIndicators.has(indicator.id)) {   // ← should be !has(...)
    this.#activityIndicators.set(indicator.id, indicator);
  }
}
```
Both sibling entities do the opposite (correct) thing: `StackChapterData.addActivityIndicator` and `VersesBundleData.addVerse` use `if (!...has(...))`. `PieceActivityService` (~line 561) loops `container.addActivityIndicator(indicator)` where the container can be an `InfoLabelData`. So for label containers the indicator bots are *shown* but the internal map stays empty — `updateIndicatorsPosition` can't position them and `clearActivityIndicators` can't release them back to the object pool. The shown bots leak and never reposition/hide.

Repro (unit-level): `const l = new InfoLabelData(...); l.addActivityIndicator({ id: "a", ... }); expect(l.activityIndicators.length).toBe(1);` → currently 0.
Fix: `if (!this.#activityIndicators.has(indicator.id))`.

**H2 — `disableLabelDates()` turns label dates ON (copy-paste inversion)**
`patterns/bible-stack/bible-stack/application/services/ScripturePiecesStateService.tsx:46`
```ts
enableLabelDates()  { this.#shouldShowLabelDates = true; }
disableLabelDates() { this.#shouldShowLabelDates = true; }   // ← should be false
```
`#shouldShowLabelDates` can never be turned back off once enabled; any caller invoking `disableLabelDates()` to hide verse-label dates shows them instead.
Fix: set `false` in `disableLabelDates()`.

**H3 — Arrangement path lookups return the wrong `arrangementIndex` when the active arrangement isn't index 0**
`patterns/bible-stack/bible-stack/application/services/ArrangementService.tsx` — `getTestamentInfoPathByName` (189-202), `getSectionInfoPathByName` (261-274), `getBookInfoPathById` (344-357)

Each method copies `getAllArrangements()`, moves the initial arrangement to the front (`splice` + `unshift`), loops the reordered array, and on a match returns `arrangementIndex = currentArrangementIndex` — the position in the **reordered** array, not the real arrangement index.

Before/after: active index = 1, arrangements `[A0, A1]` → reordered to `[A1, A0]`. A book found in `A1` sits at reordered position 0, so the method returns `arrangementIndex: 0`. Callers then do `getArrangementByIndex(0)` (e.g. `getBooksNamesBySectionName`, and chapter navigation in `PieceActivityService` / `StackPresenceNavigationService`) and resolve against `A0` — wrong book/section, or a "not found" throw. Masked only when the active index is 0 (no reorder happens).

(`getAllArrangements()` returns a fresh `[...statics, ...custom]`, so the in-place `splice`/`unshift` doesn't corrupt stored state — the bug is confined to the returned index.)
Fix: track the real index alongside each entry before reordering (e.g. `[{ arrangement, realIndex }]`) and return `realIndex` on match, or search the current arrangement first without mutating the index mapping.

---

### 🟠 Medium

**M1 — `PaintService.unpaint` skips exactly the pieces it should unpaint**
`patterns/bible-stack/bible-stack/application/services/PaintService.tsx:96`

`unpaint` reuses `paint`'s guard `if (data.paintColor === this.#color) return;`. In `paint` that correctly skips re-painting an already-correct piece; in `unpaint` it skips removal precisely when the piece is painted with the currently-active color.
Repro: active color = X; paint a verse X; call `unpaint` on that verse while X is still active → returns early, verse stays painted.
Fix: guard on absence of paint instead, e.g. `if (!data.paintColor) return;`.

**M2 — `updateNotification` shows a notification even when it decided to hide**
`patterns/bible-stack/bible-stack/application/services/PieceActivityService.tsx:622-624`
```ts
if (shouldHide) {
  const hid = this.tryHideNotification(container);
  if (hid) return;   // ← only returns when something was actually removed
}
```
When `shouldHide` is true but there's no existing notification to remove, execution falls through and builds/shows a fresh one. A chapter with `pieceActivity.length === 0` (so `shouldHide`) and no current notification gets a spurious zero-activity notification.
Fix: return unconditionally inside the `shouldHide` branch.

**M3 — `UpperCoverOpacityAdapter` aborts the whole loop instead of skipping one bible**
`patterns/bible-stack/bible-stack/infrastructure/adapters/stacks/UpperCoverOpacityAdapter.tsx:26`
```ts
for (const bibleData of bibles) {
  const upperCover = bibleData.staticBiblePieces?.upperCover;
  if (!upperCover || bibleData.currentState !== BibleStates.Open) return;   // ← should be continue
  ...
}
```
Repro: bible A (closed / no `upperCover`) inserted before bible B (open); rotate the camera → the loop hits A, `return` aborts the method, and B's cover opacity is never recomputed and stays stale. Any open bible following a closed/coverless one in insertion order silently stops responding to camera rotation.
Fix: `continue;`.

**M4 — bible-stack extension never unregisters its below-reader tool on uninstall**
`packages/bible-stack/bootstrap.tsx:34,65`

`registerBelowReaderTool` returns an unregister disposer (`BibleToolsManager.tsx:1365`). The sibling `scripture-map` extension yields it — `yield context.tools.registerBelowReaderTool({...})` (`scripture-map/di/bootstrap.tsx:83`). Here the return value is discarded and a no-op `yield () => {};` is yielded instead. So uninstalling the extension leaves the "Bible Stack" tool button in the reader. (Reinstall dedupes by id, so it isn't duplicated — but it's never removed.)
Repro: install bible-stack → tool appears below the reader; uninstall → tool remains.
Fix: `yield context.tools.registerBelowReaderTool({...});` and drop the no-op disposer.

---

### 🟡 Low

**L1 — bible-stack tool/pane show the raw slug `"bible-stack"` instead of the localized title**
`packages/bible-stack/bootstrap.tsx:40,60`

The below-reader tool and floating pane pass literal `title: "bible-stack"` (and `id: "bible-stack"`). `extension.json` defines localized `title` / `below-reader-tool` strings ("Bible Stack", and 40+ locales) that are never referenced, so the UI displays the untranslated slug. `scripture-map` passes a translation descriptor (`{ key, defaultValue, ns }`) instead.
Fix: pass the translation key/descriptor so the localized name renders.

**L2 — `ActivityIndicatorMapper.toDomain` rejects a valid `index === 0`**
`patterns/bible-stack/bible-stack/infrastructure/mappers/ActivityIndicatorMapper.tsx:29`

For non-`"regular"` indicators, `if (!bot.tags.index)` throws when `index` is 0, but `index: number` is required in the domain model (`domain/models/canvas.tsx`), so 0 is legitimate. The `"regular"` branch correctly uses `typeof bot.tags.index !== "number"`.
Fix: use `typeof bot.tags.index !== "number"` here too.

**L3 — `resetToDefault()` doesn't reset label-date state**
`patterns/bible-stack/bible-stack/application/services/ScripturePiecesStateService.tsx:24`

`resetToDefault()` restores `#arePiecesDraggable` but leaves `#shouldShowLabelDates`, so label-date state leaks across experiences (`ExperienceService.clearExperience()`).
Fix: also reset `this.#shouldShowLabelDates = SHOULD_SHOW_LABELS_DEFAULT;`.

**L4 — Type tightening leaves dead fallbacks and two tests asserting a now-impossible state**
`packages/seed-bible-utils/domain/models/arrangement.tsx`, `.../infrastructure/models/arrangement.tsx`, `test/unit/seed-bible-utils/application/services/ScriptureService.test.ts`

This PR makes `SubsetBookInfo(Config).startIndex/endIndex` required (was optional). As a result, `ScriptureService` (unchanged) still carries `startIndex ?? 0` fallbacks that are now unreachable, and two existing tests ("defaults startIndex to 0 when not provided", "treats missing startIndex as 0…") assert `startIndex: undefined` behavior — a state the types now forbid. They still pass only via the `Partial` factory override plus the dead fallback, so they give false confidence about a path that can't occur in production. Either keep the fields optional, or drop the now-impossible test cases and the dead `?? 0` fallbacks.

---

### Nits (non-blocking)
- `patterns/bible-stack/bible-stack/infrastructure/adapters/environment/ColorLerper.tsx`: builds 19 intermediate colors (`for j = 1; j < 20`) but ticks 20 times → an out-of-bounds read on the final tick. Harmless today (the `undefined`→white is overwritten by the final-color write in the same synchronous callback), but worth tidying.
- `patterns/bible-stack/bible-stack/infrastructure/functions/layout.tsx:456` `GetLabelFormAddress` → `ClosestNumber` returns the raw input for arrays with fewer than 2 entries; a single-entry `formAddresses` config would index `undefined`. Config-dependent, likely never hit in practice.

---

### What looked good
- The ported domain math (`Geometry`, `colors`, `math`, `time`), the mappers/controllers (thin, correct pass-throughs with guarded lookups), object pooling, and timer/subscription cleanup all held up under review. No leaked `@preact/signals` `effect()` disposers were found anywhere in the ported code.
- The existing `ScriptureService` unit tests are real behavioral tests, not always-green.

Given the High-severity issues (H1–H3 affect activity indicators, label-date toggling, and multi-arrangement navigation), I'd recommend fixing those before merge and adding a few unit tests around the ported services to lock in the fixes.

---

Blitzy:
> [!NOTE]

## Summary

This PR ports the **Bible Stack** extension to the refactored architecture and tightens two arrangement types, closing issues 776 and 1118. It's very large — 739 files, ~51k added lines — but almost all of it is the ported pattern under `patterns/bible-stack/`. The hand-written integration outside the pattern is only ~12 files.

**The previous round was handled well.** A review was posted on this PR on Aug 11. Commit `48196b0c` plus follow-ups resolved essentially all of it — I re-verified each item at the current head (`acc0b5a7`):

| Prior finding | Status |
| --- | --- |
| Inverted guard in `InfoLabelData.addActivityIndicator` | ✅ Fixed |
| `disableLabelDates()` setting `true` | ✅ Fixed |
| `resetToDefault()` not resetting label dates | ✅ Fixed |
| Wrong `arrangementIndex` after reorder | ✅ Fixed (real index tracked via `.map((a, index) => ...)`) |
| `PaintService.unpaint` inverted guard | ✅ Fixed |
| `updateNotification` falling through on `shouldHide` | ✅ Fixed |
| `UpperCoverOpacityAdapter` `return` vs `continue` | ✅ Fixed |
| Extension tool never unregistered on uninstall | ✅ Fixed |
| Raw `"bible-stack"` slug instead of localized title | ✅ Fixed |
| Dead `?? 0` fallbacks + now-impossible tests | ✅ Fixed in `seed-bible-utils` — **not in the pattern's copies**, see M2 |
| `ActivityIndicatorMapper` rejecting `index === 0` | ❌ Still present, see L1 |

**An important framing point up front:** a substantial slice of the ported code is **unreachable in the shipped flow**. All four presence data sources are hardcoded stubs, so activity indicators, notifications, user colors, and presence navigation are dead code today. That's defensible for a port laying groundwork, but it changes how several findings should be triaged — and it means those paths are untested by construction. I've split the findings accordingly: **2 High (live)**, **5 Medium (live)**, **4 Dormant** (correct code matters only once sharing is wired), **4 Low**.

---

## The dormancy, and why it matters for triage

Every presence input to the pattern is stubbed in `patterns/bible-stack/bible-stack/infrastructure/di/bootstrap.tsx`:

```tsx
// :740 — UserPresenceService
userPresenceProviderPort: {
  getSelectedReadingInstance: () => undefined,
  getRemotesPresence: () => new Map(),
  getCurrUserId: () => authBot?.id ?? configBot.id,   // the only real one
},
// :768
userColorStorePort: { getUserColor: () => undefined },
// :771
readingInstanceProviderPort: {
  getOwnReadingInstances: () => [],
  getRemotesReadingInstances: () => [],
},
// :1101 — StackPresenceNavigationService
presenceProviderPort: { getActiveTab: () => undefined },
```

These are the only implementations that exist. `PieceActivityService` is constructed exactly once (line 761), there is one DI bootstrap, and `packages/bible-stack/` never feeds reading-instance data into the pattern.

There's a second, independent reason nothing can be shared. `packages/bible-stack/bootstrap.tsx:37` mints a fresh CasualOS instance on every click:

```tsx
onSelect: () => {
  const dimension = "stack";
  const inst = uuid();        // ← new random inst, per click
```

The `inst` is what decides which clients share a collaborative space. With no way to supply or receive one, two users can never land in the same instance — and neither can the same user twice, so the stack doesn't persist across pane reopenings either.

Worth noting the app already has the infrastructure this would need: `SessionsManager.tsx:1542-1588` exposes `createSession` / `joinSession` over Yjs shared docs. It's simply not connected to the pattern.

**Consequence for review:** `getPieceActivity` iterates `allReadingInstances`, which is always `[]`, so `pieceActivity.length === 0` is always true. Everything gated behind that is unreachable. I'd ask that the PR description state this explicitly — both so reviewers know these paths are untested by construction, and so whoever wires the ports later knows the D-series findings below are waiting for them.

---

## 🔴 High — live today

### H1 — A single failed updater permanently freezes the entire stack

`patterns/bible-stack/bible-stack/application/services/StackUpdateService.tsx:57-102`

`updateAllStacks` sets a lock and blocks interaction, then clears both **only on the success path**. There is no `try/finally`:

```ts
this.#isUpdating = true;
this.#pieceInteractabilityPort.blockAll();
// ...build updates...
await Promise.all(updates);                    // ← any rejection escapes here
this.#pieceInteractabilityPort.unlockAll();    // never runs
this.#isUpdating = false;                      // never runs
```

**Before/after in plain terms:** normally an update runs, pieces are briefly locked, then unlocked. If any one updater throws, the pieces are locked and *stay* locked — forever. Worse, `#isUpdating` stays `true`, so every later `updateAllStacks()` call hits the early-return at line 58, sets `#isUpdateQueued = true`, and does nothing. The user is left with a visible but completely dead, unclickable stack for the rest of the session.

Reachable and unrelated to presence: the updater chain has many unguarded `throw`s — `BookStackUpdaterService.#prepareRegularBook` throws `"command.data.piece not defined"`, and `BookChaptersManagementService.showChapters` / `updateChaptersPosition` throw `"bookData.piece not defined"`.

**Repro:** open the Bible Stack pane, then trigger any state where a book's `piece` is undefined during an update (e.g. race a rapid section explode against a stack update). Once one throw lands, no piece in the pane responds to clicks again until the pane is closed and reopened.

**Fix:**
```ts
this.#isUpdating = true;
this.#pieceInteractabilityPort.blockAll();
try {
  // ...build and await updates...
} finally {
  this.#pieceInteractabilityPort.unlockAll();
  this.#isUpdating = false;
}
if (this.#isUpdateQueued) { ... }
```

### H2 — `hideLabel` throws when there's no label, and most callers neither guard nor await it

`patterns/bible-stack/bible-stack/application/services/PieceLabelService.tsx:177-184` (the throw), with callers across several services

```ts
async hideLabel(piece, pacing = "Regular"): Promise<void> {
  const labelData = this.getPieceLabel(piece);
  if (!labelData) {
    throw new Error(`PieceLabelService: labelData not found at hideLabel`);
  }
```

`showLabel` is the mirror operation and no-ops gracefully on a missing label; `hideLabel` throws. Because most call sites invoke it un-awaited, each throw becomes an **unhandled promise rejection** rather than a caught error.

The clearest case is `SectionStackUpdaterService.prepareSection` (`SectionStackUpdaterService.tsx:44-54`): it spawns a brand-new shadow — which by definition has no label attached yet — and immediately calls `hideLabel(data.shadow!)` un-awaited. That's an unhandled rejection on *every first update of a split, non-exploded section*, an entirely ordinary path.

The same shape appears throughout the deletion paths — `PieceLifecycleService.tsx:520, 550, 555, 582, 606, 629` — covering every deleted testament, section, book and chapter, almost none of which own a label.

`BookChaptersManagementService.tsx:104` already shows the correct pattern: check `getPieceLabel` first, then hide.

**Fix (preferred):** make `hideLabel` no-op on a missing label, matching `showLabel`:
```ts
const labelData = this.getPieceLabel(piece);
if (!labelData) return;
```
Otherwise guard every call site — but that's ~10 sites and easy to miss on the next one.

---

## 🟠 Medium — live today

### M1 — `#disposePool` skips roughly half the in-use bots (array mutated while iterating)

`patterns/bible-stack/bible-stack/infrastructure/adapters/environment/ObjectPooler.tsx:124-141`

`#disposePool` passes the **live** `pool.inUseObjects` array into `releaseObjects`, which iterates it with `for..of` while `releaseObject` splices out of that same array on each step.

**Concrete walkthrough** with 4 in-use bots `[A, B, C, D]`:
- Iterator index 0 → A. `releaseObject` splices A out; array is now `[B, C, D]`.
- Iterator index 1 → `C` (B shifted into index 0 and was skipped).
- Iterator index 2 → nothing; loop ends.

B and D are never released. The loop below then destroys only `pool.objectPool`, so **B and D survive `disposeAllPools()` as live CasualOS bots** — orphaned objects left in the dimension after the pane is torn down.

**Fix:** iterate a copy — `this.releaseObjects([...pool.inUseObjects], key)`.

Related, same file: `releaseObject` (line 102) dereferences `pool.inUseObjects` with no `if (!pool)` guard, unlike `getObject`. An unknown key throws a raw `TypeError` instead of a useful message.

### M2 — The `startIndex`/`endIndex` fix landed in `seed-bible-utils` but not in the pattern's duplicate copies

`patterns/bible-stack/bible-stack/infrastructure/mappers/BookInfoMapper.tsx:52-63`, `.../application/services/ScriptureService.tsx:42,55,56,67`, `.../application/services/ArrangementService.tsx:392-393`

This is directly tied to what the PR set out to do. The pattern carries its own copies of `BookInfoMapper`, `ScriptureService` and `ArrangementService`. The PR correctly updated the `seed-bible-utils` versions — removing the now-dead `?? 0` fallbacks and writing the destructured values — but the pattern copies still carry the pre-fix code, even though the pattern's own `SubsetBookInfoConfig` (`infrastructure/models/arrangement.tsx:38-39`) declares both fields as **required**.

The pattern's `BookInfoMapper` is worse than merely stale — it retains the latent bug the utils fix removed:

```ts
const startIndex = info.startIndex ?? 0;                              // defaulted local
const endIndex = info.endIndex ?? completeStaticInfo.numberOfChapters - 1;
const chaptersVerseCount = completeStaticInfo.chaptersVerseCount.slice(startIndex, endIndex + 1);
return {
  startIndex: info.startIndex,   // ← raw, possibly undefined
  endIndex: info.endIndex,       // ← raw, possibly undefined
```

A subset config missing `startIndex` yields verse counts sliced from 0 but a domain object carrying `startIndex: undefined`. `BookChaptersManagementAdapter.tsx:118` then computes `chapterInfo.number + undefined` → `NaN`, and **every chapter label in that book renders "NaN"**.

The static config supplies all 25 subset entries today, so this is latent — but it's exactly the class of bug the type tightening was meant to eliminate, and it survives in the copy that will be edited next.

**Fix:** port the same three diffs from `packages/seed-bible-utils/` into the pattern copies. Longer term, these three files are now a third copy of the same logic — per the repo's duplication convention, worth importing from `seed-bible-utils` rather than maintaining in parallel. This PR is itself a live demonstration of the copies drifting.

### M3 — `deleteChapter` leaks verse data on every path except one

`patterns/bible-stack/bible-stack/application/services/PieceLifecycleService.tsx:620-632`

```ts
deleteChapter(chapter: StackChapterData) {
  this.#pieceDataRepositoryPort.removeChapterData(chapter);
  const piece = chapter.clearPiece();
  if (chapter.isOnTheGround && chapter.isSelected) {   // ← very narrow condition
    const bundles = chapter.clearChildren();
    for (const bundle of bundles) this.deleteVersesBundle(bundle);
  }
  ...
```

Verse bundles are only torn down when the chapter is **both** on the ground **and** selected. Every other deletion drops the chapter from its repository while leaving its `VersesBundleData` and `VerseData` registered in `#versesBundleDataRepositoryPort` / `#verseDataRepositoryPort` with no owner.

The dominant path hits this: `ExperienceService.clearExperience` → `clearAllStacks` → `deleteBibles` → … → `deleteChapter` for all ~1,189 chapters, essentially none of which are on-the-ground-and-selected. That strands roughly **2,600 bundle objects and 31,100 verse objects** per teardown. Recreating the bible allocates a fresh set on top, so memory grows without bound across repeated bible resets in one session.

**Fix:** always `clearChildren()` and delete the bundles; keep the `isOnTheGround && isSelected` check only for whatever piece-releasing side effect actually needs it.

### M4 — Divide-by-zero produces `NaN` chapter positions at small book scales

`patterns/bible-stack/bible-stack/infrastructure/adapters/stacks/BookChaptersManagementAdapter.tsx:213-219`

```ts
const chapterHorizontalGap = horizontalEmptySpace / columns;      // columns can be 0
const chapterVerticalGap = verticalEmptySpace / (rows - 1);       // rows can be Infinity
```

`SelectedBookLayoutAdapter` computes `columns = Math.floor(scaleX / (chapterWidth + chapterGap * 2))`. With `ChapterWidth = 0.5` and `ChapterGap = 0.05`, any `scaleX` below `0.6` gives `columns = 0`. Then `rows = Math.ceil(n / 0) + 1 = Infinity`, `chapterHorizontalGap` becomes `±Infinity`, and `chapterVerticalGap` becomes `NaN`. Every `SetStrictTag` writes `NaN` into the dimension X/Y/Z tags and **the chapters disappear from the scene**.

For a `StackSectionBook`, `scaleX` comes from the registry's `unhoveredScales.x`, so this is reachable from configuration rather than a pathological runtime state. The loop's `column >= columns` check also advances a row per chapter when `columns` is 0.

**Fix:** clamp early — `const safeColumns = Math.max(1, columns);` — and guard `rows - 1` against 0 before dividing.

### M5 — `resetHierarchy(clearPiece)` silently stops propagating at the first level

`patterns/bible-stack/bible-stack/domain/entities/StackPieceData.tsx:386-397` and `StackData.tsx:35`

Every subclass threads a `clearPiece` flag down the chain — `StackSectionBookData:88`, `StackTestamentData:96`, `StackBookBaseData:158`, `StackBookData:103` all forward it — but the chain breaks at the bottom:

```ts
// StackPieceData
if (this.piece && clearPiece) { ... }              // honoured here
piecesToRelease.push(...super.resetHierarchy());   // ← flag dropped

// StackData — takes no parameter at all
resetHierarchy(): Piece[] {
  ...currChildData.resetHierarchy()                 // ← children default to clearPiece = true
```

**Before/after:** `SectionSelectionService.tsx:246` calls `data.resetHierarchy(false)` intending to *preserve* the pieces in that subtree. What actually happens: the section's own piece is preserved, and every book and chapter beneath it has its piece cleared and released anyway. The plumbing through four subclasses makes the intent unmistakable — it just can't take effect.

**Fix:** give `StackData.resetHierarchy` a `clearPiece = true` parameter and pass it through both the recursive call and `super.resetHierarchy(clearPiece)`.

---

## 🔵 Dormant — correct these before wiring up sharing

These are real defects in the code, but none can fire today because of the stubs described above. They are pre-planted: each will surface on the day someone connects the presence ports, which is the worst time to discover them. All four are small fixes; I'd take them now while the context is fresh rather than leave them for whoever does the sharing work.

### D1 — Activity would never appear on subset books (Psalms, etc.)

`patterns/bible-stack/bible-stack/application/services/PieceActivityService.tsx:242-300`

`getPieceActivity` correctly *detects* that a reading position falls in a subset book and resolves the arrangement path for it — then builds the match key from the **original complete** book id anyway.

Walk through the default arrangement, where Psalms is split into subsets `1PS`–`5PS` (each with `completeBookId: "PSA"`), for a reader at `PSA 50`:

1. `getBookInfoPathById({ id: "PSA" })` → not found (only subsets are in the arrangement).
2. Fallback: `getBookSubsetByCompleteId` returns subset `2PS` (`startIndex: 41`).
3. `getBookInfoPathById({ id: "2PS" })` → found; `testamentIndex` / `sectionIndex` / `arrangementIndex` all updated.
4. **But** the `path` built at lines 293-300 still uses the original loop variables: `key: bookId` → `"PSA"`, and `` key: `${bookId} ${chapter}` `` → `"PSA 50"`.

Meanwhile `bookActivityStrategy` reads the piece's own `bookId` → `"2PS"`, and `chapterActivityStrategy` builds `"2PS 9"`. The filter at line 324 compares `"2PS"` against `"PSA"` and never matches.

Once presence is live this reads as "indicators work for John but never for Psalms" — for every user, on all five Psalms subsets. `StackPresenceNavigationService` already does this remap correctly via `mapSubsetToCompleteBook`; this path just never applies the result.

**Fix:** after the successful subset lookup, carry the subset identity forward — capture `const pathBookId = bookSubset?.bookId ?? bookId` and the subset-relative chapter (`chapter - startIndex`), and build the `StackBook` / `StackChapter` keys from those.

### D2 — One rejection would permanently disable presence navigation

`patterns/bible-stack/bible-stack/application/services/StackPresenceNavigationService.tsx:132,166-170`

Same shape as H1. `update()` sets `#isThereAnOngoingUpdate = true`, awaits the navigation work, and clears the flag only after success. If the await rejects, the flag is never cleared and every subsequent `update()` takes the `shouldQueue` branch at line 132 and returns.

Today `update()` returns early at `!ativeReadingInstance` (since `getActiveTab` → `undefined`), so the flag is never even set. Once wired, the rejection path is reachable: `#navigateToChapter` → `bibleSequenceServicePort.resetBible` → `closeBible`, which throws `"upperCover not found at closeBible"` among others. The stack would silently stop following the reader's position for the rest of the session, with no error surfaced.

Compounding it, `handleSectionExploded` (line 125) calls `update()` as a floating promise with no `.catch()` — so the rejection also becomes an unhandled rejection.

**Fix:** `try/finally` as in H1, plus a rejection handler at the `handleSectionExploded` call site.

### D3 — `updateNotification` would leak a pooled bot on every refresh

`patterns/bible-stack/bible-stack/application/services/PieceActivityService.tsx:640-650`

`updateNotification` unconditionally requests a **new** notification from the object pool:

```ts
const newNotification = this.#activityNotificationAdapterPort.showNotification({
  isOwnUserInPiece, activityCount, color, direction, container,   // ← no `notification` field
});
container.attachActivityNotification(newNotification);
```

`ShowNotificationCommand` supports reusing an existing notification via its optional `notification` field (`ActivityNotificationAdapter.tsx:74`), but this call never passes it, so `getObject("ActivityNotification")` hands out a fresh bot. Then `StackChapterData.attachActivityNotification` (line 152) only assigns `if (!this.#activityNotification)` — so when one is already attached, the new bot is **silently dropped without ever being released**.

Currently unreachable: `shouldHide` is always true (activity is always empty), so the method returns at line 623.

**Fix:** pass `notification: container.activityNotification` when one is attached.

### D4 — The "+N extra users" badge would be destroyed and respawned on every update

`patterns/bible-stack/bible-stack/application/services/PieceActivityService.tsx:466-472`

The stale-indicator prune loop filters only on index, without checking `indicatorType`:

```ts
const limit = Math.min(pieceActivity.length, this.#maxIndicators);
for (const indicator of currIndicators) {
  if (indicator.index >= limit) {          // ← no indicatorType check
    this.#activityIndicatorsAdapterPort.hideIndicator(indicator);
    container.removeActivityIndicator(indicator.id);
  }
}
```

The `extraContent` / `extraBackground` indicators are created with `index = activityIndex` at lines 512-522 — exactly `maxIndicators`. With `maxIndicators = 4` and 6 users, `limit = 4`, so both extras satisfy `index >= limit`, get released to the pool, and are then immediately re-allocated a few lines later when `getDataExtraActivityIndicators` returns `undefined`. Visible flicker of the "+N" badge on every refresh plus pool churn.

Currently unreachable: `updateIndicators` returns at line 461 (`pieceActivity.length === 0`) before the prune loop.

Every other read path gets this right — `getPieceIndicatorByActivityIndex` (line 389) and `getDataActivityIndicatorByType` (line 423) both filter `indicatorType === "regular"` first.

**Fix:** `if (indicator.indicatorType === "regular" && indicator.index >= limit)`.

**Also dormant, perf only:** `updateAllNotifications` (line 661) rebuilds the full reading-instance → arrangement-path map for every chapter, and each entry runs `getBookInfoPathById`, which linearly scans the whole arrangement. Across ~1,189 chapters with 5 remote readers that's ~6,000 identical full scans per section explode. Free today (the map is empty); worth hoisting the map build to once per `updateAll*` pass before presence goes live.

---

## 🟡 Low

**L1 — `ActivityIndicatorMapper.toDomain` still rejects a valid `index === 0`** (`ActivityIndicatorMapper.tsx:29`). Carried over from the previous review, still unfixed. For non-`"regular"` indicators, `if (!bot.tags.index)` throws when `index` is `0`, but `index: number` is required in the domain model. The `"regular"` branch already does it correctly with `typeof bot.tags.index !== "number"` — use the same check in both.

**L2 — Dead ternary and an unreachable pre-flight branch** (`BookStackUpdaterService.tsx:127`). `data.type === "StackSectionBook" ? { data } : { data }` — both branches identical. More substantively, `update()` never supplies `sectionData`, so `#prepareRegularBook`'s first condition (`command.sectionData && !command.sectionData.isInExplodedView`) can never be true from this entry point; only the `selectionState === "Deselecting"` path ever hides chapters. Callers that do pass context (`SectionStackUpdaterService.tsx:60`) get the intended behavior, so the two entry points silently disagree. Also, `finalizeBook`'s `catch` at line 114 is commented "No label currently attached to the book" but wraps `showLabel`, which doesn't throw for that reason.

**L3 — All verse data is materialized up front** (`PieceLifecycleService.tsx:409`). `createChapter` eagerly builds every `VersesBundleData` and `VerseData` at bible-creation time — with `VERSES_PER_BUNDLE = 12`, ~2,600 bundles and ~31,100 verse objects allocated synchronously before the bible is displayed. Pure startup cost: `ChapterSelectionService.#prepareSelection` only attaches pieces for a chapter actually on the ground. Create bundles lazily on first selection.

**L4 — ~15k lines of dead reference code, and a 14× pattern bundle increase.** The PR adds **275 `legacy_*.txt` files (13,228 lines)** plus **22 `inspect_*.tsx` files containing zero executable statements** — every line commented out (`inspect_TrySetShape.tsx` is 293 lines of nothing; `inspect_Select.tsx`, `inspect_TryPickChapter.tsx`, `inspect_SpawnSection.tsx` and 18 more are the same).

One correction to the previous review, which waved these through as "the repo's existing convention": the only other pattern, `geo-importer`, contains **no `legacy_*.txt` files at all** — only `.aux`, `.json` and `.tsx`. This is a new convention, not an existing one. CI's build size report shows the cost: `pattern-dist` goes **73KB → 1.1MB (+1008KB)**. Since `casualos pack-aux` packs the pattern directory wholesale, this material plausibly ships in the `.aux` every user downloads when opening the pane.

**Suggestion:** delete them, or move them outside the packed pattern directory (a top-level `docs/legacy/` or an archive branch). Git history already preserves the legacy scripts.

---

## Notes on the tests

- **The test changes that are here are correct.** The only test diff is 73 lines *removed* across three files, plus one factory fix. Removing "defaults startIndex to 0 when omitted" and friends was right — those cases asserted a state the tightened types now forbid, and keeping them would mean tests passing only via the `Partial` factory override plus a dead fallback, giving false confidence about an unreachable path. One surviving assertion was properly strengthened from `expect(...).toBeUndefined()` to `expect(...).toBe(4)`. The remaining `seed-bible-utils` tests are genuine behavioral tests that would fail on a regression.
- **But there are no new tests.** ~51k added lines; CI reports `packages/bible-stack` at **0.0% statements, 0.0% functions**. That number is now partly explained by the dormancy — you can't meaningfully cover code that can't run — but H1, H2, M1 and M5 are all live, in plain TypeScript classes with injected ports, and testable in a few lines each:
  - H1: make one updater port reject → assert a second `updateAllStacks` still performs work (currently it no-ops forever).
  - H2: call `hideLabel` on a piece with no label → assert it resolves rather than rejecting.
  - M1: 4 in-use pooled objects → assert all 4 are released (currently 2 are).
  - M5: two-level hierarchy, `resetHierarchy(false)` → assert no child piece was released.
  - D1 is equally testable with a stubbed reading instance at `PSA 50` and a `2PS` chapter piece — and doing it now would lock the fix in before sharing work begins.
- If any of these get fixed, please confirm the test fails on the pre-fix code first — several are the "test passes either way" trap if the assertion is written loosely.

## What held up well

The ported domain math (`Geometry`, `colors`, `math`, `time`), the mappers and controllers (thin pass-throughs with guarded lookups), and timer/subscription cleanup all read correctly. No leaked `@preact/signals` `effect()` disposers anywhere in the ported code. The `seed-bible-utils` type tightening is safe against the static arrangement config — all 25 `subset` entries already supply the now-required fields — and the extension bootstrap now matches the `scripture-map` reference shape.

## Recommendation

**H1 is the one I'd treat as merge-blocking.** It leaves the feature permanently dead for the rest of a session with no error visible to the user, it's reachable from ordinary state, and the fix is a three-line `try/finally`. H2 is close behind — unhandled rejections on an entirely ordinary path, fixed by one `return`.

M1 through M5 are all worth taking in this PR; none are large. M2 matters disproportionately given the PR's stated purpose — the type fix landed on one of two copies.

The D-series findings shouldn't block this merge, but I'd fix them here rather than file them: they're small, the context is loaded now, and by construction nobody will be looking at that code again until it's already shipped and someone is debugging why presence half-works.

The PR is still draft with `mergeable_state: blocked`, so there's room to land these before it opens for merge.

---

Blitzy:
@soukigabriel Just reiterating what the latest code review recommends above: lets get **H1**, and **M1-5** fixed.

All of the shared session changes we can move to a new issue that makes the Bible Stack natively shared when opening in a shared seed bible session. I dont want to block merging on these since getting into a shared bible stack session is not possible in the current user flow anyway.

PR #1636, opened 2026-08-12
Record annotations offline and sync them when online

KallynGowdy:
Annotations were the least protected data in the app despite being the only user-written prose. Every write went straight to the server, so with no connection saving threw and whatever had been typed was lost. Reading was worse: a failed list request was cached as "no annotations" *and marked settled*, so an offline visit reported an empty chapter and kept reporting one for the life of the page.

Notes are now recorded on the device first and pushed when a connection is available. Creating, editing and deleting all work offline, and a count of what is still waiting is shown in the notes pane.

Conflict handling had to come with it rather than after it. A chapter's list was fetched once per page load and never revalidated, and `recordData` is a blind overwrite with no server version or etag — so two devices editing one note already clobbered each other silently. Deferring writes makes that strictly worse, since a change queued for hours is far more likely to be stale. Every write now reads the server's copy first and checks it is still the one the local edit was based on; when it isn't, the user is asked which version to keep before anything is overwritten.

Notable decisions:

- One IndexedDB store carrying its own sync metadata, not a mirror plus a separate outbox. IndexedDB skips null index values, so the index over `pendingOp` *is* the queue. Repeated offline edits then coalesce for free, and a create-then-delete offline costs no requests at all.
- Conflicts compare against a remembered base version, not "whose timestamp is larger" — timestamps are written by whichever device made the edit, so clock skew between devices makes "newer" unreliable. A content fingerprint covers records that carry no timestamp.
- A separate `seed-bible-annotations` database rather than a version bump on `seed-bible-offline`, whose open path rejects on `onblocked` and would break offline translations in a tab holding the old connection.
- Notes written while signed out go to a device-local bucket and are adopted into the account on first sign-in, cleared in the same transaction so a second account on a shared device cannot inherit them — the leak `LoginManager` already guards against for anonymous config.
- Sign-out drops synced rows but keeps unsent writing: synced notes are re-fetchable, and leaving readable prose on a shared device is what clearing the profile cache exists to prevent.
- A sync pass captures the account once and re-checks it before every write, so a queued change can never be pushed into whichever account happens to be signed in when connectivity returns (#1587).

Claude-Session: https://claude.ai/code/session_0125RRhUgDbdQFxpTAtB9V3S


Blitzy:
> [!NOTE]

## Code review: Record annotations offline and sync them when online

This PR's whole purpose is to stop annotations from being silently lost when the network is flaky, so I focused the review on exactly that: can an edit still disappear without the user being told? Two of the three findings below are races that do exactly that — one of them sits right in the everyday "edit → save → edit again" path, not an exotic corner case.

Verified by reading the actual code paths (not just the diff) in `AnnotationSyncManager.tsx` and `OfflineAnnotationStore.ts`; line numbers refer to the PR's head commit.

---

### 🟠 Medium — No. 1: A push in flight can be overwritten by clobbering the edit that happened during it, so the later edit never reaches the server and disappears locally

**Where:** `AnnotationSyncManager.tsx`, `pushRow` → `writeToServer` (around lines 313–329)

```ts
const writeToServer = async (
  owner: string,
  annotation: Annotation
): Promise<PushOutcome> => {
  const result = await os.recordData(owner, annotation.id, annotation, {...});
  if (!result.success) return classifyFailure(...);
  await store?.put(syncedRow(owner, annotation));   // <-- writes the OLD annotation
  onSynced?.(annotation, owner);
  return { status: "done" };
};
```

`runPass` reads all pending rows once via `store.listPending(owner)`, then pushes each one. `pushRow` calls `writeToServer(owner, row.annotation)` with the annotation content captured at that read. If the user edits and saves the *same* note again while `os.recordData` is still awaiting the network, the new edit is written to IndexedDB as a fresh pending row — but when the original push resolves successfully, `writeToServer` still writes `syncedRow(owner, <old annotation>)`, stamping the row back to the old content with `pendingOp: null`. The newer edit is never pushed and is gone from local storage; `onSynced` also pushes the old content back into the in-memory/UI cache, so the visible text reverts too.

**Reproduction:**
1. Go offline is not even required — this reproduces on a slow connection. Edit annotation `ann-1` to "v1" and save (a push starts: `readServer` + `os.recordData` in flight).
2. Before that network round-trip resolves, edit `ann-1` again to "v2" and save.
3. When the "v1" push resolves successfully, the local row is overwritten back to "v1" with `pendingOp: null`. "v2" is never sent to the server and is not recoverable locally — no error is surfaced to the user.

A unit test that fires two saves on the same annotation with the first `os.recordData` resolved only after the second `store.put` would fail on the current code (final stored/displayed content is "v1", not "v2").

**Suggested fix:** don't trust the closed-over `annotation` for the local write — re-read the row from the store right before writing, and only clear `pendingOp` if it still matches what was actually pushed (e.g. compare against a version/fingerprint captured when the push started, the same pattern already used for server-conflict detection). If the row changed underneath the push, leave it pending so the next pass picks up the newer edit.

---

### 🟠 Medium — No. 2: `reconcileChapter` reads existing rows before opening its write transaction, so a concurrent edit landing in that gap gets overwritten with stale server data

**Where:** `OfflineAnnotationStore.ts`, `reconcileChapter` (lines 434–459) and `applyReconcile` (lines 559–587)

```ts
const reconcileChapter = async (owner, bookId, chapterNumber, serverAnnotations, listedAtMs) => {
  const database = await openDatabase();
  const existing = await listForChapter(owner, bookId, chapterNumber); // readonly tx, awaited
  const transaction = database.transaction([ANNOTATIONS_STORE, CHAPTERS_STORE], "readwrite"); // opened after
  ...
  applyReconcile(existing, serverAnnotations, owner, { put, delete });
  ...
};
```

`applyReconcile` only skips overwriting a row when the **stale** `existing` snapshot already shows a `pendingOp`. Between the `listForChapter` await and the readwrite transaction opening, a user edit can land (writing a new `pendingOp: "upsert"` row for an id that was synced in the snapshot). Because `existing` doesn't reflect that new write, `applyReconcile` still calls `writer.put(syncedRow(owner, serverAnnotation))` for that id — overwriting the fresh edit with the old server content and clearing the pending flag, so the edit is silently dropped and never queued for sync.

**Reproduction (by code inspection):**
1. A chapter reload triggers `reconcileChapter`; `listForChapter` returns the current (synced) rows for that chapter.
2. Before the subsequent readwrite transaction is opened, the user edits one of those already-synced annotations, writing a new pending row.
3. `applyReconcile` runs against the stale `existing` list, sees no `pendingOp` for that id, and overwrites the just-written edit with the server's (older) copy.

This window is narrower than No. 1 (it's bounded by one IndexedDB round-trip rather than a full network request), but it's the same class of bug: reconciliation trusting a snapshot that can go stale before it's applied.

**Suggested fix:** re-check each row for a pending op inside the same readwrite transaction that performs the writes (e.g. `store.get(key)` per id right before `put`), rather than deciding from a separately-read snapshot.

---

### 🟢 Low — No. 3: The "N changes waiting to sync" banner in a chapter's Notes section shows the account-wide pending count, not that chapter's

**Where:** `DiscoverPane.tsx` line 901 (`const pending = annotations.sync.pendingCount.value;`), backed by `AnnotationSyncManager.tsx` line 209 (`const pendingCount = computed(() => pendingRows.value.length);`)

`pendingCount` is computed over every pending row for the signed-in account, not filtered by the chapter currently being viewed. A user with one unsynced note in Exodus who opens Genesis 1's Notes section will still see "1 change waiting to sync" there, even though nothing in Genesis is actually pending — which reads as "this chapter has an unsynced note" when it doesn't.

**Suggested fix:** either scope the banner's count to the current chapter (filter pending rows by `bookId`/`chapterNumber` the same way `getAnnotationsForChapter` does), or, if the intent is genuinely account-wide, move the banner out of the per-chapter Notes section into a more global location and reword it (e.g. "N notes waiting to sync across your account").

---

### On the tests

The new suites (`AnnotationSyncManager.test.ts`, `OfflineAnnotationStore.test.ts`) are real behavioral tests for the paths they cover — conflict detection, offline queuing, coalescing of repeated offline edits, and the account-adoption/clearing rules all have assertions that would fail if the corresponding guard were removed. However, every case exercises **sequential, non-overlapping** edits; nothing drives a second write while a push (No. 1) or a reconcile (No. 2) is still in flight, so neither race above would be caught by CI today. A test that resolves `os.recordData` only after a second `store.put` lands (No. 1), and a test that injects a write between `listForChapter` and the reconcile transaction (No. 2, may need a small test seam to simulate), would both fail on the current code and pass once fixed.

---

### Summary

| Severity | Item | Issue |
|---|---|---|
| 🟠 Medium | No. 1 | An in-flight push can overwrite a newer edit to the same note, silently dropping it |
| 🟠 Medium | No. 2 | `reconcileChapter`'s stale read-then-write gap can overwrite a concurrent edit with old server data |
| 🟢 Low | No. 3 | Per-chapter pending-sync banner actually shows the account-wide count |

No. 1 and No. 2 both undermine the core promise of this PR — that an edit, once saved, won't quietly disappear — so I'd fix those before merge. No. 3 is cosmetic/UX and can reasonably follow up separately.

---

Blitzy:
> [!NOTE]

## Code review — Record annotations offline and sync them when online

This is a substantial, well-designed change — the single-store-with-its-own-sync-metadata approach (using IndexedDB's null-key-skips-the-index behavior to make the pending-index *be* the queue) is clever, the base-version conflict model correctly avoids the clock-skew problem of "newer timestamp wins," and the offline-first rework of `AnnotationsManager` is thorough. CI is green and the new test files are extensive (819 + 416 + 327 lines) and generally exercise real behavior rather than mocks-agreeing-with-mocks.

The theme of what follows: several of the atomicity/attribution guarantees the PR description explicitly claims (account-safe writes, atomic sign-out handling, "the pass abandons its work so rows survive to be pushed after the next sign-in") hold up in the *primary* write path but not in a few secondary paths that share the same data. No linked issue is referenced in the PR body, so this is reviewed against the PR description's own stated design goals — several of which mirror the precedent set by PR #1587 ("Fix stale highlights after switching accounts"), and I checked this PR's account-switch handling against that precedent specifically.

---

### 🔴 High — A session-expiry failure permanently discards the queued edit, contradicting the code's own documented behavior

**Where:** `AnnotationSyncManager.tsx` — module doc around `RETRYABLE_ERROR_CODES` (~line 44), `recordFailure` (~line 405), `classifyFailure` (~line 729); `SessionGuard.ts` `guardRecordsClient` (~line 158).

The module's own comment states the intent clearly:

> the codes meaning "this session is over" (`session_expired`, `invalid_key`, `user_is_banned`) are absent on purpose: `SessionGuard` turns those into a sign-out, and the pass abandons its work so the rows survive to be pushed after the next sign-in.

That's not what happens. `SessionGuard.guardRecordsClient` intercepts the fatal codes with `.then()`, not `.catch()`:

```ts
return result.then((resolved) => {
  const errorCode = findFatalSessionErrorCode(resolved);
  if (errorCode && getSessionKey() === sessionKeyAtRequest) {
    onSessionInvalidated(errorCode);
  }
  return resolved; // the original {success:false, errorCode:"session_expired"} still resolves normally
});
```

So `os.getData`/`recordData`/`eraseData` still resolve (never reject) with the failure object. That flows into `classifyFailure(errorCode, message)`, and since the fatal codes are deliberately excluded from `RETRYABLE_ERROR_CODES`, they classify as `"permanent"`. `recordFailure` then does:

```ts
if (outcome.status === "permanent") {
  await store?.put({ ...row, pendingOp: null }); // wipes the queued change
  return;
}
```

**Reproduction:** Sign in, go offline, edit a note (queues an `upsert`), let the session expire (or get revoked/banned from another device), come back online. The sync pass's `getData`/`recordData` call resolves `{success:false, errorCode:"session_expired"}`; `SessionGuard` fires the sign-out as a side effect but still returns the failure object normally; `AnnotationSyncManager` treats it as a permanent server rejection and clears `pendingOp` on that row. Signing back in does not retry it — `listPending()` no longer returns it, and the edit is gone.

**Suggested fix:** Treat the fatal-session codes as a third outcome, distinct from both `retry` and `permanent` — leave `pendingOp`/`attempts` untouched and let a fresh sign-in resync it — rather than folding them into "permanent" by virtue of not being in `RETRYABLE_ERROR_CODES`. Alternatively, have `runPass` check `login.userId`/session validity immediately after a failed push and skip `recordFailure` entirely when the failure's error code is one of the fatal-session codes.

---

### 🟠 Medium/High — A save or delete can be written into the wrong account's in-memory cache if the signed-in account changes mid-operation

**Where:** `AnnotationsManager.tsx` — `upsertIntoCache` (~line 830), `removeFromCache` (~line 841), `saveEditingAnnotation` (~line 1029), `deleteAnnotationAndRefresh` (~line 1051).

This is the same class of bug PR #1587 fixed for highlights, one layer up — the *durable* store write is safe (`saveAnnotation`/`deleteAnnotation` capture `owner` once and use it consistently through both IndexedDB round trips, matching the #1587 "capture once, use consistently" pattern), but the reactive UI cache update on top of it isn't:

```ts
// saveEditingAnnotation, ~line 1029
const saved = await saveAnnotation(current); // internally captured owner = A
upsertIntoCache(saved);                      // no owner passed through

// upsertIntoCache, ~line 830
const cacheOwner = owner ?? login.userId.peek() ?? LOCAL_OWNER; // re-reads the CURRENT login, not A
```

If the signed-in account changes between the start of `saveAnnotation` (two awaited IndexedDB round trips) and `upsertIntoCache` running, the just-saved content gets inserted into the *new* account's `entries` cache — with `settled = true`, so it reads as authoritative — even though the durable row was correctly written under the original account. Concretely: Alice starts saving an edit, signs out mid-flight, Bob signs in; the durable write lands correctly under Alice's id, but Bob's reading view for that chapter now shows Alice's note as if it were his own, until a full cache purge happens to run again. The delete path (`deleteAnnotationAndRefresh` → `removeFromCache`) has the same unguarded pattern, though the effect there is milder (a missed cache removal rather than a leak, since the lookup key won't match under the wrong owner).

**Suggested fix:** Have `saveAnnotation`/`deleteAnnotation` return the owner they actually wrote under (or capture it independently in the caller before awaiting), and thread it into `upsertIntoCache`/`removeFromCache` explicitly — exactly the pattern already used correctly for the sync-driven path (`onSynced: (annotation, owner) => upsertIntoCache(annotation, owner)`).

**Test gap:** The existing account-switch test in `AnnotationsManager.test.ts` (~line 397, "reflects an account switch instead of leaking the previous account's data") only exercises the read/list path — nothing changes `login.userId` mid-flight during a save or delete, so this wouldn't be caught by the current suite.

---

### 🟠 Medium — `reconcileChapter` reads and writes across two separate IndexedDB transactions, unlike the store's other atomic operations, so a concurrent local edit in the gap can be silently overwritten

**Where:** `OfflineAnnotationStore.ts` `reconcileChapter` (~line 434), `applyReconcile` (~line 559).

`adoptLocalRows` and `clearSynced` genuinely are single atomic transactions (verified this holds up under the "a transaction stays alive across microtask continuations" behavior browsers implement) — but `reconcileChapter` isn't built the same way:

```ts
const reconcileChapter = async (owner, bookId, chapterNumber, serverAnnotations, listedAtMs) => {
  const database = await openDatabase();
  const existing = await listForChapter(owner, bookId, chapterNumber); // its own separate read-only transaction, fully commits
  const transaction = database.transaction([ANNOTATIONS_STORE, CHAPTERS_STORE], "readwrite"); // a brand-new transaction
  ...
```

`applyReconcile` decides what to overwrite based on `row?.pendingOp` from that stale `existing` snapshot. Between the two transactions there's a real gap the event loop can fill with other work — including a concurrent `saveAnnotation`/`deleteAnnotation` call.

**Reproduction:** Open a chapter (kicks off a server list fetch → `reconcileChapter`). While that's in flight, edit or create a note in the same chapter (marks `pendingOp: "upsert"` via its own `put`). If that local write lands after `existing` was read but before `reconcileChapter`'s write transaction runs, `applyReconcile` still sees the pre-edit snapshot and overwrites the row with the server's version, clearing `pendingOp` — the new edit is silently discarded and never pushed.

Every `reconcileChapter` test in `OfflineAnnotationStore.test.ts` runs against `createInMemoryAnnotationStore()`, which is fully synchronous and structurally can't exhibit a two-transaction race — so this is untested in either direction. (Worth noting: the repo has no `fake-indexeddb`-style dependency anywhere, so the real transaction-boundary code — in this store and in `OfflineTranslationStore` before it — has effectively zero test coverage; not a regression this PR introduces, but the PR description's atomicity claims are only actually verified for `adoptLocalRows`/`clearSynced`, not `reconcileChapter`.)

**Suggested fix:** Fold the `existing` read into the same read-write transaction as the write (a single `readwrite` transaction can read before it writes), so there's no gap for a concurrent local write to land in.

---

### 🟠 Medium — An in-flight push can re-mark a row "synced" (readable) after sign-out already cleared that account's synced rows

**Where:** `AnnotationSyncManager.tsx` — the account-change effect (~line 671) vs. `runPass`/`pushRow` (~line 431).

Nothing cancels or awaits an in-flight sync pass when the signed-in account changes. If a row is mid-push (e.g. awaiting `os.recordData` inside `writeToServer`) at the moment of sign-out, `clearSynced(previous)` — itself correctly atomic — can run its `getAll` + filter + delete before the in-flight push's own `store.put(syncedRow(...))` lands. `clearSynced` correctly skips the row because it still looks pending at that moment (by design, so unsent writes survive sign-out) — but the in-flight push then completes *after* sign-out and writes the now-synced, fully-readable row back into IndexedDB under the account that just signed out. Per the store's own docstring, this is exactly the shared-device content leak `clearSynced` exists to prevent, just reopened by a narrow timing window (sign out immediately after triggering a save while still online, on a slow connection).

**Test gap:** The existing sign-out test ("drops synced rows on sign-out but keeps unsent writing") only covers the two steady-state cases (already-synced, never-synced) — it doesn't simulate a push still in flight at the exact moment `login.userId` flips to null.

**Suggested fix:** Have the sign-out handler cancel/await the in-flight sync pass (or have `pushRow`'s final `put` re-check that the row's owner is still the signed-in account before writing) so a completed-after-sign-out push can't resurrect a synced row for an account that's no longer active.

---

### 🟡 Low/Medium — Rows are updated via unguarded read-then-write, so a stale background-retry write can in principle clobber a fresh same-tab edit

**Where:** `AnnotationsManager.tsx` `saveAnnotation` (~line 470); `AnnotationSyncManager.tsx` `recordFailure` (~line 406).

`OfflineAnnotationStore.put` is a blind overwrite by key with no version check, and both the user-edit path and the sync pass's own bookkeeping do "read current row, write a new row built from it" without atomicity between the two calls. This doesn't need two devices — a background retry's `recordFailure` write (built from a pre-edit snapshot: old content, incremented `attempts`) resolving concurrently with a fresh same-tab edit could in principle land last and silently revert the user's newest content, with nothing surfaced. This is a narrower window than the findings above (same-tab async interleaving rather than a routine account switch or sign-out), so flagging as a design characteristic worth being aware of rather than a must-fix-before-merge item.

---

### Test quality

The new test files mock only at the real boundary (`os.getData`/`recordData`/`eraseData`/`listDataByMarker` are spied via `vi.spyOn(os, ...)`), and use `createInMemoryAnnotationStore()` — a real alternate implementation of the store interface — the same sanctioned seam already established by `OfflineTranslationStore`'s in-memory test double, not a violation of "don't mock sibling managers." Assertions are generally disciplined: the `keep_both` conflict test specifically checks the original row survives with the server's content rather than relying on a `??` fallback that would mask a silent wipe, and the "stops mid-pass when the account changes" test (`AnnotationSyncManager.test.ts`, ~line 674) is a genuine regression test — it switches accounts *from inside* the mocked server call and asserts every call that did happen used the original owner, which would fail if the per-row account guard were removed.

The gaps that exist are scenario coverage, not weak assertions: none of the account-switch-during-save-callback race (finding No. 2), the sign-out-during-in-flight-push race (finding No. 4), or the `reconcileChapter` transaction race (finding No. 3) are exercised by the current suite, for the reasons described above.

---

### Things that look right

- The base-version conflict model (comparing against a remembered base rather than "newer timestamp wins") is the correct call given cross-device clock skew, and the content-fingerprint fallback for records with no timestamp is real, exercised code, not dead weight.
- Conflict resolution correctly updates the base pointer after `keep_mine` (so the same conflict doesn't re-fire) and correctly preserves both notes on `keep_both` (a fresh pending create plus the original row replaced with the server's synced version, not deleted).
- True create-vs-create id collisions can't organically arise (ids are randomly generated per device), and the hypothetical collision case correctly falls through to a content-fingerprint mismatch and surfaces a conflict rather than crashing.
- Deleting a note that was created offline and never synced correctly costs zero server requests (detected via `baseUpdatedAtMs === null && !baseFingerprint`) rather than queuing a pointless tombstone.
- Repeated offline edits to the same note correctly coalesce to the latest content while preserving the original base version for conflict comparison — no stale-content-wins bug there.
- `adoptLocalRows`/`clearSynced` are genuinely single atomic transactions, and the "IndexedDB skips a null key" mechanism the whole pending-queue design leans on is accurate.
- `AnnotationConflictModal` always resolves the conflict actually on screen (read fresh each render, no stale-index closure), and its tests assert resolution by exact conflict id rather than "was it called."
- `OfflineTranslationStore.ts`'s changes are a byte-for-byte behavior-neutral extraction into the new shared `indexedDbUtils.ts` — no change to the already-shipped offline translation cache.

---

KallynGowdy:
Just tested this at a basic level. You can take down notes while offline and they will automatically sync when you come back online. When you change the note from two places, it shows a basic conflicts screen that lets you choose between one or the other.

Blitzy:
> [!NOTE]

## Code review — Record annotations offline and sync them when online

This PR just moved from draft to ready for review, so I re-reviewed against the current head (`3dd93ab0`, a no-op merge from `develop` on top of `d1f3fe43`). Two earlier automated reviews on this PR (Aug 14) found 5 real races where a concurrent operation could silently drop or misattribute an annotation edit. Since then, two follow-up commits (`9689d89`, `d1f3fe43`) claim to fix all of them. I verified each fix by reading the current code and hand-tracing the interleaving, rather than taking the commit messages at their word.

### Previously-flagged races: verification

All five are genuinely **fixed**:

1. **Push race** (a slow save being clobbered by an edit made while the first push was still in flight) — fixed. `writeToServer`/`eraseOnServer` no longer write back the stale pre-push snapshot; both go through `recordPushed`, which re-reads the row and only marks it synced if it's unchanged since the push started, otherwise rebasing the newer edit onto what was just pushed. The regression test genuinely mutates the store from inside the mocked `recordData` call before it resolves, so it reproduces the real interleaving.
2. **Session-expiry data loss** (a dead session permanently discarding a queued edit) — fixed. Fatal session codes (`session_expired`, `invalid_key`, `user_is_banned`) are now their own `PushOutcome` (`"session_ended"`), checked before the retryable/permanent split, and `runPass` leaves the row untouched and stops instead of calling `recordFailure`. Tests cover all three fatal codes plus "later queued rows stay queued."
3. **`reconcileChapter` transaction race** (a concurrent edit landing between a stale read and a later write getting overwritten by old server data) — fixed for the path that's tested, with one gap. The real IndexedDB implementation now reads inside the same `readwrite` transaction it writes in, closing the gap. The in-memory test double was fixed the same way. **However**, `createIndexedDbAnnotationStore()` is only ever exercised in jsdom to confirm it returns `null` (no `indexedDB` global there) — there's no `fake-indexeddb`-backed test that actually runs the fixed transaction code path. The fix's correctness for the real backend rests on an unverified (if generally true) assumption about IndexedDB transaction lifetime across microtask-only awaits.
4. **Sign-out vs. in-flight push** (a push resolving after sign-out re-writing a readable row for an account that already left) — fixed. `recordPushed` checks the currently signed-in account against the row's owner before writing a synced copy back; if they differ, an unchanged row is deleted outright rather than written back readable. Two tests simulate signing out from inside the mocked `recordData` resolution.
5. **Account misattribution on save/delete** (a save/delete landing in the wrong account's cache after an account switch mid-flight) — fixed. `saveEditingAnnotation`/`deleteAnnotationAndRefresh` now capture the owner before awaiting the durable write and thread it explicitly into `upsertIntoCache`/`removeFromCache`, matching the pattern already used on the sync-driven path. Tests flip `login.userId` between kicking off and awaiting the save/delete and assert the cache lands under the original account.

### New issues found in this pass

**🟠 Medium — No. 1: "Keep both" leaves the newly-created copy invisible until it happens to sync**

**Where:** `AnnotationSyncManager.tsx`, `applyResolution`, `keep_both` branch.

When a conflict is resolved with "keep both," the branch marks the original row synced with the server's content (correctly calling `onSynced?.(conflict.server, owner)` so that row updates in the UI) and queues a brand-new row with `pendingOp: "upsert"` for the user's version — but nothing notifies the reactive cache about that new row. Since the chapter's cache entry is already `settled = true`, the new copy sits in `OfflineAnnotationStore` without appearing in the Discover pane until a later sync pass actually pushes it and fires `onSynced` for it.

**Reproduction:** While offline, edit a note that has a pending server-side conflict so "keep both" is offered; choose it; stay offline. The original note updates to the server's content as expected, but the new copy holding your own edit doesn't show up anywhere in the UI — it only appears once you reconnect and the queued push completes.

This isn't a regression introduced by the latest race-fix commit (the branch predates it), but it's a real, currently-unaddressed gap in the same conflict-resolution code this PR adds.

**Suggested fix:** have the `keep_both` branch also notify the cache about the newly queued row (e.g. an `onCreated`/`onSynced`-style callback fired with the still-pending row), the same way a locally-created offline note becomes visible immediately rather than waiting for its first successful push.

**🟢 Low — No. 2: The real IndexedDB `reconcileChapter` fix has no test that actually runs against IndexedDB**

Covered above under race No. 3 — flagging separately since it's a coverage gap on its own, independent of whether the fix is correct. Worth adding a `fake-indexeddb`-backed (or equivalent) test that drives the real `createIndexedDbAnnotationStore()` through the same concurrent-edit-during-reconcile scenario the in-memory double already covers, so the production code path — not just its stand-in — is what's actually verified.

**🟢 Low — No. 3: Pending-sync count is account-wide, not per-chapter (confirmed still unaddressed)**

`AnnotationSyncManager.tsx`'s `pendingCount` comes from `store.listPending(owner)` with no chapter filter, but `DiscoverPane.tsx` renders it directly under one chapter's notes ("N changes waiting to sync"). A user with one unsynced note in Exodus who opens Genesis 1 still sees "1 change waiting to sync" there. This was flagged as low-severity/follow-up in the earlier review and remains unfixed — reasonable to continue treating as a follow-up rather than a blocker, just confirming it wasn't dropped.

### Things that check out

- `AnnotationConflictModal.tsx` re-reads conflicts fresh each render with no stale-closure risk.
- The delete-confirmation flow in `DiscoverPane.tsx` only calls the already-fixed `deleteAnnotationAndRefresh`.
- `indexedDbUtils.ts` is a clean, behavior-neutral extraction (confirmed via the earlier reviews and re-checked here).
- Test mocking throughout all three test files stays at the `OsManager` boundary (`os.getData`/`recordData`/`eraseData`/`listDataByMarker`) and exercises the real in-memory store and real sibling managers — matching this repo's testing conventions. No fixed-duration sleeps; the one `setTimeout(resolve, 0)` found is a legitimate microtask flush, not a duration guess.
- The "keep both" test explicitly avoids the `??`-masking assertion pattern that let an earlier version of this same bug slip through.

### Summary

| Severity | Item | Status |
|---|---|---|
| — | Races No. 1–5 from earlier reviews | ✅ Fixed, verified by tracing the actual interleaving |
| 🟠 Medium | No. 1 — "keep both" copy invisible until synced | New, unaddressed |
| 🟢 Low | No. 2 — real IndexedDB reconcile path untested | New, coverage gap |
| 🟢 Low | No. 3 — pending count is account-wide not per-chapter | Previously flagged, still unaddressed |

The core offline-sync and conflict-resolution machinery is now solid against every race identified so far, and the fixes are backed by regression tests that genuinely reproduce the interleavings they claim to fix rather than just re-testing sequential behavior. The one new Medium (No. 1) is worth a small follow-up fix since it undercuts the same "an edit, once saved, won't quietly disappear from view" promise the PR is built around — though in this case the data is safe on disk and does eventually appear, just not until the next successful sync. The two Lows are reasonable to take as follow-ups.

---

Issue #1381, opened 2026-07-09
Scripture map: default reading history and user presence to on


Blitzy:
## What is happening

On the Scripture map, the **reading history** and **user presence** displays default to off. Users have to turn them on manually to see them.

_Default state: reading history and user presence are off._
<img width="933" height="669" alt="Image" src="https://github.com/user-attachments/assets/5dc2d49a-48d7-40fe-8412-57780ac35531" />

## What is expected

Both should default to **on**:

- Reading history shown by default.
- User presence shown by default.

_Desired default state: reading history and user presence are on._
<img width="920" height="670" alt="Image" src="https://github.com/user-attachments/assets/7d4af88d-1e45-498e-bf2f-3c9d569a71a1" />

PR #1638, opened 2026-08-12
Task-1381

Bored-Wizard:
the user history and user presense ore now turned on by default, while testing I discovered that the user presense were not syncing properly with guest reading state so fixed it too

task #1381


Blitzy:
> [!NOTE]

## Code review — PR No. 1638 (Task-1381)

### What this PR does
[Issue #1381](https://github.com/HelloAOLab/seed-bible/issues/1381) asks for one thing: make the Scripture-map **reading history** and **user presence** default to *on* instead of *off*. That part is two one-line flips and is done correctly:

- `useScriptureMapProvider.tsx` — `isUserPresenceEnabled` default `false → true`
- `scripture-map/di/bootstrap.tsx` — `initialIsReadingHistoryEnabled: false → true`

The rest of the PR (~370 of the 382 added lines) is a **separate, larger fix**: presence used to render *every* peer at the local reader's own position, because it read the session-wide `readingState` for everyone. This PR adds a per-connection `reading_positions` shared map so each participant broadcasts and is shown at their *own* chapter. That's a real improvement and the design (debounced broadcast, connectivity-gated display, dispose cleanup) is sound and well-commented.

Worth flagging that the bulk of the review risk lives in this bundled presence change, not in the tiny change the linked issue actually requested.

**Verified locally:** `pnpm vitest run SessionsManager.test.ts SeedBiblePresenceProvider.test.ts` → **70 passed**. The parse helpers (`toPositiveIntOrNull`, `toStringOrNull`) correctly reject chapter `0` and non-strings, so `parseParticipantReadingPosition` drops malformed entries as the tests claim.

---

### Findings by severity

#### 🔴 Blocker (mergeability) — merge conflict with `develop`
GitHub reports this PR as `mergeable_state: dirty`. `develop` is 35 commits ahead of the branch, and `SessionsManager.tsx` conflicts.

**Reproduce:**
```bash
git fetch origin develop
git merge-tree --write-tree HEAD origin/develop
# CONFLICT (content): Merge conflict in
#   packages/seed-bible/seed-bible/managers/SessionsManager.tsx
```
**Fix:** merge (or rebase onto) latest `develop`, resolve the `SessionsManager.tsx` conflict, and re-run the suite. The other two "changed in both" files (`SessionsManager.test.ts`, `TabsManager.test.ts`) auto-merge cleanly.

---

#### 🟡 Low — presence fallback momentarily renders un-broadcast peers *on top of you*
`SeedBiblePresenceProvider.getRemotesPresence()` (lines 66–69) falls back to `session.readingState` (the **local** reader's position) for any peer who hasn't broadcast a position yet:
```ts
const position = positions.get(user.connectionId) ?? {
  bookId: session.readingState.bookId.value,
  chapterNumber: session.readingState.chapterNumber.value,
};
```
So during the window between a peer connecting and their first broadcast arriving, that peer is drawn exactly where *you* are — which reads as "they're reading with me" when they may not be.

Two mitigating facts keep this **low**: (1) it is strictly better than the pre-PR behavior, which showed *all* peers at your position permanently; (2) the broadcast effect fires ~`PUBLISH_DEBOUNCE_MS` (150 ms) after a peer joins regardless of navigation, so the window is transient.

**Suggested decision:** consider dropping the fallback entirely (skip the peer until they've broadcast) so a peer is never briefly misplaced onto your chapter. The comment already argues "beats showing nothing" — that's a judgment call worth making explicitly rather than by default, since showing a *wrong* position is arguably worse than showing none.

#### 🟡 Low — asymmetric defaults: one toggle is plumbed, the other is hardcoded
`isReadingHistoryEnabled` reads its default from an `initialIsReadingHistoryEnabled` config prop threaded through `bootstrap.tsx`, but `isUserPresenceEnabled` is hardcoded `useState<boolean>(true)` directly in the component (line 302–303). Both toggles come from the same issue and behave as a pair everywhere else, so the split is surprising and makes the presence default un-overridable and harder to test.

**Suggested fix:** add an `initialIsUserPresenceEnabled` prop (defaulting to `true` in `bootstrap.tsx`) and feed it into the `useState`, mirroring reading history.

#### 🟢 Low / informational — `reading_positions` grows unbounded on ungraceful disconnect
`dispose()` deletes the local entry, and display is correctly gated by `connectedClients`, so stale entries never *render*. But a peer that vanishes without disposing leaves its entry in the CRDT forever (the doc "never shrinks", per the comments). This mirrors the existing `user_profiles` map, so it's not a regression — noting it only so the growth characteristic is a conscious choice rather than an oversight.

---

### Test assessment
The task for this review specifically asks whether the tests reflect real-world behavior rather than passing unconditionally. They hold up well:

- **Burst coalescing** (`"publishes a burst of local navigations as one transaction…"`) asserts *exactly one* `reading_positions.set` for three rapid chapter changes. This fails on broken code both ways: remove the debounce → 3 writes; drop the broadcast → 0. Real assertion.
- **Not-an-allowed-navigator** (two tests) were tightened from "nothing is written" to "session nav is blocked **but** the local position is still broadcast" (`toHaveBeenCalledWith(os.connectionId, { bookId: "EXO", chapterNumber: 8 })`). This pins down the intended distinction — presence ≠ navigation permission — and would fail if the broadcast were gated on `userCanNavigate`.
- Edge cases are covered, not just the happy path: malformed entries ignored, disconnected participants dropped via connectivity (not the map), and dispose deletes the entry.
- The presence-provider tests verify the current user is skipped, two peers hold separate positions, and a peer with no position anywhere is skipped.

One minor caveat, pre-existing and not introduced here: `flushPublishDebounce()` is a fixed `setTimeout(…, 250)` sleep, which `CLAUDE.md` discourages in favor of polling a condition. It works because 250 ms > the 150 ms debounce, but it's a small flake risk under CI load and every new async assertion leans on it. Not blocking — worth a follow-up to convert to a `waitFor`-style poll if these ever go flaky.

---

### Bottom line
The requested default-toggle change is correct, and the bundled presence fix is a genuine, well-tested improvement. **The one thing that must happen before merge is resolving the `develop` conflict.** The low-severity items (fallback-on-top-of-you, hardcoded vs. plumbed default) are judgment calls the author can reasonably accept or address.

---

KallynGowdy:
@Bored-Wizard I've run into a really weird issue when trying to test this: https://youtu.be/8uQZXIE9OpI

There seems to be a rotating square behind the scripture map elements for some reason.

Bored-Wizard:
> @Bored-Wizard I've run into a really weird issue when trying to test this: https://youtu.be/8uQZXIE9OpI
> 
> There seems to be a rotating square behind the scripture map elements for some reason.

I thought it was a feature 😆

Bored-Wizard:
> @Bored-Wizard I've run into a really weird issue when trying to test this: https://youtu.be/8uQZXIE9OpI
> 
> There seems to be a rotating square behind the scripture map elements for some reason.

@KallynGowdy I have been trying but the same bug is not  getting produced for me

KallynGowdy:
Strange, seems fixed now. Maybe it was my graphics card

Blitzy:
> Strange, seems fixed now. Maybe it was my graphics card

Its not just you, i have seen this before in the past as well. I havent been able to reproduce it though 😆

PR #1639, opened 2026-08-13
Added New Line For Laptop and Mobile View

### LINKED ISSUE #1623: Chat input should be multi-line and support inserting newlines
Chat input currently is forcing text to be a single line. It isn't possible to make newlines to create paragraphs or other structured text content. 

The chat input field will also need to expand in, I suggest we start with up to 5 lines and then scroll after that.

## Desktop

Newline: Shift+Enter
Submit: Enter / Click submit

## Mobile

Newline: Enter
Submit: Tap submit

Kabi4:
Fixes: #1623

### DIFF
--- a/packages/seed-bible/seed-bible/components/ChatView/ChatView.css
+++ b/packages/seed-bible/seed-bible/components/ChatView/ChatView.css
@@ -257,7 +257,7 @@ img.sb-chat-view-message-avatar {
 .sb-chat-view-compose {
   display: flex;
   gap: 0.5rem;
-  align-items: center;
+  align-items: flex-end;
 }
 
 .sb-chat-view-compose-shell {
@@ -409,20 +409,24 @@ img.sb-chat-view-message-avatar {
   background: color-mix(in srgb, var(--sb-background, #fff), transparent 5%);
   color: var(--sb-font-color);
   padding: 0.625rem 0.75rem;
+  font: inherit;
   font-size: 0.875rem;
   line-height: 1.2;
+  resize: none;
+  overflow-y: auto;
+  /* Cap at 5 lines: 5 × font-size × line-height + vertical padding. */
+  max-height: calc(0.875rem * 1.2 * 5 + 1.25rem);
 }
 
 /* Mobile type-hint: faux blinking caret while the field is unfocused.
-   Inputs can't host ::before/::after, so this is a sibling pseudo-caret. */
+   Textareas can't host ::before/::after for this, so this is a sibling. */
 .sb-chat-view-input-hint-caret {
   display: none;
   position: absolute;
   left: 0.75rem;
-  top: 50%;
+  top: 0.625rem;
   width: 1.5px;
   height: 1.05em;
-  transform: translateY(-50%);
   background: var(--sb-font-color);
   pointer-events: none;
   z-index: 1;
--- a/packages/seed-bible/seed-bible/components/ChatView/ChatView.css
+++ b/packages/seed-bible/seed-bible/components/ChatView/ChatView.css
@@ -414,8 +414,14 @@ img.sb-chat-view-message-avatar {
   line-height: 1.2;
   resize: none;
   overflow-y: auto;
+  scrollbar-width: none;
+  -ms-overflow-style: none;
   /* Cap at 5 lines: 5 × font-size × line-height + vertical padding. */
   max-height: calc(0.875rem * 1.2 * 5 + 1.25rem);
+}
+
+.sb-chat-view-input::-webkit-scrollbar {
+  display: none;
 }
 
 /* Mobile type-hint: faux blinking caret while the field is unfocused.
--- a/packages/seed-bible/seed-bible/components/ChatView/ChatView.tsx
+++ b/packages/seed-bible/seed-bible/components/ChatView/ChatView.tsx
@@ -51,6 +51,35 @@ interface JoinEvent {
   timeMs: number;
 }
 
+/** Compose field grows with content, then scrolls past this many lines. */
+const COMPOSE_INPUT_MAX_LINES = 5;
+
+/**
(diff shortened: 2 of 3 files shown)


Blitzy:
> [!NOTE]

## Code Review — Multi-line chat compose (PR #1639)

Thanks @Kabi4! This cleanly converts the chat compose `<input>` into an auto-growing `<textarea>` and wires up the Enter/Shift+Enter behavior from issue #1623 (Enter submits on desktop, Shift+Enter inserts a newline, Enter inserts a newline on mobile, capped at 5 lines then scroll). The mention-picker key handling was folded into a single `handleComposeKeyDown` without losing coverage, and the test suite was faithfully migrated to the textarea (35 tests pass locally).

Two functional bugs are worth fixing before merge, plus a few low-severity notes.

---

### Medium — Compose box stays expanded after sending a multi-line message

**Where:** `ChatView.tsx`, `handleSubmit` (around lines 725–731)

After a send, the draft is cleared and then the field is re-measured **synchronously**:

```ts
draft.value = "";
chat.setTypingStatus(false);
const input = inputRef.current;
if (input) {
  input.focus();
  resizeComposeInput(input); // runs before the DOM value is actually cleared
}
```

Clearing `draft.value` schedules a Preact re-render, but the textarea's DOM `value` is not updated until the next microtask. So `resizeComposeInput` measures `scrollHeight` against the **still-populated** textarea and sets the inline height to the old (tall) size. When the re-render then clears the text, the inline height is never recomputed, so the empty field stays tall.

This is corroborated inside the PR itself: `insertMentionText` (line 683) deliberately wraps its `resizeComposeInput` in `window.queueMicrotask(...)` for exactly this reason, and the new desktop-submit test only sees `input.value === ""` **after** an explicit `await Promise.resolve()` — direct evidence the clear is asynchronous.

**Reproduction (desktop):**
1. Open a chat.
2. Type a message spanning 3+ lines (Shift+Enter to add newlines).
3. Press Enter to send.
4. The message sends and the text clears, but the compose box remains ~3 lines tall. It only collapses back to one line once you type the next character.

**Suggested fix** — defer the resize like `insertMentionText` already does:

```ts
draft.value = "";
chat.setTypingStatus(false);
window.queueMicrotask(() => {
  const input = inputRef.current;
  if (input) {
    input.focus();
    resizeComposeInput(input);
  }
});
```

---

### Medium — Enter during IME composition submits prematurely

**Where:** `ChatView.tsx`, `handleComposeKeyDown` (line 785)

```ts
if (event.key === "Enter" && !event.shiftKey && !state.app.isMobile.value) {
  event.preventDefault();
  void handleSubmit(event);
}
```

There's no guard for an in-progress IME (Input Method Editor) composition. For CJK and other composed scripts, the user presses Enter to **confirm the candidate word**, not to send. Because this handler fires on that Enter, the message is submitted mid-composition (often with a half-finished or first-candidate string). Given the app ships 24 locales, this will hit real users.

**Reproduction (desktop):**
1. Switch the OS keyboard to a CJK IME (e.g. Japanese).
2. In a chat, type romaji so the IME shows candidates.
3. Press Enter to confirm the candidate.
4. Instead of only confirming the word, the message is sent.

**Suggested fix** — skip submit while composing:

```ts
if (
  event.key === "Enter" &&
  !event.shiftKey &&
  !event.isComposing &&
  !state.app.isMobile.value
) {
  event.preventDefault();
  void handleSubmit(event);
}
```

(`event.isComposing` covers modern browsers; some codebases also guard `event.keyCode === 229` for older engines.) The same guard is worth considering for the mention-picker `Enter` branch above it, since that path can also fire on candidate confirmation.

---

### Low — Test name claims a newline it doesn't verify

**Where:** `ChatView.test.tsx`, `"submits on Enter and inserts a newline on Shift+Enter on desktop"`

The test asserts that Shift+Enter does **not** submit (`expect(sendMessage).not.toHaveBeenCalled()`) — which is meaningful and would fail on a broken handler. But it never asserts that a newline is actually inserted; jsdom doesn't perform the browser's native textarea newline insertion on a synthetic `keydown`, so that half of the behavior is untestable here. The assertions that exist are real (not always-green), but the title over-promises. Consider renaming to reflect what's actually checked (e.g. "…does not submit on Shift+Enter…") or add a comment noting the jsdom limitation so a future reader doesn't assume newline insertion is covered.

### Low — Two sources of truth for the 5-line cap

`ChatView.css` hardcodes `max-height: calc(0.875rem * 1.2 * 5 + 1.25rem)` while `resizeComposeInput` computes the cap from live computed styles. They agree today, but a future change to the input's `font-size`/`line-height`/`padding` in CSS would need a matching edit to the comment/constant reasoning in the TS. Additionally, the JS `maxHeight` adds `borderTop + borderBottom` (~2px) while the CSS `calc()` omits the 1px top/bottom border on `.sb-chat-view-input`, so the effective caps differ by ~2px — cosmetic, but the CSS `max-height` will win and clamp slightly under a true 5 lines. Minor; noting for consistency.

### Nit — Mobile newline while a mention query is open

On mobile the mention-picker `Enter` branch runs before the mobile check, so pressing Enter while an `@query` is active selects a mention rather than inserting a newline. This is arguably the right behavior (you're mid-mention), just flagging it as an intentional interaction to confirm against the issue's "Mobile: Enter = newline" line.

---

**Summary:** The feature matches issue #1623 and the tests are genuine behavioral tests, not always-green ones. The two medium items (stuck height after multi-line send; IME Enter) are the ones I'd resolve before merge; the rest are polish.

---

KallynGowdy:
@Kabi4 I found a thing.

There's a scrollbar on the text area on desktop and mobile that should probably be hidden:
image

Kabi4:
Yes you are right @KallynGowdy For me it was somehow very thin and looking good i fixed it for mobile though

image

PR #1640, opened 2026-08-13
Close the chat if in Mobile State + Added Test cases

### LINKED ISSUE #1622: Tapping a chat verse reference on mobile should close the chat.
## What is happening

Currently on mobile, if you tape a verse reference in a chat it will navigate behind the chat window but you then have to manually close the chat window to see it.

## What I expected 

Tapping a verse reference on mobile chat should close the chat window. This is because  on mobile the chat window takes up the whole screen so it needs to close in order to see the verse that has been navigated to my clicking the verse reference link.

Kabi4:
Fixes: #1622

### DIFF
--- a/packages/seed-bible/seed-bible/components/ChatView/ChatView.tsx
+++ b/packages/seed-bible/seed-bible/components/ChatView/ChatView.tsx
@@ -1048,7 +1048,12 @@ function ChatMessage({
       <p className="sb-chat-view-message-body">
         <MessageBody
           message={message}
-          onVerseReferenceClick={(ref) => state.app.openVerseReference(ref)}
+          onVerseReferenceClick={(ref) => {
+            if (state.app.isMobile.value) {
+              state.sidebar.closeChatPanel();
+            }
+            void state.app.openVerseReference(ref);
+          }}
         />
       </p>
       {showTimestamp && (
--- a/test/unit/seed-bible/components/ChatView.test.tsx
+++ b/test/unit/seed-bible/components/ChatView.test.tsx
@@ -577,10 +577,14 @@ describe("ChatView", () => {
       getMessageAuthors: vi.fn().mockReturnValue([]),
     });
     const openVerseReference = vi.fn().mockResolvedValue(undefined);
+    const closeChatPanel = vi.fn();
     const state = {
       app: {
         openVerseReference,
         isMobile: signal(false),
+      },
+      sidebar: {
+        closeChatPanel,
       },
     } as unknown as SeedBibleState;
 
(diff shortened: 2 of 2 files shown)


Blitzy:
> [!NOTE]

## Code Review — PR No. 1640 (fixes No. 1622)

### Summary

This PR makes tapping a verse reference inside a chat message close the chat panel on mobile, so the navigated-to verse is actually visible instead of hidden behind the full-screen chat. **Before**, tapping a reference navigated the reader underneath the chat, and you then had to manually close the chat to see the verse. **Now**, on mobile the chat panel closes first and then the reference opens.

The implementation is small and targeted: `ChatMessage`'s `onVerseReferenceClick` handler now checks `state.app.isMobile.value` and calls `state.sidebar.closeChatPanel()` before `openVerseReference`. Desktop behavior is unchanged, which matches the issue exactly (the issue is mobile-specific because chat is full-screen there).

**Verdict: looks good to merge.** I found no correctness bugs. The one thing I'd call out is a pre-existing, out-of-scope observation (No. 1 below).

### Test assessment

The reviewer prompt specifically asks whether the tests exercise real behavior rather than always passing — so I checked this directly:

- Both tests render the real `ChatView`, query the real `.sb-verse-reference-link` element, and dispatch a real `click` event, so they drive the actual production click handler (only `openVerseReference`/`closeChatPanel` are mocked, which is the correct boundary — the managers, not internals).
- I reverted the fix locally and re-ran the suite: the new test **"closes the chat panel on mobile when a verse reference link is clicked"** failed (`expected "vi.fn()" to be called 1 times, but got 0 times`), then passed again once the fix was restored. So it is a genuine regression test that would catch this bug regressing — not one that always succeeds.
- The existing non-mobile test was correctly extended with `expect(closeChatPanel).not.toHaveBeenCalled()`, which locks in that desktop does **not** close the panel.
- Full suite result on the PR head: `34 passed (34)`.

Nice touch: the two tests differ only by the `isMobile` signal value, which cleanly isolates the branch being tested.

### Findings by severity

**High / Blocking:** None.

**Medium:** None.

**Low (pre-existing, out of scope — noted only because the handler was just touched):**

1. **A verse-reference tap also toggles the message timestamp.** Each message is wrapped in an `<article onClick={() => setShowTimestamp(!showTimestamp)}>` (ChatView.tsx around line 1046). `MessageBody` calls `e.preventDefault()` on the link but not `e.stopPropagation()` (around line 191–194), so a reference tap bubbles up to the article and flips the timestamp too. On mobile this is now harmless since the panel closes, but on desktop tapping a reference still flickers the timestamp. This predates this PR and isn't something I'd block on — just flagging it since it lives right next to the changed code. If you wanted to tidy it up, adding `e.stopPropagation()` alongside the existing `e.preventDefault()` in the `verse_reference` branch of `MessageBody` would scope the tap to navigation only.

_Reproduction for No. 1 (desktop, pre-existing):_ open a chat containing a verse reference → click the reference → observe the message's relative timestamp toggle on/off in addition to the reader navigating.

### CI

The latest `build (24.x)` was still in progress at review time; the prior run's `build (24.x)` and `deploy (24.x)` both succeeded, and nothing is failing. `mergeable_state` is `blocked` only because the requested review (Blitzy) hasn't been submitted yet.

---

PR #1641, opened 2026-08-13
Login Modal opens on forced logout

### LINKED ISSUE #1621: Show login screen when being forced to sign out
We should show the login screen when a forced logout occurs as a convenience for signing back in quickly. Being forced to sign out should be safe indication that the user was previously not anonymous and so showing the login screen can be assumed to be a convenience and not a hindrance.

Kabi4:
Fixes: #1621 

I have done asked thing but 

What I’d want eyes on in review:

Force-open bypasses login() — modal is driven by isLoginOpen, not the login() promise. That’s fine if dismiss + successful re-auth always clear the flag (they should now). Reviewers should verify those two close paths, since that’s the easy place to leave a stuck overlay.

Don’t expand this to voluntary logout — that would punish intentional sign-out and surprise anonymous/guest flows. The invariant “only forceSignOut opens it” is the important line to protect.

### DIFF
--- a/packages/seed-bible/seed-bible/components/ToastHost/ToastHost.css
+++ b/packages/seed-bible/seed-bible/components/ToastHost/ToastHost.css
@@ -122,7 +122,7 @@
   left: 50%;
   bottom: 6rem;
   transform: translateX(-50%);
-  z-index: 1000;
+  z-index: 10002; /* was 1000 — above every overlay, incl. the raised tutorial (10001) */
   display: flex;
   justify-content: center;
   max-width: min(90vw, 26.25rem);
--- a/packages/seed-bible/seed-bible/managers/LoginManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/LoginManager.tsx
@@ -348,6 +348,10 @@ export function createLoginManager({
    * parse — so the call could only fail; it costs a round trip on a path often taken
    * while connectivity is poor; and for a banned account it can never succeed.
    *
+   * Also opens the login screen. A forced sign-out only happens when there was a real
+   * session, so the user was not anonymous — offering sign-in again is a convenience,
+   * not a hindrance. (A deliberate `logout()` does not go through here.)
+   *
    * `sessionEnded` is left set rather than reset, which is what lets a sign-out during
    * construction still reach the toast: `SeedBibleStateManager` wires that effect much
    * later, and an effect reads its dependencies eagerly on its first run. Don't
@@ -370,6 +374,10 @@ export function createLoginManager({
       reason,
       id: ++sessionEndedCount,
     };
+    // Skip during SSR — there is no interactive login surface to show.
+    if (!import.meta.env.SSR) {
+      isLoginOpen.value = true;
+    }
   };
 
   /** Signs the user out because the server reported the session key dead. */
@@ -600,6 +608,7 @@ export function createLoginManager({
   }
 
   async function cancelLogin() {
+    isLoginOpen.value = false;
     if (loginPromise && rejectLoginPromise) {
       rejectLoginPromise(new Error("Login cancelled"));
       loginPromise = null;
@@ -668,6 +677,9 @@ export function createLoginManager({
         id: userId.value,
         email: result.email,
       };
+      // Close even when login was opened by a forced sign-out (no `login()`
+      // promise) — that path never reaches `loginCore`'s `finally`.
+      isLoginOpen.value = false;
       if (resolveLoginPromise) {
         resolveLoginPromise(userInfo.value);
         resolveLoginPromise = null;
--- a/test/unit/seed-bible/managers/LoginManager.test.ts
+++ b/test/unit/seed-bible/managers/LoginManager.test.ts
@@ -512,6 +512,44 @@ describe("createLoginManager", () => {
       expect(localStorage.getItem("connectionKey")).toBe(null);
     });
 
+    it("opens the login screen so the user can sign back in", async () => {
+      // Forced sign-out only happens when there was a real session, so offering
+      // the login UI is a convenience rather than surprising an anonymous user.
+      const manager = await signOutViaGetUserInfo("session_expired");
+
+      expect(manager.isLoginOpen.value).toBe(true);
+    });
+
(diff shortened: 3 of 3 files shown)


Blitzy:
> [!NOTE]

## Code review — Login modal opens on forced logout

**Short version:** The change is small, well-guarded, and correctly wired. It does what issue #1621 asks: after a forced sign-out, the login modal opens so the user can sign back in quickly. I found **no correctness bugs**. The two close paths the PR description asked reviewers to verify (dismiss + successful re-auth) both work and are covered by tests. I have one design/UX question about the suspended-account case, and a couple of minor test-coverage notes.

### What the change does

- `forceSignOut` now sets `isLoginOpen = true` (skipped during SSR) after clearing the session, so a forced sign-out pops the login modal.
- `cancelLogin` and the success branch of `loadUserInfo` now set `isLoginOpen = false`. These are the two close paths that do **not** flow through `loginCore`'s `finally`, so without them a force-opened modal would have no way to close — the exact "stuck overlay" the author flagged.

### Verification I ran

- `pnpm vitest run …/LoginManager.test.ts` → all 89 tests pass.
- I reverted the three source additions locally and re-ran: **4 of the new assertions turn red** (`opens the login screen…`, `closes…when the user dismisses it`, `closes…when the user signs back in`, and the corrupt-key `…isLoginOpen === true`). So these are real regression tests — they fail on the pre-fix code and pass on the fix, rather than passing either way.
- I traced every path that writes `isLoginOpen` for race conditions (background `loadUserInfo` on init, `refreshSession`, a late-arriving dead-session event after dismissal). The existing session-key guards (`sessionKey.peek()` re-checks after each `await`, and `forceSignOut`'s early return when `!sessionKey.peek()`) already protect these: once dismissed, the modal will not silently reopen, and `loadUserInfo` only runs when a session key exists, which never overlaps with a force-opened modal.
- Confirmed `LoginModal` is driven entirely by `isLoginOpen` and completes the whole email → code → submit flow without needing a `login()` promise, so force-opening produces a fully functional modal.

---

## Findings by severity

### Low — `account_suspended` (banned account) also force-opens the login modal

A banned account produces `forceLogout("user_is_banned")` → `forceSignOut("account_suspended")`, which now runs the same `isLoginOpen = true`. So a suspended user gets, at the same moment:

- a toast (from `SeedBibleStateManager`): *"Your account has been suspended."*, and
- the login modal titled *"Login to your account"* inviting them to sign in again.

Issue #1621 framed the feature specifically as *"a forced logout is a safe indication the user was previously not anonymous, so showing login is a convenience."* That reasoning fits `signed_out`, but not suspension — re-logging-in a suspended account is futile and the two surfaces send contradictory messages. The PR description also states the invariant to protect is *"only forceSignOut opens it"* — but within `forceSignOut`, both reasons open it, and the suspended one is the questionable case.

This is a UX/design call, not a correctness bug — the user can dismiss the modal — so I'm flagging it Low and deferring to the maintainers on intent.

**Reproduction**
1. Sign in normally so there is a real session.
2. Cause the next `getUserInfo`/refresh to return `user_is_banned` (ban the account server-side, or in a test set `getUserInfoMock` to `{ success: false, errorCode: "user_is_banned" }`).
3. Observe: the "account suspended" toast **and** the login modal both appear.

**Suggested fix (if unintended)** — gate the open on the reason:

```ts
// Skip during SSR — there is no interactive login surface to show.
// Only nudge re-login for an ordinary forced sign-out; a suspended
// account can't usefully sign back in.
if (!import.meta.env.SSR && reason === "signed_out") {
  isLoginOpen.value = true;
}
```

and add a test asserting `account_suspended` leaves `isLoginOpen === false`.

### Low — test coverage gaps

- **Suspended-account modal behavior is unpinned.** The `reports %s to the UI as` cases assert only `reason`; the new `opens the login screen…` test uses `session_expired` only. Whichever way you decide the point above, add an assertion so a future change can't flip the suspended-account behavior unnoticed.
- **The SSR guard has no test.** `forceSignOut`'s `if (!import.meta.env.SSR)` branch is never exercised (in Vitest, `import.meta.env.SSR` is always false). This is genuinely awkward to drive from jsdom, so I'd accept the gap — noting it only so it's a conscious choice rather than an oversight.

---

## Notes (no action needed)

- **Toast + modal now co-appear on an ordinary `signed_out`.** The toast text *"You've been signed out. Please sign in again."* pairs naturally with the modal, so this reads as complementary rather than conflicting. Worth one quick visual check on the deployed branch that the toast isn't hidden behind the modal overlay, but I don't consider it a defect.
- **The `isLoginOpen = false` in `loadUserInfo` is redundant on the normal `login()` path** (where `loginCore`'s `finally` also clears it). That's intentional and documented in the code comment — it's specifically for the forced-sign-out path that never reaches that `finally`. No change needed.
- **Test quality is good overall.** The three new dynamic tests exercise real user paths (dismiss, re-auth) through the public API rather than internals, and I confirmed they fail on the un-fixed code. The added `isLoginOpen === false` assertion on the deliberate-logout test is an invariant guard (it passes both before and after the fix) — that's appropriate here: it locks in "a voluntary logout must never open the modal," which is exactly the line the PR author asked to protect.

---

KallynGowdy:
@Kabi4  The toast message does appear behind the modal overlay. Ideally, in this case it would appear above the overlay so it is easier to read. Other than that it looks good to me

Blitzy:
Changes look good ✅

I tested the forced logout in the browser by corrupting the stored session key so the server rejects it. If anyone else wants to try, just paste this JS into your DevTools console:

```js
const p = localStorage.sessionKey.split("."); // vSK1 . userId . sessionId . secret . expiry
const s = p[3]; // p[3] is the session secret
const i = Math.floor(s.length / 2); // flip one char mid-string
p[3] = s.slice(0, i) + (s[i] === "A" ? "B" : "A") + s.slice(i + 1);
localStorage.sessionKey = p.join(".");
location.reload();
```

Seconding what @KallynGowdy mentioned. The toast is appearing behind the login modal's dim overlay. 

**Suggested fix — one line in `ToastHost.css`:**

```css
.sb-toast-host {
  z-index: 10002; /* was 1000 — above every overlay, incl. the raised tutorial (10001) */
}
```

PR #1642, opened 2026-08-14
Added a flick hanlder in case of andriod which prirotize the stopping…

### LINKED ISSUE #1505: Mobile book selector does not reliabley open with each tap
## What is happening 

The mobile book selector often does not open when tapped. Sometimes it takes 2 or 3 taps. This seems to occur most after scrolling down for a chapter a little bit and then stopping and scrolling back up the page a little. Not sure if this is the only case but seems to be the most reliable way to reproduce it.

https://youtube.com/shorts/CKB-dPcBY7c?si=D2rdMyIF3dDm867n

## What I expected

The book selector should open on a single tap.

Kabi4:
… of current scroll instead of tapping on button so pririty listens to click now

Fixes: #1505 

The new helper listens for a full press-and-release on that button (pointerdown → pointerup). If that press stays roughly in place (not a swipe), it opens the selector even when Chromium never sends click.


Blitzy:
> [!NOTE]

## Code review: PR #1642 — fling‑safe tap for the mobile book selector

**What it does.** Fixes #1505, where tapping the mobile book/chapter selector while the page is still coasting from a fast flick does nothing. On Android/Chromium the tap that *stops* a momentum scroll is treated as a scroll‑cancel: `pointerdown`/`pointerup` fire but no `click` follows, so a control wired to `onClick` alone silently ignores that first tap. The issue thread confirms this is Chromium‑specific and that WebKit/Safari is unaffected — matching the new `flingSafeTap.ts` doc comment exactly.

The new `flingSafeTapHandlers` helper activates from `pointerup` for touch/pen (which fire even when the gesture is cancelled), de‑duplicates the `click` that *does* arrive on a normal tap, and leaves mouse/keyboard on the `click` path. It's wired into the book‑title `<h2>` in `BibleReader.tsx` and the floating nav label `<button>` in `BibleReaderToolbar.tsx`.

**Overall:** solid, well‑documented, and correctly targeted at the real bug. Tests pass locally (33/33 across both touched files) and type‑check is clean. I found **no correctness‑blocking issues** — only low‑severity edge cases and one test‑quality nit. The findings below are worth a look but I don't consider any of them merge‑blockers.

Verified locally: `pnpm vitest run test/unit/seed-bible/app/flingSafeTap.test.ts test/unit/seed-bible/components/BibleReaderToolbar.test.tsx` → 33 passed; `pnpm check:ts` → 0 visible errors.

---

### Low severity

**No. 1 — Module‑global state has a single slot, so concurrent presses on two fling‑safe controls drop one.**
`flingSafeTap.ts:35-36` keeps one `pendingTap` and one `handledTap` for the whole app. If two controls that both use these handlers are pressed at once (two fingers, or a stray second touch), the second `pointerdown` overwrites `pendingTap`, and the first control's `pointerup` then fails the `press.pointerId !== event.pointerId` / `press.element !== event.currentTarget` guard and is discarded (`flingSafeTap.ts:87-93`). The result is a *missed* activation, never a wrong one, and with only two such controls today (the title and the nav label, which sit far apart) this is unlikely in practice. Flagging it because the module comment frames the global scope as deliberate — the tradeoff it doesn't mention is that the design assumes at most one in‑flight tap across *all* consumers.
*Suggested fix (optional):* key the pending/handled state by `pointerId` (a small `Map`) rather than a single slot, so overlapping presses don't clobber each other.

**No. 2 — The `click` fallback can double‑activate (or swallow) in two narrow windows. Harmless today, latent for reuse.**
`onClick` runs `onTap()` for any click that isn't matched to a `handledTap` within `CLICK_AFTER_TAP_MS = 700` (`flingSafeTap.ts:112-123`). Two edge cases fall out of that:
- A `click` that arrives **more than 700 ms** after the `pointerup` that already fired `onTap` is treated as fresh → `onTap` runs twice.
- A **quick second tap that drifts 12–~25 px** within 700 ms: `pointerup` treats it as a drag and returns without refreshing `handledTap` (`flingSafeTap.ts:96-102`), so the stale `handledTap` from the first tap suppresses the delivered `click` → that second tap is lost.

Both are currently **harmless** because both consumers call `selectorState.setOpen(true, …)` (`BibleToolsManager.tsx:664-673`, `BibleReader.tsx` header handler) — an explicit *open*, so a duplicate is a no‑op and a missed one just needs another tap. But `flingSafeTapHandlers` is exported as a general‑purpose utility with a doc comment inviting reuse; a future consumer wired to a **toggle** would surface both bugs as real misbehavior (the exact open‑then‑close symptom #1505 is about).
*Suggested fix (optional):* clear `handledTap` on `pointerdown` of the next press, and/or refresh `handledTap` even on the drag‑reject branch so a rejected tap still suppresses its own trailing click.

**No. 3 — Tests don't reset the module globals between cases, so isolation rests on identity luck.**
`flingSafeTap.ts` holds mutable `pendingTap`/`handledTap` at module scope, and neither test file resets them in `beforeEach` (`test/unit/seed-bible/app/flingSafeTap.test.ts`, `BibleReaderToolbar.test.tsx`). Today the element‑identity and `pointerId` guards keep leaked state from contaminating later cases (I traced each case — they pass for the right reasons, not by accident), but CLAUDE.md's testing guidance is explicit about "no shared mutable fixtures," and a future test that reuses an element reference or `pointerId: 1` could pass or fail based on order rather than behavior.
*Suggested fix:* export a small `__resetFlingSafeTapState()` (or dispatch a `pointercancel` + advance past `CLICK_AFTER_TAP_MS`) and call it in `beforeEach`.

---

### Nits

**No. 4 — The book‑title `<h2>` remains keyboard‑inaccessible.** The helper's doc says "keyboard activation still go through `click`," which is true only for real buttons. The `<h2>` in `BibleReader.tsx` (the reader header title) has no `role`, `tabindex`, or focus, so there's no keyboard path to it at all. This is **pre‑existing** (it was `onClick` on an `<h2>` before this PR, not a regression), but since the PR touches this exact element it's a cheap moment to add `role="button"` + `tabIndex={0}`, or leave a note that the toolbar `<button>` is the accessible entry point.

**No. 5 — `handledTap` retains a reference to a detached DOM node** until the next tap overwrites it (`flingSafeTap.ts:104`). Negligible (one element), noted only for completeness.

---

### On the tests (per the review brief: are they real, or do they always pass?)

They're real. I confirmed the positive cases would fail on the pre‑fix `onClick`‑only control (a `pointerup` with no trailing `click` produced no activation before this helper existed), and the negative cases (`ignores a press that travels away`, `ignores a press the browser turns into a scroll`, `does not activate a disabled control`, and the toolbar `leaves the book selector closed when the tap becomes a swipe`) each fail if their corresponding guard is removed — e.g. dropping the `TAP_SLOP_PX` check turns the swipe test's 140 px drag into an open. The dedup tests (`runs the action once when the click does arrive`, `activates twice for two taps`) genuinely pin the once‑per‑activation contract. Good real‑world coverage: touch, mouse, keyboard, disabled, cancel, drag, and two‑control interleave.

One gap worth a follow‑up test: the "click arrives after the 700 ms window" case in finding No. 2 is currently unguarded by any test — a case asserting single activation there would document the intended dedup window and catch a future regression.

---

KallynGowdy:
@Kabi4 it looks like this change breaks the book selector on mobile android chrome.

See this video: https://youtube.com/shorts/YCc8PyhFwWE?is=9SV154C9mY4PNOPw

It looks like it opens the book selector on pointer down, which then means that a chapter button gets a pointer up event and is clicked.

KallynGowdy:
Seems to work well now

PR #1643, opened 2026-08-14
Add analytics for playlists

KallynGowdy:
This is so that we can track playlist feature usage over time.


Blitzy:
> [!NOTE]

## Code review: Add analytics for playlists

Thanks for the PR! The instrumentation is clean, the `typeof posthog !== "undefined"` guard matches the existing pattern in `SeedBibleStateManager`/`LoginManager`, and the `playlist_created` vs `playlist_updated` and `isCreator` tests exercise real behavior. All 88 tests in `PlaylistManager.test.ts` pass locally.

The concerns below are all about **`playlist_finished` firing when the playlist wasn't actually finished** — which matters because this metric is the whole point of the PR. There's no linked issue, so I reviewed against the PR description ("track playlist feature usage over time").

---

### 🟠 Medium — No. 1: `playlist_finished` fires immediately for single-item playlists (and any playback that *starts* at the last item)

The finished detector treats "current index equals the last index" as finished:

```ts
const disposeFinishedTracking = effect(() => {
  const queueLength = playingState.queue.value.length;
  const index = playingState.currentIndex.value;
  if (hasFiredFinished || queueLength === 0 || index !== queueLength - 1) {
    return;
  }
  hasFiredFinished = true;
  capturePlaylistEvent("playlist_finished", playingState.playlists.peek());
});
```

The effect runs once as soon as the extension activates. For a **single-item queue** the starting index (`0`) already *is* the last index (`queueLength - 1 === 0`), so `playlist_finished` fires before the user has done anything. The same is true whenever playback starts at the last step — e.g. opening a shared `?playlist=<locator>&playlistStep=<lastIndex>` deep link, which lands `currentIndex` on the last item via `setState`. In both cases you get a "finished" with no actual playback, and if playback started mid-list (non-zero step) it even reports finishing playlists whose items were never shown.

A related, smaller variant of the same root cause: `removeFromQueue` clamps `currentIndex` to the new last index, so deleting the trailing items while playing can also trip the "finished" event without the user reaching the end.

**Reproduction (confirmed with a throwaway unit test):**
1. Build a playlist with exactly one item.
2. Call `startPlaying(playlist)` — do **not** call `next()`.
3. Observe `posthog.capture("playlist_finished", …)` fires exactly once immediately.

A test asserting zero `playlist_finished` captures right after `startPlaying` of a one-item playlist fails on the current code (got 1, expected 0).

**Suggested fix:** only fire when playback *advances forward into* the last item, rather than on any state where the index happens to be last. Track the previous index and require a forward move:

```ts
let hasFiredFinished = false;
let previousIndex = playingState.currentIndex.peek();
const disposeFinishedTracking = effect(() => {
  const queueLength = playingState.queue.value.length;
  const index = playingState.currentIndex.value;
  const advancedToLast =
    queueLength > 0 && index === queueLength - 1 && index > previousIndex;
  previousIndex = index;
  if (hasFiredFinished || !advancedToLast) {
    return;
  }
  hasFiredFinished = true;
  capturePlaylistEvent("playlist_finished", playingState.playlists.peek());
});
```

This skips the initial run (no forward move), skips the queue-shrink clamp (index doesn't increase), and skips single-item queues. If product *wants* opening a single-item playlist to count as finished, that should be handled deliberately rather than as a side effect of the index check.

---

### 🟠 Medium — No. 2: In a shared session, every participant reports `playlist_finished`, but only the initiator reports `playlist_played`

`playlist_played` is captured inside `startPlaying`, which runs only on the client that presses play. But `playlist_finished` is a reactive effect on `playingState.currentIndex`, and that index is synced to every participant through the inbound `data` → `setState` path (the same mechanism the file comments describe for keeping sessions in sync). So when one participant advances into the last item, the synced index update fires the finished effect on **every** participant's client, each with its own `hasFiredFinished` flag.

The result: in a shared reading session, `playlist_finished` is inflated by the number of participants, while `playlist_played` is not — so the two events aren't comparable and any "finish rate" derived from them is wrong.

**Reproduction (by code inspection, not run):**
1. Two participants join a shared session playing a multi-item playlist.
2. Participant A advances to the last item.
3. Both A and B emit `playlist_finished` (B via its inbound `setState` sync).

**Suggested fix:** report `finished` from the same place that drives local advancement (the `next`/`navigateNext` flow) rather than from a reactive effect on synced state, so it fires only on the acting client — matching how `playlist_played` already behaves. If you keep the effect, gate it so it doesn't fire when the index change originated from an inbound sync.

---

### 🟢 Low — No. 3: Inconsistent event payloads and a duplicated PostHog guard

`playlist_created` / `playlist_updated` send `{ playlistId, playlistLocator, itemCount }` (no `isCreator`), while `playlist_played` / `playlist_finished` send `{ playlistId, playlistLocator, isCreator }` (no `itemCount`). That's a defensible split, but it's worth a deliberate decision since it makes the events harder to query uniformly in PostHog. Separately, `saveEditingPlaylist` inlines its own `typeof posthog !== "undefined" && posthog` guard + `posthog.capture(...)` instead of going through the new `capturePlaylistEvent` helper — minor duplication that will drift if the guard ever changes.

---

### 🟢 Low — No. 4: Test coverage gap around the edge cases above

The added tests are good and assert real behavior — in particular the "does not re-report on revisit" test is a meaningful regression guard (it fails if the `hasFiredFinished` guard is removed). But they only cover the multi-item happy path (advancing through a 2-item queue). None cover the single-item / start-at-last case from No. 1, the shared-session double-fire from No. 2, or the `posthog`-absent branch for played/finished (other managers do test the absent branch). Adding a "single-item playlist does not report finished on start" test would have caught No. 1 and would fail on the current code.

---

### Summary

| Severity | Item | Issue |
|---|---|---|
| 🟠 Medium | No. 1 | `playlist_finished` fires without the playlist being finished (single-item, start-at-last, queue shrink) — confirmed repro |
| 🟠 Medium | No. 2 | Shared sessions report `finished` per participant but `played` only for the initiator |
| 🟢 Low | No. 3 | Inconsistent event payloads + duplicated PostHog guard |
| 🟢 Low | No. 4 | Missing tests for the edge cases above |

Nothing here breaks user-facing functionality — the impact is on the accuracy of the analytics this PR exists to collect. No. 1 and No. 2 are worth fixing before merge so the `finished` numbers are trustworthy from day one.

---

PR #1644, opened 2026-08-14
Add two-way postMessage bridge between PortalComponent and its ao.bot iframe

KallynGowdy:
## Summary
- `PortalComponent` now supports sending messages into its `ao.bot` iframe (`ref` + `sendMessage`) and receiving messages back out (`onMessage` prop), mirroring the host-side pattern from [casual-simulation/casualos#866](https://github.com/casual-simulation/casualos/pull/866) (`os.sendEmbedMessage` / `@onEmbedMessage`).
- Inbound messages are validated against both origin (`https://ao.bot`) and `event.source` (the specific iframe's `contentWindow`), so messages can't be spoofed and multiple simultaneously-open portals don't cross-talk.
- Documented the new messaging API, including the pattern-side `os.sendEmbedMessage`/`onEmbedMessage` usage and the readiness-handshake caveat, in `DEVELOPERS.md`.

## Test plan
- [x] `pnpm vitest run PortalComponent.test.tsx` — 8 tests pass, including send/receive, wrong-origin rejection, wrong-iframe (sibling portal) rejection, and listener cleanup on unmount
- [x] Confirmed the two rejection tests fail (red) when the origin/source guard is reverted, then pass again once restored
- [x] `pnpm check:ts` — 0 errors
- [x] `pnpm lint` — 0 errors (pre-existing unrelated warnings only)
- [ ] Manual end-to-end check against a live `ao.bot` deployment that supports `os.sendEmbedMessage`/`@onEmbedMessage`

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -94,11 +94,11 @@ On the pattern side (the code that runs inside the portal's iframe), use CasualO
 // send a message out to the host page
 await os.sendEmbedMessage({ type: "featureClicked", featureId });
 
+// onEmbedMessage.tsx
 // receive messages sent from the host page
-export function onEmbedMessage(data: { origin: string; message: any }) {
-  if (data.message?.type === "highlightFeature") {
-    highlightFeature(data.message.featureId);
-  }
+let data = that;
+if (data.message?.type === "highlightFeature") {
+  highlightFeature(data.message.featureId);
 }
 ```
 
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -59,3 +59,47 @@ context.panes.openPane({
   },
 });
 ```
+
+### Messaging with a Portal
+
+`query` only gets data into a pattern once, at the moment its iframe is created — changing `query` on a later render does not reach an already-open portal. To send and receive messages after the portal has opened, use `PortalComponent`'s `ref` and `onMessage`:
+
+```typescript
+const portalRef = useRef<PortalComponentHandle>(null);
+
+context.panes.openPane({
+  placement: "floating",
+  title: "My Portal",
+  component: () => (
+    <PortalComponent
+      ref={portalRef}
+      portal="map"
+      portalType="map"
+      pattern={myPattern}
+      inst={inst}
+      onMessage={(message) => {
+        // handle a message sent from inside the pattern via os.sendEmbedMessage
+      }}
+    />
+  ),
+});
+
+// later, send a message into the pattern:
+portalRef.current?.sendMessage({ type: "highlightFeature", featureId });
+```
+
+On the pattern side (the code that runs inside the portal's iframe), use CasualOS's built-in embed-message APIs:
+
+```typescript
+// send a message out to the host page
+await os.sendEmbedMessage({ type: "featureClicked", featureId });
+
+// receive messages sent from the host page
+export function onEmbedMessage(data: { origin: string; message: any }) {
+  if (data.message?.type === "highlightFeature") {
+    highlightFeature(data.message.featureId);
+  }
+}
+```
+
+The iframe loads asynchronously, so a pattern should send a `{ type: "ready" }`-style message once its `onInstJoined`/`onEggHatch` handler has run, and the host should wait to see that before calling `sendMessage` for the first time — otherwise the very first message can be sent before the pattern is listening and will be silently dropped.
--- a/packages/seed-bible/seed-bible/components/PortalComponent/PortalComponent.tsx
+++ b/packages/seed-bible/seed-bible/components/PortalComponent/PortalComponent.tsx
@@ -1,7 +1,16 @@
 import "./PortalComponent.css";
-import { useRef } from "preact/hooks";
+import { useEffect, useImperativeHandle, useRef } from "preact/hooks";
+import { forwardRef } from "preact/compat";
 
 export type CasualOSPattern = { name: string } | { aux: string };
+
+/** The origin of every `ao.bot` portal iframe — the only origin trusted for postMessage traffic in or out. */
+const AO_BOT_ORIGIN = "https://ao.bot";
+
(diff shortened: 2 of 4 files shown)


Blitzy:
> [!NOTE]

## Code Review — Two-way postMessage bridge for `PortalComponent`

Nice, focused change. The core design is sound: outbound `sendMessage` posts with an explicit `targetOrigin` (never `"*"`), and inbound messages are gated on **both** `event.origin` and `event.source`, which is exactly the pair you need so that (a) a hostile origin can't inject and (b) sibling portals mounted in other floating panes don't cross-talk. I verified the security-critical tests are real regression tests, not always-green filler:

- Reverting the **origin** check → `ignores a message from an untrusted origin` goes red. ✅
- Reverting the **source** check → `ignores a same-origin message from a different portal's iframe` goes red. ✅

So the two guards are genuinely covered. Below are the issues I found, by severity.

---

### Medium

**No. 1 — The unmount test doesn't actually test unmount cleanup; it passes even if the listener is never removed.**

`stops calling onMessage once unmounted` looks like it guards against a `window` `"message"` listener leak, but it can't detect one. After unmount, Preact sets `iframeRef.current` to `null`, so the guard `event.source !== iframeRef.current?.contentWindow` becomes `capturedWindow !== undefined` → `true` → the handler returns early **regardless of whether the listener was removed**. The source check masks the leak, so the test is green either way.

I confirmed this empirically: deleting the cleanup line

```ts
return () => window.removeEventListener("message", handleMessage);
```

leaves **all 8 tests passing**, including this one. That means a future refactor that drops the cleanup (reintroducing a real listener leak — every open/close of a portal pane would accumulate a dead `window` listener) would sail through CI unnoticed.

The production code is correct today — the cleanup *is* there. The problem is purely that the test protecting it doesn't do its job.

_Reproduction:_
1. In `PortalComponent.tsx`, delete the `return () => window.removeEventListener(...)` line from the `useEffect`.
2. `pnpm vitest run test/unit/seed-bible/components/PortalComponent.test.tsx`
3. Observe 8/8 still pass — the leak is invisible to the suite.

_Suggested fix — assert on the listener lifecycle directly, so the guard can't hide a leak:_

```ts
it("removes the same message listener it added, on unmount", () => {
  const addSpy = vi.spyOn(window, "addEventListener");
  const removeSpy = vi.spyOn(window, "removeEventListener");

  act(() => {
    render(
      <PortalComponent
        portal="map"
        portalType="map"
        inst="inst-1"
        pattern={null}
        onMessage={vi.fn()}
      />,
      container
    );
  });

  const added = addSpy.mock.calls.find(([type]) => type === "message");
  expect(added).toBeDefined();
  const handler = added[1];

  act(() => {
    render(null, container);
  });

  expect(removeSpy).toHaveBeenCalledWith("message", handler);
});
```

This ties the mounted handler identity to the one removed on unmount, so deleting the cleanup makes it fail. You could keep the existing behavioral test too (it's still a fine smoke check), but this is the one that actually protects the invariant.

---

### Low

**No. 2 — First `sendMessage` can be silently dropped, and the API gives callers no built-in way to know when the iframe is ready.**

This is documented in `DEVELOPERS.md` (the "send a `{ type: "ready" }`-style message" handshake), which is good and the guidance is correct — an `iframe` `onLoad` wouldn't help anyway, since it fires when the document loads, not when the pattern's `onInstJoined`/`onEggHatch` handler has registered its listener. So the docs land in the right place.

The residual sharp edge is that `sendMessage` before readiness is a **silent** no-op: `iframeRef.current?.contentWindow?.postMessage(...)` swallows the case where the pattern isn't listening yet, so a caller who forgets the handshake gets no error and no message — just nothing. Since there's no in-tree consumer of `sendMessage`/`onMessage` yet, this will bite the first person who wires it up.

Not blocking, but two options worth considering (either, not both):
- Have `PortalComponent` buffer messages sent before the first inbound message arrives, flushing them once the pattern is confirmed listening; or
- Leave the code as-is and add a one-line inline reminder at the `sendMessage` call site pointing at the handshake note, so the constraint is visible from the code, not only the external doc.

**No. 3 — `useImperativeHandle(ref, factory)` has no dependency array (nit).**

The handle factory re-runs and re-creates the `{ sendMessage }` object on every render. It's harmless here — the closure only reads the stable `iframeRef`, so a stale handle would behave identically — but passing `[]` (mirroring the empty-dep pattern you deliberately use for the `useEffect` right below it) makes the "created once, stable for the life of the mount" intent explicit:

```ts
useImperativeHandle(ref, () => ({
  sendMessage: (message: unknown) => {
    iframeRef.current?.contentWindow?.postMessage(message, AO_BOT_ORIGIN);
  },
}), []);
```

---

### Notes / non-issues (verified, no action needed)

- **`onMessageRef` updated during render** (`onMessageRef.current = props.onMessage`) — this is a render-phase side effect, but it's safe under Preact (no concurrent rendering) and mirrors the existing `iframeSrcRef` pattern in the same file. Fine.
- **Cross-origin `event.source` reference compare** — comparing `event.source` against `iframeRef.current.contentWindow` (both `WindowProxy` references) is the correct, MDN-recommended approach and works across origins. Good.
- **`sendMessage` positive test** — spying on `iframe.contentWindow.postMessage` and asserting the `(payload, "https://ao.bot")` args is a real assertion that would fail if the target origin were broadened or the payload mangled. Good.
- No linked issue on the PR and no `Closes #…`; the PR body references `casual-simulation/casualos#866` for the mirrored host-side API, which is sufficient context. Flagging only so the tracking side is intentional.

---

**Summary:** No correctness or security defects in the shipping code — the guards are right and well-covered. The one thing worth fixing before merge is the unmount test (No. 1), which currently can't catch the leak it's meant to prevent. No. 2 and No. 3 are optional polish.

---

KallynGowdy:
Manually confirmed that messaging works. This is good to merge

PR #1645, opened 2026-08-15
Adding First Desgin for Histroy of playlsit play

### LINKED ISSUE #1427: Track recently played playlists in your reading history
Currently, when you play a playlist, there is no history of it. This means that if someone shares you a link, you'll have to keep the playlist link for the next time you want to use it. This task is to track playing playlists in your reading history so that we can display it back to the user.

Kabi4:
Fixes: #1427 

image


Blitzy:
> [!NOTE]

## Code review — Playlist play history (No. 1645)

Thanks for tackling [issue #1427](https://github.com/HelloAOLab/seed-bible/issues/1427). This PR adds a per‑session play‑history record (`PlaylistPlayHistorySchema`), tracks it as playback runs, exposes `continueFromHistory`/`replayFromHistory`, and renders an expandable "Playlist history" section in Discover. The direction matches the issue well: a signed‑in user who opens a shared `?playlist=` link now gets a history row, so they can reopen it later without keeping the URL.

I checked out the branch and ran the two affected suites locally — **`PlaylistManager.test.ts` + `DiscoverPane.test.tsx`: 115 passed**. The tests exercise real behavior (start/stop, advance, continue/replay chaining, sign‑out clearing, the empty state, and expand → dispatch of the right method), and each of the guard tests would fail if its guard were removed, so they aren't tautological. Two coverage gaps are noted under Low below.

One process note before the findings: GitHub reports this PR as `mergeable_state: dirty` — it currently conflicts with `develop` and needs a rebase/merge before it can land. The PR description is also empty; a short summary linking #1427 would help reviewers.

---

### 🔴 High

**1. History only loads the first page from the backend, ordered by record id — not by recency — so recent plays silently disappear after a reload.**

`loadPlayHistory` calls `os.listDataByMarker` (a single page), then sorts that page by `startedAtMs`:

```ts
// PlaylistManager.tsx ~626
const records = await os.listDataByMarker(recordName, PLAYLIST_PLAY_HISTORY_MARKER);
...
return records.items
  .map((record) => PlaylistPlayHistorySchema.safeParse(record.data))
  .filter((r) => r.success).map((r) => r.data)
  .sort((a, b) => b.startedAtMs - a.startedAtMs);
```

The records controller pages and sorts by **address** (the record id, `playlist_history_<uuid>`), ascending — not by time. Because every play creates a brand‑new random‑UUID record (see No. 3), once a user accumulates more than one page of sessions the first page is effectively a *random subset by id*. Sorting that subset by `startedAtMs` can't recover the entries that never came back. Net effect: the most recently played playlist — exactly what the feature exists to resurface — may not appear in the list after a reload.

This mirrors the existing `listPlaylists`, but playlists are user‑created and few, whereas history grows one row per play. The correct pattern is already in the codebase: `ReadingPlansManager` reads its similarly‑growing progress records with `os.listAllDataByMarker` (`ReadingPlansManager.tsx:1359, 1417`).

*Reproduction*
1. Sign in and play more than one page worth of playlists (or seed &gt;N records under your user record with marker `publicRead:playlistPlayHistory`).
2. Reload the app.
3. Open Discover → "Playlist history".

*Observed:* an arbitrary subset shows; the just‑played playlist can be missing. (In a single session it looks fine because `upsertLocalPlayHistory` adds the new row to the in‑memory signal — the bug only surfaces after reload, when the list comes from the backend.)
*Expected:* recent sessions, newest first.

*Suggested fix:* load with `os.listAllDataByMarker` like `ReadingPlansManager`, then keep the N most recent after sorting by `startedAtMs`.

---

### 🟠 Medium

**2. The duration timer writes a full record to the backend every 5 seconds for the entire active session — even when idle and nothing changed.**

```ts
// PlaylistManager.tsx ~1193
effect(() => {
  const historyId = activePlayHistoryId.value;
  if (!historyId) return;
  const intervalId = setInterval(() => {
    ...
    void patchActivePlayHistory({ currentStep, totalSteps, lastItem });
  }, 5000);
  return () => clearInterval(intervalId);
});
```

`patchActivePlayHistory` → `persistPlayHistoryPatch` (~660) always calls `savePlayHistory` → `os.recordData`; there's no "did anything change?" check. So as long as a session is active — including while paused or parked on one item — the app issues a `recordData` write every 5 seconds indefinitely, each of which also `console.log`s the full payload (`OsManager.recordData`, ~321). The mirror effect (~1177) additionally writes on every `currentIndex` change, and it fires once immediately after `beginPlayHistory` already wrote — so starting a playlist is at least two writes back‑to‑back. This is a lot of avoidable network traffic and log noise for a background feature.

*Reproduction:* sign in, start a playlist, open the network panel, and leave it on one item → observe a `recordData` POST every ~5s with no user interaction.

*Suggested fix:* skip the write when the patch is identical to what was last persisted (compare `currentStep`/`lastItem`/rounded `durationMs`); and/or drop the periodic timer, lengthen its interval substantially, or only run it while playback is actually advancing. A final flush on `finalizeActivePlayHistory` (already present) covers the stop case.

**3. History grows without bound — no cap, no dedup, nothing is ever erased.**

Every `startPlaying` for a signed‑in user creates a new record (`beginPlayHistory`, ~713), and `continueFromHistory`/`replayFromHistory` each start a *new* chained row. Opening the same shared link ten times yields ten rows for the same playlist; nothing prunes old sessions. This is the root cause that makes No. 1 bite (single‑page load over an ever‑growing set) and also makes the Discover list arbitrarily long, with many duplicate rows for one playlist.

*Suggested fix:* decide the retention model — e.g. keep only the latest session per playlist (dedup by `playlistId`), or cap total retained rows and `eraseData` the oldest when writing a new one. Whatever you choose, it should bound what No. 1's loader has to read.

---

### 🟡 Low / Nitpicks

**4. "Complete" flips the instant you reach the last item, before consuming it.** `isPlaylistPlayHistoryComplete` returns true when `currentStep >= totalSteps - 1`, so advancing onto the final queue item immediately shows 100% and swaps the button to "Replay" even though the user hasn't finished that item. Confirm this is intended; if not, a session likely shouldn't count as complete until playback actually ends on the last item.

**5. "Time played" is wall‑clock tab‑open time, not listening time.** Because the 5s timer keeps extending `durationMs` while a session is active regardless of interaction, leaving the tab open on a playlist inflates the number (open a playlist, walk away 10 min → "Time played 10m"). The schema comment acknowledges this, but the user‑facing label "Time played" implies active playback. Consider relabeling (e.g. "Session length") or tracking only active time.

**6. Two untested paths — one of which hides No. 1.**
 - The failure/toast branch (`playlist-history-open-failed`, shown when the playlist was deleted) is never exercised. `continueFromHistory`/`replayFromHistory` await `loadPlaylist`, which rejects on a missing record, and the component's `.catch` shows the toast — but the DiscoverPane tests mock those manager methods to resolve, so nothing asserts the toast. Add a test where `getData` rejects and assert the toast fires.
 - No test seeds more than one page of history, so the pagination truncation in No. 1 is invisible to the suite. A manager test that has the mocked list return multiple pages (and asserts the loader reads all of them / keeps the most recent) would turn No. 1 into a red test first.

**7. Cosmetic.**
 - `resolveBookName` in `PlaylistHistorySection` resolves against the *currently selected tab's* translation books; a history row from a different translation (or before books load) falls back to the raw id and shows e.g. "JHN 3:16".
 - Unused CSS modifier `.sb-playlist-history-item--expanded` (added in JSX, never styled).
 - `.sb-playlist-history-details` sets `padding: 0 10px 10px;` then immediately `padding-top: 8px;` — the shorthand's top value is dead.

---

### i18n
Only `en.json` was touched for the new keys, per the project convention — 👍

Overall this is a solid first cut with genuinely useful tests. The main thing to resolve before merge is No. 1 (with No. 3 behind it): as written, the history list becomes unreliable for exactly the frequent‑user case the feature targets. No. 2 is worth fixing alongside it to avoid a steady stream of background writes.

---

Kabi4:
Playlist history was only the latest play per playlist, so Continue/Replay replaced the previous row.

Discover now keeps a listening log: sessions newest-first (capped at 200), grouped by day, with Continue listening for unfinished playlists. Continue within 30 minutes resumes the same session instead of adding a duplicate. You can remove a play from history, and deleting a playlist clears its history too.

KallynGowdy:
@Kabi4 I have a couple UI notes:

1. Could you remove the "Continue Listening" section?
image

2. Each playlist history item should have a play button icon that lets the user resume where they were (or restart if finished). It should be like the existing playlist items. Instead of being able to expand it, it should have a three dots context menu that has an option to remove from history. Finally, the "last item" field should be displayed in the status/description area after the completion percentage. (e.g. `2:23 PM - 70% complete - Ezekiel 2`). Remove the "Time Played" field.
image

KallynGowdy:
@Kabi4 The visuals look good now, but I think I found a bug: When resuming a playlist from a different day, it creates a duplicate history entry instead of updating the old one. I think we should make it so that a user can only ever have one history entry for a single playlist.

### Steps to reproduce
1. Get a history entry that is at least a day old. In my case, the playlist history entry was from August 19 2026.
2. Click on it to resume it.
3. Immediately stop playing it.
4. Go back to the discover page to see your playlist history.
5. See that there is now a duplicate history entry and the old one wasn't updated.
image

Kabi4:
Hmm yes it was to track the exact history day wise since you played it on odd days like not same date if i remove it from past you won't have the track @KallynGowdy let me know if you want to change this behavior though but for same day it does the trick.

but i found a bug though playlist don't care what you play it only tracks the last item played position to calculate the % of completion. This might need to track playing in deeper level let me know if seems fine to you.

KallynGowdy:
> Hmm yes it was to track the exact history day wise since you played it on odd days like not same date if i remove it from past you won't have the track @KallynGowdy let me know if you want to change this behavior though but for same day it does the trick.
> 
> but i found a bug though playlist don't care what you play it only tracks the last item played position to calculate the % of completion. This might need to track playing in deeper level let me know if seems fine to you.

Yeah so it should track on a per-playlist level, but only ever track one history entry per playlist. (at least for now) We really don't need a full history like with reading history where you can go back months to see all the playlists you've ever played. Instead, this feature is mostly about keeping track of what you've recently played and restarting/finishing playlists that you've stopped halfway through.

If I play `playlist 1`, then later I should see `playlist 1` in my history with a completion percentage. If I resume that playlist, or start it fresh, then it should just reset my history for `playlist 1` to show the most recent history and not the full history.

Blitzy:
> [!NOTE]

## Code review — Playlist play history (No. 1645)

This is a re-review after the discussion that followed the first pass on this PR. Since then the design changed significantly based on maintainer feedback in the thread: the feature moved from a growing session log to **one history row per playlist** (newest session replaces the old one), the pagination bug was fixed (loads with `os.listAllDataByMarker` instead of a single page), the periodic 5-second background-write timer was removed, and the Discover UI was reworked per [KallynGowdy's mockups](https://github.com/HelloAOLab/seed-bible/pull/1645#issuecomment-5346409774) (play button to resume/restart, a "⋮" menu to remove from history, status line showing time/percent/last item, "Continue Listening" section and "Time Played" field removed).

I checked out `feature/1427` at the current head (`6fec164`) and independently ran the two affected suites: **`PlaylistManager.test.ts` + `DiscoverPane.test.tsx` — 178 passed.** I also verified the specific bug [KallynGowdy reported on Aug 21](https://github.com/HelloAOLab/seed-bible/pull/1645#issuecomment-5371244985) (resuming a playlist from a different day created a duplicate history row instead of updating the existing one) by tracing `beginPlayHistory`/`continueFromHistory`: it reuses the existing row's `id` when a history entry already exists for that `playlistId` + `playlistRecordName`, so `os.recordData` overwrites the same backend record rather than creating a new one. That matches the maintainer's ask ("only ever track one history entry per playlist") and is covered by a dedicated regression test (`"starting a playlist again resets that playlist's history to the new session"`).

CI is green on this head commit, and the earlier High/Medium findings (pagination truncation, unbounded background writes, unbounded history growth) all look resolved with test coverage backing them.

---

### 🟡 Low

**1. Branch needs a rebase — GitHub reports `mergeable_state: dirty`.**

I confirmed this locally: merging `origin/develop` into `feature/1427` produces real conflicts in `CHANGELOG.md` and `packages/seed-bible/seed-bible/i18n/en.json` (both files have drifted — mostly reordered/re-inserted entries from `develop` moving ahead since the last sync on Aug 24, not content collisions). Not a code bug, but it'll need a merge/rebase before this can land.

*Suggested fix:* `git fetch origin develop && git merge origin/develop` (or rebase, per repo convention), resolve the two files, and push.

**2. The failure path when reopening a deleted/missing playlist is still untested.**

This was flagged in the first review round and is still open. `playFromHistory` (`DiscoverPane.tsx`) calls `continueFromHistory`/`replayFromHistory` and shows a toast (`playlist-history-open-failed`) if the promise rejects — which happens for real when `loadPlaylist` can't find the backing record (e.g., the shared playlist was deleted by its owner). Every test in `DiscoverPane.test.tsx` mocks `continueFromHistory`/`replayFromHistory` to resolve, so nothing exercises the `.catch` branch or asserts the toast fires.

*Reproduction:* mock `continueFromHistory` (or `replayFromHistory`) with `.mockRejectedValue(new Error("not found"))`, click the play button on a history row, and assert `toast` was called with the `playlist-history-open-failed` message.

*Suggested fix:* add that one test — it's a small addition and closes the last real coverage gap in this feature.

**3. Cosmetic — `resolveBookName` in `PlaylistHistorySection` still resolves against the currently-selected tab's translation.**

A history row for a playlist read in a different translation (or before books have loaded) falls back to the raw book id in the "last item" summary (e.g. "JHN 3:16" instead of "John 3:16"). Same nitpick as the first review round; low priority since it only affects the label, not functionality.

---

### i18n
New keys added only to `en.json`, per project convention — 👍. Note that the `en.json` diff also contains a lot of unrelated reordering from the stale base (see No. 1) — worth double-checking after the rebase that no keys were accidentally dropped in the merge.

---

Overall this addresses everything raised in the earlier review round and the maintainer conversation — nice iteration. Nothing here blocks merge except the rebase (No. 1); No. 2 and No. 3 are polish that could land now or in a fast follow-up.

---

PR #1649, opened 2026-08-17
fix: stub clearBranchDeviceCache in all SessionsManager tests

KallynGowdy:
The isSynced sync-status test flips synced false→true, which triggers
SessionsManager's presence-rebuild path and calls the real (unmocked)
os.clearBranchDeviceCache(). That lazily opens a genuine websocket
connection to auth.seedbible.org. Locally the connection never
completes, but on GitHub Actions' runners it can, and the resulting
async socket event fires after the test finishes, hitting an
undici/jsdom Event-realm mismatch that crashes an unrelated, later
test with "Unhandled Errors". Stubbing the method for every test in
this file (not just the presence-recovery describe block) removes the
real network call entirely.

### DIFF
--- a/test/unit/seed-bible/managers/SessionsManager.test.ts
+++ b/test/unit/seed-bible/managers/SessionsManager.test.ts
@@ -397,6 +397,12 @@ describe("SessionsManager", () => {
     getSharedDocumentMock = vi
       .spyOn(os, "getSharedDocument")
       .mockResolvedValue(mockDocument as unknown as SharedDocument);
+    // The real implementation lazily builds an inst client, which opens a
+    // real websocket connection — stub it everywhere so a sync-status test
+    // that flips false→true (which triggers a presence rebuild) can't
+    // trigger a real network connection whose async events fire after the
+    // test has finished and crash an unrelated, later test.
+    vi.spyOn(os, "clearBranchDeviceCache").mockImplementation(() => undefined);
     mockDataManager = {};
     mockLoginManager = {
       getUserProfile: vi.fn(async (userId: string) => ({


Blitzy:
> [!NOTE]

## Code Review — PR #1649: stub `clearBranchDeviceCache` in all SessionsManager tests

### Summary

This is a small, test-only change (6 additions, 1 file). It hoists a stub of `os.clearBranchDeviceCache()` from the `presence recovery after a reconnect` describe block up into the top-level `beforeEach` of `SessionsManager.test.ts`, so **every** test in the file gets the no-op stub rather than only the presence-recovery ones.

The stated problem checks out against the code:

- The real `clearBranchDeviceCache` (`managers/OsManager.tsx:266`) calls `getInstClient()`, which lazily builds the inst client and opens a real websocket to auth.seedbible.org.
- `SessionsManager`'s `rebuildRemoteClientsSubscription` (`managers/SessionsManager.tsx:1092-1098`) calls `os.clearBranchDeviceCache(null, id, "session_data")` on every sync `false → true` transition.
- The test `isSynced starts true and tracks the document's sync status updates` (line 1580) lives **outside** the presence-recovery describe block and emits `synced: false` then `synced: true` (lines 1594/1597). Before this PR that transition hit the **real** method, opening the socket. That matches the description of the CI-only crash (a late async socket event landing in an unrelated later test).

No product code changes; nothing here alters runtime behavior.

### Findings by severity

**High / Medium correctness: none found.** The fix targets the correct boundary. Per the repo's own testing guidance, `OsManager`/CasualOS SDK calls are the sanctioned mock boundary, and `clearBranchDeviceCache` is exactly such a boundary method. Stubbing it to a no-op removes only the unwanted network side effect; it does not weaken any assertion, because no test outside the presence block asserts on it.

**Low / Nit — redundant double-spy inside the presence-recovery block (optional).**
Within `describe("presence recovery after a reconnect")`, `clearBranchDeviceCache` is now spied twice per test: once by the new top-level `beforeEach` (line 405) and again by the block's own `beforeEach` (line 1633). This is harmless — `os` is recreated fresh each test (`os = CasualOSManager()`, line 359), and the inner `spyOn` is applied last, so `os.clearBranchDeviceCache` resolves to the captured `clearBranchDeviceCacheSpy` that the assertions (`toHaveBeenCalledWith`, `not.toHaveBeenCalled`) rely on. So it works and its assertions still verify real behavior. If you want to trim it, you could drop the inner `beforeEach` and capture the outer spy instead — but that adds churn to expose the reference, so keeping the local spy for readability is a reasonable call. No change required.

### On test quality (real behavior vs. always-green)

The change does not make any test trivially pass. The presence-recovery tests still assert observable behavior against the stub: that `clearBranchDeviceCache` is called with `(null, "group-abc", "session_data")` on recovery, that it is **not** called on the initial sync, and that `remoteClients.subscribe` is re-invoked the expected number of times. Those would fail if the source logic in `rebuildRemoteClientsSubscription` regressed, so the stub is a boundary mock, not a way to sidestep the assertion.

One inherent gap worth naming: there is no test that directly proves "no real socket is opened," because that is a suppressed side effect that is awkward to assert on. That is acceptable here — the fix is about removing a nondeterministic side effect, and asserting on the absence of a network call would itself be brittle.

### Verification performed

- Ran `pnpm vitest run test/unit/seed-bible/managers/SessionsManager.test.ts` → **47 passed**.
- Grepped the rest of `test/` for the same pattern: only `SessionsManager.test.ts` emits sync-status events / references `clearBranchDeviceCache`. The other files that mention `SessionsManager` (`SeedBibleStateManager`, `TabsManager`, `ChatsManager` tests) do not drive sync transitions, so the fix's single-file scope is complete — no sibling test carries the same latent risk.

### Reproduction of the original flakiness

This is a CI-only race and is not reliably reproducible locally — locally the websocket to auth.seedbible.org never completes, so the crashing async event never fires (the PR description notes this). The best available signal is CI history on GitHub Actions runners showing intermittent "Unhandled Errors" in an unrelated later test. No local repro steps can be given.

### Verdict

Looks good. Correct root cause, minimal and well-scoped fix, at the right mock boundary, with existing real-behavior assertions intact and the full file green. The only note is an optional cosmetic one (the redundant inner spy), which is safe to leave as-is.


---

PR #1650, opened 2026-08-17
Add PostHog analytics events for reading plan lifecycle

KallynGowdy:
Fires reading_plan_created/updated/deleted/started/finished plus
day_finished and session_finished, covering every path that can
complete a session (whole-session, whole-day, or last-reading marks)
via a shared completion-diff helper. Extracts the repeated
posthog.capture guard into a captureEvent helper in Utils.tsx and
switches the existing user_chapter_read/create_session/join_session
call sites to use it.

Claude-Session: https://claude.ai/code/session_01CUcrgG3D7eGZJc6gmYXrWN

### DIFF
--- a/packages/seed-bible/seed-bible/managers/ReadingPlansManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ReadingPlansManager.tsx
@@ -2217,6 +2217,7 @@ export function createReadingPlansManager(
   const deleteReadingPlan = async (plan: {
     recordName: string;
     address: string;
+    status: ReadingPlanStatus;
   }) => {
     const planId = formatReadingPlanId(plan.recordName, plan.address);
     const ownProgresses = userReadingPlanProgresses.value.filter(
@@ -2224,7 +2225,12 @@ export function createReadingPlansManager(
     );
     await os.eraseData(plan.recordName, `${plan.address}_metadata`);
     await os.eraseData(plan.recordName, plan.address);
-    captureEvent("reading_plan_deleted", { planId });
+    // A draft that was never finished never fired `reading_plan_created`, so
+    // discarding it here (see `discardEditingReadingPlan`) must not count as a
+    // delete — only a plan that actually got published is one.
+    if (plan.status === "complete") {
+      captureEvent("reading_plan_deleted", { planId });
+    }
     // Best effort: a progress that fails to erase leaves nothing broken behind
     // (it simply stops matching a plan), so it must not fail the delete.
     await Promise.all(
--- a/packages/seed-bible/seed-bible/managers/ReadingPlansManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ReadingPlansManager.tsx
@@ -185,6 +185,20 @@ export function formatReadingPlanId(
   return `rp_${recordName}_${address}`;
 }
 
+/** Session/reading totals for analytics, counted as the plan currently stands. */
+function readingPlanCounts(plan: ReadingPlan): {
+  totalSessions: number;
+  totalReadings: number;
+} {
+  return {
+    totalSessions: plan.sessions.length,
+    totalReadings: plan.sessions.reduce(
+      (sum, session) => sum + session.readings.length,
+      0
+    ),
+  };
+}
+
 /**
  * Creates a fresh progress record for a user starting a plan. `id` (unique) and
  * `nowMs` are passed in so this stays deterministic; the manager supplies them.
@@ -1885,6 +1899,24 @@ export function createReadingPlansManager(
           : [...userReadingPlans.value, metadata];
       });
       editingReadingPlanSaveError.value = false;
+      // Editing an already-published plan (`isNew: false`) autosaves the same
+      // way but isn't a "draft" in the lifecycle sense — it keeps its
+      // `"complete"` status throughout and reports via
+      // reading_plan_updated on finish, not here.
+      if (draft.isNew) {
+        captureEvent(
+          draft.persisted
+            ? "reading_plan_draft_updated"
+            : "reading_plan_draft_created",
+          {
+            planId: formatReadingPlanId(
+              draft.plan.recordName,
+              draft.plan.address
+            ),
+            ...readingPlanCounts(draft.plan),
+          }
+        );
+      }
     } catch (error) {
       console.error("Failed to save reading plan draft:", error);
       editingReadingPlanSaveError.value = true;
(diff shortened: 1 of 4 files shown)


Blitzy:
> [!NOTE]

## Code Review — Add PostHog analytics events for reading plan lifecycle

**Summary:** This is a clean, well-scoped change. The `captureEvent` helper is a sensible extraction, the completion-diff approach correctly handles the three paths that can complete a session (whole-session, whole-day, last-reading), and the new tests are genuine — I mutation-tested the two completion guards and confirmed the tests go red when the logic is broken (details below). CI is green; the `blocked` merge state is just the pending required review.

No correctness bugs found in the event-firing logic. The findings below are one analytics-accuracy concern and a few test-coverage gaps.

---

### Medium

**M1 — Discarding an unfinished draft fires `reading_plan_deleted` with no matching `reading_plan_created`.**

`deleteReadingPlan` now unconditionally fires `reading_plan_deleted` (`ReadingPlansManager.tsx:2227`). But that function is also the draft-discard path: `discardEditingReadingPlan` calls `deleteReadingPlan(draft.plan)` whenever a *persisted, still-new* draft is thrown away (`ReadingPlansManager.tsx:2270-2274`). Meanwhile `reading_plan_created` only fires from `finishEditingReadingPlan` (`ReadingPlansManager.tsx:2196`).

So a user who opens the plan editor, lets the draft autosave (which flips `persisted` to `true` at `ReadingPlansManager.tsx:1871`), then cancels — generates a `reading_plan_deleted` event for a plan that was never "created" in analytics. Since the whole point of this PR is accurate lifecycle metrics, this will inflate the delete count with phantom deletes and make created-vs-deleted funnels not line up.

*Reproduction (for a human developer):*
1. Start creating a new reading plan and add at least one reading.
2. Wait for the debounced draft autosave (draft becomes `persisted: true`).
3. Cancel / discard the draft instead of finishing it.
4. Observe a `reading_plan_deleted` event fires, with no preceding `reading_plan_created`.

*Suggested fix:* thread a flag through `deleteReadingPlan` (e.g. `deleteReadingPlan(plan, { fromDraftDiscard })`) and skip the capture on the discard path, or only fire `reading_plan_deleted` for plans that reached `status: "complete"`. Whichever you pick, please confirm the intended semantics — if counting discarded drafts as deletes is deliberate, a one-line comment at the capture site would save the next reader this same investigation.

---

### Low

**L1 — The `markDayComplete` idempotency guard is not tested for its main case.**

The `wasAlreadyComplete` guard (`ReadingPlansManager.tsx:1675-1690`) exists to stop `reading_plan_day_finished` from re-firing when the day is *already* complete. The test covers complete→fires and complete→incomplete→no-fire, but never marks an already-complete day complete a second time — so the guard's primary purpose is unverified. A malformed guard that always fired would still pass the current suite.

*Suggested test:* after the first `markDayComplete(day)`, `mockPosthogCapture.mockClear()`, call `markDayComplete(day)` again, and assert `reading_plan_day_finished` was **not** captured. (Mirrors the "re-saving must not re-fire" assertion already present for `reading_plan_finished`.)

**L2 — Marking a whole day complete emits both `reading_plan_day_finished` and one `reading_plan_session_finished` per session on that day, and no test asserts this.**

Because `markDayComplete` runs through `updateSelectedProgress`, `captureProgressCompletionEvents` also fires a `reading_plan_session_finished` for every session that transitions to complete on that day. This looks intended (a day completion does complete its sessions), but a single "mark day done" tap produces N+1 events, which matters if a PostHog funnel treats `reading_plan_session_finished` as a distinct user action. The `markDayComplete` test only asserts the day event and is silent on the session events, so the intended count isn't pinned down. Recommend adding an explicit assertion for the session-event count on the day path (whatever the intended number is) so the behavior is documented and locked in.

**L3 — The self-paced `reading_plan_started` branch is untested.**

`reading_plan_started` sends `cadenceId: progress.selectedCadenceId`, which is `null` for self-paced plans (see `ReadingPlansManager.test.ts:1993`). The analytics test only exercises `selfPaced: false` with a real cadence id, so the `cadenceId: null` path never runs. Not a bug, but a cheap branch to cover.

---

### What I verified

- **Tests reflect real behavior, not always-green.** I ran two mutations against the working tree:
  - Removing the "already-complete" guard in the session loop (firing whenever `completedAtMs` is set) → *"markSessionComplete … doesn't refire when already complete"* fails. ✅
  - Weakening the plan-finished guard from `previousPercent < 1 && next.percentComplete === 1` to just `next.percentComplete === 1` → *"completing the plan's last remaining session captures reading_plan_finished"* fails on the re-save assertion. ✅

  Both were reverted; the working tree is unchanged.
- The pure `mark*InProgress` helpers return new objects (never mutate), so capturing `previous` from the signal before overwriting it is safe.
- `percentComplete === 1` is exact here — it's `doneUnits / totalUnits` and `n / n` is exactly `1.0` in IEEE-754 — so the finished check has no floating-point edge.
- The `create_session` / `join_session` / `user_chapter_read` refactors to `captureEvent` remain covered by existing `SeedBibleStateManager` tests, including the "does nothing if posthog isn't available" SSR case.
- `posthog` is declared as an ambient global (`typings/globals.d.ts:24`), so `captureEvent` type-checks; the `undefined` guard keeps it a no-op under SSR/tests.

---

KallynGowdy:
Addressed in dc1fc657:

- **M1** — `deleteReadingPlan` now only fires `reading_plan_deleted` when `plan.status === "complete"`. A never-finished draft discard (`discardEditingReadingPlan` → `deleteReadingPlan`) goes through the same code path but never fired `reading_plan_created`, so it no longer fires the delete event either. Added a test that reproduces the exact repro steps from the review (start a draft, add a reading, let the debounced autosave persist it, discard) and asserts no `reading_plan_deleted`.
- **L1** — Extended the `markDayComplete` test to call it a second time on an already-complete day and assert `reading_plan_day_finished` is not re-captured, so the idempotency guard's main case is now covered.
- **L2** — Same test now also asserts the exact count of `reading_plan_session_finished` calls fired alongside the day event (one per session on the day, plus the one day event) — locking in that a day-complete tap fires N+1 events.
- **L3** — Added a test for the self-paced `startReadingPlan` branch, asserting `cadenceId: null` in the captured `reading_plan_started` properties.

Verified the new `deleteReadingPlan` test fails on the pre-fix code (reverted the `status` gate, confirmed red, restored it) before pushing. Full suite (4839 tests), typecheck, and lint are all green.

---

PR #1651, opened 2026-08-17
Flip client bootstrap to hydrate(), with a readiness gate and SW re-keying

KallynGowdy:
## Summary

Stacked on #1617 (needs `AppConfig.renderedAsWebKit`, `FreeUseBibleAPI.snapshotResponseCache`, and the `entry-ssr.tsx` substitution mechanism from that PR) — this is Stage 3 of the hydration-support plan: the actual `render()` → `hydrate()` flip, plus everything that had to be true first for it to be safe.

Today the client always calls Preact's `render()` (full rebuild), discarding the server-rendered DOM on every load — a visible flash/reflow the SSR investment never actually paid for. `hydrate()` fixes that, but only if the client's first render produces the *same* tree the server produced, and Preact's `hydrate()` does not diff attributes on existing DOM — a mismatch silently stays wrong instead of self-correcting.

## What changed

- **Hydration-readiness gate** (`app/hydrationGate.ts`): before hydrating, the client verifies the served HTML can be trusted — the URL matches what the server actually rendered for (`AppConfig.renderedForPath`), the SSR chapter-load timeout didn't fire (`AppConfig.ssrChapterContentSettled`, backed by a new `initialChapterLoadTimedOut` signal in `BibleReadingManager.tsx`), and the container has real content. Any failure falls back to the existing `render()` path. `app/init.tsx` now also waits on every initial tab's `chapterDataPromise` (already in flight — the same promise SSR suspends on) alongside `i18n.ready` before deciding.
- **Viewport**: seeded from the same `renderedAsMobile` guess on both server and client (never `window.innerWidth` up front); a one-time post-mount effect (`app.applyViewport()`) then corrects it to the real size.
- **Settings** (font size, UI size, toolbar, `disablePanels`, theme id): `login.localConfig` now seeds empty, matching SSR's lack of `localStorage`, instead of reading real storage synchronously before first paint; `login.hydrateLocalConfig()` applies the real value once, post-mount.
- **Theme**: moved entirely out of the Preact tree. `ThemeManager` now writes its CSS directly to a `#sb-theme-styles` tag in `document.head` from a plain signals effect, so it's never part of the diffed vnode tree. A pre-hydration inline script in `index.html` reads `localStorage` directly and patches that tag before any JS bundle loads, closing the remaining flash window. Custom theme/highlight override values (unvalidated free text) are defensively stripped of `<` before being spliced into the page server-side.
- **Service worker**: the HTML cache was keyed by one shared entry for every URL, explicitly documented as safe only because the client always rebuilt from scratch. Re-keyed per URL (Workbox's default), with a bounded `ExpirationPlugin`; an absent or stale cache entry now degrades to `render()` via the same gate rather than hydrating the wrong page.

## Known, pre-existing gaps surfaced but not fixed here

Found while building the regression test below — neither is introduced by this PR, and fixing either is a separate follow-up:

- The anonymous "Guest" avatar's icon/color is derived from a per-instance random connection id (`OsManager.tsx`'s `uuid()`), so it never matches between server and client for a signed-out visitor.
- `renderToStringAsync` only re-renders the specific subtree that threw on suspension — sibling content reacting to the same resolved promise (the first-run tutorial prompt) can lag behind in the SSR output.

Both are documented in detail in `test/unit/standalone/hydration.test.tsx`.

## Test plan

- New `test/unit/standalone/hydration.test.tsx`: runs a real SSR render, loads it into jsdom, and hydrates against it, asserting the DOM is unchanged (Preact does not warn on a hydration mismatch, so this is the actual safety net) — plus explicit tests for each gate-decline reason.
- Updated existing tests wherever the intentional seed/read timing changes (viewport, `localConfig`) required it.
- `pnpm check:ts`, `pnpm lint`, `pnpm test` all pass (218 files / 4660 tests).
- Manual (recommended before merge): `pnpm dev` with network throttled — confirm no flash/reflow on load, no chapter re-fetch after boot, no light↔dark flash on a stored dark theme; install the service worker, go offline, load a previously-visited URL (hydrates from cache) and a fresh one (falls back to `render()` cleanly).

---
_Generated by [Claude Code](https://claude.ai/code/session_01T2HcadY9uu84Lh5VSVLZ57)_


Blitzy:
> [!NOTE]

## Code review — hydrate() flip, readiness gate & SW re-keying

Reviewed the diff of this PR against its base (`claude/app-hydration-support-urpbjo`). This is a careful, unusually well-documented change, and the mechanics of the gate, the per-URL SW re-key, and the seed-then-correct pattern for viewport/settings all hold up. The `chapterDataPromise` you wait on in `init.tsx` is resolve-only and always settles (success **and** error both flip `initialChapterLoadSettled`), so there's no rejection-hang there — good.

I did find one issue that undermines a goal this PR explicitly sets out to hit, plus a couple of narrower ones. Listed by severity.

---

### 🔴 High — the theme write on boot re-introduces the light↔dark flash this PR is trying to kill

**What happens.** For a returning visitor whose saved theme is **dark**, the page bounces **dark → light → dark** during boot:

1. Server renders `#sb-theme-styles` with the default (light) guess.
2. The pre-hydration inline script in `index.html` reads `localStorage` and rewrites the tag to **dark** before any bundle loads. ✅ Page is dark.
3. The JS bundle loads. `init.tsx` calls `createSeedBibleState()` **synchronously**, which calls `createTheme()`, whose new `effect()` (`ThemeManager.tsx:1116-1128`) runs **immediately** and writes `themeStyleText.value` to the same tag. But `login.localConfig` was seeded **empty** (`LoginManager.tsx:463`), so `settings.themeId` is the default `"light"` (`SettingsManager.tsx:357`) → the effect writes **light**, clobbering the dark value the inline script just set. ❌ Page flashes to light.
4. Post-mount, `MainBody`'s `useEffect(() => state.login.hydrateLocalConfig())` finally brings in the real `themeId="dark"`, the computed theme recomputes, and the effect re-writes **dark**. ✅ Page is dark again.

The flash window is the `Promise.all([i18n.ready, chapterDataPromises])` await in `init.tsx` — precisely the window that widens under the throttled-network conditions your own "Manual" test plan calls for. Under throttling this is clearly visible.

**Why the code comment doesn't save it.** `ThemeManager.tsx:1107-1110` argues the effect "needs no deferral" because `document.head` isn't diffed by Preact. That's true for *hydration mismatch* but conflates it with *flash*: the write **target** needs no deferral, but the **value** it writes is derived from `themeId`, which *is* deferred (empty `localConfig`). So "whichever runs last simply wins" (line 1114) is the bug — the middle write (light, step 3) is wrong and paints before the last write (step 4) corrects it.

Custom (non-built-in) themes are worse off: the inline script's presets map only contains `light`/`dark` (`THEME_PRESET_STYLE_TEXT`), so step 2 is a no-op for them, and they show default-light for the whole boot until step 4.

**Reproduction.**
1. `pnpm dev`, load the app, switch to the dark theme (persists to `localStorage` under `sb-profile-config-local`).
2. Open DevTools → Network, throttle to "Slow 3G".
3. Reload. Watch `#sb-theme-styles` in the Elements panel (or just the page): dark on first paint → flips to light once the main bundle evaluates → flips back to dark a beat later once the mount effect runs.

**Suggested fixes** (any one):
- Guard the `createTheme` effect so its *first* client-side run doesn't overwrite an already-populated `#sb-theme-styles` tag with a value derived from the default seed — i.e. skip the initial write and let it take over only from the first *change* after `hydrateLocalConfig`.
- Or, since theme output lives entirely outside the diffed tree, exempt the theme-relevant keys (`themeId`, `customTheme`, `customHighlights`) from the empty-seed rule and read them synchronously on the client, keeping only the hydration-sensitive keys (fontSize, uiSize, disablePanels, toolbar) deferred.
- Or have the effect read the current tag's text as its initial value on the client and only write once the computed value actually diverges post-hydrate.

---

### 🟠 Medium — the whole app mount is now gated on every initial chapter load, with no client-side timeout

`init.tsx` previously mounted on `i18n.ready` alone; it now also awaits `waitForInitialChapterLoads()` (every initial tab's `chapterDataPromise`). On the client there is deliberately **no** timeout arming `initialChapterLoadSettled` (`BibleReadingManager.tsx:1264-1274` — SSR-only). So a genuinely hung initial chapter fetch (connection present but not completing) now blocks `hydrate()`/`render()` from ever running — meaning no interactivity at all (sidebar, menus, tab switching), where before the app shell mounted and only the reader subtree suspended behind its own Suspense fallback.

In the hydrate happy-path the SSR DOM stays on screen so it isn't a blank page, but the app is inert until the slowest initial chapter settles.

**Suggested fix.** Bound the mount wait with a generous client timeout (e.g. race `waitForInitialChapterLoads()` against a few-second timer). If it fires, fall back to `render()` (not `hydrate()`) rather than waiting indefinitely — that preserves the "first client render must match SSR" invariant while guaranteeing the app becomes interactive.

---

### 🟠 Medium — `warmAppShellCache()` won't warm the current page on a first install

`sw.ts:142` calls `self.clients.matchAll({ type: "window" })` without `includeUncontrolled: true`. `matchAll` returns only clients the worker *controls*, and during the `install` event the newly-installing worker controls nothing (control comes later, at activate/`clientsClaim()`). So on a **first install** — the exact "user installs the worker while reading `/AAB/genesis/10`" case the new comment (`sw.ts:134-138`) says it's fixing — `clients` is empty and it falls back to warming `/`.

This matters more than it looks: with per-URL keying, the current page's HTML was already fetched *before* the worker took control, so it isn't in `HTML_CACHE` yet either. The warm was the one chance to cache the current reading position for an offline-after-single-visit launch, and it misses it.

**Reproduction.**
1. Fresh profile (no existing SW). Load `/<translation>/genesis/10`.
2. After the SW installs, go offline and relaunch that exact URL → no cached shell for it (only `/` was warmed).

**Suggested fix.** `self.clients.matchAll({ type: "window", includeUncontrolled: true })`.

---

### 🟡 Low — test gap: nothing covers the dark-theme flash from finding No. 1

`hydration.test.tsx` is a strong safety net, but it (a) `localStorage.clear()`s so it only ever exercises the default light theme, and (b) compares only `#app` `innerHTML`, never the `#sb-theme-styles` tag in `<head>`. So the regression in finding No. 1 is entirely outside its coverage. Worth a focused test: with a saved dark `themeId` in `localStorage`, assert that after `createSeedBibleState()` runs (pre-mount) the `#sb-theme-styles` tag is **not** clobbered back to the light default. Reverting the fix should make it fail.

### 🟡 Low (nit) — `decideHydration` collapses two causes into one reason

`hydrationGate.ts` returns `reason: "no-ssr-content"` both for a missing container child *and* for a missing `renderedForPath`. The `console.warn` in `init.tsx` will then say "no-ssr-content" for a config that did have SSR content but no render provenance, which is misleading when debugging from logs. A distinct reason (e.g. `"no-render-provenance"`) would make the fallback path self-explanatory.

---

**Test-quality note (positive):** the normalization in `hydration.test.tsx` errs toward *false failure*, not false pass — the avatar/tutorial-prompt regexes assert on server↔client *differences*, and if they ever stop matching the exact markup the comparison fails rather than silently passing. The three normalized divergences are genuine pre-existing artifacts and are documented honestly. No "always-green" tests spotted in this diff.

---

KallynGowdy:
Still needs additional work:
- Hydration failure recovery (if hydration errors, then recover with full re-render)
- Next/previous chapter buttons don't work when hydrating stored tabs
- Needs claude code review hints for hydration mismatch issues
- Needs linting checks for hydration mismatch issues (loading data from local storage/indexeddb)

Blitzy:
> [!NOTE]

## Code review — re-check after the `ready_for_review` flip

This PR has grown substantially since the review posted here on 2026-08-17 (13 more commits: tab/chapter-fetch hydration-mismatch fixes, a hydration-error fallback to `render()`, disabling next/previous chapter buttons while hydrating, CSS inlining changes, and a new ESLint rule for localStorage/IndexedDB hydration hazards). I re-read the current head (`8c49f1b`) against that earlier feedback rather than re-deriving everything from scratch, since the maintainer's own comment on this PR ("Still needs additional work…") lists several of the same gaps.

Of the four issues the prior review raised, **none appear to have been addressed** by the follow-up commits — the follow-ups fixed different hydration problems (stored tabs, failed chapter fetches, next/prev buttons) but not these. Re-verified directly against the current code below.

---

### 🔴 High — dark‑theme flash on boot is still present (unresolved from the earlier review)

`ThemeManager.tsx` (`createTheme`, around line 1117) still writes the computed theme to `#sb-theme-styles` in a plain `effect()` that fires synchronously and unconditionally the moment `createSeedBibleState()` runs — before `login.hydrateLocalConfig()` has restored the real `themeId` from `localStorage` (`LoginManager.tsx` line 474 still seeds `localConfig` as `{}`).

For a returning visitor whose saved theme is dark:
1. Server renders the light default into `#sb-theme-styles`.
2. The pre-hydration inline script in `index.html` reads `localStorage` and patches the tag to dark before the bundle loads. ✅ dark.
3. The bundle loads; `createTheme()`'s effect fires immediately using the still-empty `localConfig` → `themeId` is the light default → it overwrites the tag back to light. ❌ flash.
4. Post-mount, `hydrateLocalConfig()` applies the real `themeId=dark`, the computed theme recomputes, and the effect writes dark again. ✅ dark.

**Reproduction:**
1. `pnpm dev`, switch the app to the dark theme (persists to `localStorage`).
2. DevTools → Network → throttle to "Slow 3G".
3. Reload and watch `#sb-theme-styles` (or just the page): dark → flips to light once the bundle evaluates → flips back to dark a beat later.

Custom (non-preset) themes are worse: the inline script's preset map only knows `light`/`dark`, so step 2 is a no-op for them and they show the light default for the entire boot window.

**Suggested fix (any one):**
- Skip the effect's first client-side run so it doesn't overwrite an already-populated `#sb-theme-styles` tag with a value derived from the still-empty seed — only take over from the first *change* after `hydrateLocalConfig()` runs.
- Or exempt `themeId`/`customTheme`/`customHighlights` from the empty-seed rule (they never enter the diffed Preact tree, so there's no hydration-mismatch risk in reading them from `localStorage` synchronously) and keep only the truly hydration-sensitive keys deferred.

---

### 🟠 Medium — mount is gated on chapter loads with no client-side timeout

`app/init.tsx` (`waitForInitialChapterLoads`, lines 38-44) awaits every initial tab's `chapterDataPromise` before `hydrate()`/`render()` is ever called, and unlike the SSR path there is no client-side timer arming `initialChapterLoadTimedOut` — that signal is SSR-only per `BibleReadingManager.tsx`. A chapter fetch that genuinely stalls (connection open, response never completes) blocks the entire app mount indefinitely — no sidebar, no menu, no tab switching — rather than the pre-PR behavior where the shell mounted and only the reader subtree suspended.

The new hydration-error fallback (commit `8520171`) doesn't help here: it only fires if `hydrate()` throws, not if the `Promise.all(...).then(...)` callback never runs at all.

**Suggested fix:** race `waitForInitialChapterLoads()` against a few-second client timer; on timeout, fall back to `render()` (never `hydrate()`, to preserve the "first client render must match SSR" invariant) so the app becomes interactive either way.

---

### 🟠 Medium — `warmAppShellCache()` still won't warm the current page on first install

`standalone/sw.ts` line 142 calls `self.clients.matchAll({ type: "window" })` without `includeUncontrolled: true`. During the `install` event a newly-installing worker doesn't yet control any clients (control comes later, at activate), so `matchAll` returns an empty array and the code falls back to warming `/` — even though the comment directly above it (lines 134-138) says the whole point of this function is to warm the page the user is actually on (e.g. `/AAB/genesis/10`) on first install.

**Reproduction:**
1. Fresh profile, no existing service worker. Load `/AAB/genesis/10`.
2. Let the SW install, then go offline and reload that same URL → no cached shell for it (only `/` was warmed).

**Suggested fix:** `self.clients.matchAll({ type: "window", includeUncontrolled: true })`.

---

### 🟡 Low — stale comment risks a future regression in `LoginManager.tsx`

Line 494's comment says "`localConfig` was just seeded from `readLocalConfig()` above," but line 474 seeds it with an empty object (`signal<Record<string, unknown>>({})`) — the `readLocalConfig()` call happens later, inside `hydrateLocalConfig()`. A maintainer reading only the comment could reasonably conclude the `isFirstLocalConfigWrite` skip-guard is just avoiding a redundant re-serialization of data that was "just read," and remove it. Removing it would make the effect's first run persist `{}` to `localStorage` before `hydrateLocalConfig()` ever runs, wiping every returning visitor's saved settings on page load.

**Suggested fix:** update the comment to match the actual seed (empty object, not a `readLocalConfig()` read), so the guard's purpose (skip the write caused by the *empty* seed, not a redundant echo of a real read) is clear on its own.

---

### 🟡 Low — no regression test covers the theme-flash issue

`test/unit/standalone/hydration.test.tsx` only checks that `#app` markup matches between SSR and hydrate; it never asserts on `#sb-theme-styles`, and doesn't exercise a non-default (dark) saved theme. None of the tests added across this PR's 13 follow-up commits cover the flash in finding No. 1 either — worth a focused test asserting that with a saved dark `themeId` in `localStorage`, `#sb-theme-styles` is never observed with the light value between mount and `hydrateLocalConfig()` completing. Reverting the eventual fix should make it fail.

---

**Positive notes:** the hydration-mismatch fixes added since the last review (saved-tabs handling, failed chapter fetch, disabled next/prev buttons while hydrating, the hydration-error → full-render fallback, and the new ESLint rule for localStorage/IndexedDB access) are solid, targeted fixes for real gaps, and the new tests for each genuinely exercise the DOM rather than mocks — I didn't find any always-passing tests in this diff.


---

PR #1653, opened 2026-08-17
Cache translations list across SSR requests

KallynGowdy:
## Summary

Adds a shared, TTL-based cache for the translations list (`available_translations.json`) in the SSR server process. This prevents redundant network fetches when multiple HTTP requests hit the same Bible API endpoint within a short window — a common scenario in a long-running server.

## Changes

- **New `TranslationsCache` interface** in `BibleDataManager.tsx`: An opaque cache contract (get/set/delete) that `getTranslations()` can use to avoid re-fetching. Only the SSR host supplies one; client behavior (per-page-load cache + localStorage) is unchanged.

- **New `ssrTranslationsCache` module** (`standalone/ssrTranslationsCache.ts`): A module-level singleton that implements `TranslationsCache` with a 1-hour TTL (configurable via `SSR_TRANSLATIONS_CACHE_TTL_MS` env var). Entries expire automatically; failed fetches are not cached (they're evicted immediately so the next call retries).

- **Updated `getTranslations()` logic** in `BibleDataManager.tsx`:
  - Checks the cache before fetching; if present and not expired, returns the cached promise.
  - On cache miss, starts a fetch and stores the promise immediately (so concurrent calls share the in-flight request).
  - Respects `refresh: true` by evicting the cache entry first.
  - Does not cache rejected promises — they're deleted so retries can succeed.

- **Wired cache into SSR entry point** (`standalone/entry-ssr.tsx`): Passes `ssrTranslationsCache` to `createSeedBibleState()`, which forwards it to `createBibleDataManager()`.

- **Comprehensive test coverage**:
  - Unit tests in `BibleDataManager.test.ts` verify cache sharing, in-flight deduplication, refresh bypass, and failed-fetch retry behavior.
  - New `ssrTranslationsCache.test.ts` tests TTL expiry, per-endpoint keying, and the reset seam for test isolation.
  - Regression test in `entry-ssr.test.ts` confirms two SSR renders sharing an endpoint fetch `available_translations.json` only once.

## Implementation Details

- The cache stores promises, not resolved values. This allows concurrent callers to await the same in-flight fetch, eliminating the "thundering herd" problem when multiple requests race on a cache miss.
- Failed fetches are immediately evicted from the cache (via `.catch()` handler), mirroring `FreeUseBibleAPI`'s own cache behavior. This ensures transient network errors don't block retries for the entire TTL.
- The `refresh: true` option (used by `getTranslations(undefined, { refresh: true })`) deletes the cache entry before fetching, forcing a fresh network call.
- Test isolation is maintained via `resetSsrTranslationsCacheForTests()`, which clears the module-level cache between tests.

https://claude.ai/code/session_019U3qc3eGNmiDSbbwFTgVjN


Blitzy:
> [!NOTE]

## Code Review — Cache translations list across SSR requests

**Overall:** This is a clean, well-documented change that solves a real problem (every SSR request re-fetching `available_translations.json`). The design is sensible — caching promises rather than resolved values to dedupe in-flight requests, evicting failed fetches, and confining the cache to the SSR host via an opaque interface so client behavior is untouched. The tests are genuine rather than always-green (see the Test Quality note below). I found no critical or high-severity issues. There are two correctness edge cases worth addressing and a couple of minor notes.

No linked issue is referenced in the PR description, so this review is against the stated summary and the code itself.

---

### Medium — Stale `.catch` handler can evict a valid, freshly-cached entry

**Where:** `packages/seed-bible/seed-bible/managers/BibleDataManager.tsx`, `getTranslations()` (the `resultPromise.catch(...)` at ~line 948).

**What:** The rejection handler deletes the endpoint key unconditionally:

```ts
resultPromise.catch(() => translationsCache?.delete(normalizedEndpoint));
```

If a slow fetch is still in flight when its entry gets replaced — either by a `refresh: true` call or by TTL expiry followed by a fresh fetch — then when the *old* promise finally rejects, its `.catch` deletes the endpoint key and evicts the *new, valid* entry that replaced it. The handler has no way to tell "the entry I stored" from "whatever is stored now."

**Impact:** Limited but real. It does not serve wrong data — it causes a spurious cache eviction, so the next request re-fetches instead of hitting the cache. It also won't trigger on the common SSR render path (which never passes `refresh`), but it is reachable via the offline update check's `refreshTranslations(endpoint, { refresh: true })`, and via a fetch that hangs longer than the TTL and then fails.

**Reproduction (confirmed locally):** With a shared cache, start call A whose fetch hangs; issue a `refresh: true` call B that succeeds and stores its promise; then reject A. After A rejects, the cache entry (B's valid promise) is gone. I verified this with a throwaway test against the real `createBibleDataManager` — the entry logs as `GONE (bug)` after A rejects.

**Suggested fix:** Only delete if the currently-stored promise is still the one that failed:

```ts
resultPromise.catch(() => {
  if (translationsCache?.get(normalizedEndpoint) === resultPromise) {
    translationsCache.delete(normalizedEndpoint);
  }
});
```

Note `get()` also applies TTL expiry, so an already-expired entry returns `undefined` and the guard correctly skips the delete.

---

### Low — `SSR_TRANSLATIONS_CACHE_TTL_MS` parsing has two silent failure modes

**Where:** `standalone/ssrTranslationsCache.ts`, line 12.

```ts
const TTL_MS = Number(process.env.SSR_TRANSLATIONS_CACHE_TTL_MS ?? 60 * 60_000);
```

Two operator-error cases fail silently (both confirmed with `node -e`):

- **Empty string** (`SSR_TRANSLATIONS_CACHE_TTL_MS=`) → `Number("")` is `0` → every entry expires the instant it is written, so the cache is silently disabled. Easy to hit if the var is declared but left blank in an env file.
- **Non-numeric string** (e.g. `"1h"`) → `NaN` → `expires = Date.now() + NaN = NaN`, and `NaN <= Date.now()` is always `false`, so entries **never expire** and are cached until the process restarts — the opposite of a safe default.

**Suggested fix:** Validate and fall back on a non-positive or non-finite value:

```ts
const parsed = Number(process.env.SSR_TRANSLATIONS_CACHE_TTL_MS);
const TTL_MS = Number.isFinite(parsed) && parsed > 0 ? parsed : 60 * 60_000;
```

(Deciding whether `0` should mean "disabled" vs "use default" is a judgment call — if "disabled" is a wanted feature, it's worth handling explicitly rather than as an accident of `Number("")`.)

---

### Nits (non-blocking)

- **Idle endpoints are never evicted.** Entries are only dropped lazily on `get()` for that same key (or via `delete`). An endpoint that stops being requested keeps its entry until process restart. This is bounded by the small number of distinct endpoints, so memory impact is negligible — noting it only for completeness.
- **Shared array reference across concurrent renders.** On a cache hit, every caller receives the same `Translation[]` instance. This is safe today because `mergeTranslations` only reads it and nothing mutates individual `Translation` objects during render. It would become a cross-request contamination bug if any future SSR code mutates a translation in place — a defensive copy or a comment pinning the read-only invariant would guard against that.

---

### Test Quality

The tests exercise real behavior rather than passing unconditionally:

- The `entry-ssr` regression test (`shares one available_translations.json fetch across multiple SSR renders`) genuinely fails on pre-fix code — I removed the `translationsCache` wiring locally and it went red (`expected [...] to have a length of 1 but got 2`), then passed again once restored.
- Failed-fetch retry, `refresh: true` eviction, and in-flight dedup are each covered with observable assertions (network call counts, updated `sha256`), not internals.
- `ssrTranslationsCache.test.ts` uses fake timers to assert the TTL boundary (just-under vs just-over), per-endpoint keying, and the reset seam.

Two additions worth considering:

1. A regression test for the Medium finding above (slow-reject-after-replace evicting a valid entry) — it would fail on the current code and pass after the identity-check fix.
2. The TTL-boundary test hardcodes `60 * 60_000`; if the default ever changes, that assertion breaks in a way that reads as a cache bug rather than a config change. Deriving the expected boundary from a shared constant would make the intent clearer.

---

KallynGowdy:
Thanks for the thorough review — addressed in 673c9201:

- **Medium (stale `.catch` eviction):** fixed. `getTranslations()`'s rejection handler now only evicts the cache entry if it's still the promise on record (`translationsCache?.get(normalizedEndpoint) === failedPromise`), so a superseded request rejecting later can't wipe out a fresher entry. Added a regression test reproducing your exact repro (hanging call A, superseding `refresh: true` call B, then A rejects) — confirmed it fails on the prior code and passes now.
- **Low (TTL parsing):** fixed. `TTL_MS` now falls back to the 1-hour default unless the env var parses to a positive, finite number, so an empty string or non-numeric value no longer silently disables or permanently defeats the TTL. Added tests covering unset/empty/non-numeric/negative/valid-override cases.
- **Nit (shared array reference):** added a doc comment on `ssrTranslationsCache` pinning the read-only invariant, so a future change that mutates a `Translation` in place has a pointer to what would break.
- **Nit (idle endpoints never evicted):** left as-is per your note — bounded by the small, fixed number of distinct endpoints, so not worth the complexity.
- **Test quality suggestions:** both taken — added the eviction regression test, and the TTL-boundary test now derives its expected value from the exported `TTL_MS` constant instead of a hardcoded literal.

---

PR #1654, opened 2026-08-18
Add "/{lang}/about" page with SEO meta tags and sitemap entries

### LINKED ISSUE #1648: Add a "About Seed Bible" page
Currently, there's no way for users or search engines to discover information about the Seed Bible beyond what's in tutorials or what they've been told by friends. We should add an "About the Seed Bible" page that gives a basic primer on what the Seed Bible is, why we're building it, and what it can do.

Ideally, this will help us place better in search results for "Seed Bible".

### implementation notes

- the page should be at /{lang}/about
- the page should be included in each languages sitemap
- it should have proper canonical URL, title, description, open graph, and twitter meta tags.

KallynGowdy:
The app has no router — every URL is either a reading position or a
legacy shape that redirects into one — so a static informational page
needed its own small parsing/guard plumbing: a new StaticPagePath
parser recognizes "/{lang}/about" without colliding with the reading
URL scheme, I18nManager resolves its language segment, and
TabsManager's URL-commit effect now skips static pages so it doesn't
overwrite the address bar with the underlying reading tab's position
on mount or language change.

SeedBibleStateManager's title/description/socialTitle/canonicalUrl
signals branch on a new isAboutPage computed, so entry-ssr's existing
meta-tag block (canonical URL, Open Graph, Twitter card) picks up
About-specific content with no changes of its own. The page itself
renders in place of the reader chrome and pulls its copy from new
translatable en.json keys. The sitemap generator gains a dedicated
pages.xml child sitemap with one "/{lang}/about" URL per supported UI
language.

Closes #1648.


Blitzy:
> [!NOTE]

## Code review — `/{lang}/about` page (closes #1648)

Nice, tidy change. The `StaticPagePath` parser is kept deliberately separate from the reading-URL scheme, the meta signals all branch off a single `isAboutPage` source of truth, and the sitemap/canonical share one builder so they can't drift. I verified the important guardrails hold: `/en/about` is **not** grabbed by `legacyReadingUrlRedirect`/`acceptLanguageRedirect` (the numeric-chapter check keeps the 2-segment About shape from colliding with the legacy `/{book}/{chapter}` shape), and it renders 200 rather than 404. I also confirmed the `TabsManager` regression test is genuine — with the new guard disabled, the URL is clobbered to `/en/AAB/genesis/1` on mount, so the test goes red as it should. `LANG_META`'s keys and the locale JSON files match exactly (77 each), so every sitemap About URL is a real supported language. Full changed-file test suite passes (435 tests) and the `build` check is green.

One real issue worth addressing before merge, plus some minor notes.

### Medium

**No. 1 — An unsupported/garbage language in `/{lang}/about` returns HTTP 200 with a self-referential canonical (soft-404 / duplicate content).**

`parseStaticPagePath` (`managers/StaticPagePath.ts`) accepts *any* string as the language segment — it only validates the page slug (`about`), never the language. So `/x-hacker/about`, `/foo/about`, `/luke/about`, etc. all satisfy `isAboutPage`, render the full About page, and return 200.

The problem is the canonical. `SeedBibleStateManager`'s `canonicalUrl` carries this comment:

> Uses the *resolved* i18n language, not the raw URL segment, so a garbage/unsupported language in the URL still canonicalizes to a real page instead of echoing back something that doesn't exist.

That isn't what happens. i18next has no `supportedLngs` configured, so `i18n.language` keeps whatever was requested (only *resource lookup* falls back to `en`). I confirmed this empirically — for `/x-hacker/about`:

```
isAboutPage = true
i18n.language = "x-hacker"
canonicalUrl  = "/x-hacker/about"   ← self-referential, not "/en/about"
```

So every `/{anything}/about` is an indexable page whose `rel=canonical` points at itself. For a feature whose whole purpose is SEO, that's the exact "soft 404 / duplicate content" pattern the codebase already goes out of its way to avoid for reading URLs — `entry-ssr.render()` returns `notFound: true` (a real 404) when a book segment doesn't resolve (`bookMatch === "unresolved"`). The About page has no equivalent guard for an unresolved *language*.

_Reproduction (human):_
1. `pnpm dev` (or hit the branch deployment).
2. Request `/x-hacker/about` (any non-language first segment).
3. Response is **200**; view-source shows `<link rel="canonical" href="/x-hacker/about">` and `<title>About the Seed Bible | Seed Bible</title>`.
4. Contrast with an unresolved reading path like `/notabook/5`, which correctly comes back as a 404 (`notFound`).

_Suggested fix (pick one):_
- **Preferred, consistent with the reading scheme:** treat an About URL whose language segment isn't a supported UI language as `notFound` in `entry-ssr.render()`, mirroring the `bookMatch === "unresolved"` branch. `StaticPagePath.ts` shouldn't pull in the i18n module, so either validate against `LANG_META`'s keys (already imported in the sitemap script) or have `parseStaticPagePath` take an `isSupportedLanguage` predicate and return `null` for unsupported languages.
- **Minimal, makes the comment true:** resolve the language to a supported one (unsupported → `DEFAULT_UI_LANGUAGE`) before building the canonical, so `/x-hacker/about` at least canonicalizes to `/en/about`. This still serves 200 at the junk URL, so it's weaker than the 404 option.

### Low / minor

**No. 2 — No test covers the unsupported-language case.** The unit tests only exercise the happy `/en/about` path — which is exactly why No. 1 slipped through. Whatever fix you choose for No. 1, please add a test asserting the chosen behavior (404, or canonical → `/en/about`) for a URL like `/x-hacker/about`. A regression test here should fail on the current code.

**No. 3 — The reader↔AboutPage swap in `app/main.tsx` isn't directly asserted.** `isAboutPage` (state) and the SSR meta output are both tested, but the `main.tsx` ternary that renders `<AboutPage>` in place of the reader chrome has no test. It's trivial, so this is just a coverage note, not a blocker.

**No. 4 — Redundant `role="main"` on `<main class="sb-about-page">`** (`components/AboutPage/AboutPage.tsx`). `<main>` already carries the `main` landmark role implicitly, and the reader's own `<main class="sb-main-content">` doesn't set it — so this is both redundant and inconsistent. Some `jsx-a11y` configs flag it (`no-redundant-roles`). Safe to drop.

**No. 5 — `canonicalUrl` / `og:url` are root-relative (`/en/about`), not absolute.** This matches the existing reading-page behavior, so it's pre-existing rather than introduced here — but it's worth flagging on an SEO page, since some crawlers/scrapers resolve `og:url` less reliably than a browser would (the meta block in `entry-ssr.tsx` already notes `og:image` is effectively unresolvable for the same root-relative reason). Not blocking; consider making the canonical absolute in a follow-up.

**No. 6 (optional) — No `hreflang` alternates.** The issue didn't ask for them, but since you now emit one About URL per supported UI language in the sitemap, `rel="alternate" hreflang="…"` links tying those per-language variants together would be the natural SEO complement. Worth a follow-up, not this PR. (Related aside: `iw` and `he` are both supported and both render Hebrew, so they'll produce two self-canonical Hebrew About URLs — a pre-existing property of the language list, not something this PR introduces.)

---

Overall this is close. No. 1 is the one I'd want resolved (or explicitly deferred with a comment) before merge, since it undercuts the SEO goal the page exists for; the rest are minor.

---

Blitzy:
This seems to work as described. I did find some issues that should probably be addressed.

**No 1. — The about page is not localized to the language its being accessed from.**

The about page always appears in English even if you access it from another locale (ie `/fr/about/` or `/ar/about`).
Ideally the about page should be localized to all of our 76 other supported languages.

_About page viewed via `/ar/about` path:_
image


**No. 2 — There is no way to access the about page from the user interface.**

The only way to reach the about page is to go there directly via the url `{locale}/about`. I think we should put a way to access it via the sidebar. The about page is great for SEO but its also nice for people too as it explains what the Seed Bible is, who we are, and even gives a direct link to CHANGELOG notes.

**No. 3 — The CHANGELOG link should utilize the Git SHA that the web app was built on.**

Currently the CHANGELOG link is hardcoded to the `main` branch. I think it should link to the CHANGELOG.md of the commit that the web app was actually built on? This is kind of a nitpick but I think it makes sense?

An alternative option would be to instead of linking to CHANGELOG.md (the file), we could link to the Releases page which is a little more public facing by nature. It shows the same information regardless but lives outside the source code itself. Not really sure what would be "better" here?

KallynGowdy:
@Blitzy 
1. It is already set for localization, we just need to run Google translate on it.
2. That is a good idea. I'm planning on adding a "info" icon to the sidebar above the current settings button for desktop, and a button on the settings screen for mobile that lets users get into the about page.
3. Yeah that should be possible

Blitzy (approved):

Issue #263, opened 2025-10-23
Tabernacle stuff


craigbbradley:
[It used to work but  it don't now](https://www.loom.com/share/c7ff9350ebb046849d729c067acc1205)

- [x] Remove outlines
- [x] Make the complete tabernacle show up across every chapter it is mentioned
- [x] Implement camera re-center feature

Note: Bug in issue #335 must be fixed to actually be able to work on this

craigbbradley:
The tabernacle and the scripture should be able to interact with each other bidirectionally:


Interacting with a piece should highlight the verse(s) where that piece is mentioned (if any).

Interacting with a verse that mentions a piece should highlight the piece (if visible).
Check [this video](

https://www.loom.com/share/0821350b36e74261ae7165a5ca988340?t=461&sid=71fd6d77-d6cd-4376-95ed-5961ef84c433

) for reference on how this should work.

soukigabriel:
@craigbbradley adding the #131 as a sub-issue for better reference

Issue #131, opened 2025-10-04
Responsive tabernacle pieces


soukigabriel:
The tabernacle and the scripture should be able to interact with each other bidirectionally:

- [x] Interacting with a piece should highlight the verse(s) where that piece is mentioned (if any).
- [x] Interacting with a verse that mentions a piece should highlight the piece (if visible).

Check [this video](https://www.loom.com/share/0821350b36e74261ae7165a5ca988340?t=461&sid=71fd6d77-d6cd-4376-95ed-5961ef84c433) for reference on how this should work.

PR #1661, opened 2026-08-19
Refactor/tabernacle

soukigabriel:
Old Tabernacle extension refactored to the new house-of-the-lord extension, working with the new seed-bible architecture.


Blitzy:
> [!NOTE]

## Code review — `refactor/tabernacle` → `develop`

**What this PR does:** ports the old Tabernacle extension onto the new Seed Bible architecture as `house-of-the-lord`, restructured into a hexagonal (domain / application / infrastructure) layout. 179 files, ~7.8k insertions, essentially all new code. `build (24.x)` is green.

**Short version:** the architecture and typing are clean and consistent, but a few pieces of the *interaction* that the linked issues describe were left unwired during the port, and there are no tests. The most important items are the two empty click handlers and the missing test coverage, because they mean the shipped behavior is narrower than what the old extension delivered.

**Context from linked issues.** This area is covered by [#131](https://github.com/HelloAOLab/seed-bible/issues/131) ("Responsive tabernacle pieces" — *bidirectional*: clicking a piece highlights its verses, clicking a verse highlights the piece) and [#263](https://github.com/HelloAOLab/seed-bible/issues/263) (remove outlines, whole tabernacle across chapters, camera re-center). Both were closed as completed against the old extension, so this refactor should preserve that behavior. Several findings below are about behavior from those issues that is disabled in the new code.

---

### 🔴 High

**H1 — Verse options in the piece context menu do nothing.**
`infrastructure/adapters/pieces/ContextMenuRendererAdapter.tsx:145-149` and `:177-181`. When you click a piece, the menu of referencing verses is built correctly, but each verse option's `onClick` is an empty body with a `// TODO: wire this verse-menu option to the controller` comment. So the piece→verse navigation that issue #131 calls for is inert — the menu renders, clicking a verse does nothing.
- *Repro:* open the tabernacle, click a piece (e.g. the Ark), then click one of the listed verse references in the popup. Nothing happens.
- *Fix:* wire the option to the scripture side, e.g. have the DI bootstrap expose a controller method and call it from the listener (`piecesInteractionController`/a scripture controller) with `bookId`/`chapter`/`verse`. Note the reverse direction is also disabled — see H3.

**H2 — Piece hitboxes are spawned, pointable, and do nothing; they likely swallow piece clicks.**
`infrastructure/adapters/pieces/HitboxLifecycleAdapter.tsx:50-54` registers an empty `onClick` on every spawned hitbox (again just a TODO referencing `piecesInteractionController?.handlePieceClick(...)`). Those hitboxes are spawned for every piece (`PiecesSetUpService.setUpPieces` → `spawnPiecesHitbox`), created `color: "clear"` and `transformer: piece.id` so they overlay each piece, and `HitboxConfigProvider.isPointable()` returns `true`. Piece clicks, meanwhile, are wired directly on the piece bots in `infrastructure/di/bootstrap.tsx:286-290`. With an invisible **pointable** hitbox sitting over each piece, the pointer will generally hit the hitbox (empty handler) rather than the piece — so piece selection may not fire at all, and at minimum there are invisible dead-click regions.
- *Repro (needs the alpha deploy):* open `https://alpha.seedbible.org/b/refactor_tabernacle`, click on a tabernacle piece, and check whether the highlight/context-menu actually triggers. Verify in the CasualOS inspector whether the click lands on the `clear` hitbox bot or the piece bot.
- *Fix:* either wire the hitbox `onClick` to `handlePieceClick(hitbox.tags.pieceKey)` and drop the direct piece-bot listeners, or make the hitboxes non-pointable if the piece meshes are the intended click target. Pick one owner of the click so the two don't fight.

**H3 — No tests, and the reverse (verse→piece) direction is commented out.**
There are zero test files in this PR. The refactor introduces logic that is exactly the kind that benefits from regression tests: the verse↔piece reference lookups (`verseReference.ts`, `VerseReferenceConfigProvider`), the color math (`domain/functions/colors.tsx`), the layer/occlusion selection in `PieceHighlightAdapter`, and the visibility state machine in `PieceStateAdapter`. On top of that, the verse→piece half of issue #131 is disabled: `ScriptureInteractionService`, `ScriptureInteractionController`, `PieceStateService`, and the reading-state `effect` are all commented out in `infrastructure/di/bootstrap.tsx:35-36, 216-221, 257-259, 270-284`.
- *Suggested minimum coverage* (pure, no CasualOS needed): `getPiecesForExperience` dedupes and preserves encounter order; `VerseReferenceConfigProvider.getVersesForPiece` splits in-chapter vs other-chapters correctly and returns empty for an unknown experience; `HexToRgb`/`RgbToHex`/`InterpolateHexColors` round-trip and clamp. Per the repo's testing guidance, write at least one regression test that fails on the current (unwired) menu/hitbox handlers so H1/H2 can't silently regress again.
- Please also confirm whether shipping #131/#263's bidirectional behavior is in-scope for this PR or deferred; if deferred, that's worth stating in the PR description since both issues are marked completed.

---

### 🟡 Medium

**M1 — Unhandled promise rejections when a highlight/animation is cancelled.**
`ColorLerper.stop()` (`infrastructure/adapters/casualos/ColorLerper.tsx:108-116`) rejects the in-flight lerp promise, and `stop()` is called at the top of every `lerp()`. In `PieceHighlightAdapter.highlightPiece` (`:146-166`) the blink chain is `.then().finally()` with **no** `.catch()`, and the cone chain (`:170-184`) is `.then().finally()` too. Note that `AnimateStrictTag` exposes an `ignoreCancellation` option precisely for this — but it is never passed at any call site (grep confirms zero usages outside the helper). So a normal re-interaction produces an unhandled rejection.
- *Repro:* click a piece, then click the same piece again within the ~1s blink (or click empty grid to blur mid-blink). Watch the console for an unhandled rejection ("Color lerp has been canceled" / "The animation was canceled").
- *Fix:* pass `ignoreCancellation: true` on the highlight/cone `AnimateStrictTag` calls, and add a `.catch()` (swallowing the cancel reason) to the `ColorLerper` blink chain — or have `stop()` resolve rather than reject, since cancellation is expected control flow here, not an error.

**M2 — `VERSE_REFERENCE_MAP` is duplicated verbatim in two packages.**
`packages/house-of-the-lord/referenceMap.ts` and `patterns/house-of-the-lord/house-of-the-lord/infrastructure/config/verseReference/referenceMap.tsx` contain the same ~260-line table (confirmed byte-identical for the tabernacle data). The extension uses its copy to decide which pieces the verse toolbar offers; the pattern uses its copy to decide highlight + verse references. They will drift, and when they do a piece can be offered in the toolbar but not highlight (or vice-versa). `TABERNACLE_PIECE_KEYS` and `EXPERIENCE_KEYS` are likewise defined twice (`packages/.../pieceKeys.ts` + `domain/models/piece.tsx`; `packages/.../experience.ts` + `domain/models/experience.tsx`). This is the exact "copies drift, fixes reach some but not others" case called out in `CLAUDE.md`.
- *Fix:* pick one source of truth (a shared module both the extension and pattern import) for the reference map and the key enums, or generate one from the other at build time.

---

### 🟢 Low / nits

**L1 — Experience-key divergence (latent crash).** The extension package defines both `TABERNACLE` and `SOLOMON_TEMPLE` (`packages/.../experience.ts`), registers a verse-toolbar tool for each (`bootstrap.tsx:27`), and has an `EXPERIENCE_META` entry for Solomon's temple — but the pattern's `ExperienceKey` only knows `TABERNACLE` (`domain/models/experience.tsx`). Solomon is inert today (its reference map is `{}`, so the toolbar entry never becomes visible), but if any verse reference is ever added for it, opening it will pass `experience: "solomon-temple"` into a pattern that has no pieces/layers/hitboxes for that key → runtime throw in `PiecesProvider.getPiece`/`LayerConfigProvider`. Consider dropping Solomon from the extension until the pattern supports it, or gate it behind a "has references" check. (Related: the `SOLOMON_TEMPLE_PIECE_KEYS.TEST` / `piece-test` placeholder.)

**L2 — Dead `legacy/manager/*` code.** The whole `infrastructure/legacy/manager/` tree (`App.tsx`, `SetBotsVisibility.tsx` ~127 lines, `TryRecenter.tsx`, `UpdateHighlightedWords.tsx`, `OnTabernacleSectionClick.tsx`, several `*.json`, etc.) is imported nowhere and is mostly commented-out bodies. If it's kept intentionally as a migration reference, a one-line note in the PR would help; otherwise it should come out before merge so it isn't mistaken for live code.

**L3 — `ExperienceService` display flag isn't reset on failure.** `#displayExperience` (`application/services/ExperienceService.tsx:45-61`) sets `#isExperienceDisplayed = true` before awaiting `displayDropSequence`. If that throws, it returns `false` but leaves the flag `true`, so a later `tryDisplayExperience()` short-circuits to `true` and never retries. Reset the flag to `false` in the `catch`.

**L4 — Non-English translations added directly.** `packages/house-of-the-lord/extension.json` ships translated `title`/`description`/piece strings for ~20 locales. `CLAUDE.md` asks to update only `en` and leave the rest to professionals — worth confirming this convention applies to extension-bundled translations too, or that these were professionally reviewed.

---

### What looks good
- Clean port to the ports/adapters layout; the strong typing on `ExperienceKey`/`ExperienceKeyMap` and the strict tag helpers (`SetStrictTag`/`AnimateStrictTag`/`ApplyStrictMod`) are a nice improvement over stringly-typed bot access.
- `getPiecesForExperience` correctly dedupes while preserving order; `getVersesForPiece` cleanly splits in-chapter vs. other-chapters.
- The layer-occlusion logic in `PieceHighlightAdapter` (show ≤ layer, translucent at layer+1, hide above) is a tidy way to unobstruct a highlighted piece, and `LAYERS_MAP` covers all 19 pieces so `getLayerNumber` won't throw.

The headline asks: please wire H1/H2 (or confirm they're intentionally deferred), add at least the pure-logic tests in H3, and decide on a single source of truth for the reference map (M2). Happy to expand on any of these.

---

soukigabriel:
@Blitzy Went through Claude's PR's review and fixed as much as I can. Important notes:

- H1: context menu options can't perform its intended behavior until we get pattern -> seed-bible communication, so I disabled the context menu for now.
- H3: Same story for the ScriptureInteractionService, will be commented out until we get seed-bible -> pattern communication.
- M2 and L1: Duplication is neccesary as we can't access modules that live outside the pattern.

Blitzy:
# House of the Lord — Extension Review

**Scope:** Manual walkthrough of all 15 menu-reachable Tabernacle pieces via the verse toolbar, plus source review of the extension package and the CasualOS pattern.

## Verdict

The data layer is sound. All 15 pieces load the correct model from the correct verse, with no mis-mappings and no failures to open. The verse→piece table, the toolbar integration, and the pane/portal handoff all work as designed. The test suite passes and the pattern packs cleanly.

The recurring problem across the walkthrough is **camera framing**. Three pieces — fence, red curtain, brown curtain — are framed so badly that what had loaded was not immediately legible. Fence is the worst of them: framed at ground level from a distance meant for tabletop furniture, it initially reads as an empty scene, and it takes a second look to recognise that the perimeter is there and simply out of frame.

**Nothing found here blocks merging this branch.** Two defects are worth fixing soon and have been filed as follow-up issues: [#1698](https://github.com/HelloAOLab/seed-bible/issues/1698) and [#1699](https://github.com/HelloAOLab/seed-bible/issues/1699). The remainder are questions about the extension's interaction model rather than bugs to patch piecemeal, and are proposed as the agenda for a team review session on the extension's direction (R1–R7).

## Walkthrough results

| Piece | Verse used | Result |
|---|---|---|
| Altar of Sacrifice | EXO 27:1 | Correct |
| Ark of Covenant | EXO 25:10 | Correct |
| Bars | EXO 26:26 | Correct; difficult to identify |
| Bronze Laver | EXO 30:17 | Correct |
| Brown Curtain | EXO 26:7 | Correct; framing too tight |
| Fence | EXO 27:9 | Correct; framing far too tight — reads as an empty scene |
| Front Curtain | EXO 26:36 | Correct |
| Front Pillars | EXO 26:37 | Correct |
| Incense Altar | EXO 30:1 | Correct |
| Inner Curtain | EXO 26:1 | Correct |
| Inner Pillars | EXO 26:32 | Correct |
| Menorah | EXO 25:31 | Correct |
| Red Curtain | EXO 26:14 | Correct; framing too tight |
| Table of Showbread | EXO 25:23 | Correct |
| Walls | EXO 26:15 | Correct |

Every piece renders. Nothing in the set is broken.

---

# Follow-up issues

Both are filed and non-blocking. Neither needs to hold this branch.

| | Issue | Title |
|---|---|---|
| F1 | [#1698](https://github.com/HelloAOLab/seed-bible/issues/1698) | Camera framing is hard-coded for all pieces — each piece should define its own |
| F2 | [#1699](https://github.com/HelloAOLab/seed-bible/issues/1699) | Hitboxes should support multiple volumes per piece |

## F1 — Each piece should define its own camera framing

**Filed as [#1698](https://github.com/HelloAOLab/seed-bible/issues/1698).**

**Problem.** `PieceHighlightAdapter.tsx:138` calls `os.focusOn` with literal values that never vary by piece:

```
rotation: { x: 1.01229, y: 0.5 },
zoom: 40,
```

The pieces vary in size by roughly 30× (incense altar `scale: 1.1` up to fence `scale: 30`), so one framing cannot serve them all. Observed: the fence sits outside the viewport at this distance and the scene appears empty until the camera is moved; the red and brown curtains fill the screen edge to edge.

A secondary effect: ao.bot does not respond to camera drags while the pointer is over a mesh. When a curtain fills the viewport there is no empty region left to drag from, so the camera becomes unmovable. Correct framing removes this.

**Proposed fix.** Add a per-piece camera config following the pattern already used three times in this codebase — `positionsMap`, `hitboxMap`, and `layersMap` each pair a per-piece config map with a Provider class under `infrastructure/config/`. Add a `camera/` config in the same shape.

Carry `{ zoom, rotation? }` rather than zoom alone: the fence needs a higher, wider angle, not merely a further-back one, and the curtains likely want a different approach vector than the tabletop furniture.

**Files.** New `infrastructure/config/camera/{cameraMap.tsx, CameraConfigProvider.tsx}`; `PieceHighlightAdapter` takes the provider as a constructor dependency and reads framing per piece; wire in `infrastructure/di/bootstrap.tsx`.

## F2 — Hitboxes should support multiple volumes per piece

**Filed as [#1699](https://github.com/HelloAOLab/seed-bible/issues/1699).**

**Problem.** Each piece gets exactly one hitbox: a CasualOS cube, which is a **solid volume, not a shell**. The walls hitbox is `scaleX: 1, scaleY: 0.3, scaleZ: 0.32` (`hitboxMap.tsx`), parented to the walls bot via `transformer`, so it spans the walls' full footprint. Because the walls enclose the tent, that invisible cube also contains the menorah, the table, the incense altar and the ark. Clicks aimed at any of them land on the walls hitbox instead, making the tent interior unreachable.

This is not incidental. The wall and curtain meshes carry `pointableDefault: false`, so the invisible box is the only way to select them at all. The same holds for bars, rings, and all four curtains.

**Proposed fix.** Support composite hitboxes — an array of volumes per piece. Walls would define four volumes, one per wall, instead of one box enclosing everything. The piece stays fully selectable from any side, and the tent interior becomes directly clickable through the gaps between them.

The change is contained, because the click plumbing already keys off the piece rather than the box. `HitboxLifecycleAdapter.spawn` creates one bot per call and tags it `pieceKey`; the click listener emits `OnHitboxClicked(pieceKey)`. N volumes sharing a `pieceKey` all resolve to the same selection, so the adapter needs no change at all.

| File | Change |
|---|---|
| `config/hitboxes/hitboxMap.tsx` | Map value type `HitboxData` → `HitboxData[]` |
| `config/hitboxes/HitboxConfigProvider.tsx` | `getHitboxData` returns `HitboxData[]` (empty array, not null) |
| `ports/out/hitboxLifecycle.tsx` | Port signature follows |
| `services/HitboxLifecycleService.tsx` | Inner loop spawning one bot per volume |
| `adapters/pieces/HitboxLifecycleAdapter.tsx` | No change |
| `test/…/HitboxLifecycleService.test.ts` | Mocks return arrays; add a multi-volume case |

The 17 existing single-volume entries migrate as one-element arrays. Walls and the four curtains need real multi-volume definitions; bars and rings can stay single. **Fence currently has no hitbox entry at all** and cannot be selected in the scene — it should be added here, as a composite perimeter.

---

# For team review session

These are not proposed as individual patches. Each one touches how the extension is meant to be used, and they interact — deciding them together will produce a more coherent result than fixing them one at a time. Code references are included so the discussion can be concrete.

### R1 — The layer peel is one-way

Selecting a piece runs an occlusion pass (`PieceHighlightAdapter.tsx:120–134`): the selected piece's layer and everything below stay shown, the layer immediately above becomes translucent, everything further above is hidden. This is the mechanism that lets you work inward through the tabernacle.

`stopHighlight()` → `#clearFocus()` (lines 188–199) only animates the piece's colour back to white. Hidden layers are never restored, so once the scene is peeled apart the only way back is closing and reopening the pane.

**For discussion:** the peel is arguably the most compelling thing the extension does, but right now it is undiscoverable and irreversible. Is outside-in traversal the intended primary interaction, and if so what surfaces it — a reset control, a breadcrumb, a layer indicator?

### R2 — Nothing on screen identifies the piece

Three gaps stack. On open, the piece is never named: the context menu is built only inside `toggleContextMenu`, which fires on click, so the pane presents a model with no caption. On click, only the piece name appears — the reading-state bridge is commented out in `infrastructure/di/bootstrap.tsx`, so the pattern never learns which chapter the reader came from and the verse list is always empty. And the verse links would not work regardless: the `onClick` handlers in `ContextMenuRendererAdapter.tsx` are `TODO` stubs.

Observed: after loading `bars`, the piece could not be identified on screen at all.

**For discussion:** the second and third gaps are covered by the existing two-way communication issue. The first is independent and small. Worth deciding what the labelling model should be overall rather than adding a caption in isolation.

### R3 — The selection highlight lasts one second

`BLINK_DURATION = 1` (`PieceHighlightAdapter.tsx:19`). The colour lerps white → cyan → white over one second; the cone fades 0 → 0.75 → 0 and is then destroyed. Afterwards nothing distinguishes the selected piece.

This compounds R2 — by the time the model has been read, the only indication of which piece was selected has already gone.

**For discussion:** should there be a persistent selection state at all, or is the blink sufficient once pieces are labelled?

### R4 — "Translucent" is 2.5% opacity

`PieceStateAdapter.tsx:116` — `const targetOpacity = 0.025`. The layer directly above the selection is intended as see-through context but is functionally invisible, which is why peeled scenes read as empty rather than layered.

**For discussion:** one number, but it defines what "context" means visually and pairs with R1.

### R5 — Perimeter pieces may need different treatment from furniture

Fence is the clearest case that the scene's conventions are tuned for objects inside the tent. Three of them work against it at once: it occupies layer 0, so selecting it pushes every other layer to hidden or 2.5% translucent and the tent effectively vanishes (R4); the shared camera framing is calibrated for tabletop objects and puts a ground-level perimeter out of frame (F1); and it has no hitbox entry, so it cannot be reselected once deselected (F2).

F1 and F2 address the second and third directly. The first is a design question.

**For discussion:** should layer-0 selection hide everything above it? For furniture, peeling upward reveals the piece. For a perimeter that surrounds the whole scene, hiding everything inside it removes the context that makes it legible — you end up looking at a boundary with nothing bounded. A perimeter may want the inverse behaviour, or none.

### R6 — The drop-in sequence reads as a pop

Each piece travels 1 unit in 0.3 s (`PieceStateAdapter.tsx:55, 96`) on a 200 ms stagger (`PiecesSequenceAdapter.tsx:7`). The walls mesh has `scale: 9.5`, so a 1-unit drop is roughly a tenth of the mesh's own height over a third of a second — closer to a flicker than a drop.

**For discussion:** polish, but it is the first thing anyone sees.

### R7 — Solomon's Temple scaffolding

Registered as a second menu with one declared piece (`test`). Its verse map is empty, so the toolbar entry never becomes visible, and the pattern's own `EXPERIENCE_KEYS` does not define the experience — a hand-edited URL would not render it either. Not a defect; noted because the multi-experience structure is a standing assumption that several decisions above depend on.

---

# For content review

All 209 mapped verses are in Exodus, across chapters 25, 26, 27, 29, 30, 36, 37, 38, 39, 40. Two distinct questions, both for Craig or anyone with the reference knowledge.

## Gaps within Exodus

Chapters 25, 36, 37, 38 and 39 are contiguous across their mapped range — no holes. The gaps are concentrated in 26, 27, 30 and 40.

**Verified against the source text, and worth a decision:**

| Verse | Text | Note |
|---|---|---|
| **40:8** | "you will put the courtyard all around and you will put the screen of the gate of the courtyard" | Names two modeled pieces — fence and front curtain. The parallel execution verse, 40:33, **is** mapped to exactly those two. This looks like an oversight. |
| **40:23** | "he arranged on it an arrangement of bread" | Refers to the table by pronoun. 40:22, the placement verse, is mapped. |
| **40:25** | "he lifted up the lamps" | Refers to the menorah. 40:24, the placement verse, is mapped. |
| **40:27** | "he made smoke on it incense of perfumes" | Refers to the incense altar. 40:26, the placement verse, is mapped. |

The 40:23 / 25 / 27 cases form one consistent pattern: the verse that *places* an object is mapped, and the immediately following verse that *uses* that object is not. That is a defensible editorial line, but it should be a deliberate one — a reader on 40:25 sees no menorah link while a reader on 40:24 does.

**Checked and reasonable to leave unmapped:**

- **26:30** — "you will set up the tabernacle according to its plan" — refers to the structure as a whole, not a piece. Note there is a commented-out `// 9: ["tabernacle"]` at EXO 25:9 in the source, suggesting a whole-tabernacle target was considered and dropped.
- **30:11–16** — the census tax, unrelated to the furniture.

**Not yet checked against the text:** 27:19, 30:22–25, 40:4, 40:9–18, 40:31–32, and the whole of chapters 28 and 31–35.

## Books beyond Exodus

A scoping question rather than a defect. The strongest candidates are Leviticus, Numbers (chapters 3–4 in particular enumerate the same furniture almost piece by piece), and Hebrews 9, which walks the tabernacle's contents explicitly.

---

# Appendix — current coverage

Generated by evaluating `referenceMap.ts` directly. Selecting any listed verse surfaces the **Tabernacle** button with that item in its menu; choosing it opens the portal with `highlightedPiece` set to the key shown.

| Menu item | `highlightedPiece` | Exodus chapters | Verses | # |
|---|---|---|---|---|
| Altar Of Sacrifice | `altar-of-sacrifice` | 27, 29, 30, 38, 39, 40 | 27:1–8 · 29:12, 16, 20–21, 37 · 30:28 · 38:1–7 · 39:33, 38 · 40:6, 29 | 25 |
| Ark Of Covenant | `ark-of-covenant` | 25, 26, 30, 37, 39, 40 | 25:10–22 · 26:34 · 30:26 · 37:1–9 · 39:33, 35 · 40:3, 19–21 | 30 |
| Bars | `bars` | 26, 36 | 26:26–29 · 36:31–34 | 8 |
| Bronze Laver | `bronze-laver` | 30, 38, 39, 40 | 30:17–21 · 38:8 · 39:33, 38 · 40:7, 30 | 10 |
| Brown Curtain | `brown-curtain` | 26, 36 | 26:7–13 · 36:14–18 | 12 |
| Fence | `fence` | 27, 38, 39, 40 | 27:9–18 · 38:9–18 · 39:39 · 40:33 | 22 |
| Front Curtain | `front-curtain` | 26, 36, 39, 40 | 26:36 · 36:37 · 39:34 · 40:28, 33 | 5 |
| Front Pillars | `front-pillars` | 26, 36 | 26:37 · 36:38 | 2 |
| Incense Altar | `incense-altar` | 30, 37, 39, 40 | 30:1–10, 27 · 37:25–29 · 39:33, 38 · 40:5, 26 | 20 |
| Inner Curtain | `inner-curtain` | 26, 36, 39, 40 | 26:1–6, 31, 33 · 36:8–13, 35 · 39:34 · 40:21 | 17 |
| Inner Pillars | `inner-pillars` | 26, 36 | 26:32 · 36:36 | 2 |
| Menorah | `menorah` | 25, 26, 27, 30, 37, 39, 40 | 25:31–40 · 26:35 · 27:20–21 · 30:26 · 37:17–24 · 39:33, 37 · 40:24 | 25 |
| Red Curtain | `red-curtain` | 26, 36 | 26:14 · 36:19 | 2 |
| Table Of Showbread | `table-of-showbread` | 25, 26, 30, 37, 39, 40 | 25:23–30 · 26:35 · 30:26 · 37:10–16 · 39:33, 36 · 40:22 | 20 |
| Walls | `walls` | 26, 36 | 26:15–25 · 36:20–30 | 22 |

**209 distinct verses, 222 verse→item pairs.**

Seven verses produce more than one menu item — these are the multi-item test cases:

| Verse | Items | Count |
|---|---|---|
| **EXO 39:33** | Ark, Table of Showbread, Menorah, Incense Altar, Altar of Sacrifice, Bronze Laver | **6** |
| EXO 30:26 | Ark, Table of Showbread, Menorah | 3 |
| EXO 39:38 | Incense Altar, Altar of Sacrifice, Bronze Laver | 3 |
| EXO 26:35 | Table of Showbread, Menorah | 2 |
| EXO 39:34 | Inner Curtain, Front Curtain | 2 |
| EXO 40:21 | Inner Curtain, Ark | 2 |
| EXO 40:33 | Fence, Front Curtain | 2 |

EXO 39:33 is the widest case in the set and the best single verse for testing menu rendering.

Selecting several verses at once yields one item per distinct piece, in ascending verse order — EXO 25:10 + 25:23 + 25:31 gives Ark Of Covenant, Table Of Showbread, Menorah.

Four pieces load in the scene but appear on no menu, because no verse maps to them: **Grey Curtain**, **Purple Curtain**, **Rings**, and **Ground**. Grey Curtain, Purple Curtain and Rings can still be clicked in the scene; Ground cannot, as it has no hitbox entry. Note that **fence** is in the same position — it appears on menus but, having no hitbox entry and carrying `pointableDefault: false` like every other piece, it cannot be clicked in the scene either (see F2).

---

# Test status

`pnpm vitest run test/unit/house-of-the-lord` — 63 tests across 12 files, all passing. The pattern packs cleanly (`pnpm pattern pack house-of-the-lord`, 88 KB `.aux`). No regressions. The F2 work will require updating `HitboxLifecycleService.test.ts`.

Blitzy (approved):

PR #1662, opened 2026-08-19
Support the `annotationRecordKey` query parameter

### LINKED ISSUE #763: Saving annotations for custom translations is broken
When publishing a translation from the Codex Seed Bible Upload extension, the extension will include a record key in the URL called `annotationRecordKey` that should be used for recording annotations. For some reason this appears to have been broken, and instead the annotations tries to read a data record and then fails.

Here's a demonstration video: https://www.loom.com/share/f7b933cc4b0946b8b6f03794737fb9e5

KallynGowdy:
In old versions of the Seed Bible, we supported an `annotationRecordKey` query parameter that could be set to a record key to allow saving all annotations to that record instead of the default user record. This allowed translations published from Codex (the VSCode-based translation editor) to be annotated and for those annotations to be viewed directly inside Codex.

Closes #763


Blitzy:
> [!NOTE]

## Summary

This PR adds an `annotationRecordKey` query parameter that, when present, routes annotation reads/writes through a caller-supplied record instead of the signed-in user's record. The plumbing is: `SeedBibleStateManager` reads the query param and passes it as a new `recordOverride` argument to `createAnnotationsManager`, and `resolveRecordName` returns that override when no explicit `query.recordName` is given.

The low-level methods (`saveAnnotation`, `deleteAnnotation`, `listAnnotationsForChapter`) honor the override correctly, and the new unit tests cover those paths well. **However, the reactive layer that the UI actually renders from does not use the override at all**, so the feature does not work end-to-end as described. Details below, by severity.

---

### High — The reactive view the UI renders ignores `recordOverride`, so the feature is broken in the app

`AnnotationsManager.tsx`

The UI does not read annotations through `listAnnotationsForChapter`. Both `DiscoverPane` (`DiscoverPane.tsx:941`, `:980`) and `BibleReader` (`BibleReader.tsx:154`, `:1300`) read from the reactive view `getAnnotationsForChapter`. That view — and the caches behind it — key off the signed-in user's id and never consult `recordOverride`:

- `getOrCreateView` (`AnnotationsManager.tsx:551-570`) returns `[]` whenever `login.userId.value` is null. With `annotationRecordKey` set but the user signed out — the exact Codex scenario the PR describes — the annotation pane and the note-count badge always render empty.
- When a user *is* signed in, `loadEntry` (`AnnotationsManager.tsx:506-530`) loads via `listAnnotationsForChapter(bookId, chapterNumber, { recordName: userId })`. Because it passes an explicit `recordName`, `resolveRecordName` short-circuits and reads from the **user's own record**, not the override record. Annotations that were written to the override never load into the view.
- `upsertIntoCache`/`removeFromCache` (`AnnotationsManager.tsx:594-622`) also bail when `login.userId.peek()` is null, so optimistic updates don't appear when signed out either.

Net effect: with the override active, writes go to the override record but the pane displays the signed-in user's annotations (or nothing when signed out). Saved notes appear to vanish on refresh. The one place it *seems* to work is a signed-in optimistic insert, and only until the next reload.

**Reproduction (signed out):**
1. Open the app with `?annotationRecordKey=some-record`.
2. Note the annotation pane / note count.
3. `getAnnotationsForChapter` returns `[]` unconditionally because there's no `userId` — no annotations from the override record are ever shown.

**Reproduction (signed in):**
1. Sign in, open with `?annotationRecordKey=some-record`.
2. Save an annotation → it's written to `some-record` (correct).
3. Reload → the pane loads from the signed-in user's record (via `loadEntry` passing `recordName: userId`), so the just-saved annotation is gone from view.

**Suggested fix:** thread `recordOverride` through the reactive layer so it becomes the cache key and the record read when an override is active. Roughly: in `getOrCreateView`, use `recordOverride ?? login.userId.value` as the effective record id and don't early-return `[]` when an override is present but the user is signed out; in `loadEntry`, drop the explicit `{ recordName: userId }` and let `resolveRecordName` pick the override (or pass the effective id); and in `upsert/removeFromCache`, key on the same effective id rather than requiring `login.userId`. The `cachedUserId` cleanup effect (`:576-587`) needs to account for the override too, otherwise it will evict the override-keyed entries.

---

### Medium — Tests pass but don't cover the path that's actually broken

`test/unit/seed-bible/managers/AnnotationsManager.test.ts`, `test/unit/seed-bible/managers/SeedBibleStateManager.test.ts`

The new tests only call the low-level methods (`saveAnnotation`, `deleteAnnotation`, `listAnnotationsForChapter`) directly. None of them exercise `getAnnotationsForChapter` — the reactive view the UI renders — which is exactly where the override is dropped (see the High finding). Because of that, the suite is green while the user-facing feature does not work. This is the "tests that always succeed" case: they assert the plumbing they added, not the observable app behavior the PR is meant to deliver.

**Suggested additions:**
- A test that sets an override, reads `getAnnotationsForChapter(book, chapter).value`, and asserts the underlying `listDataByMarker` was called with the override record — not the user id. This will fail on the current code and should pass after the High fix.
- A signed-out variant: override set, `login.userId.value = null`, assert `getAnnotationsForChapter` still surfaces the override record's annotations rather than returning `[]`.
- An end-to-end-ish variant through `createSeedBibleState`: with `?annotationRecordKey=...`, save via `saveEditingAnnotation` (the real UI entry point) and assert the note shows up in `getAnnotationsForChapter`.

Also worth confirming the regression guidance in `CLAUDE.md`: revert the `recordOverride` branch in `resolveRecordName` and confirm these tests go red.

---

### Medium — "Key" vs "record name": can listing actually work with a real record key?

`AnnotationsManager.tsx`, `OsManager.tsx`

The parameter is named `annotationRecordKey` and the PR describes it as "a record key," but the value flows into two different SDK surfaces with different expectations:
- `saveAnnotation`/`deleteAnnotation` → `os.recordData(recordKey, …)` / `os.eraseData(recordKey, …)` (`OsManager.tsx:315-338`), which pass it as `recordKey` — a record *key* is accepted here.
- `listAnnotationsForChapter` → `os.listDataByMarker(recordName, …)` → `client.listData({ recordName })` (`OsManager.tsx:283-295`), which expects a record *name*.

If a caller supplies an actual CasualOS record key (the secret `name:key…` form), the writes should work but the list call is being handed a key where a name is expected. Please confirm listing resolves correctly when a real key (not a bare record name) is passed; if it doesn't, the parameter either needs to be a record name (rename to avoid implying a secret key) or the manager needs to derive the record name from the key before listing. Right now the name and the usage disagree, which will bite whoever wires up Codex.

---

### Low — Record key placed in a URL query parameter

`SeedBibleStateManager.tsx:520-522`

If `annotationRecordKey` is genuinely a record key (a write credential), carrying it in the URL exposes it to browser history, the `Referer` header on outbound requests, and any server/access logs. This may be acceptable given the embedded-iframe Codex use case and the prior behavior, but it's worth a deliberate note in the PR. If the value is really only ever a public record *name*, this concern goes away (and reinforces the naming point above).

---

### Low — `createNewAnnotation` still forces sign-in even when an override is set

`AnnotationsManager.tsx:677-686`

`createNewAnnotation` calls `login.login()` and aborts if the user declines, regardless of `recordOverride`. If part of the intent is to let a Codex context annotate through a shared record without the normal sign-in, this path still blocks it. Not necessarily wrong (the annotation payload stamps `userId`/author fields), but flagging it so the intended auth model for the override is explicit.

---

### Nits

- The added JSDoc on `createAnnotationsManager` (`AnnotationsManager.tsx:345-353`) has empty `@param os/login/tabs/discover` and `@returns` tags. Per the repo's comment convention, either describe them or drop the empty tags and keep only the `recordOverride` line that carries information.
- `navigation.currentUrl.value.searchParams.get(...)` (`SeedBibleStateManager.tsx:520`) is read once at construction, so a runtime URL change won't update the override. Likely fine, but intentional-vs-accidental is worth a one-line confirmation.

---

Happy to dig further on the CasualOS `listData` key-vs-name question if you can point me at how Codex passes the value in practice.

---

Issue #1630, opened 2026-08-11
Implement prompt for users to download their current translation


KallynGowdy:
Currently, users have to manually go to their favorite translation and download it by pressing the download button on it.

We should automatically prompt users to download a translation in the following scenarios:

- When the user first loads the Seed Bible and is shown the welcome screen
   - We should prompt to download the current translation after we show the "Install app" screen
- When the user loads the Seed Bible (and dismisses the today screen)
    - If the user doesn't have any downloaded translations and we haven't shown this prompt before (download first translation prompt), then we should prompt them to download their current translation.
    - Otherwise, if the current translation has been in their reading history for at least a day and we haven't shown this prompt before (download current translation prompt), then we should prompt them to download their current translation

This means we have three places where we can prompt and three different prompts:

- Welcome screen prompt
- Download first translation prompt
- Download current translation prompt

When we show one prompt, we should never show any more download prompts for that session.

Here's the design to follow: https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=9351-4029&t=nuyL69PtgDeUrytd-1

Blitzy:
Closed with merging of #1663

PR #1663, opened 2026-08-19
added the translation download dialog

Bored-Wizard:
task #1630


packages/seed-bible/seed-bible/managers/OfflineTranslationsManager.tsx
```diff
@@ -39,6 +39,92 @@ import {
...
+
+/** Writes a translation-ID -> timestamp map to local storage. Best-effort. */
+function writeTimestamps(key: string, value: Record<string, number>): void {
+  try {
+    window.localStorage.setItem(key, JSON.stringify(value));
+  } catch {
+    // Storage can be full or blocked; losing the record only means the user
+    // may be offered the download again on a later visit.
+  }
+}
```
KallynGowdy, line 110: @Bored-Wizard For this can we set a limit on the number of translations that we track so that it doesn't keep growing? It should track the 5 most recent translations/timestamps. Anything over 5 should just be deleted.

packages/seed-bible/seed-bible/managers/OfflineTranslationsManager.tsx
```diff
@@ -39,6 +39,92 @@ import {
...
+export function estimateTranslationSizeBytes(
+  translation: Translation
+): number | null {
+  const verses = translation.totalNumberOfVerses;
+  if (typeof verses !== "number" || verses <= 0) {
+    return null;
+  }
+  return verses * BYTES_PER_VERSE_ESTIMATE;
+}
+
```
KallynGowdy, line 138: @Bored-Wizard This is clever, but you should be able to do an [HTTP HEAD request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD) against the complete translation to get its actual size without downloading the entire thing. I would try fetching the actual size first via `fetch(completeUrl, { method: "HEAD" })` with a short timeout, and then falling back to the estimate if it doesn't work or is too slow.

Some translations in non-latin scripts (like Chinese) will take up a lot more space with the same number of verses.

Blitzy:
> [!NOTE]

## Code review — PR No. 1663: "added the translation download dialog"

Reviewed against task #1630. This adds the prompt that offers to save the current translation for offline reading, wires the timing into `SeedBibleStateManager`, and adds a one-off "glow" tour tip that points at the download button in the translation list.

**Overall:** solid, carefully-commented work. TypeScript check passes, ESLint passes (only pre-existing baseline warnings), the branch merges cleanly into `develop`, and the two new test files (26 tests) pass. The manager-level gating tests are genuinely good — they run through the real `createBibleDataManager` and only fake the network and the storage back end, so they exercise the actual gate logic and would fail if that logic changed (I confirmed they cover the offline case, already-downloaded/downloading, one-prompt-per-session, cross-session persistence, and both sides of the 24-hour boundary). Nothing here is a blocker; the findings below are mostly medium/low.

---

### Medium

**M1 — `offerDownloadPrompt` reads download state without waiting for storage to load, so a startup race can offer a translation that is already downloaded.**

`offerDownloadPrompt` (`OfflineTranslationsManager.tsx:1031`) decides using `isDownloaded(translationId)` and `downloaded.value.size`. Both read the `records` signal synchronously — `isDownloaded` is just `records.value.has(...)` (`:709`) — but `records` starts empty and is only filled inside the async `ready` IIFE (`:662`/`:668`). Unlike the read paths (`getRecord` does `await ready` at `:704`), `offerDownloadPrompt` can't await because it must return a boolean synchronously.

The caller in `SeedBibleStateManager.tsx:1249` also doesn't wait on `offline.ready`; it fires once install/tutorial resolve and a chapter is loaded. In the normal first-run flow the user spends seconds in the tutorial/install screens, far longer than an IndexedDB read, so the race almost never bites. But a **returning** user whose tutorial is already marked seen can reach the offer effect within milliseconds of load — before `ready` resolves. In that window `records` is empty, so the manager thinks nothing is downloaded.

Consequence: it can mis-classify as "first ever offer" (`isFirstEverOffer` at `:1060` becomes true when it shouldn't), or offer to save a translation the device already has (`downloadTranslation` would then re-download and replace the existing copy).

*Reproduction (timing-dependent, easiest to force):* in `offerDownloadPrompt`, temporarily read `records` before `ready` has resolved — e.g. call it synchronously right after constructing the manager but before `await manager.offline.ready` in a test that has a download already in the injected store. `isDownloaded` returns `false` even though the store holds the translation.

*Suggested fix:* have the manager expose a "records loaded" signal (set to `true` at the end of the `ready` IIFE) and add `!recordsLoaded.value` as an early-return guard in `offerDownloadPrompt`, or gate the offer effect in `SeedBibleStateManager` on that signal so the offer is only considered once storage has been read. A cheap alternative is to add a synchronous `readyResolved` boolean the effect can check.

**M2 — The issue asks for three distinct prompts; the PR ships one.**

Issue #1630 spells out "three different prompts: Welcome screen prompt, Download first translation prompt, Download current translation prompt," with a Figma to follow. The implementation covers the three *trigger scenarios* through the gating in `offerDownloadPrompt`, but renders a single dialog with one set of copy (`offlinePrompt.title` / `offlinePrompt.body`) for all of them. Consolidating may well be fine, but it's a deliberate divergence from the written requirement and the linked design, so it should be confirmed with the design owner rather than assumed. I can't open the Figma to check whether the three prompts were meant to differ in wording or layout — flagging so a human can compare against the node linked in the issue.

**M3 — The orchestration that decides *when* to offer has no test.**

The trickiest new logic — the effect at `SeedBibleStateManager.tsx:1249` that sequences the offer behind `installOfferResolved`, `onboarding.step === "done"`, `tutorial.promptVisible/running`, `openedViaContentLink`, and a loaded chapter, plus the new `installOfferResolved` signal — is entirely untested. The manager gates are well covered, but a regression in the ordering (e.g. offering during the tutorial, or the offer never firing because `installOfferResolved` never flips) would slip through. Worth an integration test around `createSeedBibleState` that asserts the offer is withheld until install/tutorial resolve and then fires once.

---

### Low

**L1 — Dialog has `aria-modal="true"` but no focus management.** In `OfflineDownloadPrompt.tsx` (`:83`) the dialog declares `role="dialog"` / `aria-modal="true"` but nothing moves focus into it on open, traps focus within it, or restores focus to the triggering control on close. For a keyboard or screen-reader user, focus stays behind the modal. Escape and backdrop dismissal are handled, so this is a polish item: autofocus the primary button on open and restore focus on unmount.

**L2 — Minor: the glow style relies on `color-mix` (inline) and the CSS uses it too.** `GLOW_STYLE` in `BibleSelector.tsx` and `OfflineDownloadPrompt.css` trigger `css/use-baseline` warnings. These are consistent with existing usage across the codebase (many files already warn), so this is non-blocking — noting only in case wider browser support matters for this surface.

---

### Notes (not action items)

- The `formatBytes` helper was correctly moved out of `BibleSelector` into `OfflineTranslationsManager` and imported back, removing the duplicate — good cleanup, and it matches the repo's "extract on the third copy" guidance.
- The apparent large deletions in `main.tsx` / `SeedBibleStateManager.tsx` that show up in the GitHub "Files changed" tab are an artifact of the PR base being compared against the tip of `develop`, which is one merge commit ahead of where this branch forked. They are **not** part of this PR — the real diff to `main.tsx` is +7 lines, and a three-way merge into current `develop` is clean. No action needed, but rebasing/merging `develop` in would make the diff read accurately for human reviewers.
- The contextual "offline-download" tip is correctly gated by `featuresSeen`, so it won't re-appear on every open of the translation list.

---

packages/seed-bible/seed-bible/managers/OfflineTranslationsManager.tsx
Bored-Wizard: @KallynGowdy it seems like the content-length is not exposed in the header. can you look into it?
image

KallynGowdy: @Bored-Wizard Looks like this is actually an issue with CloudFront, where they don't always send the Content-Length header when returning compressed responses that are cached

KallynGowdy: @Bored-Wizard For now, let's skip this. I will plan on adding support for including the complete translation size in the available_translations.json list

PR #1664, opened 2026-08-19
Fix mobile bookmarks behavior on first tap and add unit tests

### LINKED ISSUE #1527: Mobile bookmarks dont open until the second tap.
# What is happening

On the mobile bookmark page, it takes two taps to open any bookmark.

https://github.com/user-attachments/assets/c192949a-f6af-44f2-90b5-d47a83deea50

# What I expected

Tapping any bookmark would open it and close the mobile bookmark page.

mazenemad:
fixes #1527

### DIFF
--- a/packages/seed-bible/seed-bible/components/Tabs/Tabs.tsx
+++ b/packages/seed-bible/seed-bible/components/Tabs/Tabs.tsx
@@ -1748,6 +1748,11 @@ function BookmarksSection(props: BookmarksSectionProps) {
       // initial chapter data lands the reader scrolls to the bookmarked verse.
       newTab.readingState.scrollToVerse.value = scrollVerse;
     }
+    // `addTab()` only marks the tab selected inside TabsManager — it doesn't
+    // place it in a layout slot or dismiss the sidebar. Without this the mobile
+    // bookmarks screen stays on top of the reader, so the bookmark looks
+    // unopened until a second tap takes the `existing` branch above.
+    app.selectTab(newTab.id);
   };
 
   const formatVerseRef = (
--- a/test/unit/seed-bible/components/BookmarksMobile.test.tsx
+++ b/test/unit/seed-bible/components/BookmarksMobile.test.tsx
@@ -1,0 +1,179 @@
+import { render, type ComponentChildren } from "preact";
+import { act } from "preact/test-utils";
+import { formatV1SessionKey } from "@casual-simulation/aux-common";
+import { Sidebar } from "@packages/seed-bible/seed-bible/components/Tabs/Tabs";
+import type { SeedBibleState } from "@packages/seed-bible/seed-bible/managers/SeedBibleStateManager";
+import { createTestSeedBibleState } from "../testUtils/createTestSeedBibleState";
+import { TestHost } from "./TestHost";
+
+vi.mock("../i18n/I18nManager", () => ({
+  useI18n: () => ({
+    t: (key: string, options?: { defaultValue?: string }) =>
+      options?.defaultValue ?? key,
+  }),
+}));
+
+vi.mock("../components/ContextMenu", () => ({
+  closeContextMenus: vi.fn(),
+  ContextMenuItem: ({
+    children,
+    onClick,
+    className,
+  }: {
+    children: ComponentChildren;
+    onClick?: () => void;
+    className?: string;
+  }) => (
+    <button className={className} onClick={onClick}>
+      {children}
+    </button>
+  ),
+  ContextMenuWithButton: ({
+    children,
+    buttonClassName,
+    onClick,
+  }: {
+    children: ComponentChildren;
+    buttonClassName?: string;
+    onClick?: () => void;
+  }) => (
+    <div>
+      <button className={buttonClassName} onClick={onClick}>
+        Menu
+      </button>
+      <div>{children}</div>
+    </div>
+  ),
+}));
+
+vi.mock("../components/SettingsPage", () => ({
+  SettingsPage: () => <div>Settings Page</div>,
+}));
+
+vi.mock("../components/SidebarSearch", () => ({
+  SidebarSearch: () => <div>Sidebar Search</div>,
+}));
+
+/** The width `app.isMobile` needs to see for the mobile bookmarks screen. */
+const MOBILE_VIEWPORT_WIDTH = 400;
+
+const USER_ID = "user-1";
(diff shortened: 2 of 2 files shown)


Blitzy:
> [!NOTE]

## Summary

**The fix is correct, minimal, and precisely targets the bug in #1527.** No blocking issues. I verified the behavior end-to-end (including reverting the fix to confirm the new test actually catches the regression), and I have two low-severity test-hygiene notes and a couple of nits below.

### What the fix does and why it works

The mobile bookmarks screen took two taps to open a bookmark. The root cause: `openBookmark()` in `Tabs.tsx` had two branches, and only one of them finished the job.

- **Bookmark already open in a tab** (the `existing` branch) → it called `app.selectTab(existing.id)`, which closes the sidebar, binds the tab to a layout slot, and closes fullscreen panes.
- **Bookmark not yet open** (the new-tab branch) → it called `tabsManager.addTab(...)`, which _only_ sets `selectedTabId` inside `TabsManager`. It never closed the mobile drawer or placed the tab in a layout slot.

So on the first tap of a not-yet-open bookmark, a tab was created behind the scenes but the drawer stayed on top of the reader — the bookmark "looked unopened." The second tap then hit the `existing` branch (the tab now existed) and worked. The one-line fix — adding `app.selectTab(newTab.id)` at the end of the new-tab branch — makes both branches behave identically. I confirmed `SeedBibleStateManager.handleSelectTab` closes the sidebar (`isMobileOpen → false`), calls `tabsLayout.setSelectedSlotTab`, and closes panes, and that `setSelectedSlotTab` binds the freshly-added tab into the selected slot (it's already in `tabMap` because `addTab` pushed it before `selectTab` runs). So the tab is genuinely shown, not just selected in state.

This also quietly fixes the same missing-slot-binding on desktop for the "open a bookmark that isn't already in a tab" path, which is a consistent improvement.

## Findings by severity

### Low — The four `vi.mock(...)` calls in the new test are silent no-ops

In `BookmarksMobile.test.tsx`, all four module mocks use a path relative to the **test file** that does not match what `Tabs.tsx` actually imports, so Vitest never applies them and the **real** modules load instead:

| Test mocks | `Tabs.tsx` actually imports |
| --- | --- |
| `../i18n/I18nManager` | `../../i18n/I18nManager` |
| `../components/ContextMenu` | `../../components/ContextMenu/ContextMenu` |
| `../components/SettingsPage` | `../../components/SettingsPage/SettingsPage` |
| `../components/SidebarSearch` | `../../components/SidebarSearch/SidebarSearch` |

The paths are wrong on the relative depth (`../` vs `../../`) and, for the component ones, on the final segment (`.../ContextMenu` vs `.../ContextMenu/ContextMenu`). Because the resolved IDs don't match the source imports, the mock factories are never invoked.

**How I proved it:** I replaced the `I18nManager` mock factory body with `throw new Error("MOCK_I18N_APPLIED")` and re-ran the test. It **still passed** — the factory never executed. If the mock were being applied, the throw would have blown up the import.

**Impact:** Not a correctness bug — the test still passes and still exercises real behavior (arguably it exercises _more_ real behavior, since the real `ContextMenu`/`SettingsPage`/`SidebarSearch`/`I18nManager` render). The problem is that the mocks are misleading dead code: a future reader will believe those dependencies are stubbed when they aren't, and any real coupling they introduce (e.g. `SidebarSearch` effects firing on a signed-in state) is silently in play. Worth noting this same broken pattern is copied from `SidebarSearch.test.tsx` (line 12), so it's a pre-existing repo smell, not introduced here.

**Suggested fix — pick one:**
- **Remove the mocks.** The real components render fine here, and this is closer to the project's stated testing philosophy ("mock only at the real boundary"). The one genuine boundary — the CasualOS records client (`state.os.getData` / `recordData`) — is already mocked correctly.
- **Or fix the paths** to actually stub, e.g. `vi.mock("../../components/ContextMenu/ContextMenu", ...)`. If you go this way, please re-run and confirm the test still passes with the mocks genuinely applied (it should — the bookmark button comes from the real `BookmarksSection`, which isn't mocked).

### Nit — The regression's real "teeth" is a single assertion; make sure it survives edits

In the first test, the meaningful guard against #1527 is `expect(state.sidebar.isMobileOpen.value).toBe(false)`. The line above it, `expect(state.tabs.selectedTabId.value).toBe(openedTab!.id)`, passes **with or without the fix**, because `addTab` sets `selectedTabId` on its own. I verified this by reverting the fix: the test fails **only** at the `isMobileOpen` line (`Tabs.tsx:1755` removed → `expected true to be false` at test line 153). That's fine as written — just flagging that the `isMobileOpen` assertion is the one doing the work, so please don't let it get dropped in future refactors.

### Nit — Consider asserting the tab is actually placed in a layout slot

The reported symptom was that the bookmark "looks unopened," i.e. the reader didn't show it. The test proves the drawer closes and the tab is selected, which (given the code) implies it's shown — but it doesn't assert the tab landed in the selected slot. Asserting something like `state.tabsLayout.slots.value.some((s) => s.tab?.id === openedTab!.id)` would tie the test directly to the user-visible symptom and guard against a future change to `handleSelectTab` that stops binding the slot. Optional.

### Note — The second test is a guard, not a regression test (this is fine)

"opens a bookmark that already has a tab on the first tap" passes on both the pre-fix and post-fix code, because it exercises the `existing` branch that already called `app.selectTab`. That's expected and good — it guards against the fix accidentally breaking the already-working path. Just noting it so it isn't mistaken for a second regression test.

## Verification I ran

- **New tests, as submitted:** both pass (`pnpm vitest run BookmarksMobile.test.tsx`).
- **Reverted the fix** (removed `app.selectTab(newTab.id)` from `Tabs.tsx`): the first test fails exactly at the `isMobileOpen` assertion — confirming it's a genuine regression test, not one that always passes.
- **Restored the fix and ran neighbors** (`BookmarksMobile`, `SidebarSearch`, `TabsManager`, `TabsSidebarLayout`): 76 tests pass, no regressions.
- **Traced the fix path** through `handleSelectTab` → `closeSidebarAndSettings` / `tabsLayout.setSelectedSlotTab` → confirmed the new tab is bound into the selected slot and the mobile drawer is dismissed.

## Verdict

Approve pending the (optional) test-hygiene cleanup. The production change is sound and I'd be comfortable merging it as-is; the `vi.mock` no-op is worth a quick follow-up because it will otherwise keep getting copied into new tests.

---

PR #1665, opened 2026-08-19
Fix mobile bookmark navigation and history management

### LINKED ISSUE #1338: The back button gets broken after selecting a bookmark
In mobile, selecting a bookmark messes with the history entries and causes the back button/gesture to be wrong.

### Example
https://github.com/user-attachments/assets/5510b8b4-8e48-43d0-9fa7-02ea83aa027b

mazenemad:
fixes #1338


Blitzy:
> [!NOTE]

## Code review — PR No. 1665 (fixes #1338)

### Summary

This is a well-scoped, well-reasoned fix for the broken mobile back button after opening a bookmark (#1338). The root cause is correctly identified: one user action (tap a bookmark) produces **two** URL writes — the reader's position (a `replace`, tab switch) and the sidebar dismissal (a `push`, dropping `?sidebar=open`). Left separate, the position `replace` stamps the destination over the entry that opened the sidebar, and the sidebar `push` adds a second entry for the same chapter — so "back" appears to do nothing.

The solution — a `batchWrites()` primitive in `NavigationManager` that folds every URL write in a callback into a single history operation (pushing if any write pushed, replacing otherwise), publishing each write to `currentUrl` live so the URL→state effects don't revert the very changes being batched — is a clean, general fix. The `navigateFromSidebar()` wrapper applies it consistently across `handleSelectTab`, `handleOpenInNewSlot`, `handleSelectSlot`, and `handleSelectPane`, and the bookmark path adds the previously-missing `app.selectTab(newTab.id)` so the sidebar actually dismisses on the new-tab path.

**I verified locally:**

- `pnpm vitest run NavigationManager.test.ts BookmarkHistory.test.tsx` → **20/20 pass**.
- **The regression tests are load-bearing.** Neutralizing `batchWrites()` to a passthrough turns 2 of the 3 `BookmarkHistory` tests red ("costs a single history entry…" and "returns to the sidebar over the previous chapter…"), confirming they fail on the pre-fix behavior rather than passing unconditionally.
- CHANGELOG entry is correctly placed under `## TBD` → `### 🐛 Fixed` with the issue link.

The reasoning is sound — collapsing `push` then `replace`, `replace` then `push`, etc. all yield history-equivalent results to the intended sequence, and the batch converges because `@preact/signals` effects run synchronously. **No critical or high-severity issues found.** The items below are low-severity edge cases and coverage notes.

---

### Low severity

**No. 1 — A batch that toggles a value back to its start still writes a redundant history entry.**

Each write inside a batch is checked against the *live* `currentUrl` (updated during the batch), so a sequence that changes a param and then changes it back both passes the `hasChanges` guard each time, leaving `pendingHref` pointing at the original URL. At flush, `writeHistory()` then executes a real `pushState`/`replaceState` for a URL identical to the current one — a duplicate history entry, whereas the un-batched code would have written nothing.

- **Impact:** Only reachable if a single batched action sets a URL-mirrored value and reverts it within the same callback. None of the current call sites (`navigateFromSidebar` always changes the path) do this, so it is theoretical today — but a future caller could trip it.
- **Suggested fix:** At flush time, compare `pendingHref` against the URL as it stood when the batch began and skip the write if they match. Sketch:
  ```ts
  const batchWrites = <T,>(fn: () => T): T => {
    const startHref = batchDepth === 0 ? currentUrl.peek().href : null;
    batchDepth++;
    try {
      return fn();
    } finally {
      batchDepth--;
      if (batchDepth === 0) {
        const href = pendingHref;
        const isPush = pendingIsPush;
        pendingHref = null;
        pendingIsPush = false;
        if (href !== null && href !== startHref) {
          writeHistory(href, isPush);
        }
      }
    }
  };
  ```

**No. 2 — The "already-open tab" path is not actually guarded by the regression suite.**

When I neutralized `batchWrites()`, the test *"costs a single history entry when a tab is already open at the bookmark"* still passed, while the other two failed. That path (`existing` branch in `openBookmark`) calls `app.selectTab(existing.id)` and, when a verse is present, an **async** `selectTranslationAndChapter(...)`. Because that call is `void`ed, any URL write it performs lands *after* the synchronous batch has already flushed — so it would not be batched. The current test doesn't exercise a verse on the existing-tab path (the fixture bookmark has no verse and targets the same chapter), so this gap is invisible.

- **Impact:** Low — the observed bug is on the new-tab path, which is covered. But the existing-tab-with-verse case is exactly where an un-batched async write could re-introduce a stray history entry, and nothing currently proves it doesn't.
- **Suggested fix:** Add a case that (a) opens a bookmark *with a verse* into an already-open tab and asserts the single-entry / intact-back-entry invariants, or drop a bookmark whose verse points at a chapter reachable from the existing tab. If an async escape does show up, the position/verse write in the existing-tab branch may need to be made synchronous or pulled into the batch.

**No. 3 — The full "double back" claim from the PR description isn't verified end-to-end.**

The PR body states going back "returns to the bookmarks list, then to the chapter you were reading before." The test *"returns to the sidebar over the previous chapter when going back"* verifies only the **first** back (to `?sidebar=open`) and asserts the entry underneath still reads `genesis/1`. It never presses back a second time to confirm you actually land on `genesis/1` *without* `sidebar=open`.

- **Impact:** Very low — the underlying entry is asserted intact, so the second back is highly likely correct; it's just not directly proven.
- **Suggested fix:** Extend the test with a second `window.history.back()` and assert the URL equals the pre-sidebar `genesis/1` href.

---

### Nits

**No. 4 — New `NavigationManager` tests don't `dispose()` the managers they create.** Each `createNavigationManager()` patches the shared `window.history.pushState`/`replaceState` and adds `popstate` listeners that stay live for the rest of the file. This matches the existing convention in the file (the current tests don't dispose either, and the `afterEach` only resets the URL), and the new assertions all use relative history-length deltas so accumulation doesn't make them flaky — so this is not a regression, just a standing hygiene note. If you ever see cross-test bleed here, disposing in an `afterEach` would be the fix.

---

### Reproduction (original bug, for reference)

1. On a mobile-width viewport, load the reader (lands on Genesis 1).
2. Open the sidebar and switch to the Bookmarks screen (`?sidebar=open`).
3. Tap a bookmark for a chapter you don't currently have open.
4. **Before:** the reader stays visually on the bookmarks screen and two history entries are recorded for the destination chapter, so pressing back / swiping back appears to do nothing.
5. **After:** the bookmarks screen closes and the reader lands on the chapter; back returns to the bookmarks list, and back again returns to the chapter you were reading.

---

### Verdict

Solid fix with genuine, non-tautological regression coverage and a clean, reusable primitive. The findings above are all low-severity/optional — none block merge. Nice work.

---

KallynGowdy:
@mazenemad could you address the review issues? Apart from that, it looks good to me and seems to work in my testing

mazenemad:
On it

Blitzy:
@KallynGowdy I dont see the original issue happening on my Android device in Chrome on https://alpha.seedbible.org/b/develop/ even without the changes in this PR?

https://github.com/user-attachments/assets/cf5d729f-61d7-4747-9ce1-1db7f725884e

KallynGowdy:
It was happening occasionally, but I think it might have been fixed by #1664

KallynGowdy:
@Blitzy Is there any downside to merging this?

Blitzy:
I dont believe so. Even though i couldnt get the original issue to reproduce, the batchWrite changes here make sense and i think is worth having.

Blitzy (approved):

PR #1667, opened 2026-08-19
Split extensions list into installed and available sections

### LINKED ISSUE #1666: Group extensions by installed and available
Currently, all extensions are just shown in a basic list. This can make it difficult to see at a glance what is installed and what isn't.

We should split the extensions list into two sections: Installed, and Available. All the extensions you have installed should be in the Installed section, while everything else is in the Available section.

KallynGowdy:
## Summary

Reorganized the extensions settings view to separate installed extensions from available ones, improving discoverability and reducing cognitive load when browsing the extension marketplace.

## Changes

- **Extracted `renderExtensionRow` helper** — Moved the repetitive extension row rendering logic into a reusable function to eliminate duplication and improve maintainability.

- **Split extensions into two groups** — The single flat list is now divided into "Installed" and "Available" sections, each with its own heading and empty state message.

- **Added empty state messages** — When no extensions exist in a section (e.g., "You haven't installed any extensions yet"), users now see a clear message instead of an empty list.

- **Updated translations** — Added four new i18n keys (`installed-extensions`, `available-extensions`, `no-installed-extensions`, `no-available-extensions`) to support the new UI structure.

- **Added CSS for grouping** — New `.sb-extensions-group` class provides consistent spacing and layout for each section.

- **Imported `ExtensionListEntry` type** — Added explicit type import from `ExtensionManager` for better type safety.

## Implementation Details

The refactoring filters `extensionsList` into `installedExtensions` and `availableExtensions` at render time, then maps each group through the shared `renderExtensionRow` function. This keeps the component logic clean while maintaining all existing install/uninstall behavior and state icons.

https://claude.ai/code/session_0198MKReA7kQ28rHt9wshaco

Closes #1666


Blitzy:
> [!NOTE]

## Code Review — Split extensions list into installed and available sections

Thanks for this! The change cleanly addresses issue #1666: the flat list is now split into **Installed** and **Available**, each with its own heading and empty state, and the repeated row-rendering logic is extracted into a shared `renderExtensionRow` helper. The extraction is a genuine improvement — it removes the duplicated `stateIcon`/`stateLabel`/action-button block and even drops the stray `{}` expression that was sitting before the title span. The grouping matches the issue's stated intent ("everything you've installed in Installed, everything else in Available"), the filters are disjoint so no extension can appear twice, and the empty-state messaging for each section is a nice touch.

Below are the issues I found, ordered by severity. None are blockers; the main gap is test coverage.

---

### Medium

**1. No tests cover the new grouping/empty-state behavior.**

The PR introduces user-visible logic — partitioning `extensionsList` into two groups and choosing an empty-state message per section — but adds no tests. `changed_files` is 3 (component, CSS, `en.json`), none of them a test file, and there is no existing `SettingsPage` test to extend. The behavior worth protecting is concrete and deterministic:

- A mix of installed and uninstalled entries lands the installed ones under "Installed" and the rest under "Available".
- All entries installed → the "Available" section shows `no-available-extensions`.
- Zero entries installed → the "Installed" section shows `no-installed-extensions`.
- Empty overall list → still the outer `no-extensions-available` state.

*Why it matters:* the partition key here is subtle (see No. 2) and easy to regress silently in a future refactor. A render test that asserts which section each row appears in would catch a mis-grouping that a human wouldn't notice at a glance.

*Suggested fix:* add a component render test (render `ExtensionsSettingsView` with a stubbed `extensions.extensions` signal holding a few `ExtensionListEntry` values, then assert the rows resolve under the correct heading). Keep the mock at the manager boundary per the repo's testing conventions — stub the `extensions` manager's signal, not internal helpers.

---

### Low

**2. An extension being installed ("pending") renders under "Available", not "Installed".**

The section filter keys on `entry.installed` only:

```ts
const installedExtensions = extensionsList.filter((e) => e.installed);
const availableExtensions = extensionsList.filter((e) => !e.installed);
```

But `getExtensionInstallState` derives four states, and `pending` is driven by `pendingInstallation` independent of `installed`. In `ExtensionManager.computeExtensions`, a freshly-requested install has `pendingInstallation: true` while `installed` is still `false` (it isn't in `installedExtensionIds` and isn't registered yet). So while an extension is installing, its row shows the "Installing…" icon **inside the Available section**, then jumps to the Installed section once it completes. It also shows no action button in that window (the install button only renders for `none`, uninstall only for `installed`/`downloaded`).

*Reproduction (for a human dev):* open Settings → Extensions with at least one uninstalled extension, click Install, and watch the row during the install: it stays under "Available" showing "Installing…" until it finishes, then moves to "Installed".

*Assessment:* arguably acceptable — it isn't "installed" until it finishes — so this may be intended. Flagging it so the placement is a deliberate choice rather than an accident. If you'd prefer pending items to sit under Installed, partition on the derived state instead of the raw boolean (e.g. treat `pending`/`downloaded`/`installed` as "in the Installed group").

**3. State tooltip labels remain hardcoded English inside the extracted helper.**

`renderExtensionRow` still uses raw strings for the state icon tooltip:

```ts
const stateLabel =
  installState === "installed" ? "Installed"
  : installState === "downloaded" ? "Downloaded"
  : installState === "pending" ? "Installing…"
  : "Not installed";
```

This PR's theme is i18n (it adds four `t(...)` keys for the headings/empty states), so leaving the adjacent tooltips untranslated is inconsistent. These strings pre-date this PR, but they now live in code this PR touches. Consider routing them through `t(...)` with `defaultValue` fallbacks and adding the keys to `en.json` only (per the repo convention of leaving the other locale files to the translators).

---

### Nit

**4. `sb-settings-empty-state` has no CSS rule.**

The two new empty-state blocks reuse `className="sb-settings-empty-state"`, but that class isn't defined in `SettingsPage.css` (or any other stylesheet) — it was already undefined for the pre-existing outer empty state, so the new usages just inherit the same do-nothing behavior. Not introduced by this PR, but since you're adding two more consumers it may be worth defining the rule (spacing/centering) so the empty states read as intentional design rather than bare paragraphs.

---

### Verification notes

- Grouping matches issue #1666's requirement and the sections are mutually exclusive. ✅
- The four new i18n keys (`installed-extensions`, `available-extensions`, `no-installed-extensions`, `no-available-extensions`) are all added to `en.json` and all consumed — no orphan or missing keys for the i18n lint rule. ✅
- `type ExtensionListEntry` is imported and used as the `renderExtensionRow` parameter type; the extraction preserves the original install/uninstall handlers and state icons. ✅

Overall: a solid, low-risk refactor. The one thing I'd want before merge is a test around the section partitioning (No. 1); the rest are optional polish.

---

KallynGowdy:
Thanks for the review! Pushed a follow-up commit addressing it:

- **#1 (test coverage)** — Added `test/unit/seed-bible/components/ExtensionsSettingsView.test.tsx`, rendering `SettingsPage` with `requestedSettingsView = "extensions"` and a stubbed `state.extensions` signal (manager-boundary mock, per repo convention). Covers: mixed installed/available partitioning, the "everything installed" and "nothing installed" empty states, and the outer empty state for zero extensions. Verified each new assertion actually fails against the pre-split flat-list code before restoring the fix.
- **#3 (hardcoded tooltip labels)** — `stateLabel` now goes through `t("extension-state-installed" | "-downloaded" | "-pending" | "-none", { defaultValue: ... })`, with the four keys added to `en.json` only.
- **#4 (empty-state styling)** — Added a `.sb-settings-empty-state` rule (border/background/padding matching the extension row style) now that there are three consumers of the class.

On **#2 (pending extensions under "Available")**: keeping the current behavior — partitioning on `installed` rather than the derived state. An extension mid-install isn't installed yet, so it staying in "Available" with an "Installing…" indicator until it actually finishes matches the issue's own wording ("everything you've installed" vs. "everything else"). Happy to switch to grouping by derived state if you'd rather it jump to "Installed" as soon as install is requested.

All of `pnpm check:ts`, `pnpm lint`, and `pnpm test` pass (4901 tests, including the 4 new ones).

---

Blitzy:
What if we did Available / Installed as tabs instead of one large run-on (albeit categorized) list?

KallynGowdy:
@Blitzy that could work

Blitzy (approved):
Tabs are so much better

PR #1668, opened 2026-08-20
Feature/1656

### LINKED ISSUE #1656: Move the share button to the quick toolbar
Currently, the share button is available in the toolbar or in the mobile more menu. This works, but it clutters the toolbar. This task is to move the share button from the main toolbar into the quick toolbar so that it is always visible and easy to find.

Image

Additionally, we should change the weird shared session icon on the tabs screen to be a share button as well: 

Image

Currently, it automatically starts a shared session when clicked, but ideally it shows the share modal for the current tab instead.

Kabi4:
Fixes: #1656


Blitzy:
> [!NOTE]

## Code review — PR #1668 (Feature/1656)

Nice, focused change: Share now lives in the mobile chapter header (via the quick toolbar) instead of the bottom **More** menu, and the header actions are wrapped in a flex cluster so bookmark / share / account / settings stay grouped on the right. The unit tests are genuinely reactive (they toggle `isMobile` and re-read visibility, render real components at real viewport widths, and assert Share is present/absent per surface) rather than tautological — that's the right approach.

Two things block merge, though: the branch does **not** type-check (CI `build (24.x)` is already red on `pnpm check:ts`), and the PR only implements half of #1656.

---

### 🔴 High

**No. 1 — Branch fails `pnpm check:ts`; CI `build (24.x)` is red.**

Two type errors, both introduced by this PR:

- `packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx:1201` — `openShareModal` now takes the leaner `ShareSheetContext`, whose `sharedSession` is `BibleReadingSession | null | undefined`. That value is passed straight to `<ShareModal session={context.sharedSession} />`, but `ShareModalProps.session` is `BibleReadingSession | null` (no `undefined`):
  ```
  TS2322: Type 'BibleReadingSession | null | undefined' is not assignable to type 'BibleReadingSession | null'.
  ```
- `test/unit/seed-bible/managers/BibleToolsManager.test.tsx:1200` — `openModal.mock.calls[0][0]` is possibly `undefined` under the project's index-access strictness:
  ```
  TS2532: Object is possibly 'undefined'.
  ```

Because Vitest transpiles without type-checking, the new tests pass locally (I confirmed all 90 pass) while `pnpm check:ts` still fails — so a green test run here is misleading.

**Reproduction:**
```bash
git fetch origin feature/1656 && git checkout feature/1656
pnpm install
pnpm check:ts   # → Visible errors: 2, exit code 2
```
(Also visible in the failed `build (24.x)` job for head `d16b0f1`.)

**Suggested fixes:**
- Source: coalesce at the call site (smallest change) —
  ```tsx
  session={context.sharedSession ?? null}
  ```
  or make `ShareSheetContext.sharedSession` non-optional (`sharedSession: BibleReadingSession | null`) so the two share-capable contexts agree.
- Test: assert on a non-null-checked first call, e.g.
  ```ts
  const firstCall = openModal.mock.calls[0][0];
  expect(firstCall.title).toEqual({ key: "share-sheet-title", defaultValue: "Share" });
  ```

---

### 🟡 Medium

**No. 2 — Only half of #1656 is implemented, but the PR is marked `Fixes #1656`.**

Issue #1656 has two parts:
1. Move Share out of the toolbar/More menu so it's always visible — **done** (mobile header).
2. Change the "weird shared session icon" on the tabs screen so it opens the **share modal for the current tab** instead of auto-starting a shared session.

Part 2 is untouched. The tabs-screen button still auto-creates a session:
- `packages/seed-bible/seed-bible/components/Tabs/Tabs.tsx:2346-2360` (mobile tabs header "New shared session" button) and `:820-836` (More context-menu item) both call `createSharedSessionAndCopyLink(state, t)`, which immediately spins up a shared session rather than showing the share sheet.

Because `Fixes #1656` auto-closes the issue on merge, this would silently drop the second requirement. Either implement part 2 here, or change the PR body to `Refs #1656` / `Part of #1656` and open a follow-up so the issue stays open. Worth a word from a maintainer on intended scope.

---

### 🟢 Low / observations

**No. 3 — Desktop Share was not moved to the quick toolbar.** The issue title says "move the share button to the quick toolbar," but on desktop Share stays on the main reader toolbar (the quick-toolbar Share is gated on `playlists.isMobile`). The linked screenshots are all mobile, so this may be intentional — flagging only to confirm desktop behavior is meant to be unchanged.

**No. 4 — Dead props on the desktop reader quick toolbar.** In `BibleReader.tsx:2211` the `sb-quick-toolbar-reader` instance now receives `sharedSession` / `toast` / `modals` / `app`, but its only consumer (the Share quick tool) is hidden on desktop, so these are never used there. Harmless, but slightly misleading — a short comment or leaving them off would keep the call sites honest.

**No. 5 — Silent no-op if `toast` is absent.** `openShareModal` now does `context.toast?.(...)` on "copy link." In practice both live surfaces pass `state.app.toast`, so the user always gets feedback; just noting that a future caller wiring up Share without `toast` would copy the link with no visible confirmation. The current wiring is fine.

---

### Test notes

The tests cover the visibility matrix well (mobile vs desktop, quick-toolbar vs main-toolbar vs More menu, and the `mobile-navigation-bar` surface). Two small gaps, if you want to tighten:
- No assertion that the quick-toolbar Share passes the **correct `sharedSession`** through to the opened modal (only that a modal opens with the right title).
- Nothing guards the `toast`-optional path from No. 5.

Neither is blocking.

---

**Summary:** Fix the two type errors so CI goes green (No. 1, required), and settle the scope of #1656 — either finish part 2 or downgrade the closing keyword (No. 2). The rest are optional polish.

---

KallynGowdy:
@Kabi4 This looks good. Could we also move the desktop share tool to the quick toolbar as well?
image

Blitzy (approved):
This works great. Merging!

PR #1670, opened 2026-08-20
Add MCP server management for AI chat tool integration

KallynGowdy:
## Summary

Adds support for connecting remote MCP (Model Context Protocol) servers so AI chat participants can use their tools. This ships as an **optional, installable extension** (`packages/mcp-extension/`) rather than part of the core app bundle — users opt in from Settings → Extensions, the same way `apologist-extension`/`bonfire-extension` work. Only remote, HTTP-reachable MCP servers are supported (StreamableHTTP, falling back to legacy SSE), since the whole AI chat call chain already runs client-side with no backend to proxy a local/stdio MCP server through.

## Key changes

- **`packages/mcp-extension/`** (new package): an installable extension, mirroring the existing `apologist-extension`/`bonfire-extension` layout.
  - `ext_MCP/main/MCPManager.tsx` — owns the full lifecycle of MCP server connections: persists user-configured servers to CasualOS records (kept on the default/private marker, not `publicRead`, since configs can carry auth headers), connects on login and reconnects on update, falls back from StreamableHTTP to SSE, wraps MCP tools as `AIProviderFunctionTool`s, exposes them via a `tools` signal, and handles stale in-flight connection attempts (e.g. a server removed mid-connect). A `dispose()` method cleanly tears down live connections and the login-driven load effect when the extension is uninstalled.
  - `ext_MCP/main/AIChatSettingsModal.tsx` — the "AI Chat Settings" UI: add servers (name, URL, optional auth header), toggle enabled/disabled, remove, live connection status per server (connected/connecting/error/disabled) with tool counts, and a warning when a server falls back to SSE while an auth header is configured (browser `EventSource` can't carry custom headers).
  - `ext_MCP/main/init.tsx` — registers the extension, wires `MCPManager`'s tools into `ChatsManager.addContext`/`removeContext` via a small, independently-testable `buildMcpChatContext()` helper (same pattern `audio-reader-extension` uses to keep its `init.tsx` logic unit-testable without exercising the `registerExtension` generator machinery).
  - `extension.json` carries the extension's own translations under the `mcp-extension` i18n namespace — not auto-installed, so it never reaches users who don't opt in.

- **`ChatsManager.tsx`** (core, generic): `IdentifiedLocalChatContext` gains an optional `settingsAction` (`{ label, onClick }`). This is the one small core change the refactor required — the chat header's "Active AI context" menu had no existing plug-in point for extensions, so this adds a minimal, MCP-agnostic hook that any context-contributing extension can use to surface a settings entry, with zero core knowledge of that extension.

- **`FloatingReaderPanels.tsx`** (core): renders a context row's `settingsAction` generically (gear icon, click invokes the action) instead of a hardcoded "AI Chat Settings" item. The "Active AI context" button's visibility reverts to its original condition (only shown when there's an active context) — the extension now always registers a context once installed, even with zero servers configured yet, which is what keeps its settings entry reachable.

- **Bundle isolation**: `@modelcontextprotocol/client` is a dependency of `packages/mcp-extension` only, not core. Verified directly in the production build that the core entry chunk has zero references to it — it's reachable only through the same per-extension dynamic-`import()` mechanism every other extension uses, and the MCP SDK itself is further split into its own chunk, fetched only on first real server connection.

- **Tests**: `test/unit/mcp-extension/main/` covers the manager (persistence, connection lifecycle, StreamableHTTP→SSE fallback, tool wrapping, a regression test for a load-vs-mutation race condition) and the `buildMcpChatContext` wiring helper. `FloatingReaderPanels.test.tsx` covers the generic `settingsAction` rendering.

## Verification

`pnpm check:ts`, `pnpm lint`, and `pnpm vitest run` all pass. `pnpm build:client` was inspected directly to confirm the core bundle carries no trace of MCP/the extension, and that the extension and the MCP SDK each land in their own separately-fetched chunks.

https://claude.ai/code/session_01PgFYudCRD25hWfcbyPt774


Blitzy:
> **NOTE:** This is an automated review posted by a Claude Code routine that Ryan Cook has scheduled to run against pull requests as they're marked ready for review. It is not a human review — treat it as a first pass, not a verdict.

## Review of PR No. 1670 — Add MCP server management for AI chat tool integration

**How this review was done:** single-pass, without the usual multi-agent fan-out/verification subagents (unavailable in this run) — one pass through the diff, the PR description, existing PR comments/review threads, and a search for a linked issue. No linked issue was found in the PR body or via a repo search that specifies requirements beyond the PR description itself, and there are no prior review comments or formal reviews to reconcile — the only existing comments on this PR are the automated deploy-preview and build-size bot posts.

### Findings

**1. High — Removing a server can leave its connection (and tools) live indefinitely if persisting the removal fails.**
`packages/mcp-extension/ext_MCP/main/MCPManager.tsx:359-369`

`removeServer` updates local state, then `await persist(next)`, and only *after that* closes the live connection and clears connection state:
```ts
servers.value = next;
stateVersion++;
await persist(next);
await closeConnection(id);
clearConnectionState(id);
```
If `persist` (i.e. `os.recordData`) rejects — a dropped connection, a records-server error — the last two lines never run. The server disappears from the Settings UI (`servers.value` was already updated optimistically), but its MCP client connection stays open and its tools stay in `readTools`, so they keep being exposed to every tool-calling AI chat. A user who removes a server (plausibly because they no longer trust it) would reasonably expect its tools to stop being callable immediately — not depend on a network round-trip succeeding first. `closeConnection`/`clearConnectionState` should run regardless of whether `persist` succeeds.

**2. High — MCP tool schemas are forwarded to the one tool-calling provider with `strict: true`, which most real MCP servers' schemas won't satisfy.**
`packages/mcp-extension/ext_MCP/main/MCPManager.tsx:236-242`, consumed at `packages/apologist-extension/ext_Apologist/main/init.tsx:221-229`

Each MCP tool's `inputSchema` (an arbitrary JSON Schema from a third-party server) is cast straight through as `parameters` with no normalization:
```ts
parameters: tool.inputSchema as unknown as ZodStandardJSONSchemaPayload<unknown>,
```
`apologist-extension` is the only provider in the codebase with `supportsToolCalling: true` — i.e. the only provider these MCP tools will ever actually be used by — and it sends every tool with `strict: true`. OpenAI's strict function-calling mode requires `additionalProperties: false` and every property listed as `required`; most real MCP servers define schemas with optional parameters and no `additionalProperties: false`. Connecting a typical real-world MCP server (as opposed to the hand-crafted `{type:"object",properties:{}}` schema used in the tests) will very likely get the whole chat-completions call rejected once that tool is included, breaking tool calling for the entire turn rather than failing gracefully for just that one tool.

**3. Medium — `AddMcpServerForm.handleAdd` has no error handling around `mcp.addServer`.**
`packages/mcp-extension/ext_MCP/main/AIChatSettingsModal.tsx:184-206`

```ts
try {
  await mcp.addServer({ name: trimmedName, url: validUrl, headers: … });
  setName(""); setUrl(""); setAuthHeader(""); setError(null);
} finally {
  setIsAdding(false);
}
```
There's no `catch`. If `mcp.addServer` rejects (e.g. `os.recordData` throws on a network error), it becomes an unhandled promise rejection and the user sees nothing — no error message, fields left populated, no indication anything went wrong. The component already has an `error` signal used for URL validation; a failed add just doesn't reach it. `LinkItemInput.tsx`'s `handleAdd` wraps its add logic in `try {…} catch {…}` for exactly this kind of failure — worth mirroring that here.

**4. Medium — No test coverage at all for `AIChatSettingsModal.tsx`.**
`packages/mcp-extension/ext_MCP/main/AIChatSettingsModal.tsx` (323 new lines)

The PR adds unit tests for `MCPManager` and for the `buildMcpChatContext` wiring helper, but there's no test file for the settings modal itself — the URL-validation error path, the SSE+auth-header warning banner, the enable/disable/remove wiring, and the empty-state message are all untested. This is one of the three headline files in the PR description and the most user-facing part of the feature; it's also exactly where a test would have caught finding 3 above (mock `addServer` to reject and assert an error renders).

**5. Low-medium — A new test's assertion doesn't verify what its own comment claims.**
`test/unit/seed-bible/components/FloatingReaderPanels.test.tsx`, test "does not invoke anything on click for a context with no settingsAction"

```ts
// Should not throw when clicked, and the menu content should remain
// (an inert row just prevents default, it doesn't call anything).
expect(() => {
  act(() => { item?.click(); });
}).not.toThrow();
```
The comment says the menu should remain open, but the assertion only checks that clicking doesn't throw — it never checks the menu is still rendered afterward. As written, this test would still pass even if the no-op branch accidentally closed the menu (e.g. `event.preventDefault()` dropped in a future edit), which looks like the actual regression it's meant to guard against.

**6. Low — No test exercises header pass-through to the MCP transport.**
`test/unit/mcp-extension/main/MCPManager.test.ts`

Every test adds servers without `headers`, so `transportOptions` is only ever asserted as `undefined`. Carrying an auth header safely is called out as a specific design point in the PR description (it's the reason the record is kept off the `publicRead` marker) — there's no test confirming a configured `Authorization` header actually reaches `StreamableHTTPClientTransport`'s `requestInit.headers`.

**7. Low — `updateServer` always tears down and reconnects the live connection, even for changes unrelated to connectivity.**
`packages/mcp-extension/ext_MCP/main/MCPManager.tsx:371-382`

`connectServer` unconditionally calls `closeConnection` first, and `updateServer` always calls `connectServer(updated)`. Today the only caller patches `enabled`, so this is harmless in practice, but the public `updateServer` signature also accepts `name`/`url`/`headers`, and a future "rename" UI (a likely next step) would silently drop and re-establish the connection — losing in-flight tool calls and re-listing tools — for a change that has nothing to do with connectivity. Worth gating the reconnect on whether `url`/`headers`/`enabled` actually changed.

---

None of this is a knock on the core approach — the extension/bundle isolation, the load-vs-mutation race handling in `MCPManager`, and the SSE-fallback UX all look solid and are exercised by real tests. Findings 1–3 are the ones I'd want addressed before merging; 4–7 are coverage/robustness follow-ups.

Claude-Session: https://claude.ai/code/session_013maq8JARCJAHCpXQWLyDZN

---

KallynGowdy:
Thanks for the review — all 7 findings addressed:

1. **`removeServer`** now closes the connection and clears connection state *before* `persist()`, so a removed server's tools stop being exposed immediately regardless of whether the records write succeeds.
2. **Strict-mode schema incompatibility**: added `toStrictJsonSchema()` in `MCPManager.tsx` to normalize every MCP tool's `inputSchema` before it's forwarded — `additionalProperties: false`, every property in `required`, with originally-optional properties widened to also accept `null`. `stripNullOptionalArgs()` strips those `null`s back out before the real `client.callTool()` call so the actual MCP server sees an absent key rather than an explicit `null`. Both are covered by dedicated unit tests (edge cases: `anyOf`, typeless/`$ref` schemas, nested objects).
3. **`AddMcpServerForm.handleAdd`** now has a `catch` that surfaces a "Couldn't add the server." error (new `mcp-server-add-failed` key), mirroring `LinkItemInput`'s pattern.
4. **Added `AIChatSettingsModal.test.tsx`** (previously untested): empty state, status/tool-count rendering, the SSE+auth-header warning (and its absence when no headers are set), enable/disable/remove wiring, URL validation, and the new add-failure path. This required switching the component's imports from bare `"seed-bible/*"` specifiers to the `@packages/*` deep-import convention `apologist-extension` already uses, since the bare specifiers only resolve from inside a package that declares `seed-bible` as a dependency, not from test files outside `packages/mcp-extension`.
5. Reworked the "no settingsAction" test to actually assert `event.preventDefault()` was called (confirmed it goes red without the fix) instead of just "doesn't throw."
6. Added a test confirming a configured `Authorization` header reaches `StreamableHTTPClientTransport`'s `requestInit.headers`.
7. **`updateServer`** now only reconnects when `url`/`headers`/`enabled` actually changed, not on every call.

Also merged in the `develop`-sync + build-fix commits that landed on this branch in the meantime and re-ran the full suite (typecheck, lint, all 5425 tests) on top.

Claude-Session: https://claude.ai/code/session_01PgFYudCRD25hWfcbyPt774

---

Blitzy:
> [!NOTE]

## Review of PR No. 1670 — Add MCP server management for AI chat tool integration

**Scope of this review.** I checked out the PR head (`414bc64`), installed dependencies, and ran the affected suites — `test/unit/mcp-extension` + `FloatingReaderPanels.test.tsx`, 77 tests, all green. There is no linked issue in the PR body, and a repo search turned up no open issue describing this feature, so the PR description is the only statement of intent. The one prior review (the automated pass on 2026-09-01) raised 7 findings; I re-read each fix and they are all genuinely present in the code — my notes below are mostly *new* problems, plus a few places where a fix landed without a test that would catch it regressing.

I also pulled `@modelcontextprotocol/client@2.0.0` down from npm and read its type declarations, since the tests mock the SDK entirely and nothing in the repo pins its real shape. The `Client` / `StreamableHTTPClientTransport` / `SSEClientTransport` root exports and the `new Client(info)` → `connect(transport)` → `listTools()` / `callTool({name, arguments})` usage all check out. Two findings below came out of reading those declarations.

---

## High

### 1. Tool names contain `:` and (usually) spaces, which OpenAI-compatible providers reject

`packages/mcp-extension/ext_MCP/main/MCPManager.tsx:337`

```ts
name: `${server.name}:${tool.name}`,
```

`apologist-extension` is the only provider in the repo with `supportsToolCalling: true`, so it is the only consumer these tool names will ever reach. It forwards the name verbatim into an OpenAI-compatible `/api/v1/chat/completions` call, with `strict: true` (`packages/apologist-extension/ext_Apologist/main/init.tsx:227`). OpenAI's function-name rule is `^[a-zA-Z0-9_-]{1,64}$` — letters, digits, underscore, hyphen, 64 characters max. A colon fails that on **every** tool, regardless of server name, and a space fails it for any server the user names naturally ("My Server", "GitHub MCP"). The PR's own test fixture at `MCPManager.test.ts:189` uses `"My Server"`, which would produce `My Server:search`.

Because the rejection happens on the request as a whole, one bad tool name doesn't degrade gracefully — the completions call 400s and `init.tsx:281` throws `Chat completions request failed (400)`, killing the entire chat turn. That's the same failure shape the earlier strict-schema finding was about, just via the name instead of the schema.

Two related problems in the same line:

- **Collisions.** Two servers named the same (or a name whose sanitized form collides) produce duplicate tool names. Dispatch is `chatContext.tools?.find((t) => t.name === fn.name)` (`init.tsx:364`), so a call intended for the second server silently executes against the first. Duplicate names in one `tools` array is also an API error on its own.
- **Length.** `serverName:toolName` can easily exceed 64 characters; long MCP tool names are common.

**Reproduction**

1. Install the MCP Servers extension from Settings → Extensions.
2. Add a server named `My Server` pointing at any reachable remote MCP server; wait for "Connected".
3. Open a chat with the Apologist AI participant and ask it something that would use one of that server's tools.
4. The completions request is rejected and the whole response fails, rather than the model simply not using that tool.

**Suggested fix** — sanitize to the allowed alphabet, use `_` as the separator, and de-duplicate across servers:

```ts
function sanitizeToolNameSegment(value: string): string {
  return value.replace(/[^a-zA-Z0-9_-]/g, "_").replace(/_+/g, "_");
}

/** OpenAI-compatible providers cap function names at 64 chars. */
function makeToolName(
  serverName: string,
  toolName: string,
  taken: Set<string>
): string {
  const base = `${sanitizeToolNameSegment(serverName)}_${sanitizeToolNameSegment(toolName)}`.slice(0, 64);
  let name = base;
  let suffix = 2;
  while (taken.has(name)) {
    const tag = `_${suffix++}`;
    name = `${base.slice(0, 64 - tag.length)}${tag}`;
  }
  taken.add(name);
  return name;
}
```

The de-dupe set needs to span every live connection, not just one server's tool list, since two servers can contribute colliding names. Worth a test asserting a server named `"My Server"` yields a name matching `/^[a-zA-Z0-9_-]{1,64}$/`.

### 2. A failing MCP tool call takes down the entire chat turn

`packages/mcp-extension/ext_MCP/main/MCPManager.tsx:343-350`

```ts
function: async (args: unknown) =>
  client.callTool({ name: tool.name, arguments: … }),
```

`callTool` is returned unguarded, and the consumer doesn't guard either — `init.tsx:370` is a bare `const result = await tool.function(args);` inside the streaming generator. So any rejection (server restarted, TLS/network blip, request timeout, protocol error, transport closed mid-call) propagates out of `generateResponse` and fails the user's whole message instead of being reported back to the model.

This is a new risk class for this codebase rather than a pre-existing one. The existing tools (`insertPlaylistItem`, `getPlaylistState`, …) are local, in-process functions that essentially cannot throw; MCP tools are remote calls to third-party servers that routinely will.

**Reproduction**

1. Connect an MCP server and confirm its tools appear.
2. Stop that server (or block it at the network level) while leaving the chat open.
3. Ask the AI something that makes it call one of those tools.
4. The chat response dies with the transport error instead of the model seeing "that tool failed" and carrying on.

**Suggested fix** — return an error result rather than throwing, so the model can recover within the same turn:

```ts
function: async (args: unknown) => {
  try {
    return await client.callTool({ name: tool.name, arguments: … });
  } catch (err) {
    return {
      isError: true,
      content: [
        {
          type: "text",
          text: `Tool call failed: ${err instanceof Error ? err.message : String(err)}`,
        },
      ],
    };
  }
},
```

`callTool` already returns `{ isError: true, … }` for tool-level failures, so this keeps one consistent shape. A test that makes the mocked `callTool` reject and asserts the wrapper resolves with `isError` would lock it in.

---

## Medium

### 3. The SSE fallback throws the auth header away, even though the SDK can carry it

`packages/mcp-extension/ext_MCP/main/MCPManager.tsx:311`

```ts
client = new Client(clientInfo);
await client.connect(new SSEClientTransport(url));   // no options at all
```

The premise behind the warning banner — "browser `EventSource` can't send custom headers" (`MCPManager.tsx:43-48`, and the `mcp-server-auth-header-sse-warning` string) — doesn't hold for this SDK version. `SSEClientTransportOptions` in `@modelcontextprotocol/client@2.0.0` exposes:

- `requestInit?: RequestInit` — "Customizes recurring `POST` requests to the server." In the SSE transport, the GET stream only *receives*; every JSON-RPC request the client sends (`initialize`, `tools/list`, `tools/call`) goes out as a POST. So the header could be attached to all of the outbound traffic with the same `{ requestInit: { headers } }` object already built for StreamableHTTP.
- `fetch?: FetchLike` — "Custom fetch implementation used for all network requests", which covers the stream leg too (the SDK uses the `eventsource` package, not the browser built-in).
- `authProvider?: AuthProvider` — the SDK documents `{ token: async () => myApiKey }` as the simple bearer-token path, which handles both legs.

So an SSE-only server behind auth is currently guaranteed to fail with 401, and the UI tells the user it's a browser limitation rather than a gap in the client code.

Related, in the same block: the fallback `catch` at `MCPManager.tsx:306` is unconditional. A StreamableHTTP 401 (wrong token) triggers an SSE attempt whose *different* error is what ends up in `connectionState.error`, so the surfaced message points away from the real cause.

**Suggested fix** — pass the credentials through and keep the warning only for cases the transport genuinely can't cover:

```ts
const sseOptions = server.headers
  ? { requestInit: { headers: server.headers } }
  : undefined;
await client.connect(new SSEClientTransport(url, sseOptions));
```

### 4. Nested optional arguments still reach the MCP server as explicit `null`

`packages/mcp-extension/ext_MCP/main/MCPManager.tsx:136-147`

`toStrictJsonSchema` recurses all the way down — the PR's own test at `MCPManager.test.ts:563` confirms a nested object gets `required: ["inner"]` + `additionalProperties: false`, and its optional properties get widened to accept `null`. But `stripNullOptionalArgs` only walks the top level of the arguments object, so the `null`s the strict schema *invited* at depth are forwarded to the server verbatim. The real MCP server validates against its own original schema, which doesn't allow `null` there.

I confirmed this directly against the PR head:

```ts
// schema: { filter: { type: "object", properties: { since, tag }, required: ["tag"] } }
stripNullOptionalArgs({ filter: { tag: "x", since: null } }, new Set(["filter"]))
// → { filter: { tag: "x", since: null } }   ← `since: null` survives
```

**Reproduction** — connect any MCP server with a tool taking a nested options object that has an optional field (very common: `{ filter: { tag, since? } }`), and ask the model to use the tool without that field. The server sees `since: null` and rejects it as a type error.

**Suggested fix** — drive the stripping from the original schema instead of a flat key set, recursing wherever the schema had a nested object:

```ts
export function stripNullOptionalArgs(
  args: unknown,
  schema: JsonSchemaRecord | undefined
): unknown {
  if (!args || typeof args !== "object" || Array.isArray(args)) return args;
  const properties = (schema?.properties ?? {}) as JsonSchemaRecord;
  const required = new Set(
    Array.isArray(schema?.required) ? (schema.required as string[]) : []
  );
  const cleaned: Record<string, unknown> = {};
  for (const [key, value] of Object.entries(args as Record<string, unknown>)) {
    if (value === null && !required.has(key)) continue;
    cleaned[key] = stripNullOptionalArgs(
      value,
      properties[key] as JsonSchemaRecord | undefined
    );
  }
  return cleaned;
}
```

### 5. With only this extension installed, the settings screen is unreachable

`packages/seed-bible/seed-bible/components/FloatingReaderPanels/FloatingReaderPanels.tsx:908`

```ts
{state.chats.activeContexts.value.length > 0 &&
 selectedChatHasToolCallingProvider ? ( … )}
```

The "Active AI context" button — the extension's *only* entry point — is gated on the selected chat having an AI participant whose provider sets `supportsToolCalling: true`. `apologist-extension` is the only provider in the repo that does. So the user journey is: install "MCP Servers" from Settings → Extensions → open a chat → …nothing. There's no gear anywhere until Apologist is also installed *and* added as a participant in the chat you're looking at.

`packages/mcp-extension/extension.json` declares no `dependencies`, even though the mechanism exists and several extensions use it (`bible-stack`, `scripture-map`, `transcript-ui-extension`, `twitchPub-extension`).

**Reproduction** — on a fresh profile, install only the MCP Servers extension and try to add a server. There is no reachable path to the AI Chat Settings modal.

**Suggested fix** — at minimum declare `"dependencies": ["apologist-extension"]` in `extension.json` so the requirement is enforced and visible; better still, also register an entry in the Settings page so the modal is reachable without a chat open, since the server list is per-user global config rather than per-chat state.

### 6. The enable/disable toggle reads live connection status instead of the server's own `enabled` flag

`packages/mcp-extension/ext_MCP/main/AIChatSettingsModal.tsx:126`

```ts
onClick={() => onToggleEnabled(status === "disabled")}
```

`status` is the *connection* status, and a server can have no connection-state entry at all — in which case the row renders as "Disabled" (the `statusLabel` ternary and `statusIcon` both fall through to the disabled branch) even though `server.enabled === true`. Clicking then sends `{ enabled: true }`, `updateServer` computes `connectionRelevantChange === false` because nothing actually changed, and no reconnect is attempted. The button is inert and the row is stuck.

That state is reachable today: `addServer` writes to `servers.value` before awaiting `persist`, so if the record write throws, the server is in the list but `connectServer` was never called (`MCPManager.tsx:457-461`). The user sees "Couldn't add the server." *and* a listed server that reads Disabled and won't turn on.

**Suggested fix** — drive the toggle and its label from config, and keep `status` for the status line only:

```ts
onClick={() => onToggleEnabled(!server.enabled)}
```

…with the aria-label/title/icon ternaries switched to `server.enabled` as well. A test covering "server with `enabled: true` and no connection-state entry" would cover both this and the orphan-row case.

### 7. Test gaps — two of the previous round's fixes would pass either way

The suite is genuinely good (89% statement coverage on the package, real async polling via `waitForCondition` rather than fixed sleeps, error paths covered). Three gaps, listed because they're specifically the "does this test actually test anything" kind:

- **(a) The `removeServer` ordering fix has no regression test.** I verified this by reverting the fix on the PR head — moving `await persist(next)` back above `closeConnection`/`clearConnectionState` — and re-running: **all 41 `mcp-extension` tests still pass.** The fix only matters when `persist` *fails*, and no test makes it fail. A test that mocks `recordData` to reject and then asserts `manager.tools.value` is empty and `close` was called would fail on the old ordering and pass on the new one.
- **(b) The `updateServer` "only reconnect when connection-relevant fields changed" fix has no test either.** The only `updateServer` test toggles `enabled`, which *is* connection-relevant. A test doing `updateServer(id, { name: "Renamed" })` and asserting `close` was not called would pin the new behavior.
- **(c) The add-failure test exercises the less likely failure mode.** `AIChatSettingsModal.test.tsx:327` mocks `addServer` to *reject*, which is what the new `catch` handles. But `os.recordData` is a CasualOS records call that signals most failures by *returning* `{ success: false, errorCode, … }` rather than throwing, and `persist` (`MCPManager.tsx:419`) ignores the return value entirely. So the common real-world failure — the record write being rejected — produces no error, no console warning, and a server that quietly vanishes on next login. Fire-and-forget matches `BookmarksManager`/`HighlightsManager`, so this isn't out of line with the codebase, but the new catch gives more coverage than it actually has. Checking `result.success` in `persist` (as `AnnotationsManager.tsx:481` does) would make the error path real and make the test meaningful.

### 8. Security posture worth an explicit decision: MCP tools in shared chats

Not a defect in the diff, but it should be a deliberate choice rather than a default. MCP tools run with the user's stored credentials, with no per-tool or per-call confirmation, and `ChatsManager` merges them into the context used for shared sessions too (`ChatsManager.tsx:1646`), where `messages` includes text authored by other participants. That means third-party text sits in the same prompt as a tool-calling model holding the local user's credentials — a prompt-injection surface that doesn't exist for the current local-only tools.

Options worth weighing: exclude MCP tools from shared sessions; require a one-time per-tool approval before first use; or at least surface in the settings modal which tools a server exposes, so the user sees what they're granting. Whatever the call, it's worth stating it in the PR description since this is the first feature to put user credentials behind the chat's tool loop.

---

## Low

### 9. `removeServer` / `updateServer` failures are silent

`AIChatSettingsModal.tsx:321,323` call both with `void`, so a rejected persist is an unhandled rejection with no UI feedback — the row disappears (or the toggle flips) and then comes back on next login. `addServer` got a `catch` in the last round; these two didn't. Given finding No. 7(c), the cleanest fix is for `persist` to check `result.success` and for all three mutators to surface the failure the same way.

### 10. Root `package.json` also declares `@modelcontextprotocol/client`

The PR description says the SDK is "a dependency of `packages/mcp-extension` only, not core", but the diff adds it to the root `package.json` `dependencies` as well — and the root package *is* the core app package (`"name": "seed-bible"`, the one `packages/mcp-extension` links via `workspace:*`). pnpm links the SDK into `packages/mcp-extension/node_modules/` from the extension's own manifest, and nothing under `packages/seed-bible/`, `standalone/`, or `server/` imports it, so the runtime isolation claim itself still holds.

If the root entry is there so the `vi.mock("@modelcontextprotocol/client")` id resolves from `test/`, that's a fine reason — worth a one-line note in the PR description. If not, dropping it keeps the manifest honest and means a future static import from core fails loudly instead of silently pulling 210 KB into the core chunk.

### 11. No CHANGELOG entry

`CHANGELOG.md` has a live `## TBD` section and recent feature PRs each add a line with a PR link (see the v1.7.0 entries). This adds a user-facing installable extension with nothing under `### ✨ Added`.

### 12. Small things

- **`MCPManager.tsx:320-321`** — the comment says `listTools` pagination "isn't implemented for this first version", but `Client` in SDK v2 aggregates list pages internally (bounded by its `listMaxPages` option). The comment describes v1 behavior.
- **`MCPManager.tsx:303`** — `client.connect` has no timeout or cancel. A server that accepts the connection and then hangs leaves the row at "Connecting…" indefinitely with no way out except removing it. A `setTimeout`-based abort with a clear error would be friendlier.
- **`FloatingReaderPanels.tsx:950-959`** — the gear is a `MaterialIcon` with `aria-label`/`title` but no button semantics; the whole row is the click target. Screen-reader users get an announced control that isn't focusable. Either drop the `aria-label` (it's decorative — the row does the work) or make it a real `<button>`.
- **`updateServer`'s header comparison** uses `JSON.stringify(updated.headers ?? null) !== JSON.stringify(existing.headers ?? null)`, which is key-order sensitive. Harmless today since headers are only ever set at add time, but it will report a spurious change once a header-editing UI exists.

---

## What's good

The parts the previous round flagged are properly fixed, not papered over — `toStrictJsonSchema`/`stripNullOptionalArgs` are real, well-factored, separately-tested functions rather than an inline patch, and the `anyOf`/typeless/`$ref` edge cases are all handled. Beyond that: the stale-attempt token scheme in `connectServer` is the right shape for a mutable connection list, the `stateVersion` load-vs-mutation guard is a genuinely subtle race caught with a genuinely good regression test, `addContext`'s use of `.peek()` means the `effect` in `init.tsx` doesn't self-trigger, and keeping the config record off the `publicRead` marker (with a test asserting it) is exactly right for something holding bearer tokens. Extracting `buildMcpChatContext` so the wiring is testable without the `registerExtension` generator is a nice touch, and the `settingsAction` hook on `IdentifiedLocalChatContext` is about as small and extension-agnostic as a core change for this could be.

Findings 1 and 2 are the ones I'd want fixed before merge — both break the whole chat turn rather than degrading, and both will fire on the first real server a user connects. 3–8 are worth resolving in this PR; 9–12 are fine as follow-ups.

---

PR #1671, opened 2026-08-20
Improve the draggability of the mobile verse toolbar

KallynGowdy:
- Larger dragging target area
- It is now easier to close the mobile verse toolbar by dragging it

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.css
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.css
@@ -1054,6 +1054,15 @@
   border-bottom-left-radius: 0;
   border-bottom-right-radius: 0;
   box-shadow: 0 -8px 24px var(--sb-shadow-color);
+  /* The whole panel is a drag surface for the expand/collapse/dismiss
+     gesture (see the pointer handlers on this element), not just the handle
+     below. Without this, the browser can win the race against our handler
+     on the first pointermove or two and start a native scroll instead,
+     which it then follows up with a `pointercancel` — the drag looks like
+     it starts, then snaps back to where it began. The color-swatch strip
+     opts back into horizontal panning below since it needs its own native
+     scroll. */
+  touch-action: none;
 }
 
 /* Drag target around the grab handle. The pill is only 4px tall, so the
@@ -1293,6 +1302,11 @@
   -webkit-overflow-scrolling: touch;
   scrollbar-width: none;
   overscroll-behavior-x: contain;
+  /* Opts back into native horizontal panning against the panel's
+     `touch-action: none` — this strip owns its own swipe gesture and isn't
+     part of the sheet's drag-to-expand surface (see the pointer-handler
+     ignore list in the component). */
+  touch-action: pan-x;
 }
 
 .sb-verse-toolbar-mobile .sb-verse-toolbar-swatches::-webkit-scrollbar {
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
@@ -1009,10 +1009,13 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
    *
    * The sheet follows the finger rather than snapping at a threshold: dragging up
    * grows the overflow row a pixel at a time, dragging back down shrinks it, and
-   * dragging down on an already-collapsed sheet slides the whole sheet toward the
-   * bottom of the screen to dismiss it. Releasing settles to whichever resting
-   * position the gesture ended up nearest, so a half-finished drag animates the
-   * rest of the way instead of being abandoned.
+   * once the overflow row is fully closed — whether the drag started collapsed or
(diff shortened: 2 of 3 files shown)


Blitzy:
> [!NOTE]

## Summary

This PR makes the mobile verse toolbar (bottom sheet) easier to drag in two ways:

1. **Larger drag target** — the whole panel now starts the expand/collapse/dismiss gesture, not just the thin grab handle. Buttons, inputs, and the highlight-color strip are excluded so they keep their own tap/scroll behavior.
2. **One continuous drag to dismiss** — dragging down from an expanded sheet now closes the overflow drawer *and* keeps sliding into a dismiss in a single motion, instead of needing a release and a second drag.

Overall this is a clean, well-documented change. The logic refactor (`distancePastClosed`) is correct, the code comments explain the *why* thoroughly, and the pointer handlers were consolidated onto the panel sensibly (`event.currentTarget` stays consistent because down/move/up/cancel are all wired to the same element now). No linked issue is referenced in the PR, and the only existing comment is the deploy bot.

I found no correctness bugs. The items below are UX considerations and maintainability notes — nothing blocking.

### Verification I ran

- `pnpm vitest run BibleReaderToolbar.test.tsx` → **46 passed**.
- Confirmed the two new dismiss tests are genuine regression tests: I temporarily reverted the `dismissOffset` logic to the old `startReveal === 0` formula and the "dismisses the selection from one continuous drag starting expanded" test went red, then restored it.

---

## Findings by severity

### Low

**No. 1 — The whole panel is now a tap-to-toggle surface, and `touch-action: none` disables text selection on it.**
Because the tap-vs-drag logic lives on the panel, a tap (movement under `VERSE_SHEET_TAP_SLOP`) on *any* non-excluded area now toggles the drawer — including the empty gaps between buttons and the verse-reference text (`.sb-verse-toolbar-ref-text`). Combined with `touch-action: none` on the panel, the reference text also can't be tap-selected/copied on touch anymore. This is a reasonable consequence of "make the whole panel a target," but worth a conscious confirmation that toggling on a stray tap in the reference area is desired.
- *Repro:* On a narrow viewport, select a verse to open the bottom sheet, then tap the verse-reference text (or empty space between action buttons). The overflow drawer toggles. Try to long-press/select the reference text — nothing selects.
- *Possible fix (if undesired):* add `.sb-verse-toolbar-ref` (or the ref-text span) to `VERSE_SHEET_DRAG_IGNORE_SELECTOR`, or only treat taps as toggles when they land on the handle/hint region.

**No. 2 — `touch-action: none` on the entire panel is a latent constraint on scrollable sheet content.**
Applying `touch-action: none` to `.sb-verse-toolbar-mobile` suppresses native vertical scrolling for everything on the panel except the `pan-x` swatch strip. Today the expanded overflow row is a small, fixed set of action buttons, so there's nothing to scroll and this doesn't bite. But if the expanded content ever exceeds the viewport (many registered verse-toolbar tools, very small screen, or large font scaling), the content would become unscrollable by touch. Not a bug in this PR — just a constraint to keep in mind if the sheet's contents grow. Worth confirming the expanded sheet is capped to a height that always fits.

**No. 3 — The drag-ignore selector is tag-based, so future non-`<button>` controls won't be excluded automatically.**
`VERSE_SHEET_DRAG_IGNORE_SELECTOR = "button, input, a, .sb-verse-toolbar-swatches"` correctly covers every interactive element in the sheet *today* (all verse tools and the close/clear/color controls are real `<button>`/`<input>` elements). But if someone later adds an interactive element as a `<div role="button">`, `<select>`, `<textarea>`, or `<label>`, it would fall through and start the sheet drag / have its `pointerup` captured. Consider a short maintenance comment noting that new interactive controls must be a real button/input/anchor or be added to this list, or broaden the selector (e.g. include `[role="button"], select, textarea, label`) defensively.

---

## Tests

The tests exercise real behavior rather than rubber-stamping the implementation, which is the main thing to get right here:

- The two dismiss tests ("continues straight into the dismiss slide…" and "dismisses the selection from one continuous drag starting expanded") assert on observable results — the rendered `translateY` transform, the overflow height collapsing to `0px`, and the selection actually being cleared (`readingState.selectedVerses`). I verified they fail against the pre-fix `dismissOffset` logic, so they lock in the new behavior.
- "expands the sheet when dragged from the panel background" fails on the pre-fix code (the panel had no `onPointerDown` on small screens), so it genuinely covers the new whole-panel entry point.
- "does not start the sheet drag when pressing down on a toolbar button" is a valid guard: dropping `button` from the ignore selector would make it go red (the button press would start a drag and add `sb-verse-sheet-dragging`).

### Minor test-coverage suggestions (non-blocking)

- No test covers the `isHighlightPickerOpen` early-return in `handleVerseSheetPanelPointerDown`. A test that opens the highlight picker and confirms a downward drag on the panel does **not** move/dismiss the sheet would lock in that branch (it's the reason the whole-panel drag is intentionally disabled in picker mode).
- No test covers that pressing on the `.sb-verse-toolbar-swatches` strip does not start the sheet drag (the selector's last entry). A small assertion there would protect the swatch-strip carve-out.

---

KallynGowdy:
For issue 1 I actually think this is preferrable. Users can always copy the verse and long-tapping the reference to copy it is prone to other issues (like accidentally clicking the bookmark button), so we shouldn't expect users to try to do that on their own.

Number 2 is a non-issue for now.

Same with number 3. If anyone adds another control but doesn't test it, then that is an issue there and not here.

PR #1672, opened 2026-08-21
Added Conditional Profile Picture Redering on various places

### LINKED ISSUE #1655: Change the user profile icon to be a generic profile icon instead of avatar+color unless the user is in a shared session
Currently, anonymous users (or users without a profile picture) see a randomly generated combination of avatar+color (like yellow tree) in place of where their profile picture would be. This can be confusing for new users, and it isn't clear that you have to click on this icon to login or see your profile.

Instead, when the user isn't logged in or doesn't have a profile picture, we should show a generic icon (such as `account_circle`) when they are also not in a shared session. 

When they are in a shared session or in a chat with other users it should show the avatar+color combination icon. The same principle applies for other areas, like annotations. When it is just you, show the generic icon. When other people are present, show the avatar+color combo.

Kabi4:
Fixes: #1655


Blitzy:
> [!NOTE]

## Review — PR #1672: Conditional profile picture rendering

Reviewed against issue #1655. The change delivers exactly what the issue asks for: when you have no profile picture and nobody else is around, avatars now show a generic `account_circle` icon instead of the random animal+color combo; the animal+color still appears in shared sessions, chats that include other people, and annotation lists with other authors so people can tell each other apart. The precedence order (profile picture → generic → animal) is applied consistently in the `Avatar` component, the shared helpers (`chatHasOtherPeople`, `annotationListHasOtherAuthors`) are small and well-documented, and the changelog is updated.

**Verification performed locally:** the 5 affected test files pass (274 tests), `pnpm check:ts` is clean, and — addressing the "do the tests actually test something" concern — I confirmed the new generic-icon tests are genuine regression tests: temporarily disabling the `genericFallback` branch in `Avatar.tsx` makes 5 of them fail (ChatView ×2, DiscoverPane ×1, TabsSidebarLayout ×2). They do not pass vacuously.

---

### 🟠 Medium

**No. 1 — "Other people present?" is decided from *active* participants in some places and *all* participants in others, so your own avatar can render differently across surfaces at the same time.**

`ChatSession` exposes two signals: `participants` (active only) and `totalParticipants` (active + inactive — see `ChatsManager.tsx:387-394`). The PR mixes them:

- `ChatView.getMessageAvatar` and the ChatView event avatars use `chat.participants.value` (active only).
- `Tabs.isInMultiUserIdentityContext` (which drives the sidebar/mobile self-avatar) uses `chat.participants.value` (active only).
- `FloatingChatPanel` and `ChatListAvatarCluster` in `FloatingReaderPanels.tsx` use `chat.totalParticipants.value` (all).

Because of this, in a 1-on-1 chat whose other member is currently **offline/inactive**, the same logged-in-without-a-picture user is judged "alone" by some surfaces and "with others" by others simultaneously.

Reproduction:
1. Sign in without a profile picture.
2. Have a 1-on-1 chat with another person, and make sure that person is currently inactive/offline (only you active).
3. Observe the sidebar self-avatar and your own message bubbles in `ChatView` → generic `account_circle`.
4. Observe the same chat's row in the chat list cluster (`ChatListAvatarCluster`) / the floating-panel member list → animal+color for you.

Per issue #1655 ("in a chat with other users it should show the avatar+color combination"), a 1-on-1 chat with someone who is momentarily offline is still "a chat with another user," so the more correct behavior is to treat the peer as present. Suggested fix: standardize on `totalParticipants` for the "other people present" decision everywhere (ChatView, the sidebar's `isInMultiUserIdentityContext`, and the floating panels), ideally by routing all callers through a single helper so the surfaces can't drift again. If active-only is actually intended for the message list, please make the sidebar and cluster match it — the important thing is that one user's identity looks the same everywhere at a given moment.

---

### 🟡 Low

**No. 2 — The `BibleReaderToolbar` wiring of `otherPeoplePresent` has no test.**

`BibleReaderToolbar.tsx` now passes `otherPeoplePresent={annotationListHasOtherAuthors(selectionAnnotations.value, ...)}` into `VerseToolbarAnnotationGroup`, but no test covers that path. The underlying helper (`annotationListHasOtherAuthors`) and the parallel DiscoverPane path are both tested, so the risk is low, but a small test asserting the verse-toolbar annotation author avatar switches between generic and animal based on selection authors would close the gap and guard against future regressions in this fourth call site.

**No. 3 — Your own generic icon inside a chat carries the `sb-tab-user-icon-self` ring.**

In chat surfaces, `getParticipantAvatar` sets `isSelf: participant.isSelf`, so your own generic `account_circle` renders with the `sb-tab-user-icon-self` box-shadow ring — whereas the sidebar `SelfAvatarVisual` renders the generic icon with no ring (it doesn't pass `isSelf`). Not a bug, but worth a quick visual check that the ring around a borderless/transparent generic icon looks intentional, and that "self ring" treatment is consistent between the sidebar and chat.

---

### ⚪ Nitpick

**No. 4** — In the ChatView event-avatar `.map`, `chatHasOtherPeople(chat.participants.value)` is recomputed once per participant in the group. It's cheap, but hoisting it to a single value above the `.map` (as `getMessageAvatar` already does) would be marginally cleaner and read consistently.

---

### Summary

Solid, well-tested feature that matches the issue. No correctness/crash bugs. The one thing I'd ask to be resolved before merge is **No. 1** (active-vs-total participant inconsistency) since it undercuts the very "don't confuse people about identity" goal the issue is chasing; the rest are polish.

---

KallynGowdy:
@Kabi4  this looks good to me. Just needs to fix merge conflicts

PR #1673, opened 2026-08-21
Added Ability to have text direclty into AI chat with auto height the Input box

### LINKED ISSUE #1624: Add "Ask AI" verse toolbar tool
A really handy feature I keep wanting to have is a shortcut to asking an AI agent about some verses. Currently I have to copy the verse to the clipboard, open chat, select an AI like Apologist and then paste the verses into chat and ask my question. This is a lot of steps and something that I'm finding myself wanting to do quite often and so I've thought an Ask AI tool would be handy in the verse toolbar.

Ask AI Verse Tool Flow:

**1**: Select verses

**2**: Tap "Ask AI"

**3a**: If only one agent available, open chat for agent.

**3b**: If multiple agents available, show list of available agents and let user select one.

**4:** Automatically insert verse into input field with two newlines.

**5**: User types in question. 

**6**: Press submit.

Kabi4:
Fixes #1624


Blitzy:
> [!NOTE]

## Code Review — PR #1673 ("Ask AI" verse toolbar tool)

Reviewed against issue #1624. The PR delivers the flow the issue asks for: select verses → **Ask AI** → pick an agent (only when more than one exists) → chat opens with the verses already in the compose box, followed by two blank lines, ready for the user's question. I ran the suite, type check, and lint locally:

- `pnpm vitest run` on the three touched test files — **220 passed**.
- `pnpm check:ts` — **0 errors**.
- `pnpm lint` (includes i18n key validation) — **0 errors** (only pre-existing CSS `use-baseline` warnings).

I also confirmed the new auto-height tests are **not** vacuous: I temporarily neutralized the resize effect in `ChatView.tsx` and the three height-related tests went red, then passed again once restored. Good — they exercise real behavior rather than always succeeding.

Overall this is a solid, well-tested PR. There are no correctness bugs that break the happy path. The notes below are mostly design/robustness points to consider.

---

### Medium — worth a decision before merge

**M1. "Ask AI" reuses the most recent existing chat with that agent, rather than starting fresh — this isn't in the issue spec and may surprise users.**

`openAskAiForSelectedVerses` (`packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx:572`) calls `findLocalChatForProvider` and, if a prior local chat with that agent exists, drops the new verses into that existing conversation instead of opening a new one.

Issue #1624's flow (steps 3a/3b/4) describes "open chat for agent" and "insert verse into input field" — it doesn't mention reuse. Reuse has a real trade-off: asking about *unrelated* verses appends to an old thread, so the agent answers with the earlier conversation as context, and the earlier exchange scrolls up out of view. That may be the desired continuity, or it may be confusing. This is a product call, so flagging it rather than asserting it's wrong.

- If reuse is intended, consider saying so in the PR description / CHANGELOG so it's a documented choice.
- If a fresh chat per "Ask AI" is preferred, remove the `findLocalChatForProvider` branch and always call `createLocalSession()`.

_Reproduction:_ On the alpha build (https://alpha.seedbible.org/b/feature_1624), with at least one AI agent enabled: select John 3:16, Ask AI, send a question. Then navigate elsewhere, select an unrelated verse (e.g. Genesis 1:1), Ask AI again for the same agent — note the new verse lands in the *same* thread beneath the earlier one.

---

### Low

**L1. `composerDraft` is written before a chat is secured, and isn't cleared on the early-return path.**

In `openAskAiForSelectedVerses` (`BibleToolsManager.tsx:567-576`):

```ts
const verseText = formatSelectedVerses(context.readingState);
if (context.chats.composerDraft) {
  context.chats.composerDraft.value = verseText ? `${verseText}\n\n` : "";
}

const existingChat = findLocalChatForProvider(context.chats, providerId);
const chat = existingChat ?? context.chats.createLocalSession?.();
if (!chat) {
  return; // composerDraft was already set, and is left populated here
}
```

If `chat` is falsy, the function returns with `composerDraft` still holding the verse text. Because `ChatView` consumes `composerDraft` whenever it next mounts, that stale value would then be injected into an unrelated chat the user opens later.

In practice this path is **not reachable in production** — the real `createLocalSession` (`ChatsManager.tsx:2292`) always returns a session, never null — so today this is defensive-only. But the ordering is a latent trap for a future change. Safer to set the draft only once a chat is in hand:

```ts
const existingChat = findLocalChatForProvider(context.chats, providerId);
const chat = existingChat ?? context.chats.createLocalSession?.();
if (!chat) {
  return;
}
const verseText = formatSelectedVerses(context.readingState);
if (context.chats.composerDraft) {
  context.chats.composerDraft.value = verseText ? `${verseText}\n\n` : "";
}
chat.addParticipant(providerId);
// ...
```

There's no test covering "chat creation fails after the draft was set," which is why the gap is invisible — a test that stubs `createLocalSession` to return `undefined` and asserts `composerDraft.value === ""` would lock in the safer ordering.

**L2. The prefill's correctness quietly depends on `ChatView` not being keyed per chat.**

`FloatingReaderPanels.tsx:960` renders `<ChatView chat={selectedChat} state={state} />` with no `key`, so switching chats reuses the same component instance and the local `draft` signal persists across the switch. When "Ask AI" reuses/opens a *different* chat than the one currently displayed, the prefill survives the switch partly because that `draft` carries over. If someone later adds `key={selectedChat.id}` (the natural fix for per-chat draft isolation), the reuse-existing-chat prefill could be lost. Not a bug today — just a coupling worth a comment near the `composerDraft` effect so it isn't broken by an unrelated refactor.

**L3. Every agent in the multi-agent picker shows the same icon.**

In the `getItems` branch (`BibleToolsManager.tsx`), each provider item uses `icon: AskAiIcon`, so the picker is a list of identical icons distinguished only by name. Minor, but the providers already carry an `iconUrl` (`ChatsManager.tsx:1817`) — using it would make the picker easier to scan.

**L4. i18n churn from the conflict resolution.**

`en.json` moves the whole `annotation-conflict-*` / `annotations-pending-sync*` block from the middle of the file to the end (a `Fix conflcits` artifact) in addition to adding the one new `ask-ai` key. It's harmless — I confirmed there are no duplicate keys and lint's i18n validation passes — but it inflates the diff and will read as unrelated changes to a human reviewer. Consider restoring those keys to their original position so the only `en.json` change is `+ "ask-ai": "Ask AI"`.

---

### Nits

- `AskAiIcon` (`BibleToolsManager.tsx:511`) is a one-line wrapper around `AskIcon`; you could reference `AskIcon` directly, though the wrapper is harmless and keeps the icon-component convention used by the other tools.
- The comment explaining why the tool defines both `onSelect` and `getItems` (default tools bypass `validateToolActions`) is accurate and helpful — I verified the render path (`BibleReaderToolbar.tsx:2482-2501`) does fall back to `onSelect` when `getItems` returns an empty array. Good call documenting it.

---

### Test quality

The tests are genuinely testing behavior, not tautologies:
- Height tests fail when the resize effect is removed (verified by reverting locally).
- The `ask-ai` reuse tests fail if `findLocalChatForProvider` misbehaves (they assert `createLocalSession` is / isn't called and which chat id is selected).
- Shared-chat and wrong-agent chats are correctly asserted as *not* reused.
- Mobile focus-suppression, late-arriving `composerDraft`, empty-draft-ignored, and missing-`composerDraft` cases are all covered.

One structural note: `BibleToolsManager.test.tsx` mocks the entire `ChatsManager` sibling rather than the OsManager/SDK boundary that `CLAUDE.md` prefers. This follows the pre-existing pattern in that file (the earlier `open-chat` visibility tests already mock `chats`), and the tool only touches a small, injected slice of the context, so it's a reasonable pragmatic choice — noting it only for consistency awareness, not as a blocker.

---

KallynGowdy:
Actually @Kabi4 I think it would be better if it reused the same chat if possible instead of making a new one all the time. I know the review said to make new chats, but it feels a lot more powerful to me if we reuse the same chat for different queries so that the AI can follow along with the user.

What do you think @Blitzy ?

Blitzy:
Hmmm yeah I think I agree with @KallynGowdy that by default it would be more useful to let AI agents continue a conversation with people asking about verses. I think we should continue an old thread instead of making a new one, and if the desire to quickly make new ones grows we can revisit with a UI redesign for this flow. 

However I did find one 🔴 blocking bug: the AI Chat verse toolbar doesn't seem to work at all when used in the mobile view. You tap it and it doesn't open a chat, it just sits there. @Kabi4 Can you look into making sure this is working on mobile as well?

Kabi4:
image
Adds Ask AI to the verse toolbar: select verses, pick an agent if there’s more than one, and get a chat with those verses already in the input.
Reuses an existing local chat with that agent when one exists, so follow-ups keep context.
Fixes mobile: tapping Ask AI actually opens the chat instead of doing nothing.

Blitzy:
> [!NOTE]

## Code Review — PR #1673, follow-up pass

This is a re-review after the discussion on Aug 25-26 and the follow-up commits. Since the last review:

- The team decided **chat reuse is the intended behavior** (Kallyn and the earlier reviewer agreed continuing an old thread is more useful than always starting fresh), and the CHANGELOG now documents it ("open (or reuse) a chat"). The earlier M1 note is resolved as a product decision, not a bug.
- The **blocking mobile bug** ("tapping Ask AI on mobile does nothing") has real fixes: the floating chat panel's z-index was raised above the mobile toolbar chrome, and the outside-tap dismiss listener now (a) ignores taps on `.sb-verse-toolbar` and (b) attaches after a `setTimeout(0)` so the same gesture that opened the panel can't immediately close it again.

I validated the mobile fix isn't just plausible but actually load-bearing: I reverted both changes in `FloatingChatPanel`'s pointerdown handling in `FloatingReaderPanels.tsx` and reran the new tests — both **failed** on the pre-fix code ("ignores a pointerdown on the verse toolbar..." and "does not close on an outside pointerdown that arrives before the dismiss listener attaches"), then passed again once restored. That's a real regression test, not a tautology.

I also ran the full check suite on the current head (`b4580ba`):
- `pnpm vitest run` on the four touched test files — **258 passed**.
- `pnpm check:ts` — **0 errors**.
- `pnpm lint` — **0 errors** (only pre-existing baseline-CSS warnings, unrelated to this PR).

---

### High — should be fixed before merge

**H1. "Ask AI" silently overwrites an unsent, unsaved message in the compose box.**

`openAskAiForSelectedVerses` (`packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx:587-590`) always does:

```ts
const verseText = formatSelectedVerses(context.readingState);
if (context.chats.composerDraft) {
  context.chats.composerDraft.value = verseText ? `${verseText}\n\n` : "";
}
```

`ChatView`'s consuming effect (`packages/seed-bible/seed-bible/components/ChatView/ChatView.tsx:719-744`) then does `draft.value = value` unconditionally — it replaces whatever the user has typed, with no check for existing content and no confirmation.

Now that reuse-by-default is the agreed design, this is much more likely to actually happen: the whole point of reuse is that a user comes back to the *same* agent chat repeatedly. A realistic sequence:

1. Select a verse → Ask AI → chat opens with "Apologist," verse quoted.
2. Start typing a follow-up question, e.g. "how does this connect to..." — don't send it yet.
3. Select a different verse, tap Ask AI again for the same agent (the now-intended reuse path lands you back in the same chat).
4. The partially typed question is gone, replaced by the new verse quote, with no warning.

I confirmed this directly: I rendered `ChatView`, typed text into the compose field, then pushed a new value onto `composerDraft` the same way the tool does, and the typed text was fully discarded (not merged, not preserved) — no existing test exercises this case (the only related test, "ignores an empty composerDraft so typed text is not wiped," only covers the *empty* composerDraft case).

_Fix:_ before applying an incoming `composerDraft` value, check whether the field already has non-empty content and, if so, merge instead of replace (e.g. append the verse quote after the user's typed text, or prepend it before, whichever ordering matches the intended flow) rather than silently destroying it. A test that types into the field, then pushes a second non-empty `composerDraft` value, and asserts the typed text still exists in some form, would lock this in.

---

### Low — carried over from the last review, still applies

These were flagged before merge decisions were made and weren't part of this round's changes, so re-noting for visibility rather than repeating in detail:

- **`composerDraft` is set before the chat is confirmed to exist**, and isn't cleared on the early-return path in `openAskAiForSelectedVerses` (`BibleToolsManager.tsx:587-596`). Still not reachable in practice today since `createLocalSession` never returns null, but remains a latent trap if that ever changes.
- **Every agent in the multi-provider picker uses the same generic `AskAiIcon`** rather than the provider's own `iconUrl`, so the picker is only distinguishable by name.
- **`en.json` still carries unrelated key reordering** (the `annotation-conflict-*` / `annotations-pending-sync*` block, plus new `session-host-*` keys) alongside the one intentional `ask-ai` key — likely stale-base/rebase noise rather than something introduced by this PR's own work, but it inflates the diff for reviewers.

---

### Test quality

The new tests for this round's fixes are genuine regression tests, not tautologies — verified by reverting the `FloatingReaderPanels.tsx` pointerdown changes and watching the two new tests go red, then green again on restore. The `ChatView` height/prefill tests from the earlier round were previously verified the same way and are unchanged here.

---

KallynGowdy:
@Kabi4 Could you fix the "high" importance issue for this?

Kabi4:
Fixed Instead of replace now we do join for such cases.

Blitzy:
> [!NOTE]

## Code Review — PR #1673, third-pass follow-up

This is a re-review after Kabi4's Sep 2 fix for the previously-reported High-severity issue ("Ask AI silently overwrites an unsent, unsaved message"). Current head: `d75c1ed`.

I ran the full verification pass on this exact commit:
- `pnpm vitest run` (full suite) — **5471 passed**, 231 files, 0 failures.
- `pnpm check:ts` — **0 errors**.
- `pnpm lint` — **0 errors** (only pre-existing i18n/CSS `use-baseline` warnings, unrelated to this PR).

**The previous High-severity fix is real and correctly tested.** `mergeComposerDraft` (`ChatView.tsx:80-86`) now appends the incoming verse quote after the user's typed text instead of replacing it, and the new test "merges a later composerDraft with typed text instead of replacing it" (`ChatView.test.tsx`) is a genuine regression test — I reverted the merge back to a plain replace locally and confirmed that exact test goes red, then passes again on the real code. Good fix.

However, digging into how this fix is wired up surfaced a new issue in the same area:

---

### High — should be fixed before merge

**H1 (new). When the chat panel is already open, using "Ask AI" to reuse a *different* existing chat writes the verse text into the wrong (now-hidden) chat instead of the one the user is looking at.**

`FloatingChatPanel` renders `<ChatView chat={selectedChat} state={state} />` with no `key` (`FloatingReaderPanels.tsx:986`), and only unmounts `ChatView` when the panel itself closes (`if (!isOpen) return null;`, `FloatingReaderPanels.tsx:789`). So switching `selectedChat` while the panel stays open re-renders the *same* `ChatView` instance with a new `chat` prop — it does not remount.

Inside `ChatView`, the effect that consumes `composerDraft` is registered with an empty dependency array:

```ts
// ChatView.tsx:746
useEffect(() => {
  const composerDraft = state.chats?.composerDraft;
  if (!composerDraft) return;
  return effect(() => {
    const value = composerDraft.value;
    if (!value) return;
    const merged = mergeComposerDraft(draft.value, value); // draft.value is `chat.unsentDraft` captured at MOUNT time
    draft.value = merged;
    composerDraft.value = "";
    ...
  });
}, []);
```

`draft` (`ChatView.tsx:633`, `chat.unsentDraft ?? localDraft`) is captured once, from whichever chat was mounted *first*. Because the effect never re-runs, every later `composerDraft` push — no matter which chat is now visible — gets merged into that original chat's `unsentDraft`, not the currently displayed one.

I verified this is real, not theoretical, with a standalone reproduction (rendered `ChatView` with `chat={chatA}`, re-rendered the same container with `chat={chatB}` — simulating a chat switch with the panel already open — then pushed a value onto `composerDraft`): the visible input stayed **empty**, while `chatA.unsentDraft` (no longer displayed) silently received the verse text.

_Reproduction on the real app (needs 2 configured AI agents, each with an existing chat already used at least once):_
1. Open the chat panel on Agent A's conversation (e.g. by tapping Chat in the toolbar, or a prior "Ask AI").
2. Without closing the panel, select a verse and tap **Ask AI**, then pick **Agent B** — a chat that already exists but isn't the one on screen.
3. The panel switches to show Agent B's conversation, but the compose box stays empty — the verse never appears.
4. Switch back to Agent A's chat (e.g. from the chat list): the verse text is now sitting in *Agent A's* compose box, even though you never asked about it there.

This directly undermines the reuse-by-default design the team just agreed on (Aug 25-26 discussion) — the most common way to reach this bug is exactly the "keep switching between the same couple of agent chats" flow that reuse was built for.

_Suggested fix:_ Key the component so a genuine chat switch remounts it: `<ChatView key={selectedChat?.id} chat={selectedChat} state={state} />` (`FloatingReaderPanels.tsx:986`). This is safe with the current design specifically because typed text now lives on `chat.unsentDraft` rather than component-local state — that's the same mechanism the existing "keeps typed text after ChatView unmounts and remounts" test already relies on, so a remount on chat-switch won't lose anything. (A narrower alternative: change the composerDraft effect's dependency array from `[]` to `[chat]` so it re-subscribes against the currently visible chat without remounting the rest of the component — more surgical, but only fixes this one effect.)

A regression test worth adding: render `ChatView` with `chat={chatA}`, re-render (same container, no unmount) with `chat={chatB}`, then push a value onto `state.chats.composerDraft`, and assert the visible input shows the prefill and `chatB.unsentDraft` (not `chatA.unsentDraft`) received it.

---

### Low — carried over from earlier rounds, unchanged

Not part of this round's diff, re-noting only for visibility:

- `composerDraft` is still set before the chat is confirmed to exist, and isn't cleared on the early-return path in `openAskAiForSelectedVerses` (`BibleToolsManager.tsx:595-604`). Still unreachable in practice today (`createLocalSession` never returns null), but remains a latent trap if that ever changes.
- Every agent in the multi-provider picker still uses the same generic `AskAiIcon` (`BibleToolsManager.tsx:538,1121,1137`) rather than the provider's own `iconUrl`, so the picker is only distinguishable by name.
- `en.json` still carries i18n keys unrelated to this PR (`session-host-*`, `annotation-conflict-*`, `annotations-pending-sync*`) alongside the one intentional `ask-ai` key — this looks like base-branch drift picked up from merging `develop` in, not something introduced by this PR's own commits, so not blocking.

---

### Test quality

The merge-vs-replace regression test is solid (verified red-on-revert as noted above), and the earlier mobile pointerdown regression tests from the previous round are unchanged and still pass. No existing test in the suite currently covers the H1 (new) scenario above — the closest ones ("reuses the most recent local chat," the composerDraft prefill tests) all mount a single `ChatView` against a single chat and never simulate a chat switch while the component stays mounted, which is exactly the gap that let this through.

---

Kabi4:
I think is bug if irrelevant because we cannot open chat and select verse at same time reason is chat closes whenever clicked outside.

Kabi4:
image

Blitzy:
This is in a really good spot now. Thanks for that clarification as well @Kabi4, I agree its a non-issue right now because the chat does automatically close when a click occurs outside of it, I tested this just to be sure and its true.

I'm gonna get a lot of use out of this new tool! 😄

Blitzy (approved):

Issue #1674, opened 2026-08-21
Fix the display of the Mobile Account Settings page to be consistent with desktop


KallynGowdy:
Currently, the Account Settings page is the only settings page that has a visually distinct appearance on mobile from desktop. In particular, the title has a different font and font size (serif instead of sans-serif), and the close button is missing.

This means that there's a weird layout shift when going to the account settings, and also the user isn't able to close the page directly from account settings, but instead has to press back first and then close.

### Desktop

<img width="337" height="877" alt="Image" src="https://github.com/user-attachments/assets/d31acfb8-a864-4c3c-a51b-5a07d9f49952" />

### Mobile

<img width="427" height="909" alt="Image" src="https://github.com/user-attachments/assets/0cd4efdb-ae4e-4d6f-ae3d-bdc8e0a46ab3" />

PR #1675, opened 2026-08-21
Remove conditional close button styling for account settings

KallynGowdy:
## Summary

Simplifies the settings sidebar close button by removing mobile-specific styling that hid the button on the account view. Instead, the settings heading now uses consistent styling across all settings pages (main list and subpages like Account), matching the desktop appearance.

## Changes

- **Removed conditional close button class** — The `sb-sidebar-settings-close-button-account` class and its associated `display: none` rule are no longer needed. The close button now appears consistently on all settings views, including the account page.

- **Added unified settings heading style** — New CSS rule restores plain heading styling for `.sb-sidebar-settings-view .sb-sidebar-tabs-title` on mobile, so the "Settings" heading renders the same way on mobile as it does on desktop (no longer styled as a book title).

- **Simplified component logic** — Removed the `isAccountView` check in the Settings component that conditionally applied the account-specific class. The close button now uses a single, unconditional class name.

## Implementation Details

The settings sidebar reuses the `.sb-sidebar-tabs-title` class (which normally styles book titles in the tab list), but the "Settings" heading isn't a book title. The new CSS rule explicitly overrides the inherited book-title styling (`flex`, `font-size`, `font-weight`, `font-family`, `color`) to restore the plain heading appearance, ensuring visual consistency across mobile and desktop.

https://claude.ai/code/session_012pmEzfb7EEhyZZRwkQm4JJ

### DIFF
--- a/packages/seed-bible/seed-bible/components/Tabs/Tabs.css
+++ b/packages/seed-bible/seed-bible/components/Tabs/Tabs.css
@@ -1319,12 +1319,6 @@
     display: inline-flex;
   }
 
-  /* On mobile the account panel is dismissed via the bottom bar of options
-   * (e.g. the "Bible" tab), so the redundant X close button is hidden. */
-  .sb-sidebar-settings-close-button-account {
-    display: none;
-  }
-
   /* The desktop bottom-actions row (settings cog + self avatar) is hidden
    * on mobile — the bottom is now the Bible/Tabs switcher. */
   .sb-tabs-sidebar-mobile-open .sb-sidebar-bottom-actions {
@@ -1350,6 +1344,20 @@
     color: var(--sb-book-title-font-color);
     white-space: nowrap;
     text-overflow: ellipsis;
+  }
+
+  /* The settings sidebar (main list, and every settings subpage, including
+   * Account) shares .sb-sidebar-tabs-title with the tab-list header above,
+   * but its "Settings" heading isn't a book title — restore the plain
+   * heading style so it renders the same as on desktop. */
+  .sb-tabs-sidebar-mobile-open
+    .sb-sidebar-settings-view
+    .sb-sidebar-tabs-title {
+    flex: initial;
+    font-size: 1.125rem;
+    font-weight: 700;
+    font-family: inherit;
+    color: var(--sb-font-color);
   }
 
   /* Tab list takes the body of the sidebar; leave room at the bottom for
--- a/packages/seed-bible/seed-bible/components/Tabs/Tabs.tsx
+++ b/packages/seed-bible/seed-bible/components/Tabs/Tabs.tsx
@@ -927,7 +927,6 @@ export function Settings(props: SettingsProps) {
   const { state } = props;
   const { sidebar } = state;
   const { t } = useI18n();
-  const isAccountView = sidebar.requestedSettingsView.value === "account";
 
   return (
     <div className="sb-sidebar-settings-view">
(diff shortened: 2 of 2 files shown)


Blitzy:
> [!NOTE]

## Code Review — Remove conditional close button styling for account settings

### Verdict

Looks good. This is a small, well-scoped change (15 additions, 10 deletions across 2 files) that correctly addresses both symptoms called out in #1674:

1. **Missing close button** on the mobile Account Settings page — fixed by deleting the `.sb-sidebar-settings-close-button-account { display: none; }` rule so the X button now renders on every settings view.
2. **Serif book-title font on the mobile "Settings" heading** — fixed by adding a `.sb-tabs-sidebar-mobile-open .sb-sidebar-settings-view .sb-sidebar-tabs-title` override that restores the plain heading style.

I verified the CSS mechanics hold up:

- The settings view (`<Settings>`) renders inside the `.sb-tabs-sidebar-mobile-open` aside (`Tabs.tsx:2606`, `2632`), so the new selector's ancestor is always present when it needs to be.
- The new override has specificity `(0,3,0)` vs. the book-title rule's `(0,2,0)` at `Tabs.css:1334`, and both sit inside the same `@media (max-width: 480px)` block (opened at line 1210), so the override reliably wins on mobile.
- The removed `sb-sidebar-settings-close-button-account` class has no remaining references anywhere in the codebase (CSS, TSX, or tests), so the dead-class removal is clean.
- Dropping the `isAccountView` variable leaves no unused imports or dangling references.

No critical or high-severity issues found. Two low-severity notes below.

---

### Low severity

**L1 — The heading override only partially resets the book-title styling, which can truncate long localized labels on mobile.**

The mobile book-title rule (`Tabs.css:1334`) applies these truncation properties:

```css
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
min-width: 0;
```

The new override (`Tabs.css:1353`) resets `flex`, `font-size`, `font-weight`, `font-family`, and `color`, but **not** the four properties above — so the "Settings" heading keeps `nowrap` + `overflow: hidden` + `ellipsis` on mobile. Combined with `flex: initial` (`0 1 auto`, i.e. still shrinkable), a heading that's wider than the available space would be truncated with an ellipsis on mobile only, whereas on desktop (`.sb-sidebar-tabs-title` at line 124 sets none of these) it would wrap/overflow normally.

The English "Settings" is short, so there's no visible effect today. But the app ships 24 locales, and a longer translation could trip this — which slightly undercuts the PR's stated goal of "renders the same as on desktop." This is a latent inconsistency, not a bug in the shipped English UI.

*Reproduction (for a human dev):* temporarily set the `settings` key in a locale (or `en.json`) to a long string, or artificially narrow the mobile header, then open Settings at a viewport ≤ 480px — the heading truncates with "…" on mobile but not on desktop.

*Suggested fix* — reset the inherited truncation properties in the override so the settings heading fully matches desktop:

```css
.sb-tabs-sidebar-mobile-open
  .sb-sidebar-settings-view
  .sb-sidebar-tabs-title {
  flex: initial;
  font-size: 1.125rem;
  font-weight: 700;
  font-family: inherit;
  color: var(--sb-font-color);
  white-space: normal;
  overflow: visible;
  text-overflow: clip;
}
```

**L2 — No test coverage for the behavior change.**

The purely visual/CSS part (font, media-query rendering) isn't practically unit-testable in jsdom, so leaving that untested is reasonable. However, the component-logic change *is* observable and testable: the close button now always renders with the single class `sb-sidebar-settings-close-button` regardless of `requestedSettingsView`. There's already a rendering harness for this area (`test/unit/seed-bible/components/TabsSidebarLayout.test.tsx` opens settings and asserts on "Settings" text), so a small regression test would be cheap to add — e.g. set `requestedSettingsView` to `"account"`, render, and assert the close button is present and no longer carries the `-account` modifier. To be a real regression test, confirm it fails against the pre-fix code (where the account view rendered the extra class / the button was hidden on mobile) before it passes on this branch.

This is optional given the size of the change, but it would lock in the fix for #1674 so the conditional class doesn't quietly reappear.

---

### Tests

No test files were added or modified. Per the notes above, the visual behavior is hard to assert in jsdom (acceptable), while the class-name/rendering change is testable but currently uncovered (L2). No existing tests reference the removed class, so nothing breaks.

---

PR #1676, opened 2026-08-21
Today screen: extension → core component

Blitzy:
Moves the Today screen out of `packages/today-screen/` and into `packages/seed-bible/`, and reshapes it to match the conventions of the rest of the core app. The old package is deleted.

Two things motivated this beyond the move itself. The package was laid out as hexagonal / ports-and-adapters, but only one slice actually used the pattern — `domain/` was 130 lines of pure type aliases with no behaviour, while 91% of the code sat under `infrastructure/`. The folder names promised an architecture the code didn't have, which is most of why it was hard to read. And because an extension can't import core managers at runtime, every host capability was threaded through a single 41-field `TodayConfig` props bag into one Preact context — indirection that becomes unnecessary the moment this is core code.

`components/TodayPane` is now **12 files totalling 2,914 lines**, down from 54 files and 4,461 lines. The old tree averaged 83 lines per file, with 21 files under 40 lines and the smallest at 3, because 14 hooks were paired 1:1 with a same-named component. There is now one file per section of the screen.

## Behaviour

Behaviour is preserved throughout, with three exceptions worth calling out.

**Fewer re-renders.** The 41-field config object was built *inside* the pane's render function, and five of its fields were unwrapped `.value` reads at that point — `language`, `username`, `userId`, `userProfile` and `theme`. That subscribed the pane's root component to all five, so a theme toggle or a profile change re-rendered the whole Today tree. Components read managers directly now, which makes subscriptions fine-grained by construction.

**Two real crash paths closed.** `useBook` threw for any community-reading user id missing from `userProfileMap`, and `HistoryCard` did `.get(id)!` on the same map. The map is filled by an effect while `communityReading` arrives from a separate async fetch, so the two can legitimately disagree for a frame, and an unsubscribed reader in community data threw for real. A throw inside a `useMemo` during render isn't caught by anything — it took down the pane. Unknown ids are now skipped and counted into the existing `+N` overflow.

**`Welcome` no longer uses `dangerouslySetInnerHTML`.** It splits on the `<hl>` markers and emits real nodes instead. Sanitizing was the wrong fix here: `ALLOWED_TAGS` doesn't include `hl`, so running the verse through `setSafeHtml` would have silently stripped every highlight, and adding `hl` to the allow-list would have loosened sanitization for playlist HTML and chat to serve one screen. Splitting also JSX-escapes the API-sourced fallback verse, which closes an injection path that exists today.

Defects found and fixed along the way:

- The timeline grid stretched its columns under wide month labels, which also affected Scripture Map. Most visible in `ar` and `bn`.
- The timeline legend described a color scale the grid never actually used.
- `GetPastDateInfo` carried hardcoded English weekday names, in a field neither consumer read. The field is gone.

## Verification

- `pnpm check:ts` — 0 visible errors (6 suppressed, all pre-existing).
- `pnpm test` — **203 files / 4,965 tests passing.**
- `pnpm lint` — **0 errors.**
- `pnpm build` — green.
- **31 manual browser checks pass:** language switching across RTL and wide-script locales, the reader filter, timespan windows, the chapter grid, bookmarks wrapping and mobile scroll, theming, Welcome, the resume card, tooltip placement and both click-outside dismissals, both entry points into the screen, the tutorial gate, and Scripture Map.

## Tests

The deleted package's 40 test files (6,191 lines) are replaced by a smaller set under `test/unit/seed-bible/`. Roughly 590 of the removed lines covered code that never ran — `useMicroGrid`, `usePresenceCard` and `PresenceCard` were all commented out of the tree — and about 1,400 were hook-level tests whose assertions duplicated the component test one layer down.

New coverage that didn't exist before: the pane open/close effects, the auto-open predicate, both entry points into the screen, and `todayCoversReader.test.ts`. That last one guards a construction-order invariant — Today's pane now opens synchronously inside `createSeedBibleState`, which is what allows the old `todayHasOpened` latch to be deleted outright. Its failure mode is silent (the tutorial prompt rendering beneath the Today screen), so it is guarded by a test rather than a comment.


Blitzy:
> [!NOTE]

## Code review — Today screen: extension → core component

Thorough refactor, and the write-up is excellent. I verified the three headline behavioural claims against the code: the `dangerouslySetInnerHTML` removal in `Welcome`, the unknown-reader crash guard in `SocialSection`, and the timeline grid/legend fixes all check out. The CSS grid-stretch fix (`grid-auto-columns: var(--timelineItemSize)`) is well-reasoned, and the colour interpolation is now genuinely shared between the day cells and the legend, so they can't drift.

Findings below are ordered by severity. Reproduction steps are included where a human can follow them.

---

### 🔴 High

**H1 — Broken machine-translated weekday labels ship in every non-English locale.**

The PR adds three new keys — `monday-short`, `wednesday-short`, `friday-short` — and they were auto-translated into all locale files. The results are wrong in ways that will be visible to users:

- `wednesday-short` came back as the *verb* "to marry" in multiple languages: German `"Heiraten"`, Spanish `"Casarse"`, French `"Épouser"`, Arabic `"تزوج"`. The translator read the bare English abbreviation "Wed" as the verb "wed", not the day.
- The abbreviations aren't abbreviated: German `monday-short` = `"Montag"`, `friday-short` = `"Freitag"` (full words), which also breaks the compact single-column label the grid is built around.

These strings render as the weekday row labels on the timeline (`useReadingHistoryTimeline.tsx:305-333`), so they're directly user-facing.

Two things compound here:
1. The change edits every locale JSON, which conflicts with the repo convention in `CLAUDE.md` ("only update en.json … leave those for the professionals"). The other-locale files look auto-generated by the `pnpm i18n` tooling.
2. Even setting the convention aside, routing 3-letter weekday abbreviations through machine translation is the wrong mechanism — the source is too ambiguous to translate.

**Suggested fix:** don't use translation keys for weekday abbreviations at all. The same file already localizes month names and dates via `Intl.DateTimeFormat(language, { month: "short" })` (`useReadingHistoryTimeline.tsx:403`, `ReadingHistoryTime.ts:51`). Do the same for weekdays with `Intl.DateTimeFormat(language, { weekday: "short" })`, which returns correct, properly abbreviated labels (de `"Mi"`, es `"mié"`, fr `"mer."`, ar `"الأربعاء"`) with no translation round-trip. Then drop the three keys from the locale files (and, per convention, revert the non-`en` edits).

**Repro:** set the UI language to German (or French/Spanish/Arabic), open Today → Community → the "All" timeline. The left-hand weekday labels show the wrong word for Wednesday ("Heiraten"/"Épouser"/"Casarse") and, in German, unabbreviated day names.

---

### 🟠 Medium

**M1 — The header's date and greeting never update while Today stays open.**

`Header.tsx:9-27` computes `date` and `greeting` from `new Date()` inside a `useMemo` whose dependency list is `[language, t]` — no time input — so it's frozen after the first render. `Header` also never consumes `useTimeContext()`. `TimeProvider` re-renders every 10s expressly "so that time-relative content — the header's date and greeting … stays current" (`TimeContext.tsx:12-14`), but that can't reach the header: the memo excludes time, and the header isn't a context consumer. Only the timeline actually reads `tick`.

**Suggested fix:** add `const { tick } = useTimeContext();` in `Header` and include `tick` in the memo deps, mirroring what `useReadingHistoryTimeline` already does.

**Repro:** open Today shortly before a greeting boundary (e.g. 11:59 → 12:00, or across midnight for the date) and leave it open. The greeting stays "Good morning" and the date doesn't roll over.

**M2 — The unknown-reader crash guard has no test that can fail.**

The PR calls out closing a real crash path where a reader id present in `communityReading` is missing from `userProfileMap`; the fix is `if (!profile) continue;` plus counting the skipped id into `+N` (`SocialSection.tsx:401-416`, and again in the chapter grid at `451-453`). But no test exercises it. As the component is wired, `userProfileMap` only ever holds the signed-in user and `userFilters` is kept in sync with exactly its keys, so any id that survives the upstream `userFilters` filter is always in the profile map — the branch is unreachable through the public component. The one related test (`SocialSection.test.tsx:502`) only asserts a `+N` badge is *absent*. Reverting the guard to a throwing `userProfileMap.get(id)!.name` would leave every test green, so a documented crash fix has no regression guard.

**Suggested fix:** export `Book`/`FilteredReading` for a focused unit test, or extend the stubs so `userProfileMap` and `userFilters` can be supplied out of sync. Render >7 readers (`MAX_ICONS` is 7) with one id present in `userFilters` but absent from the profile map, and assert (a) no throw and (b) the `.sb-today-filtered-reading-book-extra` badge shows the correct count. Add a chapter-grid variant for the `451-453` skip.

---

### 🟡 Low

**L1 — Negative tutorial test relies on a fixed 50 ms sleep.** `todayCoversReader.test.ts:64-71` waits `setTimeout(…, 50)` then asserts the tutorial prompt is *not* visible. `CLAUDE.md` forbids fixed-duration sleeps; because the sibling positive test needs up to 2000 ms via `waitFor` to see the prompt, 50 ms is not comfortably longer than the real latency — if the covering-pane guard regressed, a slow machine could still snapshot `false` and pass. Consider driving the positive path first to establish the offer would have fired by now, or exposing `readerVisible` for a synchronous assertion.

**L2 — `todayReactivity` proves propagation, not scoping.** The tests confirm the correct piece updates on a theme/profile change (good, and not hollow), but the headline's "doesn't re-render the whole tree" half is unverified — there's no render-count assertion that an unrelated subtree is left untouched. Add a render counter in a child and assert an unrelated signal change doesn't bump it.

**L3 — Three timeline tests are named stronger than they assert** (`useReadingHistoryTimeline.test.tsx`):
- "ignores results that resolve after unmount" (540-563) only asserts the fetch was called, not that the post-unmount signal write was skipped; removing the `isMounted` guard leaves it green.
- "ignores out-of-range and sub-minute events" (490-538) mocks the colour to a constant, so it can't distinguish a counted vs skipped event — echo `readingTimeSeconds` (as the legend test does) and assert the accumulated time.
- "yields to the main thread while summarizing many days" (591-613) only asserts `length > 0`; the yield behaviour itself is unguarded.

**L4 — "auto-opens … no reading position" doesn't test the no-position branch.** `TodayManager.test.ts:114-119`: `createTestSeedBibleState` always stamps `?today=open`, which short-circuits `todayWillAutoOpenForUrl` before the no-position heuristic. The heuristic is covered by the pure `todayWillAutoOpenForUrl` tests, so this is a misleading name / redundant coverage rather than a hole — rename it, or build a genuinely bare boot URL.

**L5 — "resets to loading on an account switch" assertion is tautological.** `createReadingHistoryState.test.ts:100-102`: account A's fetch is never resolved, so the state is already `loading` when `userId` flips to B — the assertion passes whether or not the reset at `TodayReadingHistory.ts:202-204` exists. There's no test covering the `ready → loading` transition. Resolve A to `ready` first, then switch, so the reset has something to undo.

---

### Nits

- Copy: `resume-reading` = "CONTINUE WHERE YOU LEFT" (likely "…LEFT OFF"); `today-search-verses` = "Search books, chapter, verses...." (mixed singular/plural, four dots). (`en.json`)
- Stale comment: `useReadingHistoryTimeline.tsx:456` says "Legend is currently placeholder data", but the code below now generates real, theme-derived legend swatches.
- Duplication: `managers/Colors.tsx`, `Numbers.ts`, and `Strings.ts` reintroduce colour/util logic that also lives in `seed-bible-utils` (used by Scripture Map). This is probably justified by the pattern/core import boundary, but per `CLAUDE.md`'s duplication guidance it's worth a note — the two copies can drift.

---

### Worth acknowledging

Several tests are model regression tests, not box-ticking:

- **Welcome XSS + highlight spans** (`Welcome.test.tsx:129-169`) — double-guarded (`querySelector("img")` is null *and* `textContent` keeps the literal markup); genuinely fails if `dangerouslySetInnerHTML` is reintroduced.
- **Timeline theme-recolour** (`useReadingHistoryTimeline.test.tsx:671-708`) — asserts a recolour happens on a theme switch *with no additional fetch*, pinning the exact "`.value` read inside a `useMemo` didn't subscribe" bug.
- **Stale-response tests** across Welcome, SocialSection, and SearchSection deliberately settle the superseded promise *last*, which defeats the "later write wins by ordering" false pass.
- **`createReadingHistoryState`** cross-account and stale-fetch guards, and **`loadDailyReadingHistory`**'s "out-of-range events are dropped from buckets but still counted in the total" test, both cover real edge cases.

Net: the refactor is solid and most of the suite is strong. The one item I'd treat as blocking is H1 (wrong text shipping in many languages); M1 and M2 are worth fixing before merge.

---

Blitzy:
Thanks — H1, M1 and M2 are addressed.

**H1 — weekday labels — `d065ceac7`.** Switched to `Intl.DateTimeFormat(language, { weekday: "short" })` as suggested and dropped the three keys from all 77 locale files. Fixes both halves: de now reads `Mo / Mi / Fr` instead of `Montag / Heiraten / Freitag`. Scripture Map uses the same keys but is unaffected — it resolves its own namespace first and its copies were correct, which suggests the bad strings came in when the keys were promoted into core. Added a regression test, mutation-checked.

**M1 — frozen header clock — `fe17fff52`.** `useTimeContext()` plus `tick` in the memo deps, as suggested. Two regression tests, one per symptom in your repro. `Header.test.tsx` had to move inside `TimeProvider` since it rendered standalone, and `TICK_INTERVAL_MS` is now exported — `TimeContext.test.tsx` already had five hardcoded copies of it.

**M2 — unknown-reader guard — `f81fd2a68`.** Documented rather than tested. It's not three untested branches but one feature that hasn't landed: `userProfileMap` returns exactly one entry while `SubscribedUsersProvider` is a stub, so `readerIds.length ≤ 1` and the guard, the `+N` badge and the 451-453 filter are all unreachable for the same reason. A test would have to construct state the component can't produce. The real find was that the old comment was wrong — it claimed the maps race, which they can't — so both now say it's unreachable today and name subscriptions as what changes that. Happy to add the coverage your way if you'd still prefer it.

Blitzy:
> [!NOTE]

## Follow-up review — commits since the last pass

This PR was just marked ready for review, so this is a check on the current head (`f81fd2a68`) against the review posted above. I read the three fix commits pushed since then and the author's reply, checked each fix against the actual code, and re-checked whether the earlier Low-severity items were touched.

### Verified fixed

**H1 (weekday labels), `d065ceac7`.** Confirmed correct. `useReadingHistoryTimeline.tsx` now derives Monday/Wednesday/Friday labels from `Intl.DateTimeFormat(language, { weekday: "short" })` against three fixed 2024 dates, and the three translation keys are gone from all 77 locale files (including `en.json`, per the repo's own "only edit `en.json`" convention). The new test is genuinely load-bearing, not just decorative: it asserts `labelsFor("en")` and `labelsFor("de")` differ before checking the rendered output, so it can't pass by coincidence if `Intl` ever returned the same string for both.

**M1 (frozen header clock), `fe17fff52`.** Confirmed correct. `Header` now reads `useTimeContext()` and includes `tick` in the memo's dependency array, and `Header` is wrapped in `TimeProvider` everywhere it's actually rendered (`TodayPane.tsx:44-46`), so this doesn't rely on a test-only setup that production doesn't provide. Both new tests reproduce a real symptom (date at midnight, greeting at the noon boundary) using the exported `TICK_INTERVAL_MS` rather than a re-typed literal.

**M2 (unknown-reader guard), `f81fd2a68`.** I checked the author's claim rather than taking it on faith: `userProfileMap` in `SocialSection.tsx:58-72` is built only from the signed-in `userId`, and the `userFilters` effect (`:119-132`) prunes its keys to exactly `userProfileMap`'s keys. Since `readerIds` is filtered through `userFilters`, an id that reaches the guard is guaranteed to already be in the map today — the branch really is unreachable until reader subscriptions exist, as the new comment says. Docs-only is a reasonable resolution here; no test is being skipped to get there.

No new issues in any of the three fix commits.

### Still open (not touched by these commits, so unchanged from the review above)

None of these are new — flagging that they weren't part of this round's fixes, in case they were meant to be:

- **L1** — `todayCoversReader.test.ts:69` still asserts on a fixed `setTimeout(50)`.
- **L2** — `todayReactivity.test.tsx` still has no render-count assertion for the "unrelated subtree doesn't re-render" half of the claim.
- **L3** — the three `useReadingHistoryTimeline.test.tsx` tests (now at lines 514, 564, 615) still assert less than their names claim.
- **L4** — `TodayManager.test.ts:114` ("auto-opens … no reading position") still short-circuits on `?today=open` before reaching the heuristic it names.
- **L5** — `createReadingHistoryState.test.ts:83` ("resets to loading on an account switch") is still true whether or not the reset code runs, since account A's fetch is never resolved first.
- **Nits** — the `en.json` copy issues (`resume-reading`, `today-search-verses`) and the stale "Legend is currently placeholder data" comment (now at `useReadingHistoryTimeline.tsx:465`) are unchanged.

None of these are blockers — they were "worth fixing before merge" suggestions in the original review, not the reason it was held up. With H1 addressed, I don't see anything blocking merge from this pass. CI is green on `f81fd2a68`.

---

Blitzy:
> [!NOTE]

## Follow-up review — commits since the last pass

This PR was just marked ready for review again, so this is a check on the current head (`1744635`) against the two reviews above. I diffed `f81fd2a6` (the commit the last review covered) against the current head — nine commits, all test/copy/comment fixes, no production logic changes beyond two copy strings — and read each one against the code.

### Verified fixed

Every open item from the last review is addressed, and each fix does what its commit message claims:

- **L1** (`5d33797`) — `todayCoversReader.test.ts` no longer waits a fixed 50ms. It now asserts the offer is absent while Today is open, closes Today, and polls (`waitFor`, 2000ms) for the offer to appear — so Today is proven to be the only thing holding the offer back, not just "not yet."
- **L2** (`a17a40a`) — `todayReactivity.test.tsx` adds a `renderCounts` helper built on Preact's own `options.diffed` hook and two new tests. I checked `TodayContainer`/`TodayContent` aren't just missing display names that would make the negative assertions vacuous — they're still named function declarations, and `TodayContainer` reads `today.readingHistory.value` (not `profile` or `theme`) in its own render body, so under `@preact/signals`' per-component subscription model it genuinely won't re-render on a profile or theme change. The negative assertions are real.
- **L3** (`0aae1ea`, `e72d0fc`, `328a6d3`) — all three timeline tests now assert something that can fail: the out-of-range/sub-minute test echoes seconds through the colour mock instead of a constant; the yield test moved the actual turn-counting assertion into `loadDailyReadingHistory.test.ts` and renamed the timeline test to what it can see (all 30 days coloured, not just "length > 0"); the unmount test is renamed to what it verifies (settles without warning) and a new sibling test (superseded-fetch-resolves-last) covers what the old name implied but didn't check.
- **L4** (`24e668a`) — `TodayManager.test.ts` adds `todayOpen: "fromUrl"` to the test harness so the boot URL reaches the real auto-open heuristic instead of short-circuiting on an explicit `?today=open` param. Both directions are now covered (bare URL opens, chapter URL doesn't).
- **L5** (`dab2596`) — `createReadingHistoryState.test.ts` splits into two tests: one that resolves account A to a shown card before switching to B (so the reset has something to undo), and the original stale-fetch-ordering test, renamed to what it actually proves.
- **Nits** — `resume-reading` now reads "…LEFT OFF", `today-search-verses` is pluralized and single-dotted, and the stale "Legend is currently placeholder data" comment is gone. All en.json-only, per convention.

No new issues in any of these nine commits — they're additive test coverage and two copy fixes, nothing that touches production behavior beyond the copy strings.

CI is green on `1744635` (build passing at 14:23 UTC). I don't see anything blocking merge.

---

PR #1677, opened 2026-08-23
feat: add UI language switch prompt for Bible translation selection

### LINKED ISSUE #1126: [refactor] Prompt users to switch their UI language when changing translation languages
In the refactor, we currently don't prompt users to change their UI language when they select a translation that is written in a different language. This task tracks work on this feature.

Expected flow:
1. User is currently in English UI
2. They choose a translation in a different language (e.g. Spanish)
3. They are prompted via modal to switch their UI language to Spanish
   - Header: "Switch language?"
   - Text: "Do you want to switch your language to {{nativeLangaugeName}}?"
   - Options: "Switch", "Don't Switch", "Never Ask Again"
   - The prompt should only appear once per session (use [session storage API](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage))
   - The prompt should be displayed in the language of the translation that was selected
   - If the user selects "Never Ask Again", then it should record in their user preferences (or local storage if they aren't logged in)

mazenemad:
fixed #1126


Blitzy:
> [!NOTE]

## Code Review — feat: add UI language switch prompt for Bible translation selection

Reviewed against issue #1126. The PR implements the described flow cleanly: when a user deliberately picks a Bible translation written in a supported UI language other than the current one, a modal offers to switch the interface language, with **Switch / Don't Switch / Never Ask Again**, shown once per session, with the opt-out persisted as an ordinary setting.

**Overall:** solid, well-scoped work that matches the acceptance criteria and the codebase's existing patterns. No correctness bugs found. Verification below, then a few low-severity notes.

### Verification performed
- `pnpm check:ts` → 0 visible errors.
- `pnpm lint` (includes the i18n key validator) → 0 errors.
- `pnpm vitest run I18nManager.test.ts BibleSelectorManager.test.ts` → all pass (192).
- Confirmed the tests are real, not always-green: reverting the `maybePromptUiLanguageSwitch` call in `pickTranslation` makes the "offers to switch the UI language when the user picks a translation in another language" test fail as expected. The negative tests each pin a distinct guard (`targetLanguage === language.value`, unsupported language → `null`, and the programmatic `selectTranslation` path), and the once-per-session and never-ask-again tests exercise the actual `sessionStorage` marker and preference wiring rather than mocks agreeing with each other. Good coverage of the real branches.

### Things done well
- The prompt is correctly attached to `pickTranslation` (the deliberate user click) and **not** to `selectTranslation` (selector-open sync, deep links, language-driven switch), with a test locking that distinction in.
- `confirmUiLanguageSwitch` uses `changeLanguage` + `persistLanguage` rather than `requestLanguageChange`, so confirming does **not** swap the Bible text away from the translation the user just chose — and there's a test asserting `applyTranslation` is never called.
- The "never ask again" preference rides the same profile-when-logged-in / device-when-logged-out storage as every other setting (via `saveProfileConfigValue` → `login.localConfig`), which satisfies the issue's "user preferences (or local storage if they aren't logged in)" requirement.
- Schema back-compat handled: `askToSwitchUiLanguage: z.boolean().default(true)` so settings exported before this option still import.

---

### Issues by severity

#### Low — the modal will render in English (not the target language) until translators add the keys
Issue #1126 asks for the prompt to be "displayed in the language of the translation that was selected." The infrastructure for this is correct — `t` is `i18n.getFixedT(targetLanguage, "seed-bible")`, and there's a nice `loadLanguages(...)` re-fix so the text upgrades once the locale chunk lands. But the new keys (`switch-language-title`, `switch-language-body`, etc.) were only added to `en.json`, per the project's translation convention. Until the professional translations land in the other locale files, `getFixedT("es")` falls back to English, so the dialog shows English body text with only the native language name localized (e.g. "Do you want to switch your language to Español?").

This is **expected** given the "only edit en.json" convention and is not a code defect — it will self-correct once translations arrive. Flagging it mainly so QA testing the alpha deploy doesn't read the English text as a bug. Worth confirming the follow-up translation task is tracked.

#### Low — prompt (and profile save) can fire even when the translation's content failed to load
In `BibleSelectorManager.pickTranslation`, `maybePromptUiLanguageSwitch(picked?.language)` runs unconditionally after `await selectTranslation(...)`. `handleTranslationSelect` swallows its own error (sets `error.value` and returns rather than throwing), so if the chapter/books fetch fails, `pickTranslation` still proceeds to persist the translation id and raise the switch prompt. The pre-existing `saveProfileConfigValue` call has the same shape, so this isn't newly introduced behavior, but the new prompt inherits it.

*Repro (human):* pick a translation whose `books.json` / chapter request fails (e.g. offline or a 500). The reader shows the load error, yet the "Switch language?" prompt still appears.

*Suggested fix:* gate the prompt on a successful selection — e.g. have `pickTranslation` check `error.value` (or have `selectTranslation`/`handleTranslationSelect` signal success) before calling `maybePromptUiLanguageSwitch`. Low priority since the degraded state is transient.

#### Low (a11y) — new dialog lacks focus management, `aria-labelledby`, and Escape/overlay dismissal
`UiLanguageSwitchModal` uses `role="dialog"` + `aria-modal="true"` but has no `aria-labelledby`/`aria-describedby` pointing at the title/body, no initial focus move onto the dialog, no focus trap, and no Escape-to-close or overlay-click-to-dismiss. This matches the existing `LanguageUnavailableModal`, so it's consistent rather than a regression — but since this is a brand-new user-facing dialog, associating the title via `aria-labelledby` and moving focus into it on open would be a cheap accessibility win. Consider whether it's worth addressing here or in a shared modal cleanup.

#### Nit — re-enabling the toggle mid-session won't re-show the prompt
The once-per-session `sb.uiLanguagePromptShown` marker is set when the prompt first appears and isn't cleared by re-enabling "Offer to switch language with translation" in Settings. So a user who hits "Never Ask Again", then flips the setting back on and picks another translation in the same tab, won't see the prompt again until a new session. This is arguably consistent with "once per session," so it may be intended — just calling it out in case the expectation is that toggling it back on re-arms it immediately.

#### Nit — key naming style
The new keys are flat/kebab-case (`switch-language-title`, `ask-to-switch-ui-language`) whereas the sibling feature uses a dotted namespace (`languageUnavailable.title`, `languageUnavailable.body`). Purely cosmetic and the linter is happy, but grouping under a `switchLanguage.*` prefix would read more consistently with the neighbouring block.

---

### Summary
No blocking issues. The behavior matches issue #1126, the tests are meaningful and verified to fail without the implementation, and typecheck/lint/tests are green. The notes above are all low severity — the one most worth a quick decision is No. 2 (guarding the prompt behind a successful translation load); the English-until-translated point (No. 1) is worth confirming a translation follow-up exists so QA doesn't misread it.

---

PR #1678, opened 2026-08-23
Add functionality to open book selector when Bible tab is active

### LINKED ISSUE #1525: Tapping the Bible icon on mobile should open the book selector
Currently, tapping it does nothing unless another screen is displayed. If the Bible text is already being displayed, then clicking the Bible icon should open the book selector. If it is not displayed, then clicking it should cause it to be shown (hide today screen, etc.).

mazenemad:
fixes #1525

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
@@ -1882,6 +1882,12 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
                   label={t("bible", { defaultValue: "Bible" })}
                   active={activeMobileTab.value === "bible"}
                   onClick={() => {
+                    // The Bible text is already showing, so there's nothing to
+                    // dismiss — open the book selector instead of doing nothing.
+                    if (activeMobileTab.value === "bible") {
+                      openSelectorTool.value?.onSelect();
+                      return;
+                    }
                     isMoreMenuOpen.value = false;
                     sidebar.closeSearchPanel();
                     sidebar.closeChatPanel();
--- a/test/unit/seed-bible/components/BibleReaderToolbar.test.tsx
+++ b/test/unit/seed-bible/components/BibleReaderToolbar.test.tsx
@@ -826,6 +826,79 @@ describe("BibleReaderToolbar mobile More menu", () => {
   });
 });
 
+describe("BibleReaderToolbar — mobile Bible tab", () => {
+  let container: HTMLDivElement;
+  let originalInnerWidth: number;
+
+  beforeEach(() => {
+    originalInnerWidth = window.innerWidth;
+    window.innerWidth = MOBILE_VIEWPORT_WIDTH;
+    container = document.createElement("div");
+    document.body.appendChild(container);
+  });
+
+  afterEach(() => {
+    render(null, container);
+    container.remove();
+    window.innerWidth = originalInnerWidth;
+  });
+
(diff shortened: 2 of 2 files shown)


Blitzy:
> [!NOTE]

## Code review — Add functionality to open book selector when Bible tab is active

**Summary:** This is a clean, minimal, and correct fix for #1525. When the Bible text is already on screen, tapping the Bible bottom tab now opens the book selector instead of doing nothing; when another screen covers the reader, the tap still just reveals the reader. The implementation reuses the existing `open-selector` tool rather than duplicating selector logic, which is the right call. I verified the behavior and the tests locally against the head commit (`73dee4e`).

Nothing here blocks merge. The notes below are low-severity edge cases and one observation about test coverage.

---

### What I verified

- **Behavior matches the issue.** `activeMobileTab` resolves to `"bible"` only as the fallthrough case — i.e. when no panel, drawer, settings, Today, or fullscreen pane is covering the reader (`BibleReaderToolbar.tsx:783`). So gating on `activeMobileTab.value === "bible"` is an accurate proxy for "the Bible text is already showing," and the early `return` correctly skips the dismiss logic in that case.
- **The tool call is wired correctly.** `openSelectorTool.value?.onSelect()` works with no arguments because `getToolbarTools` binds each tool's `onSelect` to its context (`BibleToolsManager.tsx:1335`), and the `open-selector` handler calls `selectorState.setOpen(true, currentSlot)` (`BibleToolsManager.tsx:712`).
- **Test 1 is a real regression test.** I reverted just the production change and re-ran: `opens the book selector when the Bible text is already showing` fails (times out) without the fix, and passes with it. The `waitFor` helper throws on timeout (`createTestSeedBibleState.ts:48`), so this test cannot silently pass — good.
- **Both tests pass on the head commit** (`pnpm vitest run ... -t "mobile Bible tab"` → 2 passed).

### Findings by severity

**Low — No. 1: Second tap while the selector is already open doesn't toggle it closed.**
While the book selector is open, `activeMobileTab` stays `"bible"` (the selector's open state isn't one of the conditions in the `activeMobileTab` computed). So tapping the Bible tab again just calls `setOpen(true, ...)` a second time — a no-op that leaves the selector open. The issue only asks for "open," so this is arguably by design, but some users may expect a second tap to dismiss it (the Search tab, by contrast, toggles). Worth a product decision rather than a required change. If a toggle is wanted, the handler could branch on `selector.isOpen.value`.

**Low — No. 2: If a user removes the "Books" tool from their toolbar, the tap silently does nothing again.**
`openSelectorTool` is looked up from `applyToolbarCustomization(...)` output (`BibleReaderToolbar.tsx:601`, `:813`). If a customization removes the `open-selector` tool, `openSelectorTool.value` is `null`, the `?.` short-circuits, and tapping the active Bible tab reverts to the original "does nothing" behavior for that user. The `?.` correctly prevents a crash, so this is only a graceful-degradation edge case — and the floating-nav "Books" button has the same dependency today — but it's worth being aware of. No change strictly required.

**Low — No. 3 (test coverage): Test 2 doesn't guard the new branch.**
`shows the Bible text instead of the selector when another screen covers the reader` passes both with and without the fix (I confirmed it stays green on the reverted code). That's expected — it exercises the `else` path, which this PR didn't change — so it's a useful guard that the new early-return doesn't leak into the cover-screen case, but it isn't a regression test for the new behavior. Test 1 covers that, so overall coverage is adequate; just noting that the two tests aren't equally load-bearing.

### Reproduction (for the fixed behavior)

1. Open the app on a mobile-width viewport with the Bible text showing (no panel/drawer/Today open).
2. Tap the Bible (seed) icon in the bottom tab bar.
3. Before: nothing happens. Now: the book selector opens.

---

PR #1679, opened 2026-08-23
Feature/1646

### LINKED ISSUE #1646: Support editing playlist descriptions
Currently, playlist descriptions can't be changed in the edit playlist UI, even though we have a field for it. This task is to add support for editing a playlist's description.

Kabi4:
Fixes: #1646


Blitzy:
> [!NOTE]

## Code review — PR No. 1679 (Fixes No. 1646: editing playlist descriptions)

**Short version:** Solid, well-scoped PR that does exactly what issue No. 1646 asks for — a description field in the playlist editor, plus a reusable `ExpandableText` "Read more/less" control for long descriptions in the list. The manager refactor (`updateEditingPlaylistMetadata`) is a clean consolidation of duplicated draft-patching logic, and the tests genuinely exercise behavior rather than rubber-stamping. Nothing here blocks merge. I found no correctness bugs — only a couple of quality/accessibility items and a nit.

I ran the affected suites and the repo checks locally on head `4adafe4`:
- `vitest` (ExpandableText, CreatePlaylistForm, PlaylistManager, DiscoverPane) — **192 passed**
- `pnpm check:ts` — **0 visible errors**
- `eslint` on the four changed source files — **clean**

### On the tests (the thing to scrutinize)
These are real tests, not always-green ones:
- `ExpandableText` overflow can't be measured in jsdom (no layout), so the suite mocks `scrollHeight` on the probe only. That's the right seam — it still asserts real DOM output and the collapsed→expanded→collapsed state transitions, the first-line-only collapse, the ellipsis, and `aria-expanded`.
- The `stopPropagation` test wraps the control in a parent `onClick` and asserts the parent handler is **not** called — remove the `event.stopPropagation()` and this test fails, so it's a meaningful guard against the "Read more" click also opening/playing the playlist.
- `PlaylistManager` covers the patch, the "nothing being edited" error path, **and** end-to-end persistence via `saveEditingPlaylist` (asserts the recorded payload). `CreatePlaylistForm` covers the whitespace-only→`null` rule and newline preservation. Good edge coverage.

---

### Medium

**1. The "Read more / Read less" toggle has no visible keyboard-focus indicator.**
`ExpandableText.css` sets `.sb-expandable-text-toggle { all: unset; }`, which strips the default focus outline, and the only interactive style added back is `:hover`. There is no global `button:focus-visible` fallback in this codebase (focus rings are defined per-component — e.g. `ReadingPlanEditor.css`, `BibleReaderToolbar.css`, `SettingsPage.css`), so this control ends up with no focus indicator at all. That's a WCAG 2.4.7 gap and inconsistent with the rest of the app.

*Reproduce:* open a playlist with a long description in the Discover list, then Tab through to the "Read more" control with the keyboard — no focus ring appears, so a keyboard user can't tell it's focused.

*Suggested fix* (in `ExpandableText.css`):
```css
.sb-expandable-text-toggle:focus-visible {
  outline: 2px solid var(--sb-primary-color);
  outline-offset: 2px;
  border-radius: 2px;
}
```

---

### Low

**2. The `maxLines` prop is only half-honored — it changes the overflow threshold but not the collapsed render.**
In `ExpandableText.tsx`, `measure()` flags overflow at `lineHeightPx(probe) * maxLines + 1`, but the collapsed body always renders a single line: `displayText = text.split(/\r?\n/, 1)[0]`, and `.sb-expandable-text-body--clamped` forces `white-space: nowrap`. So a consumer passing `maxLines={2}` gets "collapse only when taller than 2 lines, but then show just 1 line" — not "show 2 lines then clamp."

This is **not a live bug** — every current call site uses the default (`maxLines = 1`), where threshold and render agree. It's a latent trap because the component is exported from `components/index.tsx` for reuse.

*Reproduce (code-level):* render `<ExpandableText maxLines={2}>` with a 3-line string; it collapses to one line rather than two.

*Suggested fix:* either drop the `maxLines` prop until it's actually needed, or make the collapsed render honor it (e.g. `-webkit-line-clamp: maxLines` with `display: -webkit-box` for the `maxLines > 1` case) so the prop's documented contract holds.

---

### Nit

**3. Dead fallback and a small first-line edge case in `displayText`.**
`text.split(/\r?\n/, 1)[0] ?? text` — `String.prototype.split` always returns at least `[""]`, so `[0]` is never `undefined` and the `?? text` branch is unreachable (the empty-string case is already handled by the earlier `if (!text) return null`). Also worth noting: a description that *starts* with a newline collapses to an empty first line, so the collapsed view reads as just `... Read more`. Minor and unlikely in practice; only mentioning since the split takes strictly the first physical line.

---

Nice work on the manager consolidation and on writing tests that would actually catch a regression. The two items above are polish, not blockers.

---

PR #1682, opened 2026-08-24
added default translation using branding

AmitPant9999:


Blitzy:
> [!NOTE]

## Code Review — Default translation via branding

Thanks for this, @AmitPant9999. The core idea is sound and small: add an optional `branding.defaultTranslationId` and use it to seed the first tab's translation. But there is production debug logging that must be removed, a precedence bug that breaks shared/deep links on branded deployments, an orphaned server-side type change, and no tests. Details below, ordered by severity.

---

### 🔴 High

**No. 1 — Debug `console.log`/`console.error` statements left in `appConfig.tsx` (ships to production)**

`readInjectedConfig()` gained eight logging calls (lines ~65–90), including one that dumps the full parsed config (`console.log("PARSED CONFIG:", parsed)`). This function runs on **every page load** — both during SSR (server logs) and at client hydration (browser console) — so this noise (and the whole config payload) ships to real users and production logs.

There is **no `no-console` ESLint rule** in `eslint.config.mts`, so lint/CI will *not* catch this — it has to be removed by hand.

*Reproduction:* build and open any page → the browser console prints `document exists:`, `app-config element:`, `app-config text:`, `PARSED CONFIG: {…}` on load.

*Suggested fix:* remove all added `console.*` calls; `readInjectedConfig` should read as it did on `develop` plus only the `defaultTranslationId` field. (Unrelated but nearby: `SeedBibleStateManager.tsx:466` already has a pre-existing `console.log("Creating SeedBibleState with options:", options)` on `develop` — worth cleaning up in a separate change, but it is not introduced here.)

---

**No. 2 — Branding default overrides an explicitly requested translation (deep-link / shared-link regression)**

In `TabsManager.tsx:497`:

```ts
translationId: branding?.defaultTranslationId || initialTranslationId,
```

`initialTranslationId` already resolves in priority order (`getInitialTranslationId`, lines 129–141): URL path translation → `?translationId=` → `?translation=` → language default. Putting `branding?.defaultTranslationId` **in front of** that value means the branding default wins over *everything*, including a translation the user explicitly named in the URL.

This is the exact case the surrounding code takes pains to respect elsewhere — see `hadExplicitInitialUrlTranslation` (line 403) and the profile-restore effect (line 871), which deliberately refuse to override an explicit deep link. This change bypasses that distinction.

*Reproduction (on a deployment with `branding.defaultTranslationId: "BSB"`):*
1. Open a deep/shared link naming a different translation, e.g. `/en/NIV/john/3` (or `?translationId=NIV`).
2. Expected: reader opens **NIV** (the requested translation).
3. Actual: reader opens **BSB**, and `commitSelectedTabToUrl` then rewrites the address bar to the BSB path.

So every shared link and bookmark to a specific translation breaks on a branded deployment — silently.

*Suggested fix:* apply the branding default only as a fallback **below** an explicit URL translation and **above** the language default. Cleanest is to fold it into `getInitialTranslationId`:

```ts
function getInitialTranslationId(
  url: URL,
  basePath: string,
  language: string,
  brandingDefaultTranslationId?: string
): string {
  const parsed = parseReadingPath(url.pathname, basePath);
  return (
    parsed?.translationId ??
    url.searchParams.get("translationId") ??
    url.searchParams.get("translation") ??
    brandingDefaultTranslationId ??
    getDefaultTranslationForLanguage(language).id
  );
}
```

Then pass `branding?.defaultTranslationId` in at the call site (line 478) and set `translationId: initialTranslationId` (drop the inline `||`). Alternatively, gate on the existing flag: `branding?.defaultTranslationId && !hadExplicitInitialUrlTranslation ? branding.defaultTranslationId : initialTranslationId`. If the intent really is for branding to always win over explicit URLs, please confirm — but that would be a surprising break of link-sharing.

---

### 🟠 Medium

**No. 3 — `server/index.ts` adds a `branding` type field that is never populated or read (incomplete/dead wiring)**

The diff adds `branding?: BrandingConfig;` to the `RenderFn` config type (line 114) and imports `BrandingConfig`, but neither `render()` call site builds it: the config objects at lines ~423 and ~954 pass only `basePath / assetHost / renderedAsMobile / renderedAsWebKit / acceptedLanguages`, and `grep` finds no other use of `branding` in `server/`. So this change compiles but does nothing.

Worth clarifying end-to-end: `defaultTranslationId` only reaches the client through the injected `#app-config` JSON (`readInjectedConfig` → `options.config.branding` → `createTabs`). Since the in-repo server never sets `config.branding`, branding presumably arrives via an external/deploy-time config mechanism (consistent with the "config the host server injected" comment). If so, the `RenderFn` type addition here is inconsistent with how branding is actually delivered.

*Suggested fix:* either wire `config.branding` through at both render call sites (and verify it actually populates the injected JSON), or drop the type/import change from `server/index.ts` so the PR contains only the parts that do something. Please confirm how `branding` is expected to be populated so the feature is verifiably reachable.

---

**No. 4 — No tests for the new behavior**

This adds a user-facing behavior with zero test coverage, and the existing suite already has good patterns for this area (`test/unit/standalone/entry-ssr.test.ts` for config injection; `TabsManager`/`BibleReader` tests for tab seeding). Per the repo's testing guidance, tests should assert real behavior and fail on the pre-fix code. Please add cases covering:

1. `branding.defaultTranslationId` seeds the first tab when the URL names **no** translation.
2. An explicit URL translation (path or `?translationId=`) **still wins** over `branding.defaultTranslationId` — this is the regression guard for No. 2, and should fail against the current `||` implementation.
3. With `branding` / `defaultTranslationId` unset, behavior is unchanged (language default).

A quick way to confirm No. 2 is genuinely a bug: write test No. 2 against this branch as-is and watch it fail.

---

### 🟢 Low / Nits

**No. 5 — Removed blank line in `appConfig.tsx`.** The blank line between `readInjectedConfig` and the `withBasePath` doc comment was deleted. Harmless, but unnecessary churn; the pre-commit `pretty-quick` hook may or may not restore it.

**No. 6 — Invalid `defaultTranslationId` handling.** If an admin configures a `defaultTranslationId` that doesn't exist (typo, or a translation later removed), the initial tab will try to load a non-existent translation. Worth confirming the reader falls back gracefully (and ideally covering it in the tests from No. 4).

**No. 7 — `||` vs `??`.** Minor: `branding?.defaultTranslationId || initialTranslationId` treats an empty-string `defaultTranslationId` as unset. That's probably the desired behavior, but `??` would be more precise if empty string should ever be meaningful. (Moot if you adopt the `getInitialTranslationId` fix in No. 2.)

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 High | Debug `console.*` logging shipped in `appConfig.tsx` (no lint rule catches it) |
| 2 | 🔴 High | Branding default overrides explicit URL translation → breaks deep/shared links |
| 3 | 🟠 Medium | `server/index.ts` `branding` type field never populated or read |
| 4 | 🟠 Medium | No tests for the new feature |
| 5 | 🟢 Low | Stray blank-line removal |
| 6 | 🟢 Low | No handling for an invalid configured `defaultTranslationId` |
| 7 | 🟢 Low | `||` vs `??` |

No. 1 and No. 2 should block merge. No. 3 and No. 4 should be resolved before this is considered complete.

---

KallynGowdy:
@Blitzy Could we get a re-review here?

KallynGowdy (approved):

Blitzy:
> [!NOTE]

## Re-review — Default translation via branding

This PR was already reviewed once (@Blitzy's review from 2026-08-24), and two follow-up commits (`801d170`, `f19abd9`) landed since. I re-checked the current head (`f19abd9`) against that review point by point. Good news first: the two blocking issues are fixed. Two items from that review are still open, plus one new observation from the follow-up commits themselves.

---

### ✅ Fixed since the last review

- **Debug logging removed.** All the unconditional `console.log` calls in `readInjectedConfig()` are gone. The one remaining `console.error("CONFIG JSON PARSE FAILED:", error)` only fires when `JSON.parse` actually throws, so it's not noise on every page load — fine to keep (minor tone nit below).
- **Precedence bug fixed.** `branding?.defaultTranslationId` is now folded into `getInitialTranslationId()` *between* the URL-based checks and the language default, and the tab-seeding call site uses the plain `translationId: initialTranslationId` (the `||` override is gone). A deep link like `/en/NIV/john/3` or `?translationId=NIV` on a branded deployment now correctly wins over `branding.defaultTranslationId`. I also checked the profile-restore effect (`hadExplicitInitialUrlTranslation`, line ~874) and the "restore stored tabs" branch (`reconcileStoredTabs`, which still uses `defaultTranslation.id`, not the branding default) — both are consistent with only applying the branding default on a first-ever, tab-less visit, which looks like the intended scope.

---

### 🟠 Still open — Medium

**No. 1 — `server/index.ts`'s `branding` field is still dead code**

The `RenderFn` type still declares `branding?: BrandingConfig` (added by this PR), but I checked both `render({...})` call sites in `server/index.ts` (around line 421 and line 952) and neither one sets `branding` anymore — it was removed in `f19abd9` along with the hardcoded test value. So this type addition is unreachable: nothing in this repo populates it, nothing reads it besides the pass-through.

Worth noting: `AppConfig` in `appConfig.tsx` *already* had `branding?: BrandingConfig` before this PR — that's not new. `RenderFn`'s inline `config` object is a hand-duplicated subset of `AppConfig`'s shape, and this PR only added the field to that duplicate. If branding is meant to be populated by something outside this repo (a separate deploy-time mechanism, consistent with "config the host server injected"), that's fine, but then the `RenderFn` type edit in `server/index.ts` does nothing here and could be dropped from this PR. If it's meant to be wired up in this server, it still needs an actual call site.

*Suggested fix:* either wire `branding` into the two `render()` calls here (and confirm it reaches the injected `#app-config` JSON), or drop the `server/index.ts` change from this PR since it's currently inert.

**No. 2 — Still no tests for the new behavior**

Across all three commits, no test file was added or changed. This is exactly the kind of change (precedence between multiple translation-id sources) where a regression is easy to reintroduce silently — the fixed precedence bug from the first review is a good example of a case a test would have caught immediately. `test/unit/seed-bible/managers/TabsManager.test.ts` already has the harness for this (mocked fetch, `FreeUseBibleAPI`, etc.) and there's no existing coverage of `defaultTranslationId`/`branding` there. Suggested cases:

1. `branding.defaultTranslationId` seeds the first tab when the URL names no translation and there are no stored tabs.
2. An explicit URL translation (path segment, `?translationId=`, or `?translation=`) still wins over `branding.defaultTranslationId` — write this one against the current code and confirm it passes now (it would have failed before `801d170`).
3. With `branding`/`defaultTranslationId` unset, behavior is unchanged.
4. A returning visitor with stored tabs is unaffected by `branding.defaultTranslationId` (matches the `reconcileStoredTabs` path not receiving it).

---

### 🟢 Low / Nits

**No. 3 — A real client's branding config was briefly committed, then reverted, inside this PR's history**

Commit `801d170` added a hardcoded `branding` object to the dev server (`"Boa Study Bible"`, `kenboa.org`, real Cloudinary logo URLs, `defaultTranslationId: "NASB95"`) for local testing, and commit `f19abd9` removed it. None of this is secret (it's a client's own public branding), but it's a live client's real config sitting in the PR's commit history. If this repo merges with "Create a merge commit" rather than squashing, that commit stays in `develop`'s permanent history. Worth squash-merging this PR, or amending the history, so test/debug values for a specific paying client don't linger indefinitely in git log.

**No. 4 — Stray blank-line removal in `appConfig.tsx`.** Still present from the original review — the blank line between `readInjectedConfig` and the `withBasePath` doc comment stays removed. Harmless churn.

**No. 5 — Invalid `defaultTranslationId` still unhandled.** If a branding config names a `defaultTranslationId` that doesn't exist (typo, or a translation later retired), the first tab will try to seed with a translation ID that doesn't resolve. I didn't trace all the way through `createInitialTabs`/`BibleDataManager` to confirm what happens today, but it's worth confirming there's a graceful fallback rather than a broken initial tab — and covering it in the tests from No. 2.

---

### Summary

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| — | 🔴 High | Debug logging in `appConfig.tsx` | ✅ Fixed |
| — | 🔴 High | Branding overriding explicit URL translation | ✅ Fixed |
| 1 | 🟠 Medium | `server/index.ts` `branding` field still unpopulated/unused | Still open |
| 2 | 🟠 Medium | No tests for the new feature | Still open |
| 3 | 🟢 Low | Real client config committed then reverted in PR history | New observation |
| 4 | 🟢 Low | Stray blank-line removal | Still open |
| 5 | 🟢 Low | No handling for an invalid configured `defaultTranslationId` | Still open |

Nothing here should block merge on its own if the team is comfortable shipping without tests, but No. 2 (tests) is worth doing given how easily the precedence bug from the first pass slipped in — a small regression test would make that class of bug impossible to reintroduce silently. No. 1 is worth a quick decision either way (wire it up or drop the dead type change) so the PR doesn't leave inert code behind.

---

PR #1684, opened 2026-08-25
fixed issue 1680

### LINKED ISSUE #1680: The mobile highlight controls jump once you touch them
### steps to reproduce 
1. Select a verse in mobile
2. Open the highlight tool 
3. Scroll the colors list
4. See that the "Swipe to see more" hint disappears and causes the layout to shift 

### what is expected
The layout doesn't shift. The hint can disappear, but it shouldn't cause everything to shift down.

AmitPant9999:
fix: #1680

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
@@ -2827,23 +2827,29 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
                 )}
               </div>
             )}
-          {isSmallScreen.value &&
-            isHighlightPickerOpen.value &&
-            showHighlightColorSwipeHint.value && (
-              <div
-                className="sb-verse-toolbar-swipe-hint sb-verse-toolbar-swipe-hint-colors"
-                aria-hidden="true"
-              >
-                <span className="material-symbols-outlined">
-                  keyboard_double_arrow_right
-                </span>
-                <span>
-                  {t("swipe-to-see-more", {
-                    defaultValue: "Swipe to see more",
-                  })}
-                </span>
-              </div>
-            )}
+          {isSmallScreen.value && (
+            <div
+              className="sb-verse-toolbar-swipe-hint sb-verse-toolbar-swipe-hint-colors"
+              aria-hidden="true"
+              style={{
+                opacity:
+                  isHighlightPickerOpen.value &&
+                  showHighlightColorSwipeHint.value
+                    ? 1
+                    : 0,
+              }}
+            >
+              <span className="material-symbols-outlined">
+                keyboard_double_arrow_right
+              </span>
+
+              <span>
+                {t("swipe-to-see-more", {
+                  defaultValue: "Swipe to see more",
+                })}
+              </span>
+            </div>
+          )}
         </div>
       )}
     </>
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
@@ -2827,16 +2827,12 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
                 )}
               </div>
             )}
-          {isSmallScreen.value && (
+          {isSmallScreen.value && isHighlightPickerOpen.value && (
             <div
               className="sb-verse-toolbar-swipe-hint sb-verse-toolbar-swipe-hint-colors"
               aria-hidden="true"
               style={{
-                opacity:
-                  isHighlightPickerOpen.value &&
-                  showHighlightColorSwipeHint.value
-                    ? 1
-                    : 0,
+                opacity: showHighlightColorSwipeHint.value ? 1 : 0,
               }}
             >
               <span className="material-symbols-outlined">
--- a/test/unit/seed-bible/components/BibleReaderToolbar.test.tsx
+++ b/test/unit/seed-bible/components/BibleReaderToolbar.test.tsx
@@ -996,9 +996,7 @@ describe("BibleReaderToolbar — mobile verse sheet drag", () => {
     await release(hintEl, 500);
 
     expect(overflow()?.style.height).toBe(`${OVERFLOW_HEIGHT}px`);
-    // Nothing left to reveal, so the hint is invisible but remains in the layout.
-    expect(hint()).not.toBeNull();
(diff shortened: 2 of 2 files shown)


Blitzy:
> [!NOTE]

## Code review — PR #1684 (fixes #1680)

**Summary of the change:** The "Swipe to see more" colors hint used to be mounted/unmounted based on `isSmallScreen && isHighlightPickerOpen && showHighlightColorSwipeHint`. When you scrolled the color list, `showHighlightColorSwipeHint` flipped to `false`, the element was removed from the DOM, and everything below it shifted — the jump reported in #1680. The PR keeps the element mounted (whenever `isSmallScreen`) and instead toggles its **opacity** between `1` and `0`. Fading instead of unmounting is the right idea and it does fix the reported layout shift. 👍

The problems below are about the tests and the scope of "always mounted."

---

### 🔴 High — the two changed tests no longer test what they claim, and now pass no matter what the swipe-up hint does

Both modified tests live in the `mobile verse sheet drag` block and are named for the **swipe-up** hint (`sb-verse-toolbar-swipe-hint`), but after the change they actually assert on the **colors** hint (`sb-verse-toolbar-swipe-hint-colors`).

Why: the shared helper is
```js
const hint = () =>
  container.querySelector<HTMLElement>(".sb-verse-toolbar-swipe-hint");
```
`querySelector` returns the *first* match in DOM order. The colors hint carries both classes (`sb-verse-toolbar-swipe-hint sb-verse-toolbar-swipe-hint-colors`), so it also matches this selector. In these tests the highlight picker is never opened, so:

- While the sheet is collapsed, the swipe-up hint is present and `hint()` returns it.
- Once the sheet is fully expanded (`isVerseSheetExpanded` becomes true), the swipe-up hint **unmounts** (its render condition includes `!isVerseSheetExpanded.value`), and `hint()` falls through to the colors hint — which is *always* mounted at `opacity: 0` because the picker is closed.

So `expect(hint()?.style.opacity).toBe("0")` is a tautology: the colors hint is unconditionally `opacity: 0` in these tests, regardless of any swipe-up drag behavior. The assertion is decoupled from the code the test name describes, and the inline comment ("the hint is invisible but remains in the layout") is simply wrong for the swipe-up hint — that hint does *not* remain in the layout, it unmounts.

**Proof (reproduction for a human dev):** on the PR branch, change only the colors hint's style at `BibleReaderToolbar.tsx:2834` to `style={{ opacity: 1 }}` and run:
```
pnpm vitest run test/unit/seed-bible/components/BibleReaderToolbar.test.tsx -t "mobile verse sheet drag"
```
Both `expands the sheet when the swipe-up hint is tapped` and `settles open when released past halfway` fail with `expected '1' to be '0'` — even though nothing about the swipe-up hint was touched. That confirms the assertions are reading the colors hint, not the element under test. Equivalently, you can break the swipe-up hint (e.g. remove its `!isVerseSheetExpanded` gate so it stays visible after expansion) and both tests still pass — a genuine regression in the tested behavior would sail through.

Two consequences:
1. The old assertion `expect(hint()).toBeNull()` was meaningful — it verified the swipe-up hint *stands down* (unmounts) once the sheet is open. That coverage is now gone.
2. The actual bug being fixed — the colors hint staying in the layout while fading on scroll — has **no test at all** (see Low, No. 2 below).

**Suggested fix:** make the helper unambiguous and restore the original intent, e.g.
```js
const hint = () =>
  container.querySelector<HTMLElement>(
    ".sb-verse-toolbar-swipe-hint:not(.sb-verse-toolbar-swipe-hint-colors)"
  );
```
then keep `expect(hint()).toBeNull()` in both tests. That re-couples them to the swipe-up hint and makes them able to fail again.

---

### 🟠 Medium — the colors hint now reserves empty space at the foot of the sheet even when the highlight picker is closed

The new render condition is just `isSmallScreen.value && (...)`, so on every mobile verse selection the colors-hint row is in the DOM at `opacity: 0` whenever the picker is *not* open. It is a normal-flow flex element (`.sb-verse-toolbar-swipe-hint` → `display: flex; margin-top: 0.5rem`, and `.sb-verse-toolbar-mobile .sb-verse-toolbar-swipe-hint-colors { display: flex }`), with no absolute positioning, so it occupies real vertical height (~an icon + text line + top margin) at the bottom of the sheet in states where previously nothing was rendered there. That's a small but permanent dead gap for the common case (verse selected, not highlighting) — a bit against the spirit of #1680, which was about the sheet not changing size unexpectedly.

It also means an invisible element carrying `touch-action: none; cursor: pointer` sits at the bottom edge whenever the picker is closed.

**Suggested fix:** only reserve the space while the picker is actually open, since that's the only time a scroll can dismiss the hint:
```jsx
{isSmallScreen.value && isHighlightPickerOpen.value && (
  <div
    className="sb-verse-toolbar-swipe-hint sb-verse-toolbar-swipe-hint-colors"
    aria-hidden="true"
    style={{ opacity: showHighlightColorSwipeHint.value ? 1 : 0 }}
  >
    ...
  </div>
)}
```
The layout shift in #1680 happens *during* an open picker session (scrolling the swatches), so keeping it mounted across that scroll is enough — the element still never unmounts mid-session, so no jump — while closing the picker removes it and the foot goes back to clean.

---

### 🟡 Low

1. **Root cause of the test problem — the ambiguous shared selector.** `hint()` matching both hints via the base class is what let the assertions silently retarget. Even beyond the two changed tests, any future test using `hint()` after the picker opens will read whichever hint happens to be first in the DOM. The `:not(...)` selector above (or a dedicated `colorsHint()` helper) removes the ambiguity.

2. **No regression test for the actual fix.** There is nothing asserting the behavior #1680 is about: open the highlight picker on mobile, scroll the swatch list past the dismiss threshold (`scrollLeft > 4`), and verify the colors hint (a) is still in the DOM and (b) has `opacity: 0` — i.e. it faded rather than unmounting. Per the repo's testing guidance, that test should fail on the pre-fix code (where the element unmounts and would be `null`) and pass after. Worth adding so this doesn't silently regress.

3. **Misleading comments.** The `// Nothing left to reveal, so the hint is invisible but remains in the layout.` comments were added to the two swipe-up tests, but they describe the colors-hint fix, not the swipe-up hint those tests exercise. They should be removed/corrected along with the selector fix in No. 1 of the High item.

---

**Net:** the production fix is on the right track and resolves the visible jump, but (1) it broadens "always mounted" further than needed, leaving dead space when the picker is closed, and (2) the test changes weaken two existing tests into always-passing checks while leaving the real fix uncovered. I'd want the colors-hint test added and the swipe-up tests re-pointed at their own element before merge.

---

AmitPant9999:
@KallynGowdy Updated the tests to reflect the new expected behavior: the swipe hint is now hidden instead of removed from the DOM, so its layout space is preserved and the toolbar no longer shifts when the hint disappears.

Issue #1197, opened 2026-05-01
[refactor] Prompt users to switch their translation when changing language


KallynGowdy:
Currently, we don't prompt the user to switch their translation when they change their UI language. We should prompt in the following scenario:

- The user changes the UI language
- There is only one tab
- The tab translation language is different from the newly selected language
- We haven't prompted before in the same session
- If the user is logged in, then we should not prompt if the user has selected "never ask again" before.
- A complete translation exists for the selected UI language (we should not prompt if there is no complete translation for the selected UI language)

When we prompt, we should give the user the option to select "never ask again" if they are logged in. If they select "never ask again", then we should save their preference to their user profile.

KallynGowdy:
We need to update the current language -> translation switch to follow the rules listed above

PR #1685, opened 2026-08-25
Feature/issue 1197

Bored-Wizard:


Blitzy:
> [!NOTE]

## Code review — PR #1685 (Feature/issue 1197)

Reviewed against issue #1197 ("Prompt users to switch their translation when changing language"). Overall this is a careful, well-tested implementation: the gating rules from the issue (single tab, once per language, complete translation exists, "never ask again", signed-in vs. signed-out storage) are all present and covered by real regression tests. I ran the new/affected unit suites locally and all 348 tests pass, and I confirmed the SeedBibleStateManager regressions fail on the pre-fix behavior (they were adjusted to add `confirmTranslationSwitch()` calls where the switch used to happen silently).

The findings below are mostly about **consistency between the "ask" path and the "always"/silent path**, plus one UX decision worth a second look before merge. Nothing here is a hard blocker, but No. 1 and No. 2 are worth resolving.

---

### Medium

**No. 1 — The "always" (auto-switch) path can apply a _different_ translation than the prompt would, including a partial/NT-only one.**

The prompt path and the silent path resolve the target translation differently:

- The prompt offers, and `confirmTranslationSwitch` applies, `findCompleteTranslationForUiLanguage(...)` — a **complete** (66-book) translation. The prompt only appears at all when a complete translation exists.
- The `"always"` branch (and the no-context branch) in `applyBibleTranslationForUiLanguage` applies `nearest.translation` from `getNearestBibleTranslationForUiLanguage`, which goes through `findAvailableTranslationForUiLanguage` and happily returns a **partial** translation — the hardcoded default even when it's incomplete, or the first in-language entry in catalog order.

`I18nManager.tsx`:
```ts
if (preference === "always") {
  await applyBibleTranslation?.(nearest.translation); // may be partial
  return;
}
// ...
const prompt = buildTranslationSwitchPrompt(context, uiLanguage, available);
// prompt.translation = findCompleteTranslationForUiLanguage(...) // complete only
```

This is concrete, not hypothetical: several hardcoded defaults in `DEFAULT_TRANSLATIONS_BY_LANGUAGE` are NT-only, e.g. `["am", { id: "amh_amh" }]` (Amharic NT) and `["ja", { id: "jpn_loc" }]` ("New Japanese NT"). For a language whose default is partial but which also has a complete sibling in the catalog, the same "switch to X" action gives the user a **complete** Bible when they answer the prompt, but the **NT-only** default when they've opted into "always". It also means "always" will silently switch to a partial text even in cases where the "ask" flow would have stayed silent (no complete translation → no prompt → no switch).

A user opts into "always" by confirming a prompt that promised a whole Bible; getting an NT-only text on the next switch is surprising.

**Reproduction (conceptual):**
1. Catalog contains an NT-only Japanese default (`jpn_loc`) and a complete Japanese translation.
2. Answer one prompt with "Yes, switch" leaving "Never ask again" ticked (settles preference to `always`).
3. Change the UI language to Japanese.
4. Observe the tab switches to `jpn_loc` (NT only), not the complete Japanese Bible the prompt would have offered.

**Suggested fix:** make the silent path prefer the same translation the prompt would, e.g. use `findCompleteTranslationForUiLanguage(uiLanguage, available)` in the `"always"` branch and fall back to `nearest.translation` only when no complete text exists (or reuse `buildTranslationSwitchPrompt`'s resolution so the two paths can't drift). A unit test asserting "always" applies the complete sibling when the default is partial would lock this in — the current `always` tests only use catalogs where nearest already equals the complete translation, so the divergence isn't exercised.

---

**No. 2 — "Never ask again" defaults to _checked_, so any single answer permanently settles the preference — including "Choose a different translation".**

`translationSwitchNeverAskAgain` is initialized to `true` and re-set to `true` every time a prompt opens. Because `answerTranslationSwitchPrompt` saves the preference whenever the box is ticked, every button settles the global preference unless the user first unticks:

- "Yes, switch" → saves `always`
- "No, keep reading" → saves `never`
- "Choose a different translation" → saves `never`

The issue text frames this as giving the user "the option to select 'never ask again'", which reads more like an explicit opt-in than a checked-by-default. The most surprising case is the third one: a user who clicks "Choose a different translation" is actively engaging with the feature, yet (with the default-ticked box) they silently opt themselves out of all future prompts and auto-switches. The only way back is the Settings toggle, which they have no reason to look for.

**Reproduction:**
1. Change the UI language so the prompt appears (single tab, complete translation available).
2. Click "Choose a different translation" without touching the checkbox.
3. Change the UI language to another supported language later.
4. Observe: no prompt and no switch — the preference is now `never`.

**Suggested fix:** confirm with the team whether default-ticked is intended. If the goal is "answering once settles it," at minimum consider **not** treating "Choose a different translation" as a `never` settle (the user is choosing a text, not declining the feature), or default the checkbox to unticked so settling is a deliberate act. This is a product decision, so flagging rather than prescribing.

---

### Low

**No. 3 — Silent behavior change for partial-only languages and multi-tab (confirm intended).**

Before this PR, direct language support always applied the nearest translation. Now, in the default `"ask"` state, `applyBibleTranslationForUiLanguage` returns without switching whenever the prompt is gated off — including when only a **partial** translation exists for the target language, or when **more than one tab** is open. So a user on the default preference who changes the UI to a partial-only language (or who has two tabs open) will no longer see their Bible text follow the UI at all, where it used to switch silently.

The multi-tab case matches the issue's single-tab rule and is almost certainly intended. The partial-only case is a side effect the issue doesn't explicitly address (it only says "don't _prompt_ without a complete translation"). Worth a quick confirmation that dropping the silent switch entirely for those languages is desired, rather than switching silently to the partial text as before.

**No. 4 — Minor: translation-key naming inconsistency.**

The new keys are `translationSwitch.*` (dotted namespace) except `"ask-before-switching-translation"` (kebab-case) used by the Settings toggle. Harmless, but aligning it (e.g. `translationSwitch.askBeforeSwitching`) would keep the group together. Only `en.json` was updated, which is correct per the repo convention.

---

### Notes on tests

The tests read as genuine behavioral coverage, not tautologies:

- `ModalManager` gains real coverage for `onClose` firing exactly once, not firing for unknown ids, and — importantly — not recursing when an `onClose` handler closes the same modal again (the drop-before-invoke ordering). Good edge coverage.
- The `I18nManager` suite exercises each gate independently (tab count, tab-language alias equality via `he`/`iw`, partial-only, already-answered, reset semantics) and asserts `apply` is _not_ called on the negative paths, so a regression that switched silently would fail them.
- The `SeedBibleStateManager` integration tests wire the real state and drive through the modal host, including the host-close-as-dismissal round trip. The `dismissKeepingPrompts` helper correctly unticks "never ask again" so later assertions test the once-per-language rule rather than the opt-out — a subtlety that's easy to get wrong.

One coverage gap worth adding alongside the No. 1 fix: an "always"/silent test where the language's default or first catalog entry is **partial** but a complete sibling exists, asserting which translation actually gets applied.

CI checks had not yet reported on the head commit at review time.

---

KallynGowdy:
@Bored-Wizard @Blitzy I thought we already had this functionality? If I load up develop and switch my language, then it will ask if I want to switch translations as well. Does this add anything to that?

Blitzy:
@KallynGowdy Yes this was done in #1677 which itself referenced issue #1126 which appears to be nearly identical to the issue that this PR addresses #1197. Appears we just accidently logged the same issue twice a month a part. I say we close this one and keep what we already have.

KallynGowdy:
Sounds good. Thanks @Bored-Wizard. Sorry we had a duplicate task

Bored-Wizard:
> Sounds good. Thanks @Bored-Wizard. Sorry we had a duplicate task

🥲

PR #1690, opened 2026-08-25
Give Dark theme similar orange accents to the light theme

Blitzy:
This PR changes the primary and secondary colors of the Dark theme to match the style of the Light theme.

image

image
image

image
image

### DIFF
--- a/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/ThemeManager.tsx
@@ -653,10 +653,10 @@ const DARK_THEME: BibleTheme = {
   id: "dark",
   name: "Dark",
   variables: {
-    primaryColor: "#e6e6e6",
+    primaryColor: "#e07b4c",
     primaryFontColor: "#111111",
 
-    secondaryColor: "#262626",
+    secondaryColor: "#433228",
     secondaryFontColor: "#f5f5f5",
 
     tertiaryColor: "#1c1c1c",


Blitzy:
> [!NOTE]

## Code Review — Give Dark theme similar orange accents to the light theme

**Scope:** One file, two values changed in `DARK_THEME` (`packages/seed-bible/seed-bible/managers/ThemeManager.tsx`):

- `primaryColor`: `#e6e6e6` → `#e07b4c`
- `secondaryColor`: `#262626` → `#433228`

**Overall:** This is a clean, low-risk, purely cosmetic change, and the new values are sensible. The orange primary now matches the light theme's `#e07b4c`, and the dark-brown secondary (`#433228`) keeps good contrast against its `secondaryFontColor` (`#f5f5f5`). No blocking issues. A few low-severity observations below, mostly things to confirm are intentional rather than defects.

---

### 🟡 Low — `primaryFontColor` diverges from the light theme on the same orange background

Both themes now use `primaryColor: #e07b4c`, but the font that sits on top of it differs:

- Light theme: `primaryFontColor: "#fff"` (white on orange)
- Dark theme: `primaryFontColor: "#111111"` (near-black on orange) — **unchanged by this PR**

This is used wherever the primary color is a fill — e.g. Onboarding buttons, `MobileSettingsSheet` active option, `ReadingPlansPane` buttons, `AnnotationConflictModal` primary button. The practical effect is that a primary-colored button shows **white text in light mode but black text in dark mode**, even though the button background is the same orange in both.

Worth noting the accessibility angle actually favors what the dark theme already has:

- Black `#111111` on `#e07b4c` ≈ **6.4:1** contrast → passes WCAG AA.
- White `#fff` on `#e07b4c` ≈ **3.0:1** → fails WCAG AA for normal text.

So keeping black in dark mode is the *more* readable choice — I'd lean toward leaving it. The only question is whether the cross-theme inconsistency (white vs. black label on the same orange) is a deliberate call or an oversight, given the PR's stated goal of matching the light theme's style. If the intent is strict visual parity, you'd switch dark to `#fff`, but that would regress contrast. My recommendation: keep `#111111` and treat this as intentional.

**How to see it:** Open the alpha deployment (linked in the auto-comment above), toggle between Light and Dark, and compare the label color on any primary button (e.g. the Onboarding "get started" button or a Reading Plan action button).

### 🟡 Low — No CHANGELOG entry

This is a user-facing visual change, and the repo documents those in `CHANGELOG.md`. The `## TBD` section's `### 🔧 Changed` block currently has no entry for this PR. Suggest adding one, e.g.:

```
- Give the Dark theme the same orange accent colors as the Light theme. ([#1690](https://github.com/HelloAOLab/seed-bible/pull/1690))
```

### 🔵 Nit — Selected-tab text contrast is borderline (pre-existing pattern)

`selectedTabFontColor` resolves to `var(--sb-primary-color)` (orange) on `selectedTabBackground` = `secondaryColor` (`#433228`). That's roughly **4.1:1** — just under the 4.5:1 AA threshold for normal text, though fine for larger/bold tab labels. This isn't a regression: the light theme's equivalent (orange on `#faddd1`) is actually lower (~2.3:1), so the dark theme is the better of the two. No change required; flagging only for awareness if a broader contrast pass is ever done.

---

### Tests

There are no automated tests in this PR, which is appropriate — these are literal color-value constants with no behavior to assert, and a test pinning the exact hex strings would only restate the values rather than verify real-world behavior. Verification here is inherently visual, via the alpha deployment. No test concerns.

### Notes checked (no action needed)

- **`base.css` companion values:** `app/styles/base.css` defines `--sb-primary-color`/`--sb-secondary-color` at `:root` as the *light* defaults, and there's no `@media (prefers-color-scheme: dark)` block. Since `ThemeManager` injects theme values body-scoped (which win over `:root`), the dark theme reads from `ThemeManager`, so no companion edit in `base.css` is required for this change.
- `linkColor` in the dark theme was already `#e07b4c`, so links stay consistent with the new accent.

---

PR #1693, opened 2026-08-25
Implement compact discover panel

### LINKED ISSUE #1567: Implement compact discovery content panel
Utilizing the [figma design](https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=8381-3735&t=D7xWKh3SuMOR4GMX-1) from #1271, implement the discover panel using this new compact design.

Do the following:
- Visualize discovered content from reading states
    - It should be visualized by an additional view that is rendered per visible-tab.
    - It should display discovered study notes, cross references, and content.
    - In the future, it will be used for displaying user annotations (notes), and other user-generated content.
    - On desktop single tab view, the discovered content view placed to the side (right on LTR and left on RTL languages) of the scripture text. The discovered content view should be sticky to the screen, and should have its own scroll viewport so that scrolling the scripture view causes the discovered content view to stay on the screen. When using this "side contextual view", the discovered content view should only be visible if there are no panes that are also displayed using the "side" placement.
    - On desktop multiple tab view, the discovered content view should be placed after (below) the scripture text and utilize the same scroll viewport as the scripture text.
    - On mobile, the discovered content view should be placed after (below) the scripture text and utilize the same scroll viewport as the scripture text.

KallynGowdy:
Adds a compact discover panel for user notes, cross-references, and more. On mobile, it is displayed below the bible reader for each chapter (if there are user notes/content). On desktop, the user can toggle it to either be displayed below the bible reader or to the side if there is space.

### Screenshots

image

image

image

image

There's a lot of spacing between the discover panel and the end of the scripture text because we don't want it to be in the way:
image

Closes #1567


Blitzy:
> [!NOTE]

Reviewed the diff against `develop` (36 files, ~2,900 additions / ~1,100 deletions across the new `DiscoverContentPanel`, the `DiscoverPane` → `AnnotationsSection`/`DiscoveredResultsSections` split, the `BibleReader`/`TabsLayout` integration, and the supporting managers). No linked issue or prior review comments were found to cross-reference. All 126 tests in the touched files pass as-is, and `en.json` is internally consistent (the `scripture-margin*` → `scripture-width*` rename and the new `discover-content-panel`/`discover-book-title`/`show-all` keys all resolve correctly).

## High severity

**1. Leaked effect keeps mutating shared settings after a tab is disposed**
`packages/seed-bible/seed-bible/managers/BibleReadingManager.tsx` (~line 2870)

```ts
if (settingsManager) {
  let isFirstRun = true;
  effect(() => {
    const value = discoverContentPanelInline.value;
    if (isFirstRun) { isFirstRun = false; return; }
    untracked(() => settingsManager.setDiscoverContentPanelInline(value));
  });
}
```
The disposer returned by `effect(...)` is discarded instead of being pushed onto `effectDisposers` (which `disposeReadingState()` drains on teardown). Every other effect in this file does this correctly a few lines later (`effectDisposers.push(stopDiscoverEffect)`).

*Repro:* Create a reading state with a `settingsManager`, call `state.dispose()`, then set `state.discoverContentPanelInline.value = false`. The effect still fires and writes through to the shared `SettingsManager`. In the app this means every closed tab leaves a live effect closure — a memory leak that grows with tab churn, and a disposed tab can still mutate a live user setting.

*Fix:* `effectDisposers.push(effect(() => { ... }))` instead of the bare `effect(...)` call.

**2. Empty discover-panel wrapper always renders, adding dead space and defeating the "hide when empty" behavior**
`packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx` (~line 2060), `BibleReader.css` (~lines 1073-1096), `TabsLayout.tsx` (~line 714)

`DiscoverContentPanel` correctly returns `null` when there's nothing to show for a chapter, but `BibleReader.tsx` wraps it unconditionally:
```tsx
const extraContent = discoverPanel && (
  <div className="sb-bible-reader-discover-panel">{discoverPanel}</div>
);
```
`discoverPanel` is always a truthy `<DiscoverContentPanel .../>` element regardless of what it renders internally, so the wrapper `<div>` is emitted into the DOM even when the inner component renders nothing.

*Confirmed by reproduction:* temporarily asserting on the outer wrapper in the existing "renders nothing when there is nothing discovered for the chapter" test in `test/integration/seed-bible/components/TabSlotReader.test.tsx` produced `OUTER=true HTML=<div class="sb-bible-reader-discover-panel"></div>`.

*Impact:* on mobile, `.sb-bible-reader-mobile .sb-bible-reader-discover-panel { margin: 2.5rem 1rem 2rem; }` adds unwanted blank space below the chapter text on every chapter with nothing to discover — a common case for casual reading. On desktop, `.sb-bible-reader-main-content:has(+ .sb-bible-reader-discover-panel) { padding-bottom: 5rem; }` is now always true, so that padding never actually varies with content.

*Fix:* don't decide whether to render the wrapper based on prop truthiness — move the `sb-bible-reader-discover-panel` class onto `DiscoverContentPanel`'s own root (it already returns `null` when empty), or have it expose a "has content" signal that `BibleReader.tsx` checks before rendering the wrapper.

**3. No regression test for the swipe-vs-scroll fix the PR commit claims to make**
`test/integration/seed-bible/components/TabSlotReader.test.tsx`

The commit "fix: An issue where scrolling in the compact discover panel wouldn't block swiping between chapters" corresponds to a new `isInsideDiscoverPanel` early-return in `onTouchStart` in `TabsLayout.tsx`. The logic looks correct on inspection, but none of the 148 new lines in this test file exercise it — the new `describe` blocks only assert on CSS classes/DOM placement, never dispatch a touch event inside `.sb-bible-reader-discover-panel`. The file already has a `dispatchTouch(...)` helper used by ~15 existing swipe tests, so this would be straightforward to add.

*Fix:* add a test that renders the mobile `TabSlotReader` with a non-empty discover panel, dispatches `touchstart`/`touchmove` with a horizontal delta on an element inside `.sb-bible-reader-discover-panel`, and asserts the swipe track's transform doesn't change (and/or that chapter navigation isn't invoked). Per this repo's regression-test convention, confirm it fails with the `isInsideDiscoverPanel` check temporarily removed before landing it.

## Medium severity

**4. Compact discover panel gets stuck mid-scroll on mobile (unconditional `position: sticky`)**
`packages/seed-bible/seed-bible/components/DiscoverContentPanel/DiscoverContentPanel.css` (lines 5-6)

`.sb-discover-content-panel { position: sticky; top: 1rem; ... }` has no accompanying mobile override. `BibleReader.css`'s mobile rule for this class (`.sb-bible-reader-mobile .sb-discover-content-panel`) only resets `box-sizing`/`min-width`/`max-width`, never `position`.

*Repro:* on a phone-width viewport, open a chapter with annotations or discovered content so the panel renders below the scripture text, then scroll down through the swipe container (`.sb-reader-swipe-panel`, a real scroll container). Once the panel reaches ~1rem from the top, it locks there instead of scrolling away with the rest of the passage, behaving like a sticky sidebar even though it's meant to be stacked content on mobile.

*Fix:* add `position: static;` to the existing `.sb-bible-reader-mobile .sb-discover-content-panel` override, or scope `position: sticky` to the desktop side-by-side layout only.

**5. Doc comment claims bidirectional settings sync; implementation is write-only**
`BibleReadingManager.tsx` (doc ~line 417-428, impl ~lines 2866-2892)

The field doc says `discoverContentPanelInline` is "kept in sync with...the user's persisted setting," but the only wiring found is local signal → `settingsManager.setDiscoverContentPanelInline` (write-through only; no effect reads the setting back into the signal after the initial seed).

*Concrete effect:* with two tabs open simultaneously sharing the same `SettingsManager`, toggling the placement in tab A persists the change but tab B's already-constructed signal stays stale until that tab is recreated. This may be an intentional per-tab-override design, but the doc comment overstates it and should be corrected (or the sync made bidirectional, if that's the intent).

**6. A read-only UI nudge silently overwrites the user's persisted placement preference**
`BibleReadingManager.tsx` (~2871-2873, comment acknowledges this), `BibleReader.tsx` (~line 1722, `handleAnnotationVerseClick`)

`handleAnnotationVerseClick` sets `readingState.discoverContentPanelInline.value = true` purely to reveal a note when the user clicks an annotated verse marker — not an explicit settings action. Because the persistence effect (Finding 1) writes through on *every* change, clicking a note marker silently overwrites and persists a user's earlier explicit "show below content" choice to "inline," with no way to distinguish a temporary reveal from a deliberate preference change. The code comment names this tradeoff explicitly, so it may be intended — worth confirming with product intent; if not, give the nudge its own non-persisted signal.

**7. Over-mocking a sibling manager, plus a "was this called" assertion**
`test/unit/seed-bible/managers/BibleReadingManager.test.ts` (~lines 2740-2769)

The test hand-rolls a fake `SettingsManager` (`{ settings, setDiscoverContentPanelInline: vi.fn(...) }`) instead of using the real `createSettings(...)` construction that `SettingsManager.test.ts` already uses elsewhere in this PR. Per this repo's testing conventions, mocking should stop at the `OsManager`/CasualOS boundary, not at a sibling manager — this only proves "BibleReadingManager calls whatever function is passed under that name," not that persistence actually works. The primary assertion is also exactly the kind this repo's conventions warn against:
```ts
expect(settingsManager.setDiscoverContentPanelInline).toHaveBeenCalledWith(false);
```
*Fix:* build the test on a real `SettingsManager` (as the new `SettingsManager.test.ts` tests already do) and assert on the actually-persisted value.

**8. Duplicate tests in `DiscoverContentPanel.test.tsx` that don't test what their names claim**
`test/unit/seed-bible/components/DiscoverContentPanel.test.tsx`

- "still renders discovered content when the placement toggle is off" (~142-158) and "renders the discovered content" (~173-186) share byte-for-byte identical fixtures and assertions. `DiscoverContentPanel.tsx` never reads `discoverContentPanelInline` (that flag only affects placement logic in `BibleReader.tsx`), so the first test's name promises coverage it doesn't provide — it's a mislabeled duplicate.
- "renders nothing when there are no discovered results or annotations, even with the toggle on" (~160-171) and "renders nothing when there are discovered results absent but also no annotations" (~208-217) are likewise functionally identical, since the default mock fixtures already produce the empty state both tests claim to construct explicitly.

*Fix:* drop one of each pair, or repurpose the first of each pair to actually exercise `discoverContentPanelInline` through `BibleReader`/`TabsLayout`, where that flag is actually consumed.

**9. Missing coverage for the panel's core interactive behavior**
`test/unit/seed-bible/components/DiscoverContentPanel.test.tsx`

The filter-chip mechanism (`activeFilter` signal, click-to-filter, and the documented fallback to "all" when the active filter's content type disappears) has no test coverage — no test clicks a `.sb-dcp-chip`, exercises the fallback behavior, or clicks "+ Create"/"Show All" and asserts `createNewAnnotation()`/`openDiscover()` fire (both are already mocked in `createMockState()` but never asserted against). Recommend adding at least a chip-click test and one create/show-all click-through test.

## Low / Nit

**10.** `PaneLayout.css` (~lines 306-308): comment references `.sb-discover-content-panel--side`, a class that no longer exists anywhere in the codebase (leftover from an earlier iteration on this branch — see "refactor: Remove the variants for the discover content panel"). Update or remove the comment; the real mechanism is `.sb-bible-reader-content--discover-below` in `BibleReader.css`.

**11.** `DiscoverContentPanel.css` (line 7): `flex: 1 1 22.5rem;` on `.sb-discover-content-panel` is a no-op — its parent (`.sb-bible-reader-discover-panel`) isn't a flex container, so this element is never a flex item. Not currently harmful, but misleading next to the wrap-fix logic in `BibleReader.css`; worth removing to avoid confusing a future edit.

**12.** `BibleReader.tsx` (~line 1183): `readingPlanBelongs?: ComponentChildren` is declared on `BibleReaderProps` but never read, and `TabsLayout.tsx` never passes it — dead leftover from the `ReadingPlanBelongsCard` → `ReadingPlansSection` refactor (the real integration now happens inside `DiscoverContentPanel.tsx`). Safe to remove.

**13.** `BibleReader.css` (~lines 1055-1076): the comment claims `justify-self: flex-end` is needed to "hug the end of the row" on desktop, but `justify-self` has no effect on flex items — the panel ends up last purely by DOM order. Harmless no-op, but worth fixing the comment; also worth a manual check in a real browser that the narrow-phone `justify-self: auto` reset (browser support for `justify-self` on block boxes is newer/inconsistent) behaves as intended.

**14.** `BibleReadingManager.test.ts` (~lines 550-556): a signal round-trip assertion (`state.discoverContentPanelInline.value = false; expect(...).toBe(false);` and the `true` counterpart) only verifies that `@preact/signals` reads back what was written to it — not this manager's logic. The default-value assertion right above it is the meaningful part; the round-trip lines add nothing.

**15.** `BibleToolsManager.test.tsx` (~lines 689-703): "is visible when there are discovered cross references, study notes, or content" loops three cases inside one `it()` — a failure on the first case gives no case-identifying message. Prefer `it.each` or separate `it` blocks.

## Summary

The refactor of `DiscoverPane` into `AnnotationsSection`/`DiscoveredResultsSections`/`DiscoverContentPanel` is a faithful, well-executed split (verified line-for-line against the removed code), and the new highlight-and-scroll behavior is well tested. The two most important things to fix before merge are Finding 1 (the leaked effect that keeps mutating shared settings after a tab closes) and Finding 2 (the always-rendered empty wrapper, which will visibly add blank space below plenty of ordinary chapters). Finding 3 is worth closing before merge too, since it's a regression test for a bug this exact PR already had to fix once.

---

KallynGowdy:
@Blitzy this should be ready now

Blitzy:
This is fantastic @KallynGowdy 

What do you think about hiding the compact panel while the full discovery panel is open?

image

Blitzy:
@KallynGowdy what do you think about making the compact panel as wide as the scripture width when it is mounted in the footer of the reader?

I can also see just center aligning it as being more preferrable if the making it the width of scripture doesn't look great.

image

image

Blitzy (dismissed):
There really isnt anything i would say is blocking on this, this looks great and functions well. Anything like the comments i posted above can be follow ups post-merge.

KallynGowdy:
> This is fantastic @KallynGowdy
> 
> What do you think about hiding the compact panel while the full discovery panel is open?
> 
> image

I'm not sure that we need to spend time on this specifically. Long term, I kinda see the full discover panel as being replaced completely by the compact discover panel and the new profile screen that Mazen is working on

KallynGowdy:
> @KallynGowdy what do you think about making the compact panel as wide as the scripture width when it is mounted in the footer of the reader?
> 
> I can also see just center aligning it as being more preferrable if the making it the width of scripture doesn't look great.
> 
> image image

This is a good idea, and it mostly works but there is one limitation: When the compact discover panel is enabled to display to the side and it happens to wrap because of space, then it won't stretch to cover the full width. This is because we can't easily detect when it wraps and we have a max-width property to limit how much space it can take away from the rest of the bible reader.

PR #1694, opened 2026-08-26
Added Query Param to control chat first so it can appear prior

### LINKED ISSUE #1628: Support a chat-first UI
Currently, there is no way to configure the Seed Bible to be "Chat First". This task is to implement support for this.

Currently, the Seed Bible always shows the same toolbar buttons on mobile: Today, Search, Bible, Bookmarks and More.

Image

In the future, it will be reworked to show: Today, You, Bible, Discover, and More.

In either case, we should support setting a query parameter called `chatFirst=true` that makes the fourth toolbar option the "Chat" toolbar button instead of Bookmarks/Discover. This would allow an integration to load the Seed Bible and make the chat option prominent for users.

Kabi4:
- Add a boot-only `?chatFirst=true` query param so integrations can load Seed Bible with Chat prominent.
- Flag is latched from `initialUrl` (case-insensitive `"true"` only) so later navigation cannot flip the layout mid-session.
- Covers overlay edge cases (Chat ↔ Search/Today/Bible/Bookmarks) and mobile ↔ desktop resize behavior, with unit tests.

Closes #1628


Blitzy:
> [!NOTE]

## Code Review — PR #1694: `?chatFirst=true` query param

Reviewed against the requirements in issue #1628 ("Support a chat-first UI"). The PR does what the issue asks: a boot-only `?chatFirst=true` param makes Chat the fourth mobile tab (in place of Bookmarks, which moves into More) and keeps Chat prominent in the desktop/laptop labeled toolbar.

**Overall:** This is a clean, well-scoped implementation with genuinely strong test coverage. The flag is correctly latched from `initialUrl` (the frozen boot snapshot in `NavigationManager`), so later in-app navigation can't flip the layout mid-session — matching the "boot-only" intent. I verified the full suite runs green (85 passed) and that the tests exercise real behavior rather than always-succeeding (details below). No high-severity issues found. The items below are low-severity / polish.

---

### 🟡 Low

**No. 1 — The More-menu reorder changes the default (non-chat-first) experience too.**

In `MobileMoreMenu`, the item order was flipped:

```diff
- ...extraItems,
- ...(pinnedItems ?? []),
+ ...(pinnedItems ?? []),
+ ...extraItems,
```

This was needed so demoted **Bookmarks** lands at the top of More under chat-first (the test asserts `labels[0] === "Bookmarks"`). But `pinnedItems` also contains **Tabs** in the *default* build, so this silently reorders the More menu for **every** mobile user, not just chat-first ones:

- **Before:** extension-contributed tools appeared first, then `Tabs`.
- **Now:** `Tabs` appears first, then extension-contributed tools.

There is no test covering the default-case ordering, so this change is easy to miss.

_Reproduction:_ On mobile **without** `?chatFirst=true`, with an extension installed that contributes a controllable toolbar tool that overflows into More, open the More menu — `Tabs` now sits above the extension tool where it used to sit below.

_Suggestion:_ If reordering the default menu is intended, confirm it and add a small assertion for the non-chat-first order so it's a deliberate, documented choice. If it isn't intended, keep the pinned-first ordering scoped to the chat-first branch (e.g. only front-load `pinnedItems` when `isChatFirst`).

**No. 2 — Hardcoded English aria-labels on the new Chat tab's unread badge.**

The new mobile Chat tab's unread indicator uses raw English strings:

```tsx
aria-label={
  chats.wasMentioned.value
    ? "Unread mention"
    : `Unread messages: ${unreadChatIndicator.value}`
}
```

These bypass i18n, so screen-reader users on non-English locales get English. Note the *same block* localizes the typing indicator via `t("someone-is-typing", …)` and the label via `t("chat", …)`, so it's inconsistent within itself. This does mirror the existing pre-existing pattern in `MobileMoreMenu` (which hardcodes the same two strings), so it's replicating existing debt rather than introducing a new pattern — but since a new string is being added here, it's a good spot to route it through `t()` (updating only `en.json` per the repo convention).

---

### 🔵 Nits / notes

**No. 3 — `CHAT_FIRST_VISIBLE` module-level signal.** A module-scoped `signal(true)` that is never mutated is a slightly unusual way to express "always visible." It's harmless (immutable, shared, always true), but a `computed(() => true)` or an inline always-true visibility would read more clearly and avoid a shared module-level mutable-looking object.

**No. 4 — `applyChatFirstDesktopTools` also runs on mobile.** The `tools` computed is shared across mobile and desktop, so the promoted Chat tool is inserted into `tools.value` even on mobile, where it's then filtered back out (`hiddenToolIds` includes `open-chat`, and the mobile bottom bar is hardcoded). I confirmed there's no duplicate Chat render on mobile — the `flatMap` over `tools.value` is in the desktop-only branch. So this is wasted work, not a bug; worth a one-line comment noting the mobile path discards it, or gating the call on `!isMobile`.

**No. 5 — Chat tab is always shown under chat-first, even with no chat provider/chats.** This is confirmed intentional for desktop by a dedicated test ("shows Chat even when there are no providers or chats"), and is reasonable given chat-first is an explicit integration opt-in. Just flagging that tapping Chat with no provider configured opens an empty panel — worth a quick sanity check that the empty state reads acceptably.

**No. 6 — Param name casing.** The *value* is matched case-insensitively (`?.toLowerCase() === "true"`), but the param *name* `chatFirst` is case-sensitive. This matches the spec in issue #1628 exactly, so no change needed — just something integrators should know (`chatfirst=true` won't work).

---

### ✅ Test quality

The task called for scrutiny that tests reflect real behavior and don't just always pass. These hold up well:

- **Real boot path:** the test helper writes `chatFirst` onto the URL via `history.replaceState` **before** `createSeedBibleState`, so the toolbar reads it through the same `navigation.initialUrl` snapshot production uses — not a stub.
- **Negative/edge cases are genuine:** `it.each(["1", "yes", "false", "TRUE ", ""])` asserts these are *ignored*. `"TRUE "` (trailing space) failing the `=== "true"` check is a real assertion that would break if someone loosened the parse with `.trim()` — the test would catch it. The paired `chatFirst=TRUE` test confirms the case-insensitive path.
- **Isolation is correct:** a module-level `afterEach` resets the URL to `/` after every test, so a `chatFirst=true` test can't leak into a later `renderToolbar()` (no-arg) test — I verified the desktop "hides Chat without chatFirst" test runs *after* chat-first tests and still correctly sees Chat absent.
- **Observable behavior, not internals:** assertions check rendered tabs, panel-open signals, active-tab highlighting, and resize transitions — matching the repo's testing guidance.

One minor soft spot: in "keeps Chat available and prominent…", the ordering assertion is guarded by `if (searchIndex >= 0)`, so it silently no-ops if Search isn't present. `chatIndex >= 0` is asserted hard, so the test still verifies Chat renders; consider asserting the presence of the comparison anchor too.

---

### Summary

Approvable after a decision on **No. 1** (is the default More-menu reorder intended?) and, ideally, localizing the aria-label in **No. 2**. Everything else is optional polish. Nice work on the resize and overlay edge-case coverage.

---

PR #1695, opened 2026-08-26
Added Bounds in new parse refernce to handle the cases

### LINKED ISSUE #1637: Free-text verse reference matching produces false-positive links and skips chapter/verse validation
## Summary

The free-text scripture scanner (`parseVerseReferences` in `packages/seed-bible/seed-bible/managers/BibleDataManager.tsx`) links ordinary words to Bible books whenever the word happens to *begin with* a book abbreviation. In a Bible study app the worst offenders are Bible-domain words: "Isaac 24" links to **Isaiah** 24, "Judah 4" links to **Jude** 4, and "Jerusalem 70" links to **Jeremiah** 70.

Separately, the scanner never checks that the chapter or verse it parsed actually exists, so "Genesis 999" and "Genesis 1:999" both produce live links.

Every finding below was verified by calling the real functions, not by reading the code.

## Affected surfaces

Two places render auto-linked references from arbitrary text, and both go through the buggy path:

1. **Chat messages** — `ChatsManager.tsx:622` (`parseTextMessage`) scans message text, and `ChatView.tsx:188` renders each hit as a clickable `VerseReferenceLink`.
2. **Footnote bodies in the reader** — `BibleReader.tsx:2068` renders footnote text through `VerseReferenceText` (`app/verseReferenceLink.tsx:94`). This is the higher-impact surface, because footnote prose is dense with exactly the names that misfire — Isaac, Judah, Jerusalem, Joseph, Hebron — and often follows them with a number.

### Not affected

The scripture search box (`ScriptureItemInput`) is **not** affected and needs no fix. It uses a separate matcher, `matchBooksByPrefix` in `components/ScriptureItemInput/scriptureSuggestions.ts`, which only prefix-matches against real book names in the loaded translation and never calls `getBookId`. Typing "General 4" there correctly returns nothing, and it already validates chapter bounds via `bookHasChapter`.

### Root cause

`getBookId` (`BibleDataManager.tsx:629-651`) falls back to a loose prefix scan when the exact lookup misses:

```ts
if (!hadSpaces || /^\d/.test(bookLower)) {
  for (const [key, mappedId] of BOOK_ID_MAP) {
    if (bookLower.startsWith(key)) {
      return mappedId;
    }
  }
}
```

The test is backwards for abbreviation matching. It asks "does the typed word *start with* a known key?", so any word of three or more characters beginning with a three-letter book code matches. The intended case is the opposite: the typed text should be an abbreviation *of* a book name.

This compounds with the scanner's regex (`BibleDataManager.tsx:357-358`), whose book-name group captures a single word in nearly all cases. Single-word input means `hadSpaces` is false, so the loose branch is active for essentially every candidate the scanner considers.

Note that the doc comment on `resolveBookId` (`BibleDataManager.tsx:244-249`) currently asserts this cannot happen — it says prefix abbreviations are "intentionally not used here" because "short ordinary words like 'Is' / 'So' would otherwise become false-positive Isaiah / Song of Solomon links", and describes `getBookId` as "already conservative about prefixes". That holds for multi-word input only. The comment should be corrected alongside the fix.

### Confirmed false positives

Actual `getBookId` return values:

| Typed text | Links to | Category |
| --- | --- | --- |
| `Isaac` | ISA (Isaiah) | Bible-domain |
| `Judah` | JUD (Jude) | Bible-domain |
| `Jerusalem` | JER (Jeremiah) | Bible-domain |
| `Galilee` | GAL (Galatians) | Bible-domain |
| `Joseph` | JOS (Joshua) | Bible-domain |
| `Hebron` | HEB (Hebrews) | Bible-domain |
| `Jonathan` | JON (Jonah) | Bible-domain |
| `Judgment` | JUD (Jude) | Bible-domain |
| `Danielle` | DAN (Daniel) | everyday |
| `General` | GEN (Genesis) | everyday |
| `Revision` | REV (Revelation) | everyday |
| `Philosophy` | PHP (Philippians) | everyday |
| `Actually` | ACT (Acts) | everyday |
| `Number` | NUM (Numbers) | everyday |
| `Level` | LEV (Leviticus) | everyday |
| `Titles` | TIT (Titus) | everyday |
| `Jobs` | JOB (Job) | everyday |

### Reproduction

**In the app:** post any of the following as a chat message, or open a footnote containing one. The bolded span renders as a clickable verse link.

- "Isaac 24 tells us about his marriage to Rebekah." → links **Isaac 24** to Isaiah 24
- "The tribe of Judah 4 was the largest." → links **Judah 4** to Jude 4
- "Jerusalem 70 AD was when the temple fell." → links **Jerusalem 70** to Jeremiah 70
- "Galilee 3 is where he called the disciples." → links **Galilee 3** to Galatians 3
- "There were actually 5 people in the room." → links **actually 5** to Acts 5
- "See Revision 3:15 of the design doc." → links **Revision 3:15** to Revelation 3:15
- "The military rank of General 4 is held by four-star commanders." → links **General 4** to Genesis 4

**As a unit check** — each of these currently returns a match and should not:

```ts
import { parseVerseReferences } from "…/managers/BibleDataManager";

parseVerseReferences("Isaac 24 tells us about his marriage.", books);
// → [{ ref: { book: "ISA", chapter: 24 }, start: 0, end: 8 }]
```

Confirmed to reproduce both with and without a `books` list. A translation's book list is consulted for exact matches first, but "Isaac" matches no real book name, so it still falls through to `getBookId`.

## Bug 2 — No chapter or verse bounds validation

The scanner emits a reference as soon as the book name resolves, without checking the numbers against the book. Verified outputs:

| Input | Produced reference | Reality |
| --- | --- | --- |
| `Genesis 999` | `{ book: "GEN", chapter: 999 }` | Genesis has 50 chapters |
| `Genesis 1:999` | `{ book: "GEN", chapter: 1, verse: 999 }` | Genesis 1 has 31 verses |
| `Jude 5:1` | `{ book: "JUD", chapter: 5, verse: 1 }` | Jude has one chapter |
| `Genesis 0` | `{ book: "GEN", chapter: 0 }` | there is no chapter 0 |
| `Revelation 22:1-99` | `{ chapter: 22, verse: 1, endVerse: 99 }` | overruns the chapter |

`bookHasChapter` (`managers/parseVerseReference.ts:11-18`) already exists and is used by the search box, but the free-text scanner never calls it.

Adding a bounds check would also independently kill several Bug 1 false positives — "Jerusalem 70" dies because Jeremiah has 52 chapters, and "Joseph 37" dies because Joshua has 24. That makes the two fixes complementary, and bounds checking is the cheaper, lower-risk half.

The downstream click behavior for an out-of-range reference was not traced; at minimum it navigates to a chapter that does not exist.

## Bug 3 — Three misspelled keys in `BOOK_ID_MAP`, currently masked by Bug 1

This is why Bug 1 is not a one-line fix. Three canonical book names are **missing** from `BOOK_ID_MAP` because their entries are misspelled, and they resolve today *only* by falling through the buggy loose prefix:

| Line | Key present | Should be |
| --- | --- | --- |
| `BibleDataManager.tsx:441` | `laviticus` | `leviticus` |
| `BibleDataManager.tsx:473` | `ester` | `esther` |
| `BibleDataManager.tsx:518` | `zepaniah` | `zephaniah` |

I audited every entry in `BOOK_SLUGS` against `BOOK_ID_MAP`: these three are the only canonical names with no exact entry. `getBookId("Leviticus")`, `getBookId("Esther")`, and `getBookId("Zephaniah")` all work right now purely by accident.

**Any fix that tightens the loose fallback must add the correct spellings first**, or typing a full, correctly spelled book name will stop working. The misspellings can stay as extra aliases for typo tolerance — we just should not depend on them.

Minor, same area: `["lev", "LEV"]` is duplicated on `BibleDataManager.tsx:439-440`.

## Bug 4 — Two same-named parsers, one of them unused

`parseVerseReference` / `parseVerseReferences` exist as **two unrelated implementations with identical names** in two files, which makes this area easy to misdiagnose:

- `managers/BibleDataManager.tsx:347` — free-text scanner returning matches with character offsets. This is what chat and footnotes use, and it has the bugs above.
- `managers/parseVerseReference.ts:89` — anchored single-reference parser with ambiguous-prefix expansion and single-chapter verse shorthand. **It has no production caller.** Its only consumer, `scriptureSuggestions.ts`, imports just the `bookHasChapter` and `buildTail` helpers, not the matching logic. It has a full test suite, so it looks live.

Suggested: rename the two to distinguish intent (for example `scanVerseReferencesInText` versus `parseSingleVerseReference`), then decide whether the unused matcher should be adopted by the free-text path or deleted. Its ambiguous-prefix handling and chapter-bounds narrowing are arguably what the free-text scanner should be using.

## Proposed fix

**Reverse the direction of the loose prefix test** in `getBookId` so the typed text must be a prefix of a known key, rather than the key being a prefix of the typed text:

```ts
if (key.startsWith(bookLower)) return mappedId;   // instead of bookLower.startsWith(key)
```

I probed this against every confirmed case. It clears **15 of 17** false positives (Isaac, Judah, Jerusalem, Galilee, Joseph, Hebron, Jonathan, Judgment, Danielle, General, Revision, Philosophy, Actually, Level, Titles, Jobs all return `null`) while preserving legitimate abbreviations: `Genes`→GEN, `1 Chron`→1CH, `1Chron`→1CH, `Gen`→GEN, `Rev`→REV, `Phlm`→PHM, `1Kgs`→1KI, `Ezek`→EZK, `Song of Solomon`→SNG, `song-of-solomon`→SNG.

Three caveats surfaced while probing, all needing handling:

1. **`Number 5` still matches Numbers 5**, because "numbers" genuinely starts with "number". Prefix logic alone cannot fix this; it needs a short-word stoplist or a minimum-length rule. Arguably acceptable to leave as-is.
2. **`Levit` regresses to `null`**, because of the `laviticus` typo in Bug 3. Fixing the spellings resolves it.
3. **`Gen.` (trailing period) regresses to `null`**, because `getBookId` strips whitespace and hyphens but not punctuation. The free-text scanner's regex consumes the period as a separator so it is unaffected, but a direct caller passing "Gen." would break. Strip trailing punctuation in the normalizer.

Recommended order, lowest risk first:

1. Fix the three misspellings (Bug 3).
2. Add bounds validation to the free-text scanner (Bug 2).
3. Reverse the prefix direction and strip punctuation (Bug 1).
4. Rename the colliding functions (Bug 4).

## Suggested test coverage

Add to `test/unit/seed-bible/managers/BibleDataManager.test.ts`:

- Each Bible-domain false positive from the Bug 1 table returns no match when embedded in a full sentence.
- Every legitimate abbreviation in the preserved list above still resolves.
- Every canonical name in `BOOK_SLUGS` resolves via the **exact** map rather than the fallback — this locks Bug 3 shut permanently.
- Each out-of-range chapter/verse input from the Bug 2 table produces no match.

Kabi4:
Closes #1637


Blitzy:
> [!NOTE]

## Review of PR No. 1695 — bounds + prefix fix for the free-text verse scanner

Thanks for tackling issue No. 1637. The core of this PR is solid: the prefix direction is reversed correctly, the three misspelled `BOOK_ID_MAP` keys are fixed (with the typos kept as tolerant aliases), the duplicate `["lev","LEV"]` is removed, trailing punctuation is stripped, the colliding function names are disambiguated (`scanVerseReferencesInText` / `parseSingleVerseReference` / `parseVerseReferenceCandidates`), and the false-positive rejection is well covered by real tests. Type check is clean and the full suite passes locally.

The problem is with the **other half** of the fix — the new bounds validation. It rejects a set of *valid* references because the static verse-count data it relies on is wrong, and because it applies that Protestant-canon data even when a different translation is loaded. Details below, most severe first.

---

### 🔴 High — Valid verse references are silently dropped because the static verse-count table is wrong

`bookChapterVerseCounts.ts` (copied from `seed-bible-utils/infrastructure/data/booksStaticInfo.tsx`) undercounts the verses in many Psalms. The data is faithful to its source, so this is not a transcription mistake — but the source itself is wrong, and this PR is the first thing to make those errors *user-visible*, because `isVerseReferenceInBounds` now rejects any verse number above the (too-low) count.

Three cases are objectively wrong in every versification — none of these Psalms has a superscription, so there is no "counting the title as verse 1" debate:

| Reference | Data says max verse | Reality |
| --- | --- | --- |
| `Psalm 117:2` | 1 | Psalm 117 has 2 verses (it is the shortest chapter in the Bible) |
| `Psalm 119:176` | 175 | Psalm 119 has 176 verses (22 stanzas × 8) |
| `Psalm 150:6` | 5 | Psalm 150 has 6 verses ("Let everything that has breath praise the LORD") |

Beyond these, roughly 40 more Psalms are off by one in the same direction — consistent with a source that does not count Hebrew superscriptions as separate verses. I won't assert the exact list here (the superscription cases are genuinely versification-dependent), but the three above are unambiguous and are enough to show the data is unreliable as a hard gate.

**Effect (before → after this PR):** Before, typing or footnoting `Psalm 150:6` produced a clickable link. After this PR, it produces **no link at all** — the reference is treated as out of range and thrown away. This hits both surfaces the issue cares about (chat messages and reader footnotes), and it fails silently: there is no error, the link just disappears.

**Reproduction:**
```ts
import { scanVerseReferencesInText } from ".../managers/BibleDataManager";

scanVerseReferencesInText("Psalm 150:6");   // → []  (should link PSA 150:6)
scanVerseReferencesInText("Psalm 119:176"); // → []  (should link PSA 119:176)
scanVerseReferencesInText("Psalm 117:2");   // → []  (should link PSA 117:2)
```
In the app: send "Read Psalm 150:6" as a chat message, or open a footnote citing Psalm 150:6 — the reference renders as plain text instead of a link.

**Suggested fix:** Correct the verse counts. Since the data is shared with `seed-bible-utils/booksStaticInfo.tsx`, fixing it at the source and having this module continue to mirror it is cleanest (and fixes any other consumer of that table too). At minimum Psalms 117, 119, and 150 must be corrected; the ~40 off-by-one Psalms deserve a full audit against a reference versification before this ships. A regression test asserting that a set of known last-verses (`Psalm 150:6`, `Psalm 119:176`, `Genesis 50:26`, `Revelation 22:21`, …) still resolve would lock this shut — see the test-coverage note below.

---

### 🟠 Medium — Verse bounds always use Protestant-canon static data, even when a non-Protestant translation is loaded

`isVerseReferenceInBounds` uses the current translation's chapter range for the *chapter* check, but for the *verse* check it always falls back to the static table (`getChapterVerseCount`), because `TranslationBook` carries no per-chapter verse counts. That means the Protestant versification is imposed on every translation.

For any translation whose chapters legitimately contain more verses than the Protestant canon — Catholic/Orthodox editions with the Additions to Daniel and Esther, or simply a translation with different verse divisions — valid references are rejected.

**Reproduction** (Daniel 3 in a translation that includes the Song of the Three Young Men has ~100 verses; the static table has 30):
```ts
const danBook = [{ id: "DAN", numberOfChapters: 12, firstChapterNumber: 1, /* … */ }];
scanVerseReferencesInText("Daniel 3:57", danBook); // → []  (valid in that translation)
```

**Suggested fix:** When a translation is loaded but has no per-chapter verse data, consider gating only on the chapter (which we *can* validate from `numberOfChapters`) and letting the verse through, rather than rejecting it on Protestant assumptions. That keeps Bug 2's intent (kill `Genesis 1:999`, `Genesis 999`) for the common case while not silently breaking alternative canons. This is a smaller concern than the High above and can reasonably be a follow-up, but it should be a conscious decision rather than an accident of the fallback.

---

### 🟡 Low — Test coverage confirms rejection but never confirms preservation

The new bounds tests (`rejects chapter and verse numbers outside the book`, and the chat equivalent) only assert that clearly-invalid references (`Genesis 999`, `Genesis 1:999`, `Jude 5:1`) are dropped. Nothing asserts that a *valid* boundary reference still links after bounds were added. That is exactly the gap that let the High-severity regression above through — a single assertion like `expect(scanVerseReferencesInText("Psalm 150:6")).toHaveLength(1)` would have caught it and gone red.

Separately, the two new modules — `verseReferenceBounds.ts` (`isVerseReferenceInBounds`, with its translation-vs-static and range branches) and `bookChapterVerseCounts.ts` — have no direct unit tests; they are exercised only indirectly through the scanner. Given how much branching logic lives in `isVerseReferenceInBounds` (start/end chapter, start/end verse, unknown-book fallthrough), a focused test file would be worthwhile.

**Suggestion:** Add "valid boundary reference still resolves" cases for a handful of books, and a small direct test suite for `isVerseReferenceInBounds` covering: in-bounds, chapter over, verse over, cross-chapter range, unknown book (accepts ≥ 1), and translation-vs-static precedence.

---

### 🟢 Low / by-design — worth a conscious call, not necessarily a change

- **A partly-valid range is rejected whole.** `Revelation 22:1-99` produces no link at all, even though `Revelation 22:1` is valid — the out-of-range end kills the entire match. The issue listed this input as a bug, so rejecting it is defensible; just flagging that the valid prefix is lost too, rather than clamped.
- **`Number`/`Numbers` residual.** `getBookId("Number")` still resolves to `NUM` (because "numbers" genuinely starts with "number"), so `Number 5` still links to Numbers 5. Issue No. 1637 already called this an acceptable known limitation — noting it here only for completeness, since it is the one Bug 1 case the prefix reversal does not clear.
- **Dead alias.** `export const parseVerseReferences = scanVerseReferencesInText` is marked `@deprecated`, and every in-repo caller was migrated in this PR, so the alias has no remaining users. Fine to keep for external compatibility, but it could simply be dropped.

---

### Summary

The prefix reversal, misspelling fixes, punctuation handling, and function renames are good and well-tested, and they resolve Bug 1, Bug 3, and Bug 4 from the issue. The blocker is the bounds half (Bug 2): the static verse data it depends on is wrong for a well-known set of references, and it is applied across all translations, so this PR trades a set of false-positive links for a set of false-negative ones. Correcting the verse-count data (at least Psalms 117/119/150, ideally a full audit) plus a "valid boundary still links" regression test would clear the High finding; the translation-versification question (Medium) can be a follow-up if scoped explicitly.

---

Kabi4:
Psalm verse counts were wrong, so real refs like Psalm 150:6 stopped linking — those counts are fixed. When a translation is loaded, we no longer reject verses just because they don’t match Protestant numbers. Tests now check that valid last-verses still become links.

KallynGowdy (approved):

PR #1696, opened 2026-08-26
Removed AI generated translations.

soukigabriel:


Blitzy:
> [!NOTE]

## Code Review — Removed AI generated translations (#1696)

### Summary

This PR deletes the 76 machine-translated locale entries (`af` through `zu`) from `packages/bible-stack/extension.json`, leaving only the `en` entry. It's a data-only change: one file, 380 deletions, no code or tests touched. It directly reverts the translation variants that were added in commit `99fde58` ("Added description for bible-stack extension.json file and translation variants."), which lines up with the repo policy in `CLAUDE.md` — *"only update en.json … leave those for the professionals."*

**Verdict: safe to merge as-is.** I traced every consumer of these entries and none breaks. The findings below are caveats and follow-up questions, not blockers — the most important one (No. 1) is that this deletion is **not durable** and will silently come back the next time the translation tooling runs.

I verified the following, so reviewers don't have to re-check:

- **Schema stays valid.** `ExtensionMetaSchema` in `script/lib/extension.ts:318` only requires `translations.en` (`z.object({ en: ... }).catchall(...)`). An `en`-only file passes, so `casualos pack-aux` / upload still work.
- **Runtime degrades gracefully.** `fallbackLng` is `"en"` (`I18nManager.tsx:233`, `DEFAULT_UI_LANGUAGE` in `ReadingUrlPath.ts:14`), and the one call site — `bible-stack/bootstrap.tsx:43`/`:68` — passes `defaultValue: "Bible Stack"`. A French user now sees the English title/description instead of a localized one; nothing crashes. `ExtensionManager.tsx:954` even documents that missing translations must not take the Settings page down.
- **Lint stays clean.** The `translation-extension-incomplete-translations` rule (`eslint.config.mts:201`) only checks locales that are *present* against `en`; it does not require every core language to exist. With no non-English locales left, it emits zero warnings.
- **No tests reference these strings** — a grep across `test/` and `script/` for the bible-stack titles/descriptions found nothing, so no snapshots or assertions go stale.

---

### Findings by severity

#### Medium — the deletion is not durable; `pnpm i18n` will regenerate all of it

`translateExtensionKeys` in `script/i18n.ts:348-352` builds each extension's target language set as the **union of existing locales and every core app language**:

```ts
const supportedLanguages = [
  ...new Set([...existingLanguages, ...coreLanguages]),
]
```

Because bible-stack now has no non-English locales, every core language counts as "missing," so the next run of `pnpm i18n` (either the "missing" or "all" mode) re-adds and re-translates all ~76 locales and writes them back to `extension.json` (`script/i18n.ts:386-403`). The machine translations this PR removes would quietly reappear.

Mitigating factor: `pnpm i18n` is a **manual** command backed by the Google Translate API (`package.json:28`, `@google-cloud/translate`), and no CI workflow invokes it (grep of `.github/workflows/` found none). So it won't be undone automatically on merge — but it will be undone the first time any maintainer regenerates translations for unrelated work.

1. Check out this branch; confirm `packages/bible-stack/extension.json` has only `en`.
2. Run `pnpm i18n` (with translation credentials configured).
3. Observe that `bible-stack/extension.json` is rewritten with all core-language locales again.

**Suggested fix (pick one, and probably out of scope for this PR — worth a follow-up issue):**
- Add an opt-out mechanism so specific extensions are excluded from AI translation — e.g. a `"skipAutoTranslate": true` flag on the extension manifest that `translateExtensionKeys` honors when computing `supportedLanguages`, or a hardcoded skip-list in `script/i18n.ts`.
- Or, if the long-term intent is to stop AI-translating **all** extensions, retire/guard the extension branch of the i18n tooling entirely.

Without one of these, this PR is a one-time cleanup that the tooling is designed to reverse.

#### Low — scope: only bible-stack is cleaned, 12 other extensions still ship AI translations

Every other extension manifest still carries 77–78 machine-translated locales (`apologist`, `audio-reader`, `bonfire`, `compare`, `locations`, `scripture-map`, `transcript`, `transcript-ui`, `twitchPub`, `twitchSub`, `seed-bible-utils`, and the refresh example). The PR title ("Removed AI generated translations") reads as a general cleanup, but only bible-stack is touched.

This is likely intentional — bible-stack's variants were only just added in `99fde58`, so this is a targeted revert rather than a project-wide sweep — but it's worth stating the intent explicitly. **Is the plan to clean the other extensions too (in which case a tracking issue would help), or is bible-stack a deliberate one-off** because its professional translations are coming and the others' are staying?

#### Low / Info — no PR description or linked issue

There's no description on the PR and no linked issue, which makes the intent hard to verify (the reasoning above is reconstructed from the diff and commit history). A one-line description — "removing the just-added machine translations for bible-stack so professional translations can replace them" — plus a link to any driving issue would make the change self-explanatory and help whoever revisits this after the tooling regenerates the entries (see No. 1).

---

### On tests

There are no tests in this PR, and for a pure data deletion that is reasonable — there's no new behavior to assert, and the runtime fallback path is already covered by existing i18n handling. The one thing genuinely worth a test is a **guard against regeneration** (No. 1): if you add an opt-out flag, a small unit test on `translateExtensionKeys` asserting that a flagged/`en`-only extension is *not* expanded to the core languages would be a real-world regression test — and it would fail against the current tooling, which is exactly the property we want from a regression test.

---

### Bottom line

Mergeable and low-risk on its own. Before considering the underlying goal "done," decide how to stop `pnpm i18n` from re-adding these entries (No. 1); if that's a separate effort, a tracking issue and a note on this PR would close the loop.

---

Blitzy (approved):

PR #1701, opened 2026-08-27
appName from branding

AmitPant9999:


Blitzy:
> [!NOTE]

## Code Review — appName from branding (PR #1701)

Thanks for tackling this! The goal — letting a deployment override the "Seed Bible" name via `branding.appName` — is a good one, and centralizing it in a single `getBrandedAppText` helper is the right instinct. The core issue is the chosen mechanism: doing a `String.replace(/Seed Bible/gi, appName)` on **already-translated** text. That works in English but breaks down across the localized strings, and it changes the default (unbranded) app in every non-English locale. Details below, by severity.

---

### 🔴 High

#### No. 1 — Custom branding silently fails (or applies inconsistently) in most non-English locales

`getBrandedAppText` replaces the literal string `"Seed Bible"` in the *resolved* translation. But the translations only contain that literal in some locales — many localize the product name — so the replacement misses them. Worse, within a single locale, some strings contain the English literal and adjacent ones don't, so branding gets applied to one line and not the next.

Concrete, from the current locale files:

| Locale | `tutorial.promptTitle` | `tutorial.promptBody` |
|---|---|---|
| `zh` | `欢迎来到种子圣经…` (no "Seed Bible" → **not** rebranded) | `…了解 Seed Bible 的…` (has literal → **rebranded**) |
| `ja` | `シードバイブルへようこそ…` (**not** rebranded) | — |

Same story in Arabic (`سيد بايبل`) and Hebrew (`תנ"ך הזרעים`) onboarding strings — no `"Seed Bible"` substring, so the custom brand never appears.

**Reproduction:**
1. Set `branding.appName = "My Bible"` in the app config.
2. Switch UI language to Chinese (`zh`).
3. Open the tutorial prompt.
4. Expected: title and body both read "My Bible". Actual: the body reads "My Bible" but the title still reads "种子圣经" — inconsistent branding in the same dialog.

**Suggested fix:** Brand via an interpolation placeholder instead of a post-hoc string replace. Make the translation strings carry `{{appName}}` (e.g. `en.json`: `"onboarding.installBodyPre": "Add {{appName}} to your "`) and pass `appName` to `t(...)`. This is locale-independent and was, in fact, the direction the *old* code was already reaching for — it passed `appName:` as an interpolation value (the strings just never had the matching `{{appName}}` placeholder, so that param was dead). The trade-off is that the placeholder has to land in every locale file, which the "leave other locales to the professionals" rule (CLAUDE.md) makes a coordination task rather than a one-file change. If touching locale files now is off the table, replacing against the *localized* name `t("seed-bible")` rather than the English literal would at least cover locales that reuse their own localized name (zh's `种子圣经`, etc.) — but only the placeholder approach is fully correct.

#### No. 2 — The default (no-branding) app changes appearance in every localized build

`getAppName` falls back to `t("seed-bible")` — the *localized* value — rather than the literal `"Seed Bible"`:

```ts
function getAppName(t, branding?) {
  return branding?.appName ?? t("seed-bible", { defaultValue: "Seed Bible" });
}
```

So even with **no** branding configured, the running-text brand name gets overwritten with the localized `seed-bible` value. Translators deliberately kept the English proper noun "Seed Bible" in some sentences while translating the standalone `seed-bible` key differently — this erases that choice.

**Reproduction:**
1. Do **not** configure any branding.
2. Set UI language to Spanish (`es`).
3. Open onboarding.
4. Expected (pre-PR): `"Agrega Seed Bible a tu …"`. Actual (post-PR): `"Agrega Biblia de semillas a tu …"` — the stock app's wording changed for a feature nobody opted into.

**Suggested fix:** Fall back to the literal string, `branding?.appName ?? "Seed Bible"`, so that with no branding the replace is a no-op in every locale and only an explicit `appName` changes anything.

---

### 🟡 Medium

#### No. 3 — No tests for the new helper, despite an existing `I18nManager.test.ts`

`getBrandedAppText`/`getAppName` are new exported functions with non-trivial behavior (regex replacement, fallback resolution, case-insensitivity) and six call sites, but the PR adds no tests. `test/unit/seed-bible/i18n/I18nManager.test.ts` already exists as a home for them. Per the repo's test guidance, these should assert real behavior and cover the edge cases that are actually where this breaks — a locale whose string lacks the literal, a configured `appName`, and the no-branding default. Written honestly, a test for the branded-Chinese-title case (No. 1) would fail against this implementation, which is exactly the point of writing it.

**Suggested tests:** given `branding.appName = "X"`, a string containing "Seed Bible" is rebranded; a localized string *without* the literal is **not** (documents the gap); with no branding, English output is unchanged; and an `appName` containing `$` is handled (see No. 4).

#### No. 4 — `appName` is used as a replacement string, so `$` sequences corrupt output

`text.replace(/Seed Bible/gi, appName)` treats `appName` as a replacement pattern. Special sequences (`$&`, `$'`, `` $` ``, `$1`, `$$`) are interpreted rather than inserted literally.

**Reproduction:** `branding.appName = "A$&B"` turns `"Add Seed Bible to your "` into `"Add ASeed BibleB to your "` instead of `"Add A$&B to your "`.

**Suggested fix:** use a function replacer, which never interprets `$`: `text.replace(/Seed Bible/gi, () => appName)`.

---

### 🟢 Low / Nits

#### No. 5 — Wrong translation key `"seed bible"` (space) in `SeedBibleStateManager.tsx`
Both new call sites pass `"seed bible"` where the actual key is `"seed-bible"` (hyphen). `t("seed bible", …)` never matches, so it always returns the default `"Seed Bible"` and only recovers the localized name by accident through `getAppName`'s replace. It happens to work today but is a dead lookup — use `"seed-bible"`.

#### No. 6 — Over-broad rebranding of extension title/description (`SettingsPage.tsx`)
Wrapping arbitrary extension `title`/`description` in `getBrandedAppText` means any (potentially third-party) extension whose text contains "Seed Bible" gets silently rewritten. Fine for first-party strings; worth confirming that's the intent for all installed extensions.

#### No. 7 — Style nit
The new `TranslationFn` type and `getAppName`/`getBrandedAppText` block lost the surrounding blank lines (the old trailing blank line before the comment block was removed). Minor, but `pnpm format` / a blank line above the block keeps it consistent.

---

### Summary

The single-helper approach is sound, but replacing on post-translation text is the wrong layer: it under-applies for custom branding in localized UIs (No. 1) and over-applies for the default build (No. 2). Moving the substitution to a `{{appName}}` interpolation placeholder resolves both at once; failing that, at minimum fall back to the literal `"Seed Bible"` (No. 2), fix the replacer for `$` (No. 4) and the `"seed bible"` key (No. 5), and add tests that exercise a locale without the English literal (No. 3).

---

AmitPant9999:
@KallynGowdy could you please review it

PR #1717, opened 2026-08-27
Add a first pass on including default content

KallynGowdy:
Currently, the Seed Bible has no content sources at all. This PR, when merged, adds some (stale) Bible Project content as well as the ability for us to include content that we've sourced from other sources (like YouTube).


KallynGowdy:
image

Blitzy:
NOTE: This is an automated review posted by Ryan Cook's scheduled Claude Code routine, triggered by this PR's `pull_request.ready_for_review` event. It was produced as a single-pass review (the Agent tool for multi-agent fan-out/subagent verification wasn't available in this run), so treat it as one careful pass rather than the full multi-agent review.

## Code Review — PR #1717 (`feature/default-discover-content` → `develop`)

### 1. Leftover debug logging floods the console (and error trackers) on every chapter view — High
**File:** `packages/default-content-extension/ext_DefaultContent/discoveredContent.ts`, lines 197, 209, 212

```ts
if (!containsCurrentChapter) {
  console.warn("No match", { context, item });
  continue;
}
console.error("Match", { context, item });
```

`bibleProject.json` has 279 entries, and `findBibleProjectContentForChapter` runs this loop on every `discover()` call (i.e. every chapter the reader opens). For a typical chapter, ~275+ entries don't match, so this fires ~275 `console.warn` calls (each serializing a full nested `item` object) plus one `console.error("Match", ...)` per real hit — on every single page view. Two problems: (1) the console-spam/serialization cost on a hot path, and (2) `console.error` for a successful match is mislabeled — any error-monitoring tool that captures `console.error` as breadcrumbs/events (Sentry and similar commonly do) will record a false "error" on every normal chapter navigation. This reads like debug logging left in from development; it should be removed (or dropped to a single dev-only trace at most).

### 2. Test doesn't exercise real behavior — it tests markup that no longer exists — High
**File:** `test/unit/default-content-extension/init.test.tsx`, lines 149–172 (`"does not open the modal when the card's 'Watch' link is clicked"`)

The test renders `results[0].content` and looks for a `.sb-default-content-link` element, then asserts clicking it doesn't bubble to a card `onClick`. But the real `discover()` implementation in `init.tsx` never sets a `content` field on the returned `DiscoverContentResult` for either provider — the component that used to render that link (`DiscoveredContentBody`, with the `.sb-default-content-link` anchor) is commented out, and nothing in the codebase defines `.sb-default-content-link` anymore (confirmed via grep — it only appears in this test and in the commented-out block). So `results[0].content` is always `undefined`, `container.querySelector(".sb-default-content-link")` is always `null`, and `link?.dispatchEvent(...)` is a no-op — meaning `expect(cardOnClick).not.toHaveBeenCalled()` passes unconditionally, regardless of whether any stopPropagation logic exists. This is a tautological/dead test: reverting any real fix here wouldn't turn it red, because there's nothing left to click. It should either be removed or rewritten against the actual current behavior (the whole card's `onClick` opening the video modal).

### 3. Stale comment claims description is left empty, but the code sets it to real text — Medium
**File:** `packages/default-content-extension/ext_DefaultContent/init.tsx`, lines 68–70 and 102–104

```ts
// The description renders inside `content` via ExpandableText
// instead, so the shared wrapper's plain (un-truncated) span
// stays empty.
description: item.description,
```

This comment (duplicated for both discover providers) says the description is left empty because it's rendered elsewhere via `content` — but the very next line sets `description: item.description` to the real text, and there's no `content` field being returned at all (see finding 2). The description *does* render through the generic path (`ContentResultsList`'s `<ExpandableText>` in `DiscoveredResultsSections.tsx`), so the current behavior is correct, but the comment is now false and looks like a leftover from an earlier draft (before `DiscoveredContentBody`/`content` was removed). Left as-is, it risks misleading a future maintainer into "fixing" this to match the comment (setting `description` to `""`), which would actually break the UI (no description would render anywhere).

### 4. Dead, commented-out component left in the file — Low (cleanup)
**File:** `packages/default-content-extension/ext_DefaultContent/init.tsx`, lines 9–53 (roughly)

A ~45-line commented-out `DiscoveredContentBody` component (props interface + implementation) is left in the file, referencing CSS classes (`sb-default-content-item`, `sb-default-content-author`, `sb-default-content-link`) that exist nowhere else in the codebase. It's superseded by the shared `ContentResultsList`/`ExpandableText` rendering path added in `DiscoveredResultsSections.tsx`. Recommend deleting it outright rather than leaving it commented out (it's also the reason for the dead test in finding 2).

### 5. Non-null assertion can silently produce `imageUrl: undefined` despite the field's declared type — Low
**File:** `script/lib/discoveredContentList.ts`, line 292 (interface at line 26)

```ts
imageUrl: linkPreview.meta["og:image"] ?? linkPreview.imageUrl!,
```

`DiscoveredContentItem.imageUrl` is declared as a required `string` (line 26), but the `!` here just forces past whatever the actual (presumably optional) type of `linkPreview.imageUrl` is. If a URL's link preview has neither `meta["og:image"]` nor `imageUrl`, this assigns `undefined` to a field typed as always present. It doesn't crash today (nothing in the current `discoveredContent.json` hits this — checked via a quick scan of all 50 entries), and `JSON.stringify` drops the `undefined` key on write, so the consumer's `imageUrl?: string` (correctly optional in `discoveredContent.ts`) tolerates it — but the generator script's own type is misleading and the `!` should be replaced with a real fallback (or the field should be marked optional here too) so a future row with no image doesn't rely on this coincidence.

---
*Findings 1 and 2 are the ones I'd block on before merge; 3–5 are worth a follow-up but not blocking.*


---

PR #1718, opened 2026-08-28
Improve how branding configs are handled

KallynGowdy:
Previously, the branding config was just directly injected into the Seed Bible code whenever it was needed.

Now, you can specify a `seed-bible.branding.json` file at the root of the repo that includes the branding config that should be used.

Here's an example:

```
{
  "appName": "My Seed Bible",
  "shortName": "Bible",
  "logo": "https://example.com/my-logo.png",
  "icon": "https://example.com/my-icon.png",
  "websiteUrl": "https://example.com",
  "disabledToolbarTools": ["open-discover"],
  "defaultTranslationId": "eng_kjv"
}
```

### DIFF
--- a/packages/seed-bible/seed-bible/app/appConfig.tsx
+++ b/packages/seed-bible/seed-bible/app/appConfig.tsx
@@ -19,6 +19,10 @@ export interface BrandingConfig {
   disabledToolbarTools?: string[];
   defaultTranslationId?: string;
 }
+
+// Injected from Vite
+declare const __BRANDING_CONFIG__: BrandingConfig | undefined;
+
 export interface AppConfig {
   /**
    * Path prefix this deployment is mounted under, e.g. "/d/branch-develop".
@@ -54,6 +58,7 @@ export const DEFAULT_APP_CONFIG: AppConfig = {
   renderedAsMobile: false,
   renderedAsWebKit: false,
   acceptedLanguages: [],
+  branding: import.meta.env.VITEST ? undefined : __BRANDING_CONFIG__,
 };
 
 /**
--- a/standalone/entry-ssr.tsx
+++ b/standalone/entry-ssr.tsx
@@ -19,6 +19,7 @@ import {
   stripBasePath,
 } from "@packages/seed-bible/seed-bible/managers/ReadingUrlPath";
 import { getPreferredSupportedLanguage } from "@packages/seed-bible/seed-bible/i18n/I18nManager";
+import { DEFAULT_APP_CONFIG } from "@packages/seed-bible/dist/app/appConfig";
 
 /** A single chunk record from a Vite client manifest. */
 interface ManifestChunk {
@@ -364,17 +365,20 @@ export async function render(
   | { redirectTo: string; redirectStatus?: number; vary?: string }
   | string
 > {
-  const { config } = options;
+  const { config: injectedConfig } = options;
 
-  const redirectTo = legacyReadingUrlRedirect(options.path, config.basePath);
+  const redirectTo = legacyReadingUrlRedirect(
+    options.path,
+    injectedConfig.basePath
+  );
   if (redirectTo) {
     return { redirectTo };
   }
 
   const languageRedirectTo = acceptLanguageRedirect(
     options.path,
-    config.basePath,
-    config.acceptedLanguages
+    injectedConfig.basePath,
+    injectedConfig.acceptedLanguages
   );
   if (languageRedirectTo) {
     return {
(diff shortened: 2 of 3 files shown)


Blitzy:
> [!NOTE]

## Code review — Improve how branding configs are handled

Reviewed at head `aa9dfed`. The overall approach is sound: read `seed-bible.branding.json` at build time, inject it via Vite `define` as `__BRANDING_CONFIG__`, expose it through `DEFAULT_APP_CONFIG.branding`, and merge it with the per-request config during SSR so the app receives branding without the host server having to thread it through on every request. I traced the flow and the merged `config` correctly reaches `createSeedBibleState` and `<Main config={config}>`, so branding does propagate once the blocker below is fixed.

There is one blocking bug that breaks the build, the dev server, and the test suite, plus a coverage gap and some minor cleanups.

---

### 🔴 Blocker

**1. `entry-ssr.tsx` imports `DEFAULT_APP_CONFIG` from a non-existent `dist/` path — breaks the SSR build, `pnpm dev`, and the whole `entry-ssr` test suite.**

`standalone/entry-ssr.tsx:22`

```ts
import { DEFAULT_APP_CONFIG } from "../packages/seed-bible/dist/app/appConfig";
```

This points at a build-output directory (`packages/seed-bible/dist/...`) that does not exist anywhere in the repo and is never produced by any build step (the client build outputs to `standalone/dist/client`, the SSR build to `standalone/dist/server`, and the server bundle to `server/dist`). Note that the `type AppConfig` import on line 3 of the same file already uses the correct source path — only this value import points at `dist`. My best guess is a stale local `dist/` folder made it resolve on the author's machine, so it slipped past the "fix: Build errors" commit.

Impact — all three of these fail on a clean checkout:
- `pnpm build:ssr` → `[UNRESOLVED_IMPORT] Could not resolve '../packages/seed-bible/dist/app/appConfig'` (production build / CD is broken).
- `pnpm dev` → `server/index.ts` loads this module via `vite.ssrLoadModule("/standalone/entry-ssr.tsx")`, which hits the same resolution failure.
- `pnpm vitest run test/unit/standalone/entry-ssr.test.ts` → the suite fails to load with `Failed to resolve import`, so **0 of its 71 tests run** (they show as passing-by-absence in aggregate, which is worth watching for).

Reproduction:
```bash
pnpm install
pnpm vitest run test/unit/standalone/entry-ssr.test.ts   # FAIL: Failed to resolve import ".../dist/app/appConfig"
pnpm build:ssr                                            # error during build: UNRESOLVED_IMPORT
```

Suggested fix — import the value from source, ideally folding it into the existing type import on line 3:

```ts
import {
  type AppConfig,
  DEFAULT_APP_CONFIG,
} from "../packages/seed-bible/seed-bible/app/appConfig";
```

Verified: after changing line 22 to the source path, `pnpm vitest run test/unit/standalone/entry-ssr.test.ts` passes **71/71** and the import resolves cleanly.

---

### 🟡 Medium

**2. No test coverage for the new branding behavior.**

This PR adds a new SSR merge path (branding from `DEFAULT_APP_CONFIG` combined with the injected config) and no test exercises it. Two things compound this:
- The `import.meta.env.VITEST ? undefined : __BRANDING_CONFIG__` guard (`appConfig.tsx:61`) forces `DEFAULT_APP_CONFIG.branding` to `undefined` under Vitest, so the default-config branding path is invisible to tests by construction.
- A test for this path would also have caught blocker No. 1 — the suite that *should* cover `render()` currently doesn't load at all.

`render()`'s `config` option accepts an optional `branding`, so this is testable without touching the build-time define: pass a branding config through the render options and assert it ends up in the merged config / serialized `<!--CONFIG_JSON-->` output, and assert that omitting it leaves branding `undefined` rather than throwing. Please make sure any such test genuinely fails against the pre-merge code (revert the merge locally, confirm red) so it's testing real behavior and not passing unconditionally.

---

### 🟢 Low / nitpicks

**3. Leftover debug `console.log` in `vite.config.ts`.**

`vite.config.ts:54-56`

```ts
if (brandingConfig) {
  console.log("[vite.config.ts] Using branding config:", brandingConfig);
}
```

From the "chore: Log the branding config during build" commit. It prints the full branding object on every build that has a `seed-bible.branding.json`. Harmless, but consider removing it or reducing it to a one-line confirmation to keep build logs clean.

**4. `seed-bible.branding.json` is undocumented and unlisted in `.gitignore`.**

The build silently reads this file from the repo root, but there's no example file, no README/docs entry, and it's not in `.gitignore`. Worth a short note in the docs plus a decision on whether the main repo should `.gitignore` it (so a white-label fork adds its own without the main deployment ever accidentally committing one). The PR description's example JSON would be a good starting point for that doc.

**5. Explain the `VITEST` guard.**

`appConfig.tsx:61` — the `import.meta.env.VITEST ? undefined : __BRANDING_CONFIG__` guard exists because Vite `define` with an `undefined` value (`JSON.stringify(undefined) === undefined`) leaves `__BRANDING_CONFIG__` unreplaced, which would otherwise throw a `ReferenceError` in the test environment. That "why" isn't obvious from the code; a one-line comment (per the repo's comment convention for non-obvious constraints) would help the next reader.

---

PR #1720, opened 2026-08-29
Added Gallary Module + Upload image in Playlist / reading plan

### LINKED ISSUE #1647: Add support for hero images in playlists and reading plans
Currently when you create a playlist or reading plan, the only way you can style it is by setting a title and description. Ideally, users would also be able to set a hero image (e.g. 1024x768) that is displayed when looking at a playlist or reading plan. This is inspired by YouVersion. For example:

Kabi4:
Fixes: #1647


Blitzy:
> [!NOTE]

## Code review — Gallery module + hero images for playlists & reading plans

Reviewed against issue [#1647](https://github.com/HelloAOLab/seed-bible/issues/1647) ("Add support for hero images in playlists and reading plans"). The issue asks for an optional hero image (≈1024×768) that reuses the **same upload path as user avatars** and is **shown whenever someone views a playlist or reading plan**.

**Overall:** solid, well-tested work. The upload path correctly matches the avatar path (`recordFile(userId, file, { marker: "publicRead" })`, same as `LoginManager`), covers persist through a schema-versioned optional field so legacy records still parse, and the new `UserGalleryManager` has genuinely good tests — error paths, the signed-out case, the 24-item cap, cache fallbacks, and string-coerced timestamps are all exercised, not just the happy path. Local test run: all 8 touched test files, 398 tests, pass.

The findings below are about polish and a couple of behavior regressions in the **list/detail views**, not the core upload feature. Nothing here is a blocking correctness bug.

---

### Medium

**No. 1 — Reading-plan list rows lost their status icons.**
In `ReadingPlansPane.tsx`, the status-specific tiles were each replaced by `<HeroImageThumb url={...heroImageUrl} />`:
- draft ("resume") rows used `edit_note` (was line ~570),
- available/active plans used `menu_book` (~648, ~774),
- completed plans used a `check` with `sb-rp-card-tile-done` styling (~691).

`HeroImageThumb` renders a generic "No image" placeholder (`hide_image` icon + "No image" label) whenever there's no cover. Since **every existing plan has no cover**, all of these rows now show the same "No image" box, so the at-a-glance distinction between draft / active / completed plans is gone.

*Reproduction:* open the Reading Plans pane with one completed plan and one active plan, neither with a cover image → both rows show an identical "No image" tile instead of the check / book icons.

*Suggested fix:* render `HeroImageThumb` only when `meta.heroImageUrl` is set, and otherwise keep the existing status icon; or extend `HeroImageThumb` to accept a fallback icon (`fallback="menu_book"` / `"check"` / `"edit_note"`) so each row keeps its semantic icon when there's no cover.

**No. 2 — An empty "No image" banner is shown to viewers of every cover-less playlist and plan.**
`HeroImageBanner` renders a placeholder box ("No image", `role="img"`) when `url` is null, and both `PlayPlaylistView` and `ReadingPlanDetail` render it **unconditionally**. The cover is optional (per the issue), so the large majority of existing playlists/plans will now display an empty "No image" banner to the reader. A placeholder makes sense inside the editor field (where it invites you to add one), but showing "No image" to someone just viewing a plan is likely unintended clutter.

*Reproduction:* play any playlist that has no cover → an empty "No image" banner appears above the queue; open any reading plan without a cover → same banner at the top of the detail view.

*Suggested fix:* in the read/play views, render the banner only when `url` is present (or have `HeroImageBanner` return `null` for the empty state and keep the placeholder only for the editor field).

---

### Low

**No. 3 — Dead code: `resolveCoverImageUpload` and `uploadCoverImageForUser`.**
Both are exported from `managers/uploadPublicFile.ts` and covered by `uploadPublicFile.test.ts`, but neither is called anywhere in production — the components use `uploadPhotoToGallery` plus the managers' `uploadHeroImage`. Their tests pass but protect code that never ships. This also adds to a fairly crowded set of overlapping upload helpers (`uploadPublicFile`, `uploadCoverImageForUser`, `resolveCoverImageUpload`, `savePhotoToGallery`, `uploadPhotoToGallery`, plus `savePhoto`/`uploadHeroImage` on the managers), which the repo's own guidance on duplication cautions against.

*Suggested fix:* delete the two unused helpers and their tests, or wire them into the actual upload path if they were meant to be the single entry point.

**No. 4 — The image-URL schema disagrees between the gallery and the playlist/plan.**
`GalleryPhotoSchema.url` deliberately uses a lenient `z.string().min(1).max(2048)` (with a comment noting "CasualOS file URLs vary"), but `PlaylistSchema.heroImageUrl` and `ReadingPlanMetadataSchema.heroImageUrl` use a strict `z.url().max(2048)`. So a URL the gallery happily stores and displays could throw when it's parsed into a playlist (e.g. in `clonePlaylist`, or when persisting). In practice CasualOS file URLs are absolute `https` URLs — the avatar path stores `result.url` raw in `LoginManager` without validation and works — so this probably won't bite today, but the two schemas contradict each other on what a valid image URL is.

*Suggested fix:* pick one rule. If avatar-style URLs are always strict URLs, make the gallery schema strict too (and drop the "varies" comment); otherwise relax the hero fields to match the gallery.

**No. 5 — A single cover upload writes the gallery record twice.**
In `HeroImageField.openCropModal`, after `onUpload(cropped)` returns the URL it calls `gallery?.rememberPhoto(url)`. But the `onUpload` handlers (`CreatePlaylistForm` / `ReadingPlanEditor`) call `uploadPhotoToGallery`, which — when a gallery is present — already records the photo via `gallery.savePhoto`. Net result per upload: 1 `recordFile` + **2** `recordData` writes (and two front-prepends of the same URL). It's harmless because the list dedupes by URL, but it's a redundant network write on every cover upload.

*Suggested fix:* have `HeroImageField` skip the extra `rememberPhoto` when the `onUpload` path already returns a URL it has recorded (or make `onUpload` not record, leaving `HeroImageField` as the single place that remembers).

**No. 6 — The crop/encode logic has no test.**
`HeroImageCropModal.tsx` (`canvasToHeroBlob` — the draw-to-1024×768 + JPEG-encode step that actually produces the stored cover) has no unit test; the component tests stub `HeroImageField` out entirely. Given how central the 4:3 crop is to this feature, a small test around `canvasToHeroBlob` (correct output dimensions, and the "no 2d context" / "toBlob returns null" rejection paths) would be worth adding.

**No. 7 — Gallery records grow unbounded (follow-up, not a blocker).**
`MAX_RECENT_GALLERY_PHOTOS` (24) only caps what's displayed and cached; `listAllDataByMarker` lists **all** `userGallery` records and nothing is ever deleted. Over time a user accumulates unbounded gallery data records, and a file upload that succeeds followed by a failed `recordData` leaves an orphaned uploaded file. Fine to ship, but worth a follow-up to prune old records / clean up orphans.

---

*Reproduction steps above assume the alpha deploy for this branch: https://alpha.seedbible.org/b/feature_1647*

---

KallynGowdy:
This looks good overall and it works. @Blitzy What do you think about the scope of this? It adds a gallery manager that stores all the photos that a user uploads so that they can be selected again. I like the idea and I'm sure it will become more important in the future, but I'm not entirely sure if its something we want in the first bundle.

KallynGowdy:
After reviewing #1721, I think having a gallery manager is going to be really important so that users can easily see all the photos that they've uploaded and manage them.

Blitzy:
@Kabi4 @KallynGowdy I agree i think the gallery feature is a good inclusion. This is the first time we are really having users upload image content and so we are going to need a way for them to manage it from the app-level. I think its reasonable to include as part of this PR.

Aside from the agent code review issues above, here are the issues I found while testing:

## Blocking

**No 1. (🔴 High) — The hero image can get squashed when viewing it in play mode.**

The hero image for the playlist (which recommend a 4:3 aspect ratio with dimensions at least 1024 x 768) can become squashed when viewing the playlist in play mode. This occurs if the contents of the playlist are long enough, it causes the image to be made shorter. The reason this is a problem is because lots of playlist hero images will very likely contain text, and we do not want to clip that text.

_Here is the original hero image for my playlist:_
image

_Now in play mode, notice how the hero image is shorter and getting clipped:_
image

**No 2. (🔴 High) — For reading plans specifically, the hero image does not wait for the user to "save changes" when the image is changed / removed to commit.** 

For reading plan hero images specifically, when you change the image it commits to the reading plan immediately instead of waiting for the user to "save changes" to the reading plan. This means that I can edit a reading plan, remove the image and then backout without saving changes and the image is now gone from the reading plan when it should still be there because the changes were not saved.

I verified that the playlist hero images behave correctly, the hero image change / removal is only ever persisted when the user saves the changes for the playlist.

## Not blocking (can be follow ups post-merge if we want)

**No. 3 (🟡 Low) — The remove button on the hero image is a trash can which implies delete but the image is not deleted its just removed as the hero image for the plan / playlist.**

The trash can probably shouldn't be used in the context of "remove image as plan/playlist hero". Instead it should be an icon that represent removal instead of deletion. If it makes more sense the button could also be converted into a text button that just says "Remove".

**No. 4 (🟠 Medium) — There is no standalone gallery manager screen that lets you manage (add / delete) images.**

This is mostly a completeness issue. The inclusion of a gallery picker for hero over images is great but it now leaves the door open a more robust set of media management features that users will expect. 

I think we should include a dedicated gallery management tool that users can access from the settings sidebar (for now atleast until we have the Your Content / Profile screen merged #1721)

This tool would let users view all of their uploaded gallery content and also let them delete them. Deletion should come with a warning about any images that may be directly referenced by reading plans / playlists.

> [!CAUTION]
> Removing an image that is actively in use by a user's reading plan or playlist should pop a warning first that the user must acknowledge.

> [!IMPORTANT]
> An image URL that fails to load should be handled by both the Hero component and the Hero thumbnail component gracefully. Similar to how it handles there being no image at all set.

## Next steps

Overall I really like these changes. As far as what I would like to see before we merge this lets address from the agent code review: **No. 2, 3, 4, 5, and 6.** I would also like to see the two "Blocking" high severity issues from this review addressed **No. 1 and 2**.

KallynGowdy:
For number 1 specifically YouVersion actually does something similar:


The original image obviously doesn't cut off the "1", but the one displayed does. In any case I think it is important to fix, I just thought it was interesting to note

Kabi4:
The cover lived in a flex column with the queue. When the queue was long, the banner shrank below 4:3 and object-fit: cover clipped the image (and any text on it).

Editing a published plan used the same autosave path as a new draft, so changing or removing the cover was written right away. Cancel/back still left that write in place.

Kabi4:
FYI since there are now lot of data there in unsaved playlsit i also supports the unsaved drafts in it.

KallynGowdy:
I can confirm that 1 and 2 are fixed now. I'm gonna merge this and then we can make issues for 3 and 4

PR #1721, opened 2026-08-29
feat: Implement YourContentManager for managing user annotations and …

### LINKED ISSUE #1552: Implement Profile screen
## Context

Building off the design discussion in #1456, we need a proper Profile screen for users.

The Profile screen is a new, distinct screen from Account Settings. The distinction:

- **Account Settings** - for changing aspects of your account.
- **Profile screen** - the home for a user's created content: playlists, annotations, highlights, and the like. Friends/subscribers/circles management is planned for later.

Circle/Friends management is intentionally out of scope for now; we don't have explicit designs for it yet.

## Design

Figma: https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=9086-4561&t=D7xWKh3SuMOR4GMX-1

## Scope

This issue tracks the overall rollout. The work is broken into sub-issues:

- [ ] Profile screen (base profile screen/shell)
- [ ] "Your Content" section (playlists, annotations, highlights)
- [ ] Profile screen entry points (mobile toolbar, desktop sidebar)

mazenemad:
…highlights

- Added YourContentManager to handle loading, filtering, and managing user content.
- Introduced ContentFilter and ContentLoadStatus types for better state management.
- Implemented sorting for annotations by recency and utility function for extracting plain text from annotations.
- Created unit tests for YourContentManager to ensure correct functionality and error handling.

test: Add unit tests for ProfilePane and YourContentPane components

- Developed comprehensive tests for ProfilePane to validate user profile display and interactions.
- Created tests for YourContentPane to ensure correct rendering of user content sections and filtering functionality.
- Included tests for handling empty states and error scenarios in both components.

test: Add unit tests for YourContentManager

- Implemented tests for YourContentManager to verify loading behavior, error handling, and state management.
- Ensured annotations and highlights are loaded correctly and that deleted annotations are removed from the state.

fixes #1552


Blitzy:
> [!NOTE]

## Code review — PR #1721: `YourContentManager` + Profile / Your Content screens

Thanks for this — it's a large, well-organized change and the comments explaining the *why* throughout are genuinely helpful. The refactor of `openReadingPlansPane` out of the toolbar so the Profile card and the reader share one pane is a nice touch, and the new manager tests cover the error paths (failed load, malformed payloads, signed-out) rather than only the happy path. Everything I ran passes: the 48 new manager/component tests are green, and I confirmed the findings below against the actual code.

The change closes the "Your Content" sub-issue of #1552. Circles/friends being deliberately out of scope matches the issue description.

Below, issues are ordered by severity. Line references are to the PR head (`9d8c533`).

---

### Medium

**M1. Selecting a filter chip for an empty section shows a blank screen with no message.**

In `YourContentPane`, the "nothing to show" message is gated on `nothingAtAll`, which is true only when **every** section is empty — it ignores which chip is active:

```tsx
const nothingAtAll =
  status === "ready" &&
  visibleAnnotations.length === 0 &&
  visibleHighlights.length === 0 &&
  visibleBookmarks.length === 0 &&
  visiblePlaylists.length === 0;
```
(`YourContentPane.tsx:460`)

So if a user has, say, annotations but no highlights and taps the **Highlights** chip: the Annotations section is hidden (`showing("annotations")` is false), the Highlights section is hidden (nothing to render), and `nothingAtAll` is false because `visibleAnnotations.length > 0`. The result is a search box and chips over an empty void — no "empty" and no "no matches" message.

*Reproduction:* Open Your Content as a user who has at least one annotation (or bookmark/playlist) but no highlights. Click the **Highlights** chip. The area below the chips is blank.

I confirmed this with a rendering test against the PR code (filter = `highlights`, one annotation, no highlights): zero sections render and no `.sb-content-status` element is emitted.

*Suggested fix:* base the empty message on the sections actually being shown, e.g.

```tsx
const showingNothing =
  (!showing("annotations") || visibleAnnotations.length === 0) &&
  (!showing("highlights") || visibleHighlights.length === 0) &&
  (!showing("bookmarks") || visibleBookmarks.length === 0) &&
  (!showing("playlists") || visiblePlaylists.length === 0);
const nothingAtAll = status === "ready" && showingNothing;
```

A regression test worth adding: render with a non-`all` filter whose section is empty while another section has items, and assert the empty-state message appears.

---

### Low

**L1. Optimistic annotation delete has no rollback if the server delete fails.**

In the delete handler the note is removed from the list first, then the server delete is fired and its result is discarded:

```tsx
yourContent.removeAnnotation(annotation.id);
void annotations.deleteAnnotationAndRefresh(annotation);
```
(`YourContentPane.tsx:563`)

`deleteAnnotationAndRefresh` awaits `deleteAnnotation`, which throws on a failed record call. Because it's fired with `void`, a failure is swallowed and the note stays gone from *Your Content* even though it still exists on the server; it only reappears after a forced reload. Removing-immediately is the right call for perceived speed (the comment explains this well) — the gap is only the missing rollback on rejection. Consider re-loading (`load({ force: true })`) or re-inserting the item in a `.catch`.

**L2. The whole user record is swept twice on each open.**

`YourContentManager.runLoad` issues `listAllAnnotations()` and `listAllHighlights()` in parallel, and each calls `os.listAllData(recordName)` on the *same* per-user record — so the entire record (annotations, highlights, bookmarks, playlists, reading-plan progress, …) is paged down from the server twice. For a heavy user this doubles the transfer for the screen. It's a reasonable first cut given the two managers own their own parsing, but it may be worth a single shared sweep that both filter, if this screen turns out to be hot.

**L3. Test coverage gap: the active reading-plan branch of `ProfilePane` is untested.**

`ProfilePane.test.tsx` covers the signed-out prompt, initials, picture, routing, and the *no plan* fallback — but never the populated `findActivePlan` path (progress bar width, `Day X of Y`, streak, and the "furthest-along, still-unfinished" selection among multiple plans). That selection/`summarizeCalendar` logic is the most intricate part of the component and currently has no test asserting it renders correctly. Given the emphasis on tests exercising real behavior, a case with one or two started plans and a stubbed `summarizeCalendar` would be valuable.

**L4. `playlist-item-count` renders "1 items" for a single-item playlist.**

`t("playlist-item-count", { count: playlist.items.length, defaultValue: "{{count}} items" })` passes `count` but there's only one form of the key, so i18next has no singular to fall back to. A one-item playlist reads "1 items". If singular/plural matters here, add the `_one`/`_other` variants (English needs both).

---

### Minor notes (non-blocking)

- **`createYourContentManager` takes a `login` option it never uses** — the function destructures only `annotations` and `highlights`. Either drop `login` from `CreateYourContentManagerOptions` or use it. Relatedly, `listAllAnnotations` resolves its record via `resolveRecordName()` (which honors a `recordOverride`) while `listAllHighlights` reads `login.userId.value` directly; consistent with each manager's own design today, but worth a note if a record override ever needs to reach highlights.
- **`parseChapterHighlightsAddress` accepts chapter `0` and negatives.** It checks `Number.isInteger(chapterNumber)` but not positivity, so `highlights:BSB/GEN/` parses as chapter `0`. Harmless because the app never writes such addresses, but `chapterNumber > 0` would make the parser stricter and match the bookmark schema (`chapterNumber: z.number().int().positive()`).
- **`annotationPlainText` decodes `&amp;` before `&lt;`/`&gt;`**, so a double-encoded `&amp;lt;` collapses to `<`. Only affects search/preview text, never stored data — very low impact.

---

Nice work overall; M1 is the only functional issue I'd want fixed before merge, and it's a small change plus a regression test.

---

KallynGowdy:
@mazenemad  This looks pretty good. Here's a couple requests:

1. The "Your content" screen should have a back button that takes the user back to their profile page:
image

2. The user's avatar should be editable by clicking it. It should open the select picture modal directly. It doesn't need to take the user to another page first.
image

3. When the user has an avatar set, the "+" should actually be a pencil icon to indicate that it can be edited. If the user doesn't have an avatar, then it should remain a "+"
image

4. We should also display the user's email address below their avatar but above their name in a smaller font. We should already have this information available in the LoginManager.

5. We should also display the location and description that the user has set, if they have one. If they have set a location, display it next to the email address with a little `location_on` icon. If they have a description, display it below the name in a smaller font as an ExpandableText so that it doesn't take up too much space.

6. Could we also get rid of the "Account settings" button and replace it with a pencil icon in the upper right corner (or left for RTL languages) of the profile card?

Instead of going to the current account settings page, we should take the user to a new "Edit profile" page that has a back button that takes the user back to their profile page. The new edit profile page should look like the profile page except it doesn't show the content buttons and has fields for `name`, `location` and `description` (in that order). The user should be able to click the pencil or click on their name to open the new page.

image

Untitled

Blitzy:
## Follow-up work — 2026-09-10

Fourteen commits on top of the `develop` merge (`e4e3cfa3`), 31 files, +2,822/−1,436. Suite went from 5,942 to **5,998 tests, all passing**; `check:ts` and `lint` clean throughout.

### Merged `develop` in

Three conflicts, all additive — nothing dropped. `CHANGELOG.md` and `SeedBibleStateManager.tsx` were both sides appending in the same place. `en.json` looked worse than it was: the merge base had 44 accidentally duplicated keys and each branch deduplicated them differently, so the line-level merge saw shuffling as a conflict. At the key level it was clean — 32 profile keys vs 16 about-page keys, no overlap. Resolved with the repo's own `script/mergeI18nJson.ts` driver, which `pnpm install` hadn't yet registered in a fresh container.

### Profile screen

- **Mobile bottom bar back to five tabs** — `Today, You, Bible, Search, More`. The new "You" tab had made it six by pushing Bookmarks to fifth rather than demoting it. Bookmarks now lives in the More menu next to Tabs. Consequence worth flagging: **`?chatFirst=true` no longer swaps a bottom tab.** Chat is in the More menu on mobile (it had to stop being hidden there, or it would have had no way in at all); desktop is untouched.
- **"Read more" on the profile description** — was shown based on a guessed line height (`fontSize × 1.2`) rather than a real one. Now measures the collapsed line's actual width against its available width. Also fixed the collapsed text sitting left-aligned while the expanded text was centred.
- **Sign-out button** matches account settings' — same border, radius, icon, and the label reuses the existing `sign-out` key, which is translated in all 77 locales (`log-out` was English-only).
- **Picture editor** titled "Change profile picture"; **"Take a photo" now only appears where a camera will actually open**, and the two identical file-picker entries collapsed into one, "Upload from device" — named to avoid colliding with the in-app Recent uploads gallery, which the app was also calling "gallery".

### Your content

- **Playlists share Discover's row** — same look, same actions (play, plus share/edit/delete). Previously the whole tile just started playback with no way to edit or delete, on a screen reached from a card subtitled "Create and manage your content".
- **Bookmarks got the sidebar's options menu** (edit/remove). Note the difference: the sidebar's Remove takes a bookmark out of *one folder*; this list is flat, so Remove removes the bookmark, same as the edit modal's "Remove from all folders".
- **Highlights can be cleared** from a kebab menu matching the annotation card beside them, with the same optimistic-then-roll-back pattern.
- **Search was matching book name only**, so the reference printed on a row found nothing when typed back — `"John 3:16"` matched zero. All sections now match the reference as their row prints it, plus the raw book id. **And highlighted verse text is now searchable**, read back from the chapters since nothing stores the wording.

### Bugs found along the way

- **Today didn't close when another screen opened over it.** `PanesManager` clears other panes only for a pane that fills the reader area; a `side` pane replaces just the previous side pane. Discover and Reading plans are both side panes, so they opened *behind* Today. Fixed in Today's own effect rather than in `PanesManager`, because "side panes displace fullscreen panes" would also apply to the About page, which `b7c6f045` deliberately protects.
- **Dead code removed**: `uploadCoverImageForUser` and `resolveCoverImageUpload` had no production callers left — only their own tests, which is why nothing flagged them.

### Refactors

- One shared `PhotoCropModal` for profile pictures and playlist/plan covers, replacing two near-identical copies.
- Discover's playlist row extracted to `PlaylistRow.tsx`, used by both screens.
- `BibleDataManager.getCachedTranslationBookChapter` — a synchronous peek at already-fetched chapters, mirroring the existing `getCachedTranslationBooks`, so verse-text search costs nothing for chapters already in hand.

### Known gaps

- **`resetFilters()` still has no caller.** Search and chips persist across closing and reopening the screen. Left deliberately.
- The profile picture stays out of the shared Recent uploads gallery, as decided.
- `TodayManager.getVerseText` still has its own copy of the chapter-text extraction that `extractContentText` does.
- The bookmark SVG path is copy-pasted six times across five files even though `icons.tsx` exports a `BookmarkIcon` for it.


---

Blitzy:
## Code review follow-up

All three findings from the review are addressed in `0043fc7d`. Two were regressions introduced inside this PR; one was a duplicated helper.

### Your content went stale after the first visit (medium)

**Before:** you opened "Your content", went back to the reader, highlighted a verse or wrote a note, then returned — and the new item wasn't there. It only appeared after a full page reload.

**Why:** the screen asked the manager to load when it opened, but the manager deliberately skips a load once it already has content (so that flipping between screens doesn't re-fetch every time). After the first successful load, every later open was a no-op.

**Now:** the screen forces a refresh on every open.

The obvious version of that fix would have made things worse. Forcing a reload also set the status to "loading", which blanks the lists and shows a spinner — so reopening the screen would have wiped out what you were just reading for as long as the request took. So a forced refresh over content that's already loaded is now **quiet**: the lists stay on screen and the status stays `ready` while it runs, and a quiet refresh that fails leaves the old lists up instead of replacing them with an error page. A first load, and a retry after a failed first load, still show the spinner as before — there's nothing on screen to protect in those cases.

### A reading plan's day played with a blank cover (medium)

Extracting the shared `openReadingPlansPane` (so the profile screen and the reader toolbar land on the same pane) dropped `heroImageUrl` from the throwaway playlist a plan's day is played through. The player takes its cover art from the playlist it's handed, so a plan with a hero image played with nothing.

Restored, and the mapping from plan → playlist is now one named function, `readingPlanDayPlaylist`, with tests on it, so the field can't quietly go missing again during a future refactor.

### Duplicated verse-number helper (low)

`YourContentPane` carried its own `annotationVerses`, a near-copy of `AnnotationsManager.annotationVerseNumbers`. The two disagreed on an inverted range such as `5-3`: the canonical helper returns no verses (the annotation reads as whole-chapter), the local copy returned just the start verse. The local copy is gone, so the screen now labels an annotation exactly as the reader does.

### Verification

- `pnpm test` — 6049 tests across 254 files, all passing
- `pnpm check:ts` — 0 visible errors
- `pnpm lint` — 0 errors; the touched files lint completely clean
- Both new quiet-refresh tests and the hero-image test were confirmed red against the pre-fix code and green after, per the repo's regression-test convention

### Two judgement calls worth flagging

**No changelog entries.** All three findings were regressions inside this PR rather than against `develop` — `develop` already passes `heroImageUrl` through, and the "Your content" screen doesn't exist there at all. Nothing changes for a user comparing this branch to `develop`, so there's nothing to log.

**No new test for the duplicated helper.** It's a pure de-duplication; the canonical helper already has its own tests, and the only behaviour that differs is for corrupt inverted-range data. A test pinning that down would be codifying an edge case rather than protecting real behaviour.


---

Issue #1703, opened 2026-08-27
Verse toolbar (desktop & mobile) should be dismissable by clicking off of it


Blitzy:
Currently the verse toolbar on both desktop and mobile cant be closed / dismissed by clicking off of it on an empty space on the page. We should change this so that if you click on anything outside of the verse toolbar (except for another verse of course) that it should close the the verse toolbar on both desktop and mobile.

PR #1723, opened 2026-08-31
Clear verse selection only when tapping empty chapter space

KallynGowdy:
## Summary

Changed the verse selection dismissal logic to distinguish between tapping on a verse and tapping on empty space within the chapter. Previously, any tap outside the entire `.sb-chapter-content` container would clear the selection; now, only taps on empty space (padding, gaps between verses, section headings) clear it, while taps directly on a verse are left to that verse's own click handler.

## Changes

- **Updated pointer event listener** in `BibleReaderToolbar.tsx`: Changed the exclusion check from `target.closest(".sb-chapter-content")` to `target.closest(".sb-verse")`. This allows the listener to catch taps on empty space within the chapter while standing aside for the verse's own `onClick` handler.

- **Clarified intent in comments**: Expanded the comment explaining why we exclude only `.sb-verse` rather than the whole chapter container, making it clear that empty space (padding, gaps, headings) is exactly what this listener is meant to catch.

- **Added test coverage**: Two new unit tests verify the behavior:
  - Tapping in empty chapter-content space clears the verse selection
  - Tapping on a verse itself does not clear the selection

## Implementation Details

The change is deliberate: a verse's own `onClick` already handles toggling that verse's selection, so this document-level listener must not interfere. By checking only for `.sb-verse`, we let the verse handle its own interaction while still dismissing the selection when the user taps on the surrounding empty space—the actual "click off of it" behavior the listener exists to provide.

https://claude.ai/code/session_01EbAPb2nae1TdGwGcUjh8fo

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
@@ -872,6 +872,17 @@ function renderChapterContent(
           className={verseClassName}
           data-verse-number={value.number}
           onClick={(event: MouseEvent) => {
+            // Poetry lines are `display: block` so each one spans the full
+            // content width — a tap in the blank margin past a short line's
+            // last word still lands inside this outer span even though it's
+            // nowhere near the verse's actual text. Only a tap that reaches
+            // an actual `.sb-verse-decorator` (the inline span the rendered
+            // words themselves sit in) should count as selecting the verse;
+            // anywhere else in the block is empty space, which the reader's
+            // outside-click handling (`BibleReaderToolbar.tsx`) needs to be
+            // free to treat as a dismiss instead.
+            const target = event.target as HTMLElement | null;
+            if (!target?.closest(".sb-verse-decorator")) return;
             onVerseClick(verse, event);
           }}
           style={{
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
@@ -589,6 +589,22 @@ function renderInlineContent(
   return null;
 }
 
+/**
+ * Pointer type of the most recent `pointerdown` on a poetry verse's outer
+ * span, read by that same verse's `onClick` guard below to decide how
+ * forgiving its tap region is. Module scope rather than a ref: `renderVerseNode`
+ * is a plain helper re-created on every call, not a component, so it has
+ * nowhere of its own to persist state between the pointerdown and the click
+ * that follows it (same reasoning as the module-scope state in
+ * `app/flingSafeTap.ts`).
+ */
+let lastVersePointerType = "";
+
+/** Clears the module-scope pointer-type state so tests cannot leak it between cases. */
+export function resetLastVersePointerTypeForTests() {
+  lastVersePointerType = "";
+}
+
 function renderChapterContent(
   chapterData: TranslationBookChapter | null,
   onVerseClick: (verse: BibleSelectedVerse, event: MouseEvent) => void,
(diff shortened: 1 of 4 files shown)


Blitzy:
> [!NOTE]

## Code Review — PR No. 1723

**Verdict: Looks good.** The change correctly implements what issue #1703 asks for, the reasoning is well documented in the expanded comment, and — importantly — the two new tests are genuine regression tests, not tautologies. I found no correctness bugs. A couple of minor test-hardening suggestions and one behavioral note are below.

### What the change does

Before: any tap inside `.sb-chapter-content` was excluded, so tapping the empty space *within* a chapter (padding, gaps between verses, headings) did **not** dismiss the verse selection — you had to tap fully outside the reader. Now: only taps that land on (or inside) a `.sb-verse` are excluded, so a tap on empty chapter space dismisses the selection while a verse's own `onClick` still handles verse-to-verse toggling. That matches the issue's "click off of it on an empty space … except for another verse."

### Verification performed

- Ran the full test file on the PR head: **59 passed**.
- Reverted the one-line source change back to `.sb-chapter-content` and re-ran the two new tests: **both failed (red)**. This confirms they exercise the real behavior change and won't silently pass on the pre-fix code.
- Confirmed the `.sb-verse` selector token cannot accidentally match `.sb-verse-toolbar` or `.sb-verse-number` (different class tokens), and that footnote markers/verse numbers rendered *inside* a verse are still correctly excluded because `closest(".sb-verse")` walks up to the enclosing verse span (`BibleReader.tsx` `renderVerseNode`).

### Findings

#### Low — `does not clear the verse selection when a tap lands on a verse` can pass vacuously if the toolbar ever stops rendering in this setup

`test/unit/seed-bible/components/BibleReaderToolbar.test.tsx` (the second new test, ~lines 198–217) asserts the selection is *unchanged* after tapping a `.sb-verse`, but unlike its sibling it never asserts the toolbar is actually visible first. The document `pointerdown` listener only attaches while `isVerseToolbarVisible` is true, so if a future refactor caused the toolbar not to render in this setup, the listener wouldn't attach and the test would still pass (selection stays at length 1) — testing nothing. It is **not** vacuous today (I confirmed it goes red on the pre-fix code), so this is hardening, not a defect.

*Suggested fix:* add the same guard the first new test uses, right after `renderToolbar()`:

```ts
expect(container.querySelector(".sb-verse-toolbar")).not.toBeNull();
```

#### Low (nit) — no test covers the desktop path, though the issue calls out "desktop & mobile"

Both new tests run under the mobile viewport (`MOBILE_VIEWPORT_WIDTH = 400`, set in the `describe` block's `beforeEach`). The dismissal logic itself is viewport-independent — the only difference between desktop and mobile is how the toolbar chrome renders, not the listener — so functional risk is low and I wouldn't block on this. Worth a one-line comment or a desktop-viewport variant if you want the "both platforms" guarantee reflected in the suite.

#### Informational — headings, subtitles, and line breaks inside the chapter now dismiss the selection (by design)

Because the exclusion is now `.sb-verse` rather than the whole container, taps on the non-verse siblings that live inside `.sb-chapter-content` — `.sb-chapter-heading` (h3), `.sb-subtitle` (p), and `.sb-line-break` (div) — now clear the selection. That's consistent with the issue ("anything outside … except another verse"), so this is expected, not a bug. One narrow edge to be aware of: a footnote marker rendered inside a Hebrew subtitle (`.sb-subtitle`, via `onOpenFootnote(noteId, null)` in `BibleReader.tsx`) is *not* inside a `.sb-verse`, so tapping it will now both open the footnote and clear the verse selection. This is rare and arguably harmless (a subtitle footnote isn't tied to the selected verse), but flagging it in case you'd prefer footnote markers everywhere to stand aside.

### Summary

| Severity | Count |
|---|---|
| Critical / High / Medium | 0 |
| Low | 2 |
| Informational | 1 |

No changes are required to ship this correctly; the Low items are optional test-robustness improvements.

---

PR #1724, opened 2026-08-31
Style clear button with error color and consistent hover state

### LINKED ISSUE #1705: Highlight clear button on mobile should use the same erase icon that the desktop highlight modal uses and the same pink coloring.
Currently the mobile clear highlight button is a grey x (close) icon. It should instead match the clear button from the desktop highlight modal that uses an eraser icon and the same pink coloring the desktop clear button and icon use.

Desktop: 
Image

Mobile: 
Image

KallynGowdy:
## Summary
Updated the "Clear" button in the mobile verse toolbar to use the error color scheme instead of the font color, and made the hover state consistent across screen sizes by removing the conditional icon swap.

## Changes
- **Color scheme**: Changed the clear button background and text from `var(--sb-font-color)` to `var(--sb-error-color, #d64545)` with appropriate fallback values. This makes the destructive action visually distinct.
- **Hover state**: Replaced `filter: none` with `filter: brightness(0.95)` on hover to provide a subtle visual feedback that works consistently regardless of the underlying color.
- **Icon consistency**: Removed the conditional logic that showed a "close" icon on small screens and "ink_eraser" on larger screens. Now always displays "ink_eraser" for a unified appearance.
- **Color mix adjustments**: Fine-tuned the transparency values in the `color-mix()` function (85% and 78%) to work better with the error color than the previous values (88% and 80%) did with the font color.

## Implementation Details
The error color with a light transparency creates a subtle red background that clearly signals a destructive action, while the brightness filter on hover provides feedback without needing to change the entire color scheme. This approach is more accessible and semantically clearer than using the generic font color for a destructive button.

https://claude.ai/code/session_01K2qKhUP1SCnEwXJN8v1EYP

Closes #1705

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.inline.css
@@ -1348,13 +1348,21 @@
 }
 
 .sb-verse-toolbar-mobile .sb-verse-toolbar-clear {
-  background: color-mix(in srgb, var(--sb-font-color), transparent 88%);
-  color: var(--sb-font-color);
+  background: color-mix(
+    in srgb,
+    var(--sb-error-color, #d64545),
+    transparent 85%
+  );
+  color: var(--sb-error-color, #d64545);
 }
 
 .sb-verse-toolbar-mobile .sb-verse-toolbar-clear:hover:not(:disabled) {
-  filter: none;
-  background: color-mix(in srgb, var(--sb-font-color), transparent 80%);
+  filter: brightness(0.95);
+  background: color-mix(
+    in srgb,
+    var(--sb-error-color, #d64545),
+    transparent 78%
+  );
 }
 
 .sb-verse-toolbar-mobile
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
@@ -2451,9 +2451,7 @@ export function BibleReaderToolbar(props: BibleReaderToolbarProps) {
                   })}
                   title={t("clear", { defaultValue: "Clear" })}
                 >
-                  <span className="material-symbols-outlined">
-                    {isSmallScreen.value ? "close" : "ink_eraser"}
-                  </span>
+                  <span className="material-symbols-outlined">ink_eraser</span>
                   <span className="sb-verse-toolbar-action-text">
                     {t("clear", { defaultValue: "Clear" })}
                   </span>


Blitzy:
> [!NOTE]

## Code review — PR No. 1724

**Verdict:** Looks good. This is a small, focused change that correctly does what issue [#1705](https://github.com/HelloAOLab/seed-bible/issues/1705) asked for. No correctness bugs found — the notes below are all low-severity / informational.

### What the PR does (and why it's correct)

Issue #1705 asks the mobile "clear highlight" button to stop being a grey `close` (X) icon and instead match the desktop highlight modal's clear button — an `ink_eraser` icon with the same coloring.

The PR does exactly that:

- **Icon:** `BibleReaderToolbar.tsx` previously rendered `isSmallScreen.value ? "close" : "ink_eraser"`, so mobile got the X and desktop got the eraser. It now always renders `ink_eraser`, so mobile matches desktop. I confirmed `isSmallScreen` is still used in ~30 other places, so removing this one reference doesn't create a dead variable.
- **Color:** The mobile rule (`.sb-verse-toolbar-mobile .sb-verse-toolbar-clear`) moves from `--sb-font-color` to `var(--sb-error-color, #d64545)`. I checked `origin/develop`: the desktop rule (`.sb-verse-toolbar-clear`) **already** used `var(--sb-error-color, #d64545)` at `transparent 85%` base / `transparent 78%` hover with `filter: brightness(0.95)`. The new mobile values are an exact match of those, so the PR body's "fine-tuned 85%/78%" is really "matched the desktop values" — which is the right call.

CI: the prior `build (24.x)` passed and the current one was still in progress at review time; nothing failing.

---

### Low severity / informational

**No. 1 — Missing CHANGELOG entry.**
`CHANGELOG.md` keeps a `## TBD` section where merged PRs are linked (there's even a `changelog-update` skill in the repo). This user-visible tweak has no entry. Suggest adding one under `### 🔧 Changed`, e.g.:
> - Style the mobile "clear highlight" button to match the desktop eraser icon and color instead of a grey X. ([#1724](https://github.com/HelloAOLab/seed-bible/pull/1724))

**No. 2 — "Pink" in the issue vs. the red that actually renders (worth a sanity check, not a blocker).**
The issue text says "the same pink coloring." The color that actually renders is `#d64545` — a red/salmon, not pink. That's because `--sb-error-color` is undefined app-wide (see No. 3), so the `#d64545` fallback always wins. This still satisfies the issue's real intent — "match the desktop clear button" — since desktop uses the identical value. But the app does define a genuine pink (`--sb-highlight-pink-color: #f48fb1` in `app/styles/base.inline.css`), so if the reporter literally meant that pink, this isn't it. Recommend confirming the desktop red is the intended look before closing #1705.

**No. 3 — `--sb-error-color` is never defined anywhere in the codebase.**
Grepping the repo, `--sb-error-color` is only ever *read*, never *set* — it's absent from both `ThemeManager.tsx` and the base CSS. Two consequences, both pre-existing and not introduced by this PR, but this PR extends the pattern:
- The fallback is inconsistent across call sites: `#d64545` here and in `DiscoverPane`, but `#dc2626` in `SettingsPage.tsx`. So "error" UI isn't one color.
- Because it's undefined, the button color is not theme-aware — the same `#d64545` renders in both light and dark mode (the `color-mix(... transparent 85%)` background becomes ~15%-opacity red on the dark sheet, which is quite subtle in dark mode).

Not in scope for this PR, but flagging it: defining `--sb-error-color` in `ThemeManager`'s `LIGHT_THEME`/`DARK_THEME` (per the theming note in `CLAUDE.md`) would make every clear/error affordance consistent and dark-mode-tuned in one place.

**No. 4 — No test for the icon change.**
Per the review brief, I looked specifically at test coverage: there are no tests in this PR. For the pure CSS parts that's fine (CSS isn't meaningfully unit-testable here). But the icon swap *is* observable behavior — before, mobile rendered `close`; now it renders `ink_eraser` — and it's the one thing that could silently regress. A small render test asserting the clear button shows `ink_eraser` (and not `close`) when `app.isMobile` is true would lock it in. Optional given the cosmetic nature, but it's the kind of assertion that would actually fail on the pre-fix code, so it'd be a real regression test rather than one that always passes.

---

Issue #1704, opened 2026-08-27
After applying a highlight, the users verse selection should automatically clear.


Blitzy:
Currently when you highlight verses the highlighting tool stays open and the verses stay selected. Lets change this so that after performing a highlight on verses the users verse selection is automatically cleared and the verse toolbar is automatically closed. This should apply for both desktop and mobile highlighting.

Edit:

When clearing highlights, we should also be clearing the user's verse selection automatically.

PR #1725, opened 2026-08-31
Clear verse selection after applying or clearing highlights

KallynGowdy:
## Summary

Applying or clearing a highlight now automatically clears the verse selection and closes the verse toolbar, instead of leaving them open after every highlight action.

## Changes

**Source code:**
- Modified `applyHighlightWithSession()` in `BibleReaderToolbar.tsx` to call `rs.clearSelectedVerses()` after applying a highlight (both for signed-out users and those with active sessions)
- Modified the "clear highlight" button handler to also call `rs.clearSelectedVerses()` after unhighlighting verses
- Updated the JSDoc comment on `applyHighlightWithSession()` to document this behavior

**Tests:**
- Refactored the test helper `openPickerAndHighlight()` to extract a new `openPicker()` helper, since the picker auto-closes whenever the selection is cleared
- Updated all existing tests in the "clearing highlights" suite to re-select the verse and re-open the picker after highlighting, since that action now clears the selection
- Added two new test cases:
  - "clears the verse selection and closes the toolbar after applying a highlight" — verifies the selection and toolbar are gone after highlighting
  - "clears the verse selection and closes the toolbar after clearing a highlight" — verifies the selection and toolbar are gone after clearing

**Changelog:**
- Added entry documenting the new behavior

## Implementation Details

The fix addresses issue #1704, where the verse toolbar would remain open after every highlight action, forcing users to manually dismiss it. Now both `applyHighlightWithSession()` paths (signed-out and signed-in) and the clear highlight handler call `clearSelectedVerses()` to close the toolbar automatically.

The picker (and the "Clear" button inside it) auto-closes when the selection is cleared, so tests now must re-select and re-open the picker before clicking "Clear" in subsequent assertions.

https://claude.ai/code/session_01TudS19YJkcvKE3cpvu1nj1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### 🔧 Changed
 
+- Applying or clearing a highlight now clears the verse selection and closes the verse toolbar automatically, instead of leaving the selection and toolbar open after every highlight action.
+
 ### 🐛 Fixed
 
 ### 🗑️ Removed
--- a/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx
@@ -393,9 +393,13 @@ function removeSharedHighlightsFromSelection(
  *   covers it for as long as it lives (the reader draws a decoration highlight
  *   over a saved one) and it reappears when the broadcast expires.
  *
- * Either way, the verse selection is cleared once the highlight is applied —
+ * By default the verse selection is cleared once the highlight is applied —
  * the selection and its toolbar were otherwise left sitting open after every
- * highlight, forcing an extra dismiss (#1704).
+ * highlight, forcing an extra dismiss (#1704). `clearSelection` lets a caller
+ * opt out: the custom-color picker's live-drag commits pass `false` so the
+ * selection survives while the color dialog is still open, letting the user
+ * keep tweaking the shade instead of losing the selection after the first
+ * settled color.
  */
 function applyHighlightWithSession(
   rs: BibleReadingState,
@@ -405,26 +409,28 @@ function applyHighlightWithSession(
     customColor?: string;
     customFontColor?: string;
   },
-  isSignedIn: boolean
+  isSignedIn: boolean,
+  clearSelection = true
 ): void {
   if (!session || !session.userCanDecorate(session.localSessionId.value)) {
     // A participant who can't broadcast used to match neither branch here, so
     // highlighting silently did nothing for them. Saving is the only thing this
     // can mean, so a signed-out user is asked to sign in before it applies.
     void rs.highlightSelectedVerses(details);
+  } else {
+    const duration = session.options.value.highlightDurationSeconds;
+    const isTransient = duration !== null && duration > 0;
+
+    if (!isTransient && isSignedIn) {
+      void rs.highlightSelectedVerses(details);
+    }
+
+    broadcastDecorationToSession(session, rs, details);
+  }
+
+  if (clearSelection) {
     rs.clearSelectedVerses();
-    return;
   }
-
-  const duration = session.options.value.highlightDurationSeconds;
-  const isTransient = duration !== null && duration > 0;
-
-  if (!isTransient && isSignedIn) {
-    void rs.highlightSelectedVerses(details);
-  }
-
-  broadcastDecorationToSession(session, rs, details);
-  rs.clearSelectedVerses();
 }
 
 /**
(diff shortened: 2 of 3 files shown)


Blitzy:
> [!NOTE]

## Code Review — PR #1725: Clear verse selection after applying or clearing highlights

**Verdict:** Solid, well-scoped fix that correctly implements issue #1704 (including the edit that added "clear highlights should also clear the selection"). The tests are genuine regression tests — I verified they fail on the pre-fix code. There is one narrow behavioral regression worth a look (custom-color live tweaking) and a small style nit; neither blocks merge.

### What I checked
- Every place a highlight is applied now clears the selection. All apply paths route through `applyHighlightWithSession()` (preset colors, existing custom swatches, and the debounced custom-color commit), and the "Clear" button handles the un-highlight path. There are no other highlight entry points that were missed.
- **Desktop and mobile are both covered.** The issue asks for both. The color picker markup is shared across breakpoints, and clearing the selection also closes the mobile verse sheet via the existing `useEffect` at `BibleReaderToolbar.tsx:1302` that resets `isHighlightPickerOpen` / sheet state when `hasVerseSelection` goes false.
- **No selection/clear race.** `clearSelectedVerses()` is called synchronously right after `void rs.highlightSelectedVerses(...)` / `void rs.unhighlightSelectedVerses(...)` / `broadcastDecorationToSession(...)`. That is safe because each of those reads `selectedVerses.value` synchronously (before any `await`) — `highlightSelectedVerses`/`unhighlightSelectedVerses` capture `verseNumbers` at the top of the async function (`BibleReadingManager.tsx:2257` and `:2292`), and `broadcastDecorationToSession` reads the selection at `BibleReaderToolbar.tsx:322` before its work. So the verses are captured before the selection is emptied.

### Test verification
I reverted the three `clearSelectedVerses()` source changes locally and re-ran `pnpm vitest run BibleReaderToolbar.test.tsx`: **9 tests failed** (both new cases plus the 7 existing "clearing highlights" cases that were updated to re-select/re-open). Restoring the fix → **59 passed**. The new cases assert real observable state (`highlights` length, `selectedVerses` length, and the `.sb-verse-toolbar` DOM node), so they exercise the actual behavior rather than trivially passing.

One subtlety worth noting for the record: the updated existing tests fail without the fix partly because `selectVerse()` is a *toggle* (`BibleReadingManager.tsx:1730`) — re-selecting an already-selected verse deselects it, closing the toolbar and making `openPicker()` throw. That's an indirect coupling, but the two **new** cases assert the behavior directly, so coverage is sound either way.

---

## Issues by severity

### Low–Medium — Iterative custom-color tweaking silently stops after the first commit

The native custom-color `<input type="color">` fires `onInput` live as the user drags (in Chromium/Firefox the dialog emits input events continuously), and both `onInput` and `onChange` call `commitCustomColor()` (`BibleReaderToolbar.tsx:2419`–`2426`). `commitCustomColor` debounces 300ms and then calls `applyHighlightWithSession()`, which now clears the selection.

Consequence: the **first** time the user pauses for 300ms, the highlight applies and the selection is cleared. If the picker is still open and they keep adjusting the color, the next debounced commit runs with an empty selection, so `highlightSelectedVerses` returns early (`verseNumbers.length === 0`) and `broadcastDecorationToSession` returns early — the further adjustments do nothing. Before this PR the user could iterate freely within one picker session.

**Reproduction steps (desktop, Chromium):**
1. Select a verse, open the highlight picker, click "Add custom color".
2. In the OS color dialog, drag to a color and pause ~0.5s (highlight applies, selection clears).
3. Without closing the dialog, drag to a different shade and pause again.
4. Observe the verse stays on the first color; the second adjustment is lost. To change it you must close the dialog, re-select the verse, and reopen the picker.

**Suggested fix:** the debounced *live-input* path shouldn't clear the selection — only discrete "I'm done" actions should. Two clean options:
- Move the `clearSelectedVerses()` calls out of `applyHighlightWithSession()` and into the discrete click handlers (the preset button at `:2325`, the custom-swatch button at `:2352`, and the clear button), leaving `commitCustomColor` to not clear; or
- Have `commitCustomColor` skip the clear (e.g. an optional `clearSelection` flag on `applyHighlightWithSession`, defaulting true), and clear only on the color input's final `onChange` rather than every `onInput`.

This is Low-ish because the preset-color and single-pick custom flows (the common cases) work correctly; it only bites iterative tweaking within one open dialog.

### Nit — Duplicated `clearSelectedVerses()` in `applyHighlightWithSession()`

`applyHighlightWithSession()` calls `rs.clearSelectedVerses()` twice — once in the no-session/no-permission branch before its early `return` (`:415`) and once at the end (`:427`). If the custom-color concern above is addressed by keeping the clear inside this function, consider consolidating (e.g. clear once at the end and drop the early `return`, or clear at the single tail). Purely cosmetic.

### Nice-to-have — No test for the custom-color path

The new tests cover preset-color apply and clear, but not the custom-color (`commitCustomColor`) path. If you adjust the custom-color clearing behavior per the Low–Medium item, a test that a custom-color highlight applies and then can still be re-tweaked (or that the selection persists across live input) would lock in the intended behavior.

---

Nothing here is a correctness bug in the primary flows, and the changelog entry accurately describes the change. Once the custom-color interaction is confirmed acceptable (or adjusted), this looks good to merge.

---

KallynGowdy:
Addressed the review feedback in fdc32879:

**Custom-color live tweaking (Low–Medium):** Went with a variant of the suggested `clearSelection` flag on `applyHighlightWithSession()` (default `true`), but had to change how "final" is detected. The suggested approach (clear only on the input's `change`, not `input`) doesn't actually work in this app: `preact/compat` rewrites `onChange` to also listen for the native `input` event for anything but file/checkbox/radio inputs (matching React's semantics), so `onChange` and `onInput` both end up bound to the same native `input` event here and fire together on every drag tick — there's no native `change` event reaching either handler to key off. Instead, the debounced live commits now always pass `clearSelection: false`, and the selection is cleared for real in a new `finishCustomColor()` that runs on the input's `blur` (closing the OS color dialog is the reliable "I'm done" signal). It also flushes a still-pending debounced pick immediately on blur so a fast pick-and-close isn't lost to the 300ms delay, and leaves the selection untouched if the dialog closes without picking a color at all.

**Duplicate `clearSelectedVerses()` nit:** Consolidated into a single call at the end of `applyHighlightWithSession()`.

**Custom-color test coverage:** Added three tests covering live-drag (selection stays open, color keeps updating), the pending-flush-on-blur race, and closing the dialog without picking a color (selection untouched).

Full suite (5274 tests), typecheck, and lint all pass.

---

PR #1726, opened 2026-08-31
Keep the Today screen open through startup restores

Blitzy:
## What's wrong

Opening the app without a link to a specific chapter auto-opens the Today screen, but it appeared and then vanished a moment later.

**Before:** you open `https://alpha.seedbible.org/`, Today shows for a fraction of a second, then closes and dumps you in the reader. 

**Now:** Today opens and stays up until you choose where to go.

There are two separate triggers with one shared cause, and this PR fixes both.

## Root cause

A rule in `SeedBibleStateManager` closes any fullscreen screen (Today included) whenever the book or chapter in the address bar changes, so that navigating always reveals the reader. It only sees the URL — never *why* the URL changed.

Two things move the URL after the page has already loaded, and neither is the reader going anywhere:

1. **Saved tabs.** Just after the first render, the app restores your last reading position from browser storage and writes it into the address bar. The boot tab commits Genesis 1, the restore then commits wherever you actually left off, and that second write read as "the user navigated." This hit every returning reader whose last chapter wasn't Genesis 1 — a first-time visitor has nothing stored, which is why it looked intermittent.
2. **Saved translation.** For a signed-in reader, the app restores the translation saved in their profile once it arrives over the network. If that translation doesn't contain the book they're on (an NT-only translation while you're in Genesis), it falls back to the translation's first book — changing the book in the address bar, and closing Today by the same rule. An out-of-range chapter gets clamped the same way.

A pure translation swap never triggered it: the rule compares only book and chapter, deliberately ignoring the translation segment of the path.

## The fix

Both restores are now marked as restores while they write to the URL, so the rule re-baselines the reading position instead of counting it as a move away from one. In both cases the baseline is updated *before* the check, so the restored position becomes what everything after it is compared against — a real navigation afterwards still closes fullscreen screens exactly as before.

The two windows have different shapes because the restores do:

- **Saved tabs** — `restoringStoredState` wraps the whole `batch(...)` in `hydrateFromStorage`. That works because `batch()` flushes its queued effects synchronously before returning, so the reader's URL commit lands inside the window.
- **Saved translation** — `restoringProfileTranslation` wraps only the single `commitSelectedTabToUrl` call at the end of `applySavedTranslation`. This restore is asynchronous (it awaits the tab going idle, then a network fetch for the translation's book list), so a flag spanning the whole function would be true across those awaits and could swallow a genuine navigation. A one-call window is enough because `selectTranslationAndChapter` is called with `{ updateUrl: false }`, which returns before emitting any navigation — that final commit is the only URL write in the whole restore.

`TabsManager` exposes its flag as `isRestoringProfileTranslation()`. It's a getter rather than a signal on purpose: it's only meaningful read synchronously from inside the effect it informs, and making it reactive would invite subscribers that re-run on a value already back to `false` by the time they saw it.

## Testing

Two regression tests in `test/unit/seed-bible/managers/todayStaysOpenOnBoot.test.ts`, each confirmed **red before its fix and green after**.

Both boot at `/` with no `?today=` param, so the real auto-open heuristic decides rather than a test flag. Each waits for the restore to actually land — asserting the book moved *and* the pathname changed — before checking that Today is still up, so neither can pass for the boring reason that nothing moved the reader at all.

- `stays open while the saved reading position is restored` — seeds stored tabs pointing at Exodus 2, asserts the URL reaches `/en/AAB/exodus/2` with Today still open.
- `stays open while the profile's saved translation is restored` — boots on Genesis, then sets a profile whose saved translation is the Matthew-only `NIV` fixture, asserts the URL reaches `/en/NIV/matthew/1` with Today still open.

Full suite: **226 files, 5271 tests, all passing.** `pnpm check:ts` reports 0 visible errors (6 suppressed, pre-existing). ESLint clean on all changed files.

Also verified by hand in the browser against the manual repro for the saved-translation case: sign in, click a partial translation in the selector so it persists to your profile, clear site data so the next boot starts on the Genesis 1 default, then load the bare URL.

## Trade-offs worth a reviewer's attention

**These flags are exactly as strong as the synchronous window they cover.** Nothing structurally enforces that the URL write stays synchronous — if someone later debounces `commitSelectedTabToUrl`, or makes `hydrateStoredTabs` await something, a flag will already be back to `false` when the write lands and the flicker returns. Both call sites carry comments saying so, and the two regression tests are the real guard rail.

**A structurally stronger fix was considered and rejected as too broad for this bug.** Real navigations push a history entry while both restores commit a `replace`, so the effect could distinguish them that way instead of by flag. But back/forward arrives via `popstate` rather than a push, so "replace means not a navigation" isn't complete on its own — it's a genuine refactor of every fullscreen-pane-closing path, with its own edge cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@
 
 ### 🐛 Fixed
 
-- Fix the Today screen appearing and then immediately disappearing when opening the app without a link to a specific chapter. Returning readers have their last chapter restored from the browser just after the page appears, and the app was mistaking that restore for the reader navigating away, which closed Today the moment it opened. The restore is now recognised as part of the same page load, so Today stays up until you choose where to go.
-
 ### 🗑️ Removed
 
 ## v1.6.0 — 2026-08-24
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### 🐛 Fixed
 
+- Fix the Today screen appearing and then immediately disappearing when opening the app without a link to a specific chapter. Returning readers have their last chapter restored from the browser just after the page appears, and the app was mistaking that restore for the reader navigating away, which closed Today the moment it opened. The restore is now recognised as part of the same page load, so Today stays up until you choose where to go.
+
 ### 🗑️ Removed
 
 ## v1.6.0 — 2026-08-24
--- a/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
@@ -767,11 +767,14 @@ export function createSeedBibleState(
     if (previous === null || previous === location) {
       return;
     }
-    // The saved reading position arriving from `localStorage` is still part of
-    // this load, not a navigation away from it — see `restoringStoredState`.
-    // The baseline above is updated first, so the restored position becomes
-    // the position everything after it is compared against.
-    if (restoringStoredState) {
+    // A restore is not a navigation: the reader hasn't gone anywhere, the app is
+    // catching the URL up to state that only arrived after the load — the saved
+    // tabs from `localStorage` (see `restoringStoredState`) or the profile's
+    // saved translation, which reaches this effect at all because it can fall
+    // back to a different book (see `isRestoringProfileTranslation`). The
+    // baseline above is updated first in both cases, so the restored position
+    // becomes the position everything after it is compared against.
+    if (restoringStoredState || tabs.isRestoringProfileTranslation()) {
       return;
     }
     panes.closeFullscreenPanes();
--- a/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/SeedBibleStateManager.tsx
@@ -726,6 +726,14 @@ export function createSeedBibleState(
     discover,
     chats
   );
+  // True only while `hydrateFromStorage` below is applying the saved tab state.
+  // Restoring the tabs replaces the URL-seeded boot tab, and the reader commits
+  // the restored book/chapter to the address bar — a URL move that looks exactly
+  // like a navigation to the effect below, but is the tail end of the same page
+  // load. Without this, a returning visitor whose last chapter wasn't the boot
+  // default watched Today open and then immediately close again.
+  let restoringStoredState = false;
+
   // Close any fullscreen pane when the book/chapter in the URL path changes,
   // so navigating reveals the reader (every navigation path writes this
   // position into the path — see `commitSelectedTabToUrl` in TabsManager).
@@ -757,6 +765,13 @@ export function createSeedBibleState(
     lastReadingLocation = location;
 
     if (previous === null || previous === location) {
+      return;
+    }
+    // The saved reading position arriving from `localStorage` is still part of
+    // this load, not a navigation away from it — see `restoringStoredState`.
+    // The baseline above is updated first, so the restored position becomes
+    // the position everything after it is compared against.
+    if (restoringStoredState) {
       return;
     }
     panes.closeFullscreenPanes();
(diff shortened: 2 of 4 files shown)


Blitzy:
> [!NOTE]

## Summary

This is a solid, well-scoped fix for the Today-screen flicker. The root-cause writeup is accurate, the two fixes are minimal, and the regression tests are genuine — I verified they actually fail on the pre-fix code (details below). **I recommend approving.** No blocking issues found; a couple of low-severity observations follow, most of which the PR already anticipates.

## What I verified

**The regression tests are real, not always-green.** I neutralized the guard in `SeedBibleStateManager` (forced the `restoringStoredState || isRestoringProfileTranslation()` early-return to never fire) and re-ran `todayStaysOpenOnBoot.test.ts`. Both tests went red — and importantly, each failed at the *meaningful* assertion (`expect(state.today.isOpen.value).toBe(true)` on lines 96 and 127), **after** the "restore actually landed" preconditions passed (the pathname reached `/en/AAB/exodus/2` and `/en/NIV/matthew/1` respectively). So neither test can pass for the boring reason that nothing moved the reader — the "assert the book moved *and* the pathname changed before checking Today" pattern does its job.

**The two tests independently guard their two fixes.** Test 1 seeds a stored tab and sets no profile, so it exercises only `restoringStoredState`. Test 2 stores no tab (the boot tab stays on Genesis, so storage hydration moves nothing) and sets a profile, so it exercises only `restoringProfileTranslation`. Good separation — a regression in either fix is caught by its own test.

**The synchronous-window assumption the fix depends on actually holds in the code.** The whole approach rests on "the URL write and every effect that reacts to it run synchronously, inside the flag window." I traced it: `commitSelectedTabToUrl` → `writeUrl` → `navigation.updatePathAndQueryParams` → `replace()` → `window.history.replaceState`, which `NavigationManager` has patched to call `syncCurrentUrl()` synchronously, which writes `currentUrl.value` synchronously, which re-runs the fullscreen-pane effect synchronously (no batch around it). For the tabs case, `batch()` flushes queued effects synchronously before returning. So both windows genuinely cover the effect. The trade-off note in the PR body about this fragility is the right call-out.

**The profile-restore test models the real trigger.** The effect that calls `applySavedTranslation` (TabsManager, ~line 952) keys only on `login.profile.value`, not on `login.userId`, so the test setting `state.login.profile.value = {...}` directly fires the exact production path rather than a test-only seam.

**No collateral damage.** `TabsManager.test.ts`, `TabsPersistence.test.ts`, `TodayManager.test.ts`, `NavigationManager.test.ts`, and `todayCoversReader.test.ts` all pass unchanged (113 + 2 tests).

## Issues by severity

### Low

**No. 1 — The flags are exactly as strong as the synchronous window, and nothing structurally enforces that window (already disclosed).** If someone later debounces `commitSelectedTabToUrl`, or makes any step between the tab-hydration/translation-restore and the URL write `await` something, the flag will be back to `false` when the write lands and the flicker silently returns. The PR body calls this out explicitly, both call sites carry comments saying so, and the two regression tests are the guard rail — which is a reasonable place to leave it. Flagging only so it's on the record for a future reviewer: the safety net here is the tests, not the type system. No change requested.

**No. 2 — Consider a dev-only assertion as a cheaper tripwire than the rejected refactor (optional).** The PR reasonably rejects the "replace vs. push" structural fix as too broad for this bug. A lighter-weight middle option, if you want earlier warning than "a user reports the flicker again": in the fullscreen-pane effect, when a book/chapter move is about to close panes *and* it's still within the same load window (e.g. shortly after boot, no user gesture recorded), `console.warn` in dev. Genuinely optional — the tests already cover the two known paths, and this would only help catch a *new* async URL write nobody wrote a test for. Fine to skip.

### Nit

**No. 3 — Test 2 sets a profile with no `userId`.** `state.login.profile.value = { name: "", config: { translationId: "NIV" } }` represents a state (profile present, user not signed in) that can't occur in production. It's harmless here because the effect under test gates on profile presence rather than auth, so it's a faithful unit-level trigger — but a one-line comment noting "profile is set directly to fire the profile-sync effect; auth state is irrelevant to this path" would save the next reader the same trip through TabsManager I just took. Purely cosmetic.

## Reproduction (for reference)

The tests already encode the two repros. For the saved-translation case, the PR's manual repro is the real-world one: sign in, pick a partial (e.g. NT-only) translation so it persists to your profile, clear site data so the next boot starts on the Genesis 1 default, then load the bare URL — before the fix, Today opens and vanishes; after, it stays.

---

PR #1727, opened 2026-09-01
Tell apart the host disconnecting from your own connection dropping

### LINKED ISSUE #1715: When the host leaves a shared session the toast message can be confusing
Whenever the host leaves a shared session, the message "Reconnecting to the session" is shown. This is confusing wording because it implies that you lost your connection, when in reality the host lost their connection. We should distinguish between when we lost connection (all other users dissapear) compared to when only the host looses connection.

In particular, instead of showing "Disconnected" and "Reconnected" messages, we should show messages specific to the host:

- "The host disconnected from the session"
- "The host reconnected to the session"

Kabi4:
Fixes #1715

## What this changes

A shared session used to say "Reconnecting to the session…" whenever the host went away, which reads as though *you* were the one having trouble. The session now names who actually dropped:

| Message | When |
| --- | --- |
| `The host disconnected from the session` | The host dropped; you are fine |
| `The host reconnected to the session` | The host came back |
| `The host left — you were removed from the session` | Host stayed gone past the 30 s grace |
| `You lost connection to the session` | Your own connection dropped |
| `You rejoined the session` | Your own connection came back |

## Notes for review

**A test's expectation flips in this diff, deliberately.** `shows a you-lost-connection toast instead of a host toast when every other user disappears` asserted that everyone else vanishing meant *our* connection had dropped. That turned out to be wrong, and it was holding a real bug in place: with a host and two guests, closing one guest and then the host told the remaining guest they had lost their connection while it was perfectly healthy. The test is now `blames the host, not our own connection, when the other guest leaves first and the host follows`, and it fails against the previous implementation.

**The self-disconnect path never actually ran before the last few commits.** It required our own entry to still be in the presence list, but a real drop clears the whole list including us, so neither "you lost connection" nor "you rejoined" could ever fire outside of tests that kept self in the list by hand. Several tests modelled a drop that way; they now clear the list, which is what the SDK really does.

**Net effect on the manager is a simplification.** Working out who dropped comes down to two signals — whether we are synced, and whether the presence list still contains us — so `sessionsWhereOtherGuestsWereSeen`, `sessionHasRemoteUsers` and `sessionHasNonHostRemoteUsers` are gone.

## Testing

Unit tests cover each case and the regression tests fail against the prior code. Manually verified on the alpha deployment across a desktop and an Android phone: own drop and recovery, host drop and recovery, host departure and removal, the guest-then-host ordering above, and a host with two devices.

Worth knowing if you test this yourself: DevTools → Network → Offline does **not** exercise any of it, because it leaves an already-open WebSocket connected. Use airplane mode or switch Wi-Fi off to test your own drop, and close the other person's tab to test theirs.

## Known, not fixed here

Three things came out of testing that are not this PR's to solve, filed separately:

- #1819 — someone who drops ungracefully stays visible to everyone else more or less indefinitely, because nothing detects a connection that stopped responding without closing. Pre-existing, and probably belongs in CasualOS.
- #1820 — that person then appears twice if they reload after coming back. Follows from #1819.
- #1829 — toasts replace one another rather than queueing, so the two presence messages above can stomp on each other when they fire close together.

https://claude.ai/code/session_01MKjMpgmrCRx5LPrGjTgMYP


Blitzy:
> [!NOTE]

## Code Review — PR #1727 (fixes #1715)

### Summary

This PR does what issue #1715 asked for: it stops showing the misleading "Reconnecting to the session…" toast when it is the **host** who dropped, and it separates that case from the case where **your own** connection drops. Concretely:

- `session-host-reconnecting` ("Reconnecting to the session…") → renamed to `session-host-disconnected` ("The host disconnected from the session").
- `session-host-reconnected` value changed from "Reconnected to the session" → "The host reconnected to the session".
- Two brand-new keys, `session-disconnected` ("You lost connection to the session") and `session-reconnected` ("You reconnected to the session"), cover the case where *you* drop and everyone else vanishes.

The manager logic in `SeedBibleStateManager.tsx` adds two new presence sets (`sessionsWhereOtherGuestsWereSeen`, `sessionsWhereWeLostConnection`) and an `ourConnectionDropped` heuristic to tell the two failures apart. The implementation is careful, well-commented, and the tests are solid.

**Overall:** In good shape. I found no correctness blockers. The one real discussion point is the translation-convention question below; the rest are low-severity notes.

---

### Verification I ran

- **Tests pass:** `pnpm vitest run test/unit/seed-bible/managers/SeedBibleStateManager.test.ts` → 82 passed.
- **Tests are genuine (not always-green):** I reverted `SeedBibleStateManager.tsx` to `develop` and re-ran the suite. The five new/changed tests — host-disconnected toast, "only the host leaves and other guests remain", "every other user disappears", "other users reappear", and "own sync drops and recovers" — all **fail** against the old implementation and pass against this PR's. They assert on observable output (`currentToast.value.message`, tab presence, `originalDispose`), not internals, and the "host-disconnected" vs "you-lost-connection" tests are mutually discriminating (swapping the logic would fail one or the other). Good regression coverage.
- **Merge state:** GitHub currently shows the PR as not-mergeable ("dirty"), but against the exact base commit the PR reports (`44881639`) the branch merges into `develop` cleanly with **zero conflicts** locally. This looks like GitHub had not finished computing mergeability when the PR opened rather than a real conflict — see Low No. 3.

---

### Issues by severity

#### 🟠 Medium

**No. 1 — All 76 non-English locale files were hand-edited, which the project convention reserves for professionals.**

`CLAUDE.md` states: *"When adding or updating any translation key, only update en.json … don't update other translation files, leave those for the professionals."* This PR writes real translations into every locale (e.g. `es.json` → "Perdiste la conexión con la sesión", `fr.json` → "Vous avez perdu la connexion à la session").

There is a genuine tension worth calling out: the `translation-incomplete-translations` ESLint rule (`script/eslint/i18nIncompleteTranslationsRule.ts`) reports an error for **every** locale that is missing a key present in `en.json`. So once the three new keys were added to `en.json`, `pnpm lint` would fail for all other locales unless the keys were added there too — touching the other files was unavoidable to keep lint green. The convention is really about the *values*, not the *presence* of the keys.

The risk is translation quality: 76 languages of unverifiable (likely machine-generated) strings are being merged as if final, which is exactly what the "leave it to the professionals" rule is trying to avoid.

- **Reproduction:** `git checkout develop -- packages/seed-bible/seed-bible/i18n/` then edit only `en.json` to add the new keys and run `pnpm lint` — it fails with `Locale 'xx' is missing key 'session-disconnected'` for each locale.
- **Suggested fix:** Confirm the intended workflow with the maintainers. The typical way to satisfy the lint rule without pretending to translate is to add the new keys to every locale with the **English string as a placeholder** (so a human translator later replaces them), rather than committing machine translations. If the team is fine keeping these translations, no change is needed — but that decision should be explicit, not implicit in the diff.

#### 🟢 Low

**No. 2 — "You reconnected" can appear without a preceding "You lost connection" toast.**

In `SeedBibleStateManager.tsx`, when `ourConnectionDropped` becomes true the session is always added to `sessionsWhereWeLostConnection`, but the toast is suppressed if `justResumedFromBackground.value` is true:

```ts
if (!sessionsWhereWeLostConnection.has(session.id)) {
  sessionsWhereWeLostConnection.add(session.id);
  if (!justResumedFromBackground.value) {
    toast(t("session-disconnected", …));
  }
}
```

If the drop happens during the post-resume grace window (toast suppressed) and recovery happens after it (`sessionsWhereWeLostConnection.delete(...)` returns true, `justResumedFromBackground` now false), the user sees "You reconnected to the session" without ever having seen "You lost connection". Minor UX oddity, not a correctness bug.

- **Suggested fix:** Only emit the reconnect toast if the disconnect toast was actually shown (e.g. track whether the disconnect toast fired, or skip adding to the set while suppressed). Optional — the team may consider "reconnected" harmless on its own.

**No. 3 — Confirm mergeability before merging.**

As noted above, GitHub reports the PR as not-mergeable, but it merges cleanly against the same base locally. Worth a quick refresh (re-fetch / update-branch) so the reported state reflects reality; I could not reproduce an actual conflict.

**No. 4 — Two-person sessions (just you + the host) still can't distinguish "I dropped" from "the host dropped".**

The `ourConnectionDropped` heuristic relies on either `!session.isSynced.value` or having previously seen *other* guests (`sessionsWhereOtherGuestsWereSeen`). In a session with only you and the host, if your connection degrades in a way that empties the presence list to just yourself **without** `isSynced` flipping to false, the code falls through to the host-disconnect path and would tell you "The host disconnected". In practice this is mitigated because the host-disconnect timer only arms when `sessionPresenceIsTrustworthy(session)` is true (which requires `isSynced`), so a real local drop should set `isSynced=false` and be caught by the first clause. Flagging as a known limitation of the heuristic rather than a defect in this PR — it is inherent to the ambiguity the issue itself describes.

---

### Nits (non-blocking)

- The PR title, "Changed Both Cases message to a genuine message," doesn't describe the change well. Something like "Distinguish host disconnect from own disconnect in shared-session toasts (fixes #1715)" would read better in the history.
- Consider linking the PR to the issue with a closing keyword (e.g. "Closes #1715") in the description so it auto-closes on merge.

---

### What's done well

- The two-different-failures comment block (lines ~1559–1570) clearly explains the heuristic and why it exists — future maintainers will thank you.
- Toast emission is properly de-duplicated (guarded by set membership / `clearPendingHostDisconnect` returning true), so the effect re-running on every signal change won't spam toasts.
- The host-only-drop vs. everyone-vanished test pair genuinely exercises the branch that this PR adds, and both fail on the pre-fix code.

---

KallynGowdy:
I ran into two things while testing:

1. If the host joins with multiple devices and one device disconnects, it will show "You disconnected from the session" on the other device.
2. If you enable airplane mode and then disable airplane mode, when you rejoin the session it still shows "The host rejoined the session" even though it was technically you. In cases where it looks like both us and the host rejoined, it should say "You rejoined the session"

KallynGowdy:
So now for some reason I'm not getting any messages when a user loses connection.

Reproduction steps:

1. Open the Seed Bible on device 1 (desktop) and start a shared session.
2. Join that shared session with device 2 (mobile).
3. Turn on airplane mode on device 2.
4. Wait for a little bit.
5. See that the session avatars leave, but there is no toast message. It should say "You have disconnected from the session" on device 2.
6. Device 1 doesn't need to show a message (because they are the host).
7. Turn off airplane mode on device 2.
8. Wait for a bit.
9. See that the session avatars return and you are back into the session, but again there is no toast message. It should say "You have reconnected to the session"

KallynGowdy:
Investigated @KallynGowdy's repro (no toast on either side of a real connection drop for a 2-person session). Root cause found — it's a real bug in `ourConnectionDropped`, not a flake.

**What's happening:** `SessionsManager.tsx` has this comment on `rebuildRemoteClientsSubscription` (line 1222-1225):

> The OS reports every peer as disconnected when our own connection drops, but on reconnect it silently suppresses the re-sent peer list, so presence would otherwise stay empty forever — **including our own entry**.

So when your own network drops, `connectedUsers` doesn't just lose the other participants — it goes fully empty, self included. But the detection logic in `SeedBibleStateManager.tsx:2097-2103` requires the opposite:

```ts
const weSeeOurselves = session.connectedUsers.value.some(
  (user) => user.isSelf
);
const ourConnectionDropped =
  !sessionWeAreHost(session) &&
  weSeeOurselves &&                                    // <-- never true on a real drop
  !sessionHasRemoteUsers(session) &&
  sessionsWhereHostWasSeen.has(session.id) &&
  (!session.isSynced.value ||
    sessionsWhereOtherGuestsWereSeen.has(session.id));
```

`weSeeOurselves` can never be true at the exact moment a real disconnect happens, since self drops out of the list along with everyone else. So `ourConnectionDropped` never fires, no "you lost connection" toast shows, and `sessionsWhereWeLostConnection` never gets the session added — which is also why no "you rejoined" toast shows on reconnect either (the toast on rejoin is gated on having recorded the drop). It also means `sessionPresenceIsTrustworthy` (`isSynced && weSeeOurselves`) is false the whole time, so the host-disconnect grace timer never arms either — you don't get incorrectly kicked, you just get silence in both directions, exactly as reported.

The existing test at `SeedBibleStateManager.test.ts:917` ("shows you-lost-connection then you-rejoined toasts when this client's own sync drops and recovers") doesn't catch this because it manually sets `session.connectedUsers.value = [selfConnectedUser]` alongside `isSynced.value = false` — i.e. it keeps self in the list, which doesn't match what the real SDK does on a genuine drop.

**Proposed fix:** stop requiring `weSeeOurselves`, since a real drop clears it right along with everyone else:

```ts
const ourConnectionDropped =
  !sessionWeAreHost(session) &&
  sessionsWhereHostWasSeen.has(session.id) &&
  (!session.isSynced.value ||
    (!sessionHasRemoteUsers(session) &&
      sessionsWhereOtherGuestsWereSeen.has(session.id)));
```

This keeps the two independent signals the comment above the effect already describes ("isSynced goes false, and/or every other user disappears at once") as an OR, rather than gating both behind a `weSeeOurselves` check that's only true in the mocked-test world, not in production.

I didn't push this myself — this is core, already-hardened presence logic (grace timers, mobile-resume window, multi-device host handling, the #1346 false-kick fix), and I want a second pair of eyes on the interaction before changing it. Also worth adding a regression test that mirrors the real SDK behavior (clearing self too), since the current one doesn't test the failure mode it was fixing.

---

Kabi4:
i think the bug is already solved

Blitzy:
@KallynGowdy your diagnosis was right, and the silence you reported is fixed. Here is the second pair of eyes you asked for, plus what changed since.

## Your proposed fix went in, then got simplified further

Applied as written first, and it did fix the silence. But manual testing then turned up a case where it gives the wrong answer.

With a host and two guests, close one guest and then the host. The remaining guest is told **"You lost connection to the session"** while their connection is perfectly healthy. The `sessionsWhereOtherGuestsWereSeen` clause is what does it: once another guest has been seen at any point, an empty remote list reads as "my connection dropped" — which is exactly what it looks like when the other people simply leave one after another.

The discriminator was already available and simpler than either version. A real drop clears the whole list, our own entry included, so:

```ts
const ourConnectionDropped =
  !sessionWeAreHost(session) &&
  sessionsWhereHostWasSeen.has(session.id) &&
  (!session.isSynced.value || session.connectedUsers.value.length === 0);
```

If we can still see ourselves, the presence channel is demonstrably working and the others really did leave. Whether they left one at a time or all at once stops mattering, so `sessionsWhereOtherGuestsWereSeen`, `sessionHasRemoteUsers` and `sessionHasNonHostRemoteUsers` are all gone with it — the manager is smaller than before this started.

## A second bug in the same area

Found while in there, independent of the above. The "you rejoined" toast was being suppressed by the five-second post-resume window **even when the matching "you lost connection" toast had already been shown**. The flag was consumed either way, so there was no retry — you were told you had dropped and then never told you were back. On a phone that is the ordinary path: drop, lock the screen, come back as the connection restores.

It now tracks whether the drop was actually announced, and lets that decide the rejoin toast. Recovery also requires positive evidence now — synced *and* a non-empty list — because an empty list right after a resume is presence still catching up, not being back.

## The regression test you asked for

You were right that the existing test was the reason this shipped: it simulated a drop as `[selfConnectedUser]`, keeping self in the list. The self-drop tests now clear the list entirely, which is what the SDK actually does.

Two host-path tests had to change as well, and one of them had been **asserting the guest-then-host bug as correct behaviour** — worth knowing when you read the diff, since a flipped expectation otherwise looks alarming. Both new regression tests fail against the previous implementation.

Full suite is green at 6215 tests, typecheck and lint clean.

## Verified on the alpha build

Desktop plus an Android phone, not just unit tests: your airplane-mode repro in both directions, host drop and recovery, host departure and removal after the grace period, the guest-then-host ordering above, and a host with two devices.

One thing that cost us time and is worth passing on — **DevTools → Network → Offline does not exercise any of this**. It leaves an already-open WebSocket connected, so the session carries on as though nothing happened. Airplane mode or Wi-Fi off to test your own drop; close the other person's tab to test theirs.

## Found but not fixed here

- #1819 — someone who drops ungracefully stays visible to everyone else more or less indefinitely. Nothing in the stack detects a connection that stopped responding without closing: no ping/pong in the client SDK, and `SharedDocument` exposes no awareness channel. Pre-existing, and probably CasualOS's to solve rather than ours.
- #1820 — that person then appears twice if they reload after coming back. Follows from #1819.
- #1829 — toasts replace one another instead of queueing, so when the rejoin and host-disconnected messages land close together one can be cut short or missed entirely. Cosmetic, but it made the messages hard to trust while testing.

---

Blitzy (approved):

PR #1728, opened 2026-09-01
Handled Verse range addition + click along with chapter for future case

### LINKED ISSUE #1713: Make adding verses from your verse selection add a single item with a verse range instead of always multiple items
Currently, when you select a bunch of verses and then add them to your editing playlist, it will add one item for each verse.

Image

Image

### What it should do instead

Ideally, it would add a single item for each range that you have selected. So, Exodus 26:1-11 would get a single item in the example above. If you have multiple ranges selected (like Exodus 26:1-11 and v15-17, then it would add two items, one for the first range and one for the second range.

Kabi4:
Fixes: #1713

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@
 
 ### 🔧 Changed
 
+- Add selected verses to a playlist as one item per contiguous range (for example Exodus 26:1-11) instead of one item per verse.
+
 ### 🐛 Fixed
+
+- Keep the verses a playlist item points at at full brightness when you tap it, instead of dimming the whole chapter including those verses.
 
 ### 🗑️ Removed
 
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
@@ -995,6 +995,13 @@
   .sb-chapter-heading {
     animation: diminish-in-out 3s ease-in-out;
   }
+
+  /* Nested so this beats `.sb-chapter-decoration-diminish .sb-verse` above
+     (equal specificity would otherwise let the dim animation win). */
+  .sb-verse:has(.sb-verse-decoration-diminish) {
+    animation: none;
+    opacity: 1;
+  }
 }
 
 @keyframes diminish-in-out {
--- a/packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/BibleToolsManager.tsx
@@ -33,7 +33,10 @@ import {
   ReadingPlansPaneLeading,
   ReadingPlansPaneTitle,
 } from "../components/ReadingPlansPane/ReadingPlansPane";
-import type { PlaylistManager } from "./PlaylistManager";
+import {
+  groupVersesIntoPlaylistItems,
+  type PlaylistManager,
+} from "./PlaylistManager";
 import type { AnnotationsManager } from "./AnnotationsManager";
 import { i18n, useI18n } from "../i18n";
 import {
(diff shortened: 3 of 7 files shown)


Blitzy:
> [!NOTE]

## Review of No. 1728 — Verse-range addition to playlists (fixes No. 1713)

**Overall:** This is a clean, well-tested implementation that solves exactly what issue No. 1713 asked for. Selecting `Exodus 26:1-11` now adds **one** range item instead of eleven, and a gapped selection (`1-11` and `15-17`) adds two items — matching the issue's stated expectation. I ran the suite locally: the three touched test files pass (218 tests) and `pnpm check:ts` is clean. The new `groupVersesIntoPlaylistItems` tests are thorough and test real behavior, and the `add-to-playlist` tests fail on the pre-fix code (which produced one item per verse), so they're genuine regression tests, not always-green ones.

I found **no correctness bugs reachable in the current UI**. The notes below are one forward-looking limitation and two minor cleanups.

---

### 🟡 Low — Cross-chapter grouping only collapses when the boundary's earlier chapter is the active one

`BibleToolsManager` supplies `groupVersesIntoPlaylistItems` a `verseCountOf` that only knows the **currently-loaded** chapter's verse count:

```ts
(bookId, chapter) =>
  chapterData?.book.id === bookId &&
  chapterData.chapter.number === chapter
    ? chapterData.numberOfVerses
    : undefined
```

`isContinuousPlaylistSelection` needs `verseCountOf(prev.bookId, prev.chapter)` to fuse a run that ends on a chapter's last verse into verse 1 of the next chapter. Because only the active chapter has a known count, a contiguous selection crossing a chapter boundary would collapse into one range only when the **earlier** chapter is the active one, and split into two adjacent items otherwise.

**Why this is Low, not a blocker:** the reader keeps a single active `chapterData` (verse selection is bound to it; the next chapter is only a non-selectable preview), so a cross-chapter selection isn't reachable in today's UI — this is the "future case" the PR title calls out. The failure mode is also graceful: you'd get two adjacent range items rather than wrong data. Worth flagging so it isn't a surprise when whole-chapter / multi-chapter selection gets wired up.

**Suggested fix (for when that lands):** source verse counts from `BibleDataManager` for an arbitrary `(bookId, chapter)` rather than only the active chapter, so cross-chapter fusion is deterministic regardless of which chapter is on screen. Note the sibling `add-to-reading-plan` tool sidesteps this entirely by taking first→last of the whole selection — a different trade-off (it never splits on gaps), so the two tools intentionally differ here.

### 🔵 Nit — The pre-existing top-level CSS rule is now fully superseded

The new nested rule fixes the dim-through bug correctly (higher specificity beats `.sb-chapter-decoration-diminish .sb-verse`). But the pre-existing rule in `BibleReader.inline.css` (~line 576):

```css
.sb-verse:has(.sb-verse-decoration-diminish) { animation: none; opacity: 1; }
```

ties at `(0,2,0)` with the chapter-dim rule and loses on source order — which is precisely the bug being fixed, i.e. it was already never winning. Since `sb-verse-decoration-diminish` is always applied together with the `sb-chapter-decoration-diminish` container class, the new rule now covers every case the old one aimed at. Not introduced by this PR and harmless, but you could delete the old rule to avoid two rules that read as if they do the same thing. Your call — leaving it is defensible as defensive CSS.

### 🔵 Nit — The CSS fix has no automated coverage (expected)

The brightness behavior is pure CSS/animation, so a unit test isn't practical — no action needed on that. Recommend a quick manual pass on the alpha deploy (https://alpha.seedbible.org/b/feature_1713) to confirm the fix visually.

**Repro to verify the fix:**
1. Open a playlist for editing and add a verse-range item (e.g. `Exodus 26:1-11`).
2. Tap that item to play it.
3. **Before:** the whole chapter dims, including verses 1-11. **After:** verses 1-11 stay at full brightness while the rest of the chapter dims. ✅

---

Nice work — the grouping algorithm's edge-case coverage (reverse selection, dedup, single-chapter books, adjacent-book/adjacent-psalm false positives, the "must merge / must not merge" tables) is especially good and is what gives confidence the range logic is right.

---

PR #1729, opened 2026-09-01
Fix CSS specificity issue in chapter decoration diminish rule

KallynGowdy:
## Summary

Fixed a CSS specificity bug where decorated verses were being incorrectly dimmed by the chapter decoration animation, defeating the purpose of the verse-level decoration.

## The Problem

The `.sb-chapter-decoration-diminish` rule was applying a `diminish-in-out` animation to all `.sb-verse` elements. A later rule tried to override this for verses with `.sb-verse-decoration-diminish` using `:has()`, but both selectors had equal specificity (two classes each). In CSS, when specificity ties, the later-declared rule wins — which meant the override rule would always win, but only by accident of declaration order. This created a fragile situation where the very verses meant to stay bright were being dimmed instead.

## The Fix

Changed the selector from `.sb-verse` to `.sb-verse:not(:has(.sb-verse-decoration-diminish))`. This explicitly excludes decorated verses from the animation at the point of declaration, rather than relying on a later override rule to win a specificity tie.

**Before:**
```css
.sb-chapter-decoration-diminish {
  .sb-verse,
  .sb-chapter-heading {
    animation: diminish-in-out 3s ease-in-out;
  }
}
```

**After:**
```css
.sb-chapter-decoration-diminish {
  .sb-verse:not(:has(.sb-verse-decoration-diminish)),
  .sb-chapter-heading {
    animation: diminish-in-out 3s ease-in-out;
  }
}
```

## Implementation Details

- Added a detailed comment explaining why the `:not(:has())` approach is necessary, to prevent future maintainers from "simplifying" this back to the buggy version
- The fix ensures decorated verses remain bright during the chapter decoration animation, as intended

https://claude.ai/code/session_01Xyx28saQvm6N57aSvUH4ew

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
@@ -991,7 +991,13 @@
 }
 
 .sb-chapter-decoration-diminish {
-  .sb-verse,
+  /* Excluding the decorated verses here (rather than relying on the
+     `.sb-verse:has(.sb-verse-decoration-diminish)` override above to win)
+     matters because both selectors have equal specificity (two classes) —
+     a tie the later-declared rule would otherwise win regardless of which
+     one is meant to take precedence, silently dimming the very verses this
+     decoration exists to keep bright. */
+  .sb-verse:not(:has(.sb-verse-decoration-diminish)),
   .sb-chapter-heading {
     animation: diminish-in-out 3s ease-in-out;
   }


Blitzy:
> [!NOTE]

## Code Review — PR #1729

**Verdict:** The code change is correct, minimal, and low-risk. It genuinely fixes the described bug. I found no correctness issues in the CSS. The notes below are documentation accuracy, test coverage, and one optional cleanup — none block merge.

### What the change does (verified)

I traced the two competing rules and the DOM they apply to:

- `.sb-verse` is the outer verse span; the diminish decoration class (`.sb-verse-decoration-diminish`) lands on the inner `.sb-verse-decorator` span (`BibleReader.tsx` around lines 888–900), so it *is* a descendant of `.sb-verse` — which is what makes `:has(.sb-verse-decoration-diminish)` match. Good.
- **Before:** `.sb-chapter-decoration-diminish .sb-verse` (specificity `0,2,0`, declared line 993) tied with the override `.sb-verse:has(.sb-verse-decoration-diminish)` (specificity `0,2,0`, declared line 576). On a tie the **later-declared** rule wins, and the dimming rule is declared later — so a decorated verse got `diminish-in-out` and was dimmed. Confirmed bug.
- **After:** `.sb-verse:not(:has(.sb-verse-decoration-diminish))` simply stops matching decorated verses, so they only ever pick up the line-576 `animation: none; opacity: 1`. Non-decorated verses still match and still dim. Correct.

The `:not(:has())` selector relies on `:has()`, which is already an established pattern in this stylesheet (line 576 and elsewhere), so it introduces no new browser-support concern.

### Issues by severity

**Low — PR description is self-contradictory and mislabels the rule order (docs only, code is fine)**

The "The Problem" section says the `:has()` override *"would always win, but only by accident of declaration order"* and in the same paragraph says *"the very verses meant to stay bright were being dimmed instead."* Those two statements contradict each other — if the verses were being dimmed, the override was **losing**, not winning. It also calls the `:has()` override *"a later rule,"* but the override is declared **earlier** (line 576) than the dimming rule (line 993); that earlier-vs-later ordering is the whole reason the override lost the tie.

Worth fixing because the reasoning is the main artifact a future maintainer will read before touching this. Note the **inline code comment is accurate** — it correctly says "the override above" and "a tie the later-declared rule would otherwise win… silently dimming" — so only the PR body needs the correction.

- Repro: read the "The Problem" paragraph against the actual line numbers in the diff.
- Suggested fix: reword to "The override rule (declared earlier, at line 576) *lost* the specificity tie to the later-declared chapter-dimming rule, so decorated verses were dimmed."

**Low / Observation — no automated test (acceptable here, but calling it out)**

There is no test. That is defensible: the bug lives entirely in the CSS cascade, and this repo's test stack is Vitest + jsdom, which does not compute animations or resolve stylesheet specificity — so a real regression test would need a browser (Playwright), which the project doesn't currently run for this. Importantly, a jsdom-level test that just asserts the rendered class string contains `.sb-verse-decoration-diminish`, or that the CSS text contains `:not(:has(...))`, would pass regardless of whether the cascade actually keeps the verse bright — a tautology that "always succeeds," exactly the kind of test to avoid. So **no test is better than a fake one** here. If verse-range deep-link dimming regresses again in the future, the right home for a guard is a Playwright check on computed opacity of a decorated verse inside a `.sb-chapter-decoration-diminish` container, not a jsdom unit test.

**Nit — the line-576 override is now belt-and-suspenders (optional, fine to keep)**

Since decorated verses no longer match the dimming rule at all, the `&:has(.sb-verse-decoration-diminish) { animation: none; opacity: 1; }` block at line 576 is no longer strictly required to defeat this particular rule. It's harmless, defensive against any other animation that might target `.sb-verse`, and removing it risks reintroducing subtle issues — so I'd **leave it as-is**. Mentioning only so it isn't mistaken for load-bearing against this specific rule.

### Manual verification steps for a human reviewer

1. Open a chapter and trigger a verse-range decoration that applies `sb-chapter-decoration-diminish` to the chapter and `sb-verse-decoration-diminish` to the target verse(s) — e.g. a verse-range deep link, or the "Today" / search-result / compare flows that call it (`TodayManager.tsx`, `SidebarSearch.tsx`, `BibleReadingManager.tsx`).
2. Confirm the surrounding (non-targeted) verses briefly dim via the `diminish-in-out` animation while the **targeted** verses stay at full opacity.
3. On `develop` (pre-fix) the targeted verses dim along with everything else; on this branch they stay bright.

---

PR #1731, opened 2026-09-01
Prevent hydration when SSR and client bundles differ

KallynGowdy:
## Summary

This change adds a build identity check to the hydration gate. When a page's DOM was rendered by a different branch's (or commit's) SSR bundle than the one about to hydrate onto it, hydration is now declined. This prevents subtle bugs where Preact's hydration would patch mismatched component code onto the existing markup.

## Problem

The host server can render pages through a fallback branch's bundle when the requested branch isn't in the `ALLOWED_SSR_BRANCHES` whitelist (see `server/index.ts`). The URL might match exactly, but the DOM was produced by different component code. Preact's `hydrate()` only diffs tree shape, not component code identity, so it would happily patch onto markup a different version of the app — leading to runtime errors or silent behavioral mismatches.

## Changes

- **`appConfig.tsx`**: Added `renderedByCommit?: string` field to `AppConfig`. This stores the SSR bundle's own `__GIT_COMMIT__` build constant (not derived from the request URL). Absent for older server builds, treated as "no mismatch" to avoid blocking hydration on pure version skew.

- **`hydrationGate.ts`**: 
  - Added `clientCommit: string` to `HydrationGateContext` — the client bundle's own `__GIT_COMMIT__`.
  - Added `"build-mismatch"` as a new `HydrationDeclineReason`.
  - Added a check: if `config.renderedByCommit` is defined and differs from `clientCommit`, decline hydration with reason `"build-mismatch"`.

- **`entry-ssr.tsx`**: When building the `clientConfig` for SSR, set `renderedByCommit: __GIT_COMMIT__` to record which bundle produced the HTML.

- **`init.tsx`**: Pass `clientCommit: __GIT_COMMIT__` when calling `decideHydration()` in the client initialization.

- **`hydration.test.tsx`**: Updated all existing test calls to include `clientCommit: __GIT_COMMIT__`. Added two new tests:
  - One that verifies hydration is declined when `renderedByCommit` differs from `clientCommit` (the reported bug scenario).
  - One that verifies hydration succeeds when both commits match.

## Implementation Details

The check is conservative: if `renderedByCommit` is absent (older server build), it's treated as a match rather than a decline. This allows gradual rollout — old servers won't block new clients, and new servers won't block old clients, only mismatches between different branches' bundles are caught.

https://claude.ai/code/session_01Gr2KJ8VwAM3ECHZQnqR7Bc

### DIFF
--- a/packages/seed-bible/seed-bible/app/appConfig.tsx
+++ b/packages/seed-bible/seed-bible/app/appConfig.tsx
@@ -81,10 +81,13 @@ export interface AppConfig {
    * gate (`app/hydrationGate.ts`) compares this against the client's own
    * `__GIT_COMMIT__` and refuses to hydrate on a mismatch.
    *
-   * Absent for the same reasons `renderedForPath` can be absent, and (like
-   * `ssrChapterContentSettled`) treated as "no mismatch" when absent so an
-   * older server build that predates this field doesn't block hydration on
-   * pure version skew.
+   * Can also be absent for the same reasons `renderedForPath` can be
+   * absent — an older server build that predates this field, for instance.
+   * Unlike `renderedForPath`/`ssrChapterContentSettled`, an absent value
+   * here is treated as a mismatch rather than given the benefit of the
+   * doubt: an SSR document with no verifiable build identity is exactly the
+   * case this field exists to catch, so the gate declines to hydrate onto
+   * it.
    */
   renderedByCommit?: string;
 
--- a/packages/seed-bible/seed-bible/app/appConfig.tsx
+++ b/packages/seed-bible/seed-bible/app/appConfig.tsx
@@ -68,24 +68,25 @@ export interface AppConfig {
   renderedForPath?: string;
 
   /**
-   * The branch whose SSR bundle actually produced this HTML — that bundle's
-   * own `__DEPLOY_BRANCH__` build constant (see `vite.config.ts`), not
-   * anything derived from the request URL.
+   * The commit whose SSR bundle actually produced this HTML — that bundle's
+   * own `__GIT_COMMIT__` build constant, not anything derived from the
+   * request URL.
    *
-   * Ordinarily equal to the requested branch. Differs only when the
-   * requested branch isn't in the host server's `ALLOWED_SSR_BRANCHES`
-   * whitelist and `DEFAULT_SSR_BRANCH` rendered the page instead (see
-   * `server/index.ts`) — the DOM was then built by a different branch's
-   * component code than the one the client is about to hydrate with. The
-   * hydration gate (`app/hydrationGate.ts`) compares this against the
-   * client's own `__DEPLOY_BRANCH__` and refuses to hydrate on a mismatch.
+   * Ordinarily equal to the requested branch's own latest commit. Differs
+   * when the requested branch isn't in the host server's
+   * `ALLOWED_SSR_BRANCHES` whitelist and `DEFAULT_SSR_BRANCH` rendered the
+   * page instead (see `server/index.ts`) — the DOM was then built by a
+   * different branch's (and near-certainly different commit's) component
+   * code than the one the client is about to hydrate with. The hydration
+   * gate (`app/hydrationGate.ts`) compares this against the client's own
+   * `__GIT_COMMIT__` and refuses to hydrate on a mismatch.
    *
    * Absent for the same reasons `renderedForPath` can be absent, and (like
    * `ssrChapterContentSettled`) treated as "no mismatch" when absent so an
    * older server build that predates this field doesn't block hydration on
    * pure version skew.
    */
-  renderedByBranch?: string;
+  renderedByCommit?: string;
 
   /**
    * False only when the SSR-only initial-chapter-load timeout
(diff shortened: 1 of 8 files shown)


Blitzy:
> [!NOTE]

## Review — Prevent hydration when SSR and client bundles differ

**Verdict: looks correct and safe to merge.** No blocking issues. The change is small, conservative, and does what it claims. I traced the mechanism end-to-end and verified the new tests actually catch the bug (details below). The only findings are two minor test-coverage gaps.

No linked issue is referenced in the PR, and the only PR comment is the deployment bot, so there was no prior conversation to reconcile against.

### How I verified the mechanism

The fix hinges on `clientCommit` (from the client bundle) differing from `config.renderedByCommit` (from the SSR bundle) **exactly** in the buggy scenario and matching everywhere else. That holds:

- `__GIT_COMMIT__` is baked at build time (`vite.config.ts` → `resolveGitCommit()`), so a single `pnpm build`'s SSR bundle and client bundle always share the same commit. Within one deployment they can never diverge.
- In the fallback path (`server/index.ts` ~L770–786), the **requested** branch's pre-rendered HTML — which carries that branch's own client `<script>` tags, so `clientCommit` = branch B — is rendered through the **default** branch's `render()`, which substitutes `<!-- CONFIG_JSON -->` with `renderedByCommit` = branch A (`entry-ssr.tsx` L528–535). A ≠ B → `build-mismatch` → full `render()`. Exactly the intended catch.
- No false declines: asset URLs are namespaced by build id (`vite.config.ts` L34–37), so a page's HTML always loads its own matching client bundle. The markup and the client bundle only diverge in the cross-branch fallback case, which is precisely what this gate targets. A service-worker / deploy-skew false positive isn't possible for that reason.

### Tests

Both new tests exercise real behavior, and I confirmed the primary one is a true regression test. I temporarily removed the `build-mismatch` block in `hydrationGate.ts` and re-ran the suite: the test **"declines to hydrate when the SSR HTML was rendered by a different commit's bundle"** went red (pre-fix code returns `{ hydrate: true }` because the URL matches and `ssrChapterContentSettled` defaults to `true`). So it genuinely guards the fix.

The second test (**"hydrates when the rendering commit matches"**) passes with or without the check, so on its own it is not a regression test — but it does still catch an inverted comparison (`===` instead of `!==`), so it earns its place.

`pnpm check:ts` is clean, all 14 hydration tests pass, and CI `build (24.x)` is green.

---

### Findings by severity

**No High / Medium issues.**

#### Low — No. 1: The SSR-side injection of `renderedByCommit` is not asserted anywhere

`entry-ssr.tsx` now sets `renderedByCommit: __GIT_COMMIT__` on the injected config, but `test/unit/standalone/entry-ssr.test.ts` only asserts `renderedForPath` and `ssrChapterContentSettled` (see the test at ~L492, "injects the exact request path as renderedForPath..."). If that line were dropped or renamed, nothing would fail. The whole gate is inert if the field never reaches the config, so it's worth pinning down on the producing side.

_Suggested fix_ — extend the existing renderedForPath test:

```ts
const config = JSON.parse(injected as string) as {
  renderedForPath: string;
  renderedByCommit: string;
  ssrChapterContentSettled: boolean;
};
expect(config.renderedForPath).toBe(path);
expect(config.renderedByCommit).toBe(__GIT_COMMIT__);
expect(config.ssrChapterContentSettled).toBe(true);
```

#### Low — No. 2: The "absent `renderedByCommit` ⇒ still hydrate" backward-compat path has no direct test

The gate deliberately treats an absent `renderedByCommit` as "no mismatch" so an older server build doesn't block hydration (`hydrationGate.ts` L70–75, and the `AppConfig` doc comment). Every existing `hydrate: true` test, though, gets its config from `ssrRender`, which now always sets `renderedByCommit` — so `renderedByCommit` is defined-and-equal in all of them, never `undefined`.

The consequence: if someone later dropped the `config.renderedByCommit !== undefined` guard (leaving just `config.renderedByCommit !== clientCommit`), an old-server config (field absent) with a defined `clientCommit` would start declining with `build-mismatch`, and no test would catch the regression.

_Reproduction (for a human dev):_ delete `config.renderedByCommit !== undefined &&` from the gate, run `pnpm vitest run test/unit/standalone/hydration.test.tsx` — still green, despite the backward-compat contract now being broken.

_Suggested fix_ — add a small gate test:

```ts
it("hydrates when renderedByCommit is absent (older server build, no build identity)", () => {
  const container = document.createElement("div");
  container.innerHTML = "<div>Verse 1</div>";
  const config = {
    ...DEFAULT_APP_CONFIG,
    renderedForPath: "/en/AAB/genesis/1",
    ssrChapterContentSettled: true,
    // renderedByCommit intentionally omitted
  };
  const decision = decideHydration({
    config,
    pathname: "/en/AAB/genesis/1",
    search: "",
    container,
    clientCommit: "commit-feature-branch-def456",
  });
  expect(decision).toEqual({ hydrate: true });
});
```

---

### Nits (optional)

- The new comment in `hydrationGate.ts` reads "...patch onto markup a different version of the app produced." — the trailing "produced" seems like a stray word; "...patch a different version of the app onto that markup" would read cleaner. Purely cosmetic.

Nice, tightly-scoped fix — the conservative "absent ⇒ match" handling and the commit-vs-commit (rather than branch-name) comparison are the right calls for gradual rollout across old/new server and client builds.

---

PR #1732, opened 2026-09-01
feat: Cache one shared app shell for HTML instead of per-URL entries

KallynGowdy:
Switches the HTML route to StaleWhileRevalidate against a single cache entry shared by every navigation, so a service-worker-controlled visit loads instantly with no network round-trip, and refreshes in the background for next time. Uncontrolled requests (first visit, no SW) are unaffected and still hydrate() against exact per-path SSR HTML.

### DIFF
--- a/standalone/sw.ts
+++ b/standalone/sw.ts
@@ -24,7 +24,7 @@
  *  | What                                   | Strategy                        |
  *  |----------------------------------------|---------------------------------|
  *  | Core assets (entry JS, vendor, CSS, …) | Precache (install time)         |
- *  | Page HTML (any non-`/b/` navigation)   | NetworkFirst, 3s timeout        |
+ *  | Page HTML (any non-`/b/` navigation)   | StaleWhileRevalidate, one shared shell entry |
  *  | Other assets of *this* build           | CacheFirst                      |
  *  | Google Fonts / Fontshare stylesheets   | StaleWhileRevalidate            |
  *  | Google Fonts / Fontshare font files    | CacheFirst                      |
@@ -39,12 +39,12 @@ import {
   type PrecacheEntry,
 } from "workbox-precaching";
 import { registerRoute } from "workbox-routing";
+import { CacheFirst, StaleWhileRevalidate } from "workbox-strategies";
 import {
-  CacheFirst,
-  NetworkFirst,
-  StaleWhileRevalidate,
-} from "workbox-strategies";
-import { isAppShellNavigation, isCacheableStaticAsset } from "./swRouting";
+  getAppShellCacheKey,
+  isAppShellNavigation,
+  isCacheableStaticAsset,
+} from "./swRouting";
 
 // `declare let` (rather than `const`) so this shadows the `self: WorkerGlobalScope`
 // that lib.webworker declares, instead of colliding with it.
@@ -69,6 +69,22 @@ const FONT_FILE_CACHE = "seed-bible-font-files";
 
 /** Absolute form of `__ASSET_BASE_URL__`, so it can be compared against request URLs. */
 const ASSET_BASE_HREF = new URL(__ASSET_BASE_URL__, self.location.href).href;
+
+/**
+ * Fixed cache key every navigation's HTML is stored/read under, regardless of
+ * which path was actually requested. See the comment above the HTML route
+ * below for why one shared entry — not one per path — is the goal now.
+ */
+const APP_SHELL_URL = getAppShellCacheKey(self.location.origin);
+
+/**
+ * Rewrites this route's cache reads/writes to the single `APP_SHELL_URL` key
+ * instead of Workbox's default (the request URL), so every navigation shares
+ * one cache entry.
+ */
+const appShellCacheKeyPlugin = {
+  cacheKeyWillBeUsed: async () => APP_SHELL_URL,
+};
 
 /**
  * Origins the fonts come from. The stylesheet hosts are fetched as CSS; the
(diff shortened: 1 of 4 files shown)


Blitzy:
> [!NOTE]

## Review: Cache one shared app shell for HTML instead of per-URL entries

Thanks for the very thorough comments — they made this easy to follow. Overall this is a clean, well-reasoned change and I found **no correctness bugs**. The shared-key wiring is done carefully: `warmAppShellCache()` was updated to write under `APP_SHELL_URL` (not the response URL) to stay consistent with the route, and the `hydrate()` → `render()` fallback for a mismatched shell already exists and is exercised by the existing gate machinery. The findings below are one test-quality issue and a couple of behavioral things worth confirming rather than blockers.

I also checked the one interaction I was worried about — whether `ExpirationPlugin({ maxEntries: 1 })` would thrash because it tracks per-path request URLs while the cache holds a single shared entry. It does **not**: Workbox applies `cacheKeyWillBeUsed` first and passes the rewritten `effectiveRequest` (the shared `APP_SHELL_URL`) to `cacheDidUpdate`, so the expiration bookkeeping and the cache both key on the one shared entry. Confirmed against `workbox-strategies@7.4.1` `StrategyHandler.cachePut`. Good.

---

### Medium

**1. The `getAppShellCacheKey()` "returns the same key regardless of which path" test can't fail.**

`test/unit/standalone/swRouting.test.ts`:

```ts
it("returns the same key regardless of which path is being cached", () => {
  expect(getAppShellCacheKey(ORIGIN)).toBe(getAppShellCacheKey(ORIGIN));
});
```

`getAppShellCacheKey(origin)` takes only an origin — it has no path parameter — and is a pure `new URL(APP_SHELL_CACHE_PATH, origin).href`. So the assertion is `f(x) === f(x)`, which is true for any deterministic function and can only fail if the function becomes non-deterministic (it can't). The test name promises path-invariance, but since path isn't even an input the test is structurally incapable of catching a path-related regression — it's the kind of test that always succeeds. The second test in the same block (`toContain(ORIGIN)` + differs across origins) is genuinely useful; it's only this first one that's hollow.

Suggested replacement — assert the concrete value and the property that actually matters (the key is a fixed synthetic path that can't collide with a real reading URL):

```ts
it("keys the shell under a fixed synthetic path, scoped to the origin", () => {
  expect(getAppShellCacheKey(ORIGIN)).toBe(`${ORIGIN}/__app-shell`);
});
```

That version fails if the constant or the origin-scoping ever changes, which is the behavior the shared-shell design actually depends on.

_Reproduce:_ open `standalone/swRouting.ts` and change `APP_SHELL_CACHE_PATH` to any other string, or hardcode `getAppShellCacheKey` to ignore its argument and return a constant — the existing first test still passes; the suggested one fails.

---

### Low / worth confirming (not blockers)

**2. After a deploy, the first controlled load now runs the *previous* shell.**

Under the old `NetworkFirst` (3s), an online returning visitor got the fresh, current-build HTML on load. Under `StaleWhileRevalidate`, every controlled navigation serves the cached shell first and only revalidates in the background, so the load right after a deploy renders from the *previous* shell (with the previous build's asset references) and the fresh shell only takes effect on the *next* load. Online this is harmless because content-hashed assets are still fetchable from the asset host. The case worth confirming is offline-immediately-after-an-update: `registerType: "autoUpdate"` + `skipWaiting()`/`clientsClaim()` activate the new worker eagerly, and `cleanupOutdatedCaches()` drops the old precache — so a stale shell can reference entry/vendor chunks that are no longer precached. If the visitor then goes offline before the shell is revalidated, that load can't fetch those chunks. This is largely a pre-existing edge (the old strategy also served a stale shell when offline), and it may be perfectly acceptable for this app — just flagging that SWR widens the window in which the shell and the active build can disagree.

**3. Coverage gap: the actual behavior change isn't unit-tested.**

The one new test only covers the pure key helper. The parts this PR actually changes — the `cacheKeyWillBeUsed` rewrite that funnels every path to one entry, and `warmAppShellCache()` now writing under `APP_SHELL_URL` — aren't exercised by any test. That mirrors the pre-existing situation (the SW route was never unit-tested, since it needs a worker/Workbox environment), so it's not a regression and I wouldn't block on it. If you wanted to close it, `appShellCacheKeyPlugin.cacheKeyWillBeUsed` is a trivially testable pure-ish function (`await plugin.cacheKeyWillBeUsed({ request: new Request("https://seedbible.org/AAB/genesis/10") })` should equal `getAppShellCacheKey(ORIGIN)`), which would lock in the "every path collapses to one key" invariant that No. 1's test only gestures at.

---

### Nit

**4. Comment slightly overstates plugin ordering.** The route comment says `appShellCacheKeyPlugin` is placed first "so they see (and gate) that same key." `CacheableResponsePlugin.cacheWillUpdate` gates purely on response status and never looks at the cache key, so ordering relative to it doesn't affect gating. Ordering *does* matter for `ExpirationPlugin` (its bookkeeping keys off the effective request), so the placement is correct — the "gate" wording just implies more than it needs to. Cosmetic only.

---

_CI note: `build (24.x)` was still in progress at review time — worth a glance before merge, though this change is comment-heavy and low-risk._

---

KallynGowdy:
Addressed the review from [this comment](https://github.com/HelloAOLab/seed-bible/pull/1732#issuecomment-5496717471) in e4018846:

**#2 (the main ask) — post-deploy staleness while offline.** `self.skipWaiting()` is no longer called unconditionally. It now only fires once `warmAppShellCache()` confirms the shared shell entry was actually refreshed to match the build that's installing. If that fetch fails (offline, flaky connection), the *previous* worker stays active — its own shell and precache are still mutually consistent, so offline continues to work exactly as before the deploy — instead of activating and letting `cleanupOutdatedCaches()` drop chunks that a still-stale shell references. The browser's normal update polling retries the install (and gets another chance to activate) the next time it's online. This closes most of the window the review flagged, rather than just narrowing it.

**#1 — tautological test.** Replaced `getAppShellCacheKey(ORIGIN) === getAppShellCacheKey(ORIGIN)` with an assertion against the concrete, hardcoded path (`${ORIGIN}/__app-shell`, not built from the `APP_SHELL_CACHE_PATH` constant it's checking), so it actually fails if the key's shape changes. Verified locally by temporarily changing the constant and confirming the new test fails where the old one didn't.

**#4 (nit) — plugin-ordering comment.** Reworded to say `ExpirationPlugin`'s bookkeeping is what depends on cache-key ordering; `CacheableResponsePlugin` only looks at response status.

**#3 (coverage gap)** — left as-is per your note that it mirrors a pre-existing gap and isn't worth blocking on; closing it would mean exporting `appShellCacheKeyPlugin` out of `sw.ts` purely for testability, which felt disproportionate to the ask.

Ran `pnpm test` (5271 tests), `pnpm check:ts`, and `pnpm lint` — all clean.

---

PR #1733, opened 2026-09-01
Defer Today's URL-derived open state until after mount

Blitzy:
## Summary

Fixes a hydration bug where loading (or refreshing) a URL that opens the Today screen — e.g. one with `?today=open`, or a bare `/` with no reading position — could corrupt the page: a broken bottom toolbar, a stray leftover chunk of chapter text, an unexpected extra selector panel.

## Problem

The app renders on the server first (fast first paint), then the client "attaches" to that markup instead of rebuilding it, on the assumption that the client would have built the exact same thing. `TodayManager`'s `isOpen` broke that assumption: SSR always forces it `false` (Today closed — deliberately, so a crawler indexing a cold `/` request sees real scripture instead of the Welcome screen), while the client computed its *initial* value from `todayWillAutoOpenForUrl(...)` reading the live URL. Whenever those disagreed, the client's first attach pass fought a structurally different SSR document instead of matching it, and Preact's `hydrate()` doesn't diff structure the way a full re-render does — it just silently patches, incorrectly.

## Changes

- **`managers/TodayManager.tsx`**: `isOpen` now seeds `false` unconditionally — client and server alike — instead of branching on `import.meta.env.SSR`. The real URL-derived value, and the two-way URL↔`isOpen` sync (previously wired up at construction time), are now applied by a new idempotent `hydrateAutoOpen()` method instead.
- **`app/main.tsx`**: calls `state.today.hydrateAutoOpen()` from a post-mount `useEffect(() => {...}, [])`, alongside the three existing effects that defer other SSR-seeded state (viewport, local settings, saved tabs) until right after the first commit.
- **`test/unit/seed-bible/testUtils/createTestSeedBibleState.ts`**: the test fixture now calls `hydrateAutoOpen()` too, mirroring `hydrateLocalConfig()`/`hydrateFromStorage()`, so it continues to represent a fully-loaded app.

## Alternatives considered

- **Decline hydration whenever the mismatch is detected** (fall back to a full `render()`): fixes the corruption, but trades away the fast-attach path for that page load, causing a visible flash. Ruled out — not acceptable to lose hydration for this.
- **Have SSR render Today open for an explicit `?today=open`**: no flash at all, but the "bare `/` also opens Today" case is trickier — SSR deliberately never opens it there for SEO, and would need care not to regress that. More invasive than necessary here.
- **What shipped**: seed both sides to the same safe default (`false`) so there's never a disagreement, then apply the real value one render later — the same pattern already used for restoring saved tabs (`TabsManager.hydrateStoredTabs`), local settings, and viewport size. No flash, no lost hydration, no SEO risk, and it reuses an established pattern instead of introducing a new one.

## Test plan

- [x] `TodayManager.test.ts`, `todayCoversReader.test.ts`, `todayStaysOpenOnBoot.test.ts` — all passing
- [x] Full suite (228 files, 5336 tests) — passing
- [x] `pnpm eslint` / `pnpm check:ts:client` — clean

### DIFF
--- a/packages/seed-bible/seed-bible/app/main.tsx
+++ b/packages/seed-bible/seed-bible/app/main.tsx
@@ -161,6 +161,15 @@ function MainBody({
     state.app.hydrateFromStorage();
   }, []);
 
+  // Deferred real read, same reason as the three above: `isOpen` seeds
+  // `false` to match SSR (which always renders Today closed, for crawler
+  // SEO — see `TodayManager`), so the first hydrate pass can't disagree with
+  // it. Apply the URL's real open/closed state, and start the live URL sync,
+  // once right after mount.
+  useEffect(() => {
+    state.today.hydrateAutoOpen();
+  }, []);
+
   if (typeof document !== "undefined") {
     useSignalEffect(() => {
       document.title = state.app.title.value;
--- a/packages/seed-bible/seed-bible/managers/TodayManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/TodayManager.tsx
@@ -121,6 +121,18 @@ export interface TodayManager {
   getTranslationBooks: (translation: string) => Promise<TranslationBooks>;
   open: () => void;
   close: () => void;
+  /**
+   * Applies the URL's real open/closed state and starts the live URL <->
+   * `isOpen` sync.
+   *
+   * `isOpen` seeds to `false` unconditionally, on the client as well as
+   * during SSR, so the first hydrate pass can't disagree with the server
+   * over whether Today is showing. Call this once from a post-mount effect
+   * (see `MainBody` in `app/main.tsx`) so the real value -- and the two-way
+   * URL sync -- arrive as a normal diffed re-render instead, the same
+   * pattern `TabsManager.hydrateStoredTabs` uses for saved tabs. Idempotent.
+   */
+  hydrateAutoOpen: () => void;
   /** Tears down the internal effects. The app never calls this; tests do. */
   dispose: () => void;
 }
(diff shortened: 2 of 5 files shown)


Blitzy:
> [!NOTE]

## Code review — Defer Today's URL-derived open state until after mount

**Verdict:** The fix itself is correct and well-reasoned. Seeding `isOpen` to `false` on both server and client, then applying the real URL-derived value from a post-mount effect, is the right shape and it mirrors the established `hydrateFromStorage` / `applyViewport` / `hydrateLocalConfig` pattern faithfully. The idempotent guard, the deferred URL sync, and the `disposeUrlSync` handoff all hold up. CI is green and I re-ran `TodayManager.test.ts`, `todayStaysOpenOnBoot.test.ts`, `todayCoversReader.test.ts`, and `hydration.test.tsx` locally — 32 passing.

The one substantive issue is in the **tests**, not the fix: nothing added or existing actually reproduces the hydration bug this PR closes.

---

### Medium — no test fails on the pre-fix code (the regression is unguarded)

**What's wrong.** The bug this PR fixes is a client/server hydration *mismatch*: the client used to seed `isOpen` from the live URL (`todayWillAutoOpenForUrl(...)`) while SSR always forced it `false`, so `hydrate()` silently patched a structurally different document. The fix's core invariant is therefore: **`isOpen` must be `false` until `hydrateAutoOpen()` runs.** No test asserts that invariant, and no test exercises the divergent scenario end-to-end:

- **The unit tests never see the pre-hydrate state.** `TodayManager.test.ts`, `todayStaysOpenOnBoot.test.ts`, and `todayCoversReader.test.ts` all build state through `createTestSeedBibleState`, which now calls `state.today.hydrateAutoOpen()` before returning (`createTestSeedBibleState.ts:288`). So every assertion observes the *post*-hydrate value only. Whether `isOpen` seeded `false` or URL-derived beforehand is invisible to them — the end state is identical either way.
- **The real hydration test dodges the divergent case.** `hydration.test.tsx` renders SSR at `PATH = "/en/AAB/genesis/1?useFreeBibleAPI=true"` — a canonical reading path. `todayWillAutoOpenForUrl` returns `false` for a reading path in *both* the old and new code, so Today is closed on server and client regardless of the fix. The byte-for-byte DOM comparison there can never catch the Today mismatch, because Today never opens in that scenario. The bug only manifests on a bare `/` or an explicit `?today=open`, and neither URL is hydrated in any test.

**Verification (confirmed, not theoretical).** I reintroduced the exact pre-fix seeding in `TodayManager.tsx`:

```ts
const isOpen: Signal<boolean> = signal(
  import.meta.env.SSR
    ? false
    : todayWillAutoOpenForUrl(navigation.initialUrl, navigation.basePath)
);
```

and re-ran all four files: **32/32 still pass.** A regression test that stays green on the code it's meant to guard isn't testing anything — which is exactly the "tests that always succeed" risk worth flagging here.

**Suggested fix (pick either, ideally both):**

1. *End-to-end, highest value* — add a case to `hydration.test.tsx` that renders SSR at a bare `/` (Today auto-opens there), hydrates, and asserts the DOM is byte-identical across the boundary (i.e. Today is **closed** in the hydrated tree, matching SSR), then that Today opens only after the post-mount `hydrateAutoOpen()` runs (via `act`, the same way the saved-tabs case at `hydration.test.tsx:272` drives `hydrateFromStorage`). On the pre-fix code this fails, because the client tree would carry the whole open Today screen the served HTML lacked.
2. *Focused, cheap* — a unit test that builds state for a bare-`/` URL and asserts `state.today.isOpen.value === false` **before** calling `hydrateAutoOpen()`, then `true` after. This directly pins the seed-then-correct invariant and fails on the old construction-time seeding. It would need `createTestSeedBibleState` to expose the pre-hydrate window (e.g. an option to skip the internal `hydrateAutoOpen()` call), or a smaller test that constructs `createTodayManager` directly.

**Repro for a human developer:** check out the branch, apply the pre-fix `isOpen` seeding above, run `pnpm vitest run TodayManager.test.ts todayStaysOpenOnBoot.test.ts todayCoversReader.test.ts hydration.test.tsx` — everything passes, demonstrating the gap.

---

### Low — `render()`-fallback path now flashes Today closed→open

When `decideHydration` declines and the client does a fresh `render()` instead of `hydrate()` (`hydration.test.tsx` covers several decline reasons: `url-mismatch`, `build-mismatch`, etc.), there is no SSR markup to match, yet `isOpen` still seeds `false` and only flips in the post-mount effect. On an auto-open URL that produces a brief reader-then-Today transition that the old construction-time seeding rendered open in one pass. This is inherent to deferring the read unconditionally and is almost certainly an acceptable trade — but the PR description's "no flash" claim holds only for the true-hydration path, not the decline-and-render path (and, strictly, the SSR-closed-for-SEO decision makes a one-frame open unavoidable for a bare `/` even under clean hydration). Worth a sentence in the PR body so the trade-off is on record; no code change required.

### Nitpick (pre-existing, not introduced here) — stale doc comment

`TodayManager.tsx:68-69` says the predicate is "Exported so `readerVisible` can ask the same question without keeping its own copy." `readerVisible` (`SeedBibleStateManager.tsx:796`) actually derives from fullscreen-pane presence, not `todayWillAutoOpenForUrl` — the predicate has no caller outside `TodayManager`. Not caused by this PR, but you're editing the surrounding block, so it's a cheap cleanup while you're here.

---

**Bottom line:** ship the fix — it's sound — but add a test that actually reproduces the mismatch before considering the bug closed. As it stands, a future refactor could silently reintroduce the exact corruption this PR removes and the suite would stay green.

---

KallynGowdy:
@Blitzy  Confirmed that this fixes the issue on `develop`. It would be nice to have a working regression test for the hydration mismatch though, since they are difficult to find and prevent

KallynGowdy (dismissed):

Issue #1734, opened 2026-09-01
Dismissing the login prompt breaks the action that asked for it


Blitzy:
## What is happening

Several features offer a sign-in when you use them while signed out. Dismissing that prompt doesn't return you to what you were doing — it breaks the action that opened it, silently and with no error shown.

The cause is in `LoginManager`, not in any one feature. `cancelLogin()` rejects the pending `login()` promise instead of resolving it, so `await login.login()` throws and every line after it in the calling function is skipped. Callers are all written expecting `login()` to resolve `null` when there is no account, so their signed-out handling never runs.

What that looks like in practice:

- **Annotations** — the editor never opens. (How this was found.)
- **Playlists** — the editor never opens. The AI playlist tool's "sign-in was required and did not complete" message is unreachable; the tool call throws instead.
- **Bookmarks** — the bookmark modal stays open, because `onClose()` comes after the throw.
- **AI transcript** — throws out of its sign-in handler, leaving `isLoggedIn` unset.
- **Transcript UI** — throws a `TypeError` reading `currentPane.id`.
- **Highlights** — its signed-out guard is unreachable dead code, so this becomes an unhandled rejection rather than a clean no-op.
- **Settings page sign-in button** — unhandled promise rejection on dismiss.

## What is expected

Dismissing the login prompt should mean "carry on without an account", not throw. `login()` already declares `Promise<UserInfo | null>`, so `null` is the answer every caller is written to handle.

Found while investigating #1591 (offline and anonymous highlighting), which needs this same path to work.

PR #1735, opened 2026-09-01
Make highlights work signed out and offline

### LINKED ISSUE #1591: Support offline and anonymous highlighting, with sync when reconnected
## What is happening

Highlighting currently requires being logged in and online. Two related gaps:

- An anonymous (not logged in) user invited into a shared session is prompted to log in when they try to highlight a verse.
- Even a logged-in user cannot highlight while offline. Highlighting should work offline and then sync once the connection is back, rather than requiring a live connection to work at all.

_Evidence:_ Loom walkthrough (timestamp 1:15): https://www.loom.com/share/b24a81faa78e44ad943ccc58585a23ac

## What is expected

Highlighting should not depend on being logged in or online at the moment you make the highlight:

- **Anonymous users** should be able to highlight without being prompted to log in, the same way anonymous users are already supported for things like display and theme settings.
- **Offline highlighting** should work for any user, anonymous or logged in. Highlights made while offline should be saved locally and synchronized automatically once the connection is restored.

This applies both inside shared sessions and generally.

VictorUvarov:
## Summary

- Highlights now work signed out and offline: writes go to a local IndexedDB mirror and are pushed by a background sync pass when signed in and online. The sign-in prompt on highlight is gone; anonymous highlights are adopted into the account on sign-in and merged with what the server holds.
- Generalized the annotations-only offline engine into `OfflineRecordStore<T>` / `RecordSyncManager<T>`, parametrized by a per-feature `SyncDomain<T>`. Annotations moved onto it with no behavior change; the shipped `seed-bible-annotations` database migrates from v1 to v2 on first open (the row's server base is rebuilt from what v1 remembered, so queued offline edits still push cleanly).
- Highlight conflicts never show a dialog: `mergeChapterHighlights` does a per-verse three-way merge (local wins where both sides touched a verse, server wins where local left it alone) and collapses back into ranges. Annotations keep their conflict modal.
- Adoption on sign-in pushes against a null base (a union for highlights, a create for annotations); keeping the account row's base let a signed-out edit silently replace the server's whole chapter.

Closes #1591

## Test plan

- [ ] `pnpm vitest run` passes (5365 tests; new suites `OfflineRecordStore.test.ts`, `RecordSyncManager.test.ts`, and the `#1591` block in `HighlightsManager.test.ts`)
- [ ] Migration with a real v1 database: on `develop`, sign in, write an annotation, go offline, edit it, close the tab. Switch to this branch and reload: DevTools > Application > IndexedDB shows `seed-bible-annotations` at version 2 with `records` and `listed` stores and the pending row carrying `address`, `collection`, `payload`, `base`. Go online: the edit pushes with no conflict modal.
- [ ] Same migration check with a second tab left open on the old version: it recovers after reload and annotation saves in the meantime fall through to the server instead of failing.
- [ ] Signed out: highlight a verse in the reader. No login modal; highlight visible; `seed-bible-highlights` has a row under owner `local`.
- [ ] Sign in: the highlight survives, the row moves under the user id, and after a moment `pendingOp` is null. If the account already had highlights in that chapter, both sets are visible.
- [ ] Signed in, go offline (DevTools Network), highlight another verse, reload: both visible. Go online: one `recordData` request in the Network tab carrying both verses.
- [ ] Two devices highlight different verses in the same chapter while one is offline; after reconnecting both verses are on both devices.
- [ ] Annotations regression: create, edit offline, delete, and resolve a conflict with the modal; all behave as on `develop`.


KallynGowdy:
@VictorUvarov I really like this! It currently only works for highlights though? It currently still seems to require a login for saving annotations (notes).

Also, there was one thought I had while testing, which is if I'm using a shared device then I wouldn't want it to overwrite or add to my account's highlights and notes. If highlights and/or notes were recorded while not signed in, and the user signs in, could we trigger a popup modal asking them if they want to adopt the highlights from the device to their account? For example, it could say:

- Title: `Sync device content?`
- Body: `This device is storing {highlights/notes/etc.} that are not associated with any account. Do you want to add them to your account?`
- Options:
  - `Add to account`
  - `Don't add`

@Blitzy do you have anything to add?

Blitzy:
The login required issue should be resolved when #1736 gets merged in.

As far as the shared device scenario, I agree that is an important use-case we have to consider. I'm not sure that a generalized modal is going to be enough to handle that case as it could be likely that some of the annotations and highlights are yours and some are not so you would need to be able to pick the ones to sync? Starts getting pretty in the weeds then though.

VictorUvarov:
> @VictorUvarov I really like this! It currently only works for highlights though? It currently still seems to require a login for saving annotations (notes).
> 
> Also, there was one thought I had while testing, which is if I'm using a shared device then I wouldn't want it to overwrite or add to my account's highlights and notes. If highlights and/or notes were recorded while not signed in, and the user signs in, could we trigger a popup modal asking them if they want to adopt the highlights from the device to their account? For example, it could say:
> 
> * Title: `Sync device content?`
> * Body: `This device is storing {highlights/notes/etc.} that are not associated with any account. Do you want to add them to your account?`
> * Options:
>   
>   * `Add to account`
>   * `Don't add`
> 
> @Blitzy do you have anything to add?

Annotations are on the same engine and do save locally while signed out. What you're seeing is #1734: the create flow offers a sign-in, and on develop dismissing that modal throws, so the editor never opens. Blitzy's #1736 fixes that. Once it's in, dismiss the prompt and you get a local draft that gets adopted on sign-in. Highlights skip the prompt entirely. Happy to drop it for annotations too if you'd rather they match.

The shared-device modal makes sense to me. There's already a single hook in RecordSyncManager where adoption runs on owner change, so a "Sync device content?" prompt with add / don't add slots in cleanly. "Don't add" would discard the signed-out rows so they don't get offered to the next person who signs in.

One thing that softens the risk in the meantime: adoption can only add. Highlights merge as a union with what the account already has and signed-out notes are created as new notes, so nothing on the account gets overwritten. The worst case today is a few extra highlights you didn't make.

I can add the modal

VictorUvarov:
> The login required issue should be resolved when #1736 gets merged in.
> 
> As far as the shared device scenario, I agree that is an important use-case we have to consider. I'm not sure that a generalized modal is going to be enough to handle that case as it could be likely that some of the annotations and highlights are yours and some are not so you would need to be able to pick the ones to sync? Starts getting pretty in the weeds then though.

Agreed that per-item picking is where it gets messy. I'd ship the two-option modal first since it covers the common case where someone else used the device and you don't want any of it.

Blitzy:
@VictorUvarov do you think you can add the modal? I'd love to get this in soon! We have some stuff that I think would be good to build off of the new `OfflineRecordStore` pattern

VictorUvarov:
I added the modal @Blitzy 
Screenshot 2026-09-08 at 11 21 47 PM

KallynGowdy (approved):
This seems to work well. I've checked:

- saving highlights without an account
- Logging in to adopt highlights
- Logging in to not adopt highlights
- Saving highlights while logged in but offline (it adds them once the user comes back online and survives reloads)

PR #1736, opened 2026-09-01
Dismissing the login prompt no longer breaks the action that asked for it

### LINKED ISSUE #1734: Dismissing the login prompt breaks the action that asked for it

Blitzy:
Fixes: #1734

## Summary

`LoginManager.cancelLogin()` rejected the pending `login()` promise instead of resolving it, so dismissing the sign-in modal threw out of whatever the user was doing. This makes dismissal resolve `null` — the value every caller was already written to expect.

## Problem

Signed out, open a chapter and try to add an annotation. The login modal appears. Dismiss it and nothing happens: the annotation editor never opens, and no error is shown.

The prompt is meant to be an offer rather than a requirement — a signed-out note is kept on the device and adopted into the account on first sign-in. But `await login.login()` threw on dismissal, so every line after it was skipped, including the two that build the draft and switch the Discover pane to the editor. The call site is `void annotations.createNewAnnotation()`, so the throw surfaced only as an unhandled rejection in the console.

The signed-out handling was already written correctly and simply never ran. `createNewAnnotation` checks `if (!userId && !recordOverride && !store)` and only gives up when there is nowhere local to write — it never got that far.

This was never caught because the manager tests mock `login.login` as resolving `null`, which is what the callers expect but not what `LoginManager` did. The mock and the caller agreed with each other while disagreeing with the real implementation.

## The Fix

`cancelLogin()` now resolves the pending promise with `null` rather than rejecting it. `rejectLoginPromise` had no other use and is removed, and the `login()` / `cancelLogin()` doc comments now state the contract.

Dismissing a prompt is an answer — "carry on without an account" — not a failure. `login()` already declared `Promise<UserInfo | null>`, so `null` was always the intended way to say "no account".

## Blast radius

All nine `login()` call sites were checked. Every one of them already assumed a resolved value, so this fixes more than annotations:

| Call site | Before | After |
|---|---|---|
| `AnnotationsManager.createNewAnnotation` | editor never opens | opens a local draft |
| `PlaylistManager.createNewPlaylist` | editor never opens; the AI tool's "sign-in was required and did not complete" message was unreachable | returns `null`, message reachable |
| `BookmarksManager.addBookmark` | bookmark modal stays open, because `onClose()` came after the throw | modal closes cleanly |
| `transcriptionManager` (AI transcript) | throws out of its sign-in handler, `isLoggedIn` left unset | sets `isLoggedIn` to false |
| `transcript-ui` `init.tsx` | `TypeError` reading `currentPane.id`, because its `catch` nulled `currentPane` first | closes the pane cleanly |
| `HighlightsManager` (×2) | guard clauses unreachable, unhandled rejection | clean no-op with a warning |
| `SettingsPage` sign-in button | unhandled rejection on dismiss | resolves quietly |
| `twitchPubManager` | already correct — its `catch` and its `if (!userInfo)` do the same thing | unchanged; the `catch` is now redundant |

## Testing

- `LoginManager.test.ts` — "can cancel the login flow" now asserts `resolves.toBeNull()`. Confirmed red against the pre-fix code.
- `AnnotationsManager.test.ts` — new test covering the reported case: signed out, **online**, prompt dismissed, local store present. The editor opens and the note is written to the `local` bucket without touching the server. Existing coverage only had the offline case and the no-store case, so this exact combination was untested.
- Full suite: 5376 tests across 229 files, all passing. `pnpm check:ts` clean, Prettier clean.

**Manual check:** sign out, open a chapter, Discover → Annotations → new note, dismiss the sign-in modal. The editor should open, and the note should still be there after a reload.

## Notes

Found while investigating #1591 (offline and anonymous highlighting), which needs this same path to work — highlighting signed out currently hits the same rejection.

This changes a documented, tested contract on `LoginManager`. I went this way rather than adding `try`/`catch` at nine call sites because the return type already advertised `null` and every caller assumed it, but it's worth a look during review.

### DIFF
--- a/packages/seed-bible/seed-bible/managers/LoginManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/LoginManager.tsx
@@ -127,6 +127,11 @@ export interface LoginManager {
 
   /**
    * Attempts to login the user.
+   *
+   * Resolves with the signed-in user, or `null` when no account was obtained —
+   * including when the user dismisses the prompt. Dismissal is deliberately not
+   * a rejection: callers offer sign-in as a convenience and carry on without an
+   * account, so throwing would abandon whatever the user was actually doing.
    */
   login: () => Promise<UserInfo | null>;
 
@@ -156,6 +161,8 @@ export interface LoginManager {
 
   /**
    * Cancels an in-progress login attempt, if one exists. This is useful to abort a login flow if the user navigates away or closes the login modal before completing authentication.
+   *
+   * Any pending {@link login} call resolves with `null`.
    */
   cancelLogin: () => Promise<void>;
 
@@ -465,7 +472,6 @@ export function createLoginManager({
 
   let loginPromise: Promise<UserInfo | null> | null = null;
   let resolveLoginPromise: ((value: UserInfo | null) => void) | null = null;
-  let rejectLoginPromise: ((err: Error) => void) | null = null;
   let currentLoginPromise: Promise<UserInfo | null> | null = null;
 
   // const userId = os.userId;
@@ -652,11 +658,16 @@ export function createLoginManager({
 
   async function cancelLogin() {
     isLoginOpen.value = false;
-    if (loginPromise && rejectLoginPromise) {
-      rejectLoginPromise(new Error("Login cancelled"));
+    if (loginPromise && resolveLoginPromise) {
+      // Resolves null rather than rejecting. Dismissing the prompt is an
+      // answer ("carry on without an account"), not a failure, and every
+      // caller is written to read that answer from the resolved value. A
+      // rejection instead threw out of the caller mid-way, which is what
+      // stopped a dismissed prompt from leaving a signed-out user able to
+      // write a note (#1591).
+      resolveLoginPromise(null);
       loginPromise = null;
       resolveLoginPromise = null;
-      rejectLoginPromise = null;
     }
   }
 
@@ -726,7 +737,6 @@ export function createLoginManager({
       if (resolveLoginPromise) {
         resolveLoginPromise(userInfo.value);
         resolveLoginPromise = null;
-        rejectLoginPromise = null;
         loginPromise = null;
       }
 
@@ -739,9 +749,8 @@ export function createLoginManager({
   async function loginCore(): Promise<UserInfo | null> {
     if (!sessionKey.value) {
       if (!loginPromise) {
-        loginPromise = new Promise((resolve, reject) => {
+        loginPromise = new Promise((resolve) => {
           resolveLoginPromise = resolve;
-          rejectLoginPromise = reject;
         });
       }
 
--- a/test/unit/seed-bible/managers/AnnotationsManager.test.ts
+++ b/test/unit/seed-bible/managers/AnnotationsManager.test.ts
@@ -1148,6 +1148,28 @@ describe("AnnotationsManager", () => {
       expect(manager.editingAnnotation.value).not.toBeNull();
     });
 
+    it("starts a signed-out draft online after the sign-in prompt is dismissed", async () => {
+      // Online, so sign-in is offered. Dismissing it means "carry on without an
+      // account", and a device with somewhere to put the note must honour that:
+      // the editor opens and the note is kept locally. Before #1591 the
+      // dismissal threw out of `createNewAnnotation`, so the editor never
+      // opened and the user was simply stuck.
+      login.userId.value = null;
+      login.login.mockResolvedValue(null);
+      const manager = createOfflineManager();
+
+      await manager.createNewAnnotation();
+
+      expect(login.login).toHaveBeenCalled();
+      const draft = manager.editingAnnotation.value;
+      expect(draft).not.toBeNull();
+
+      await manager.saveEditingAnnotation();
+
+      expect(await store.get(LOCAL_OWNER, draft!.id)).not.toBeNull();
+      expect(recordDataMock).not.toHaveBeenCalled();
+    });
+
     it("shows signed-out drafts in the chapter listing", async () => {
       login.userId.value = null;
       const manager = createOfflineManager();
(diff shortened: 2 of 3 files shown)


Blitzy:
> [!NOTE]

## Code Review — PR #1736

**Verdict: the source change is correct and well-reasoned; one of the two new tests doesn't actually guard the fix.**

I read the change against issue #1734 (the bug this closes) and #1591 (the follow-up that depends on this path). The core fix — having `cancelLogin()` resolve the pending `login()` promise with `null` instead of rejecting it — is the right call. I traced all nine `login.login()` callers in the repo to confirm the reject→resolve switch doesn't break any of them:

- The in-repo managers (`AnnotationsManager`, `PlaylistManager`, `BookmarksManager`, `HighlightsManager`) all re-check `login.userId.value` or the returned `userInfo` after awaiting, so a `null` resolution flows through their existing signed-out handling.
- The two extension callers (`twitchPubManager`, transcript UI) are double-guarded — a `.catch()` plus an `if (!userInfo)` fallback — so they behave correctly whether the promise resolves `null` or rejects.

There are no lingering references to the removed `rejectLoginPromise`, and no other test still asserts the old `"Login cancelled"` rejection. `LoginManager.test.ts` line 240 (`resolves.toBeNull()`) is a valid regression test — it fails on the pre-fix code where cancel rejected. Good.

---

## Findings by severity

### Low — the new `AnnotationsManager` test passes identically on the pre-fix code

**File:** `test/unit/seed-bible/managers/AnnotationsManager.test.ts` (the new test at line 1151, "starts a signed-out draft online after the sign-in prompt is dismissed")

The test's comment claims: *"Before #1591 the dismissal threw out of `createNewAnnotation`, so the editor never opened."* But the test never exercises the reject→resolve change it describes. In this test file `login` is a full `Mocked<LoginManager>` (line 77), and the test sets `login.login.mockResolvedValue(null)` (line 1158) — so it drives a mock that resolves `null` directly, never the real `cancelLogin`.

`AnnotationsManager.tsx` is **not touched by this PR** and already did `userId = userInfo?.id ?? null`. So with `login.login` mocked to resolve `null`, this test passes byte-for-byte on the pre-fix `LoginManager` (where cancel rejected). Revert the `LoginManager.tsx` change and this test stays green — which is exactly what the repo's own rule warns against:

> Regression tests must fail on the pre-fix code — revert the fix locally, confirm red, then restore it. A regression test that passes either way isn't testing anything. — `CLAUDE.md`

The actual reject→resolve fix is only genuinely covered by `LoginManager.test.ts` line 240. That's fine as coverage — but this new test doesn't add a regression guard, and its comment overstates what it verifies.

1. On the PR branch, `git stash` or revert only the `LoginManager.tsx` hunk (leave both test files as-is).
2. Run `pnpm vitest run AnnotationsManager.test.ts`.
3. The new test still passes — it never touched the reverted code.

**Suggested fixes (pick one):**
- **Make it a real guard:** in this test, drive the actual dismissal contract rather than a hardcoded mock — e.g. mock `login.login` to *reject* (`login.login.mockRejectedValue(new Error("Login cancelled"))`) and assert the editor still opens and the note is written locally. That version goes red on the pre-fix code (the reject propagates out of `createNewAnnotation`) and green after, which is what the comment promises.
- **Or reframe it honestly:** keep the `mockResolvedValue(null)` but drop the "Before #1591 the dismissal threw" framing and describe it as what it is — a coverage test for the online + signed-out + local-store combination (which the PR notes was previously untested). The reject→resolve regression itself is then explicitly owned by `LoginManager.test.ts`.

The first option is stronger because it ties the `AnnotationsManager` behavior to the real `LoginManager` contract instead of a mock that happens to agree with it — which is the very "mock and caller agreed while disagreeing with the implementation" trap the PR description itself calls out as why the bug went uncaught.

---

## Notes (non-blocking)

- The PR description's blast-radius table is accurate against what I traced — nice write-up. The `twitchPubManager` row correctly notes its `.catch()` is now redundant; leaving it in place is fine (harmless defense-in-depth), no change needed.
- Doc comments on `login()` and `cancelLogin()` now state the `null`-on-dismissal contract clearly. This is the right place to pin it down given nine callers depend on it.

**Bottom line:** the fix is correct and safe to merge on its behavior. I'd address the Low finding before merge so the regression is actually guarded at the level the test claims — but it's a test-quality issue, not a correctness blocker.

---

Blitzy:
AnnotationsManager test fixed with one that tests the real behavior of the LoginManager. Test now verifies that the annotation editor becomes visible after the login modal is dismissed. Verified red without changes, verified green with changes.

PR #1738, opened 2026-09-02
Record audio listening time separately from screen time

Blitzy:
## Summary

Audio playback now records listening time independently from screen time. When a phone is locked or the app is backgrounded, the audio element's own clock keeps advancing while the page is frozen and timers stop running. This change captures that listening by measuring the audio element's clock and writing it back when the page wakes up, rather than relying on the app's own timer which can't run while the screen is off.

## Key Changes

- **New `attachListeningRecorder` function** in the audio reader extension that monitors an audio element's playback and credits time to reading history based on how far the audio clock advanced, not wall time. It handles:
  - Playback that continues while the page is frozen (locked screen, backgrounded tab)
  - Seeks and speed changes (only credits actual audio advancement, capped by wall time)
  - Periodic saves during playback so long listens survive page termination
  - Re-anchoring to the audio clock after seeks

- **New `saveReadingHistorySpan` function** that records a stretch of time already spent on a chapter (start and end times), rather than just marking "read now". It extends existing events when a span continues them, but only if the span starts within a join threshold—time nothing was watching opens a new event rather than being back-filled.

- **Reading history manager now exposes `saveReadingSpan`** alongside the existing `saveReadingHistory`, allowing callers to credit measured time rather than just the present moment.

- **Reader's reading-time ticker now respects visibility state**: it only credits time while the app is actually visible. When the page goes into the background, the ticker stops; when it returns to the foreground, it resumes. This prevents the reader from claiming to have watched chapters while the user was doing something else.

- **Tests added** for both the listening recorder (handling frozen pages, seeks, speed changes, detachment) and the span-saving logic (joining events, never moving end times backwards, respecting join thresholds).

## Implementation Details

The listening recorder works by tracking two clocks:
- **Wall time** (Date.now): the real passage of time
- **Audio time** (element.currentTime): how far the narration has advanced

When playback stops or the page wakes up, it calculates how much audio advanced since the last anchor point, caps that by wall time (to handle playback at double speed), and writes a span. The audio element's clock is re-anchored after seeks so that skipping ahead doesn't credit the skipped portion as listening time.

The reader's ticker now uses `saveReadingSpan` to credit one tick at a time, with a cap on how much any single tick can claim (to handle pages that sleep without reporting themselves hidden). Visibility changes pause and resume the ticker, so backgrounded tabs don't accumulate reading time.

https://claude.ai/code/session_01AusbajD9FqZ9zywYkuFQA6


Blitzy:
> [!NOTE]

## Code Review — Record audio listening time separately from screen time

I reviewed the diff, the full surrounding code in `ReadingHistoryManager.tsx`, `SeedBibleStateManager.tsx`, and the audio extension, and traced how the new span writes interact with the existing summary math. No linked issue is referenced in the PR; the only comment on the thread is the deploy bot, so this review is based on the code and CHANGELOG intent alone.

**Overall: this is a well-designed change and I did not find any correctness bugs.** A few things I specifically checked and confirmed are safe, followed by lower-severity notes.

### What I verified is correct (no action needed)

- **No double-counting when listening with the screen on.** The reader ticker and the audio recorder both write spans for the same chapter, but `getReadingHistorySummary` sums `event.end - event.start` per event, and both writers land on and extend the *same* event (the audio span's start sits within the 30s join threshold of the reader event's end while both are live). One event, counted once.
- **No race between concurrent saves.** `saveReadingHistorySpan` is `async`, but the critical section — `findMostRecentReadingEvent` → `createMap`/`push` — contains no `await`. Concurrent calls (a reader tick landing next to an audio save) await the same cached document promise and then run their find→push synchronously in microtask order, so the second call always sees the first call's pushed event. No duplicate events.
- **Skip-forward can't over-credit.** Even if a `timeupdate` were to fire with the jumped `currentTime` before `seeked` re-anchors, the wall-clock cap (`Math.min(run.startWallMs + playedMs, now())`) bounds the credit to real elapsed time. The re-anchor is belt-and-suspenders on top of that.
- **The regression tests are real** — each new test exercises a brand-new function or the new `saveReadingSpan` path and would fail on the pre-fix code (e.g. "never moves an event's end backwards" and "extends an event even one older than the join threshold" both flip to failures if their respective guards are removed). Good.

---

### Findings by severity

#### Low — listening during a background freeze is only persisted on wake

`attachListeningRecorder` writes on `visibilitychange`, `pause`, `ended`, and periodically via `timeupdate`. While a page is *frozen* (screen locked), none of those fire, so the stretch that advanced while frozen is only written when the page **wakes**. If the OS terminates the frozen page before it ever wakes, that stretch is lost — which is precisely the "phone in a pocket" scenario this PR targets.

Mitigating factors: the `visibilitychange → hidden` handler saves everything up to the moment of backgrounding, so only audio that advances *after* backgrounding and before an eventual kill is at risk; and an actively-playing media element often keeps the page alive (and `timeupdate` firing) in the background, so real-world exposure may be small.

- **Suggested fix:** also flush on `pagehide` (and optionally `freeze`), which browsers deliver before discarding a page:
  ```js
  window.addEventListener("pagehide", onVisibilityChange);
  ```
  Remember to remove it in the returned detach function.

#### Low (test quality) — the basic ticking tests no longer assert anything about the span contents

In `SeedBibleStateManager.test.ts`, the pre-existing tests ("saves first history event after 5 seconds", "saves history once for each additional 5 seconds", the reset tests) were converted to `anySpanFor(bookId, chapter)`, which matches `[bookId, chapter, expect.any(Number), expect.any(Number)]`. That means the normal foreground path no longer verifies the timestamps make sense — a regression that credited a zero-length span (`from === to`) or an oversized one would still pass these tests.

The two new freeze tests do assert magnitudes (`to - from <= 10`, `from >= resumedAtSeconds`), which is good — but the everyday path deserves the same. **Suggested fix:** in at least the "first history event after 5 seconds" test, assert `to > from` and that `to - from` is roughly one tick (≈5s), so the common path is pinned down.

#### Low (test quality) — "saves periodically … survives the page being killed" only proves a single save

```js
it("saves periodically so a long listen survives the page being killed", () => {
  h.el.play();
  h.listen(20);              // one timeupdate for 20s
  expect(h.saveSpan).toHaveBeenCalledWith("psalms", 23, START_SECONDS, START_SECONDS + 20);
});
```
`listen(20)` dispatches a single `timeupdate`, so this asserts one save — it doesn't actually exercise the `SAVE_INTERVAL_MS` cadence the name describes. A real long listen fires many `timeupdate`s and should produce repeated saves. **Suggested fix:** drive several shorter `listen()` steps across the interval and assert `saveSpan` was called more than once with a growing `end`.

#### Informational — seek test doesn't cover the `timeupdate`-before-`seeked` ordering

"credits a skip forward with only what was heard before it" sets `currentTime += 600` then dispatches `seeked`, without a `timeupdate` in between. Since the wall-clock cap is what actually prevents over-crediting a skip (not the re-anchor alone), consider a test that fires `timeupdate` *after* the jump and *before* `seeked` while wall time is held, to lock in that the cap — not just event ordering — is doing the protecting.

---

### Reproduction for the Low finding (freeze-then-terminate)

1. Sign in, open a chapter with audio, tap **Listen**.
2. Lock the phone and let narration play for several minutes with the screen off.
3. Force-kill the app from the OS task switcher **without** unlocking/returning to it first.
4. Reopen and check reading history: the minutes that played after the screen locked may be missing, because they were never flushed to a wake or `pause`.

(Hard to reproduce deterministically since it depends on whether the platform freezes an actively-playing media page; the `pagehide` flush closes the gap regardless.)

---

None of these block merging in my view — the core mechanism is sound and well-tested. The two test-quality notes are the ones I'd most encourage addressing, since they weaken coverage of the everyday path.

https://claude.ai/code/session_015vXAzzNuYxQuChsJbrk37C

---

Blitzy:
> [!NOTE]

## Code Review — Record audio listening time separately from screen time

This PR was already reviewed once (comment above, 2026-09-02) while it was still a draft. I re-reviewed the current head (`ca713a89`) against that feedback and against the diff itself.

**The earlier review's findings look addressed:** the recorder now flushes on `pagehide`/`freeze` in addition to `visibilitychange`/`pause`/`ended`, the reading ticker is visibility-gated, the previously-weak `anySpanFor(...)` assertions in `SeedBibleStateManager.test.ts` were tightened to check actual span magnitudes, and the "survives the page being killed" test now drives multiple ticks across `SAVE_INTERVAL_MS` and asserts repeated saves with a growing `end`, instead of asserting on a single `timeupdate`. Good follow-through.

I traced the rest of the diff line-by-line (join-threshold matching, the concurrent-writer path, seek/speed-change handling) and didn't find further problems there. One new gap did turn up, below.

---

### Finding — Low severity: a listening/reading span that crosses a UTC year boundary can be double-counted

**Where:** `packages/seed-bible/seed-bible/managers/ReadingHistoryManager.tsx:149-166` (`saveReadingHistorySpan`), combined with `packages/audio-reader-extension/ext_audioReader/host/init.tsx`'s `attachListeningRecorder`, which never re-anchors `run.startAudioSeconds`/`run.startWallMs` after a successful periodic save (it keeps replaying the *original* start of the run on every 15s tick — see the `save` closure, roughly line 96 in that file).

```js
const year = new Date(endTimeSeconds * 1000).getUTCFullYear();
const doc = await getReadingHistoryDocument(os, recordName, year, marker, name);
...
const event = findMostRecentReadingEvent(array, userId, bookId, chapter, startTimeSeconds - joinThresholdSeconds);
```

The document (and therefore the array `findMostRecentReadingEvent` searches) is selected purely by **`endTimeSeconds`'s** year. That's harmless as long as one listening run's saves all land in the same year's document — but the audio recorder deliberately keeps `startTimeSeconds` fixed at the run's original start on every periodic save (that's what lets it always resolve to the same, extendable event). If the run happens to straddle a UTC year rollover, that assumption breaks:

**Reproduction (audio playback held open across UTC midnight, e.g. Dec 31 → Jan 1):**
1. A user starts listening at `2026-12-31T23:59:50Z`. The first periodic/flush save has `start=23:59:50`, `end=23:59:59` (both 2026) → looked up in the **2026** document → no match → creates a new event `{start: 23:59:50, end: 23:59:59}` there.
2. Narration continues past midnight. The *next* save still uses the run's original, un-re-anchored `start=23:59:50`, but now `end=00:00:10` (2027) → `year` is computed from `endTimeSeconds` → **2027** → `getReadingHistoryDocument` opens/creates the 2027 doc → `findMostRecentReadingEvent` searches that doc's events, which is empty for this user/book/chapter → no match found → a **second** event is created in the 2027 doc: `{start: 23:59:50 (a 2026 timestamp), end: 00:00:10}`.
3. Net effect: the 2026 doc has an event covering `23:59:50–23:59:59`, and the 2027 doc has a second, independent event covering `23:59:50–00:00:10` — the ~9 seconds before midnight are now credited **twice** across two different yearly summaries, and the second event's `start` timestamp doesn't even belong to the year of the document it's stored in.

This is a narrow edge case (it needs sustained listening activity spanning the exact UTC year boundary), so it's Low severity, and I confirmed there's no existing test covering a span that crosses a year boundary — the `saveReadingHistorySpan` test suite mocks a single shared document per test and never exercises the multi-document lookup path (`test/unit/seed-bible/managers/ReadingHistoryManager.test.ts:672` onward).

**Suggested fix:** either look up the candidate event by the year of **both** `startTimeSeconds` and `endTimeSeconds` when they differ (check the previous year's document too, and only fall back to creating a new event if neither has a match), or avoid the calendar-year document boundary being reachable mid-span at all by having `attachListeningRecorder` re-anchor `run.startWallMs`/`run.startAudioSeconds` to the just-saved point after every successful periodic save (this also has the side benefit of making each periodic save a cheap, fixed-size delta instead of replaying the whole run's span every 15s). Re-anchoring alone would still split the event at the boundary rather than merge it, but it removes the double-count, since the second document's event would then start at `00:00:00`-ish rather than replaying the pre-midnight start.

---

None of this blocks merging — it's a narrow edge case layered on top of an otherwise sound and well-tested design. Worth a follow-up test + fix given it silently inflates reading-history stats when it does trigger.

---

PR #1739, opened 2026-09-02
Add SSR support for Customizations

KallynGowdy:
## Summary

A shared `?customization=<recordName>.<id>` link lets someone send a styled version of Seed Bible, but the customization's colors previously only applied **after** the client hydrated and re-fetched the record — the server-rendered page always showed the visitor's own default theme first, then flashed to the customization's colors. This stacked PR fixes that by making the server wait for the customization to load before it finishes rendering.

- **`CustomizationsManager.tsx`**: exposes `initialCustomizationLoadPromise` / `initialCustomizationLoadSettled`, resolving once the `?customization=` load reaches a terminal outcome (loaded, failed, or — SSR only — a 5s timeout backstop). The promise never rejects, so it's safe to throw from a component.
- **`main.tsx`**: `ExternalResourceDependencies` (the component that emits the theme `<style>` tags) throws that promise during SSR until it settles, following the codebase's existing "throw a promise to suspend" pattern (the same one `BibleReadingManager`'s `chapterDataPromise` uses). The gate is scoped to just this component rather than to `MainContent` as a whole — `preact-render-to-string` doesn't block later siblings on an earlier one suspending, so a narrower gate leaves the rest of the tree's first-render timing untouched.
- **`entry-ssr.tsx`**: also now blocks on the initial tab's own `chapterDataPromise` alongside `state.i18n.ready` before attempting any render. This turned out to be necessary: any additional SSR-blocking wait introduced before the first render attempt can outrun the initial chapter fetch (which starts at state construction), and `preact-render-to-string` never calls `options._catchError` for a component that throws to suspend — so if `BibleReader` itself ends up needing to suspend on its own `chapterDataPromise`, the signal write that would resolve it never flushes, and the request hangs forever. That fetch has always finished by the time rendering reaches `BibleReader` today purely because nothing made the render wait on anything else first; explicitly waiting for it keeps that race unreachable regardless of what else (the customization load, or anything added later) delays the first render.

## Test plan

- [x] `pnpm vitest run CustomizationsManager.test.ts entry-ssr.test.ts` — 172 tests pass, including new coverage for the settlement latch/timeout and an end-to-end SSR render with a linked customization
- [x] Confirmed the new SSR test actually catches the regression: reverting the `ExternalResourceDependencies` guard makes it fail (missing `--sb-primary-color` in the HTML) — the customization response is deliberately delayed so it can't slip past `render()`'s other await first
- [x] `pnpm check:ts` — clean
- [x] `pnpm lint` — 0 errors (only pre-existing unrelated warnings)
- [x] `pnpm test` — full suite (234 files / 5608 tests) passes

https://claude.ai/code/session_01K8axGRHv1cGfzmkPnzQs4t

---
_Generated by [Claude Code](https://claude.ai/code/session_01K8axGRHv1cGfzmkPnzQs4t)_

### DIFF
--- a/packages/seed-bible/seed-bible/app/main.tsx
+++ b/packages/seed-bible/seed-bible/app/main.tsx
@@ -55,16 +55,31 @@ import { OfflineDownloadPrompt } from "../components/OfflineDownloadPrompt/Offli
  * the server but not on the client, causing a hydration mismatch; Preact also
  * never diffs `dangerouslySetInnerHTML` during hydration, so this stays
  * inert even if the two sides' CSS text does legitimately differ.
+ *
+ * The SSR suspend below is deliberately scoped to just this component rather
+ * than gating `MainContent` as a whole: `_renderToString`'s array-of-children
+ * traversal doesn't block later siblings on an earlier one suspending, so
+ * gating only here keeps everything else's first-render timing (most notably
+ * `BibleReader`'s own chapter-load suspend) exactly as it is without a
+ * `?customization=` link in play.
  */
 export function ExternalResourceDependencies({
   themeCssVariables,
   themeCssClasses,
   googleFontFamilies,
+  initialCustomizationLoadPromise,
+  initialCustomizationLoadSettled,
 }: {
   themeCssVariables: ReadonlySignal<string>;
   themeCssClasses: ReadonlySignal<string>;
   googleFontFamilies: ReadonlySignal<string[]>;
+  initialCustomizationLoadPromise: Promise<void>;
+  initialCustomizationLoadSettled: ReadonlySignal<boolean>;
 }) {
+  if (import.meta.env.SSR && !initialCustomizationLoadSettled.value) {
+    throw initialCustomizationLoadPromise;
+  }
+
   return (
     <>
       <link
@@ -256,6 +271,12 @@ function MainContent(props: {
           themeCssVariables={theme.themeCssVariables}
           themeCssClasses={theme.themeCssClasses}
           googleFontFamilies={theme.googleFontFamiliesToLoad}
+          initialCustomizationLoadPromise={
+            state.customizations.initialCustomizationLoadPromise
+          }
+          initialCustomizationLoadSettled={
+            state.customizations.initialCustomizationLoadSettled
+          }
         />
         <Sidebar state={state} />
 
--- a/packages/seed-bible/seed-bible/managers/CustomizationsManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/CustomizationsManager.tsx
@@ -614,6 +614,18 @@ export interface CustomizationsManager {
    */
   linkedCustomization: ReadonlySignal<SeedBibleCustomization | null>;
   /**
+   * Resolves once the initial `?customization=...` load (if any) reaches a
+   * terminal outcome: loaded, failed, or (during SSR only) exceeded a
+   * deadline. Resolves immediately if there was no `?customization=` param.
+   * Throw this in a component to suspend SSR rendering until then.
+   *
+   * Never rejects — a rejected promise thrown during `renderToStringAsync`
+   * surfaces as a render exception and takes down the whole SSR document.
+   */
+  initialCustomizationLoadPromise: Promise<void>;
+  /** True once the initial `?customization=` load has settled — see above. */
+  initialCustomizationLoadSettled: ReadonlySignal<boolean>;
+  /**
    * The local, unpersisted draft of the customization currently open in the
    * editor settings pages, or null when none is open. Edits accumulate here
    * and are only written to CasualOS by `saveEditingCustomization`.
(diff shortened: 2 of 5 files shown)


Blitzy:
> [!NOTE]

## Review of "Add SSR support for Customizations"

**Overall: this is a solid, carefully-reasoned change and I recommend it for merge after considering the notes below.** None of the findings are blocking. The core mechanism is correct, the comments do an unusually good job of explaining the tricky SSR/signals interactions, and the most important test is a genuine regression test rather than one that passes regardless of the fix.

### What I verified

- Ran the two touched test files: **172 tests pass** (`entry-ssr.test.ts` + `CustomizationsManager.test.ts`).
- Confirmed the key test is real, not vacuous: I temporarily neutered the SSR suspend (`if (false && import.meta.env.SSR && ...)` in `ExternalResourceDependencies`) and re-ran. The test **"includes the linked customization's theme colors in the SSR'd HTML" flipped to red**, exactly as a regression test should. Restored afterward.
- Traced the resolution paths: `initialCustomizationLoadPromise` always settles during SSR (load resolves → `settleInitialCustomizationLoad`, or the 5s backstop timer fires), and `chapterDataPromise` — newly awaited in `entry-ssr.tsx` — has its own 5s SSR backstop in `BibleReadingManager` (`SSR_INITIAL_CHAPTER_TIMEOUT_MS`), so neither new `await` can hang the request open unboundedly. Good.

---

### Findings by severity

#### Low

**No. 1 — A copied comment describes client behavior that doesn't exist for this promise.**
`CustomizationsManager.tsx` (around lines 818–820) says the SSR timer is:
> Not armed on the client: there the promise only gates a Suspense boundary...

That sentence is accurate for `chapterDataPromise` (its comment in `BibleReadingManager.tsx`), because `BibleReader` throws `chapterDataPromise` **unconditionally** (`BibleReader.tsx:1596`), so it really does gate a client Suspense boundary. But `initialCustomizationLoadPromise` is thrown in only one place — `ExternalResourceDependencies` in `main.tsx:79` — and that throw is guarded by `import.meta.env.SSR`. On the client the promise is never thrown, so it gates nothing. A future maintainer reading this comment would reasonably (and wrongly) believe a client-side Suspense boundary exists for the customization load.

*Repro:* `grep -rn "initialCustomizationLoadPromise" packages/` shows the only `throw` site is SSR-guarded; the comment claims a client Suspense boundary.

*Suggested fix:* reword to reflect that the promise is consumed only during SSR — e.g. "Not armed on the client: the promise is only thrown (to suspend) during SSR, so on the client it is never awaited and a slow load simply applies the customization late, as it did before this change."

**No. 2 — One of the two SSR tests does not guard the fix.**
The task flagged watching for tests that always succeed. The positive test is genuine (verified above). Its companion, **"renders normally, with no customization CSS, when the record isn't found"**, passes on both the pre-fix and post-fix code — with the suspend disabled it stayed green while the positive test went red. It's still a worthwhile test (it proves a not-found record doesn't crash SSR or inject phantom colors), but it does not exercise the suspend/settle path the way the positive test does, so it shouldn't be counted as coverage for the fix itself. Optional: strengthen it, or just be aware the positive test is the one doing the real work.

---

### Nits / observations (non-blocking)

- **Every SSR render now awaits `chapterDataPromise`, not just `?customization=` renders.** The added `Promise.all([state.i18n.ready, ...chapterDataPromise])` in `entry-ssr.tsx` is a behavior change for all server renders. It's intentional and beneficial (it closes the race described in the comment and guarantees chapter data is in the SSR HTML), and it's bounded by the chapter's own 5s backstop, so this is fine — just noting it's broader than the customization feature.

- **`initialCustomizationLoadTimer` has no teardown path.** `CustomizationsManager` has no `dispose()`, so unlike `BibleReadingManager` (which clears its timer via `effectDisposers`), a leaked SSR timer would hold its closure until it fires. In practice the SSR render awaits the promise that clears the timer, so the window is negligible, and this matches the manager's existing (dispose-free) style. Mentioning only for completeness.

- **Meta-tag render after a customization suspend is untested.** `render()` does a second `renderToStringAsync` for the `og:*`/`twitter:*` meta after the main tree suspended and resumed on the customization promise. The customization test asserts on body content (`--sb-primary-color`, "Verse 1") but not that the meta block still renders correctly following a suspend. Given how much of this fix rides on `preact-render-to-string` + `@preact/signals` internals, a small assertion that a `?customization=` render still emits correct `og:title`/`description` would add cheap insurance against a future library upgrade quietly breaking the resume path.

- **Fragility, well-managed.** The fix depends on subtle, library-internal behavior (that `preact-render-to-string` doesn't call `options._catchError` for suspend-throws, and the resulting `@preact/signals` write freeze). The comments document this thoroughly and the design sidesteps it correctly — customization resolution is signal-independent, and `chapterDataPromise` is pre-resolved before any suspend. No change requested; the positive test is the guardrail that would catch a library regression here, which is exactly why keeping it effective (see No. 2) matters.

---

### Process note

This is a stacked PR — its base is `claude/seed-bible-customization-center-l6jfo7`, not `develop`. Reviewers should make sure the base lands first, and re-target/rebase as needed so this doesn't merge ahead of its foundation.

---

KallynGowdy:
Just tested and this does work (just not from the branch deployment because it requires the bundle to run on the server)

PR #1741, opened 2026-09-02
Make the reader header's translation a real button on both layouts

Blitzy:
## Summary

Makes the desktop reader header's translation control match the mobile one: the same muted pill, opening the translation picker directly. Both pills are now real buttons, so each is keyboard reachable and announces what it does, and the letters inside them sit optically centred.

## Problem

On desktop the translation was plain text inside the chapter title — `Genesis 1 / BSB` — and the whole `<h2>` opened the *book* list. So nothing signalled the text was a control, there was no way to reach the translation picker from the header at all, and it looked nothing like the pill mobile users see. It also printed the raw `translationId`, which for a translation served from a non-default endpoint is a full URL that would have splattered across the header.

Both pills were `<span>`s with an `onClick`. Neither was in the tab order, Enter and Space did nothing, and assistive tech announced them as ordinary text with no hint they were actionable.

The pill text also sat visibly high. A line box reserves room for descenders that an all-caps abbreviation like `BSB` never uses, and Newsreader — the default book-title font — is an extreme case: it leaves 0.265em below the baseline but only 0.065em above the caps, putting the ink 0.2em (4px at the desktop size) above centre. The other five selectable fonts are all within a pixel, so this reads as "something's subtly off" rather than as an obvious bug.

## Changes

- **`components/BibleReader/BibleReader.tsx`**: the desktop header's title and translation now sit in a `.sb-bible-reader-heading` flex row, with the translation as a `<button>` that calls `openTranslationSelector()`. The title's `<h2>` switches from calling `setOpen` directly to the existing `openBookSelector` helper — without that, opening the picker and dismissing the panel would leave the next title click showing translations instead of books. The mobile chip becomes a `<button>` too, and both take their visible short name and `aria-label` from one pair of values computed beside those two handlers.
- **`components/BibleReader/BibleReader.inline.css`**: one shared rule holds the pill's look for both layouts, with only size per layout, so the two can't drift. Adds hover/`:focus-visible` on desktop and `:active`/`:focus-visible` on mobile, and a `@supports (text-box: trim-both cap alphabetic)` block that trims the line box to the caps and compensates the padding so the pill keeps its size. Trimming is font-agnostic, so it holds for whatever fonts get added later, and browsers without it (Firefox, currently) render exactly what they render today. The block sits after both chips' own rules, since `@supports` adds no specificity.
- **`i18n/en.json`**: adds `change-translation`. English only, per the repo convention.
- **`test/unit/seed-bible/components/BibleReader.test.tsx`**: three regression tests, plus an optional `selectorState` argument to `createMobileState()` so the mobile chrome's own entry points into the selector can be exercised. Existing callers pass nothing and behave exactly as before.

## Test plan

- [x] Three new regression tests, each confirmed red against the pre-change component before being kept
- [x] `BibleReader.test.tsx` 76/76, `TabSlotReader.test.tsx` 31/31, `TutorialManager` + `BibleSelector` 39/39
- [x] `check:ts:client` clean; ESLint and stylelint clean on the touched files; Prettier matches
- [x] Desktop: pill against the title at each reader font size, narrow split pane, dark mode
- [x] Mobile: pill height and press feedback
- [x] Firefox: untrimmed fallback still looks right
- [x] Keyboard: Tab reaches both pills, Enter/Space opens the picker

### Desktop

_Before:_
image

_After:_
image

### Mobile

_Before:_
image

_After:_
image

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
@@ -77,7 +77,7 @@
   flex: 1 1 auto;
   min-width: 0;
   display: flex;
-  align-items: flex-start;
+  align-items: center;
   flex-wrap: wrap;
   gap: 0.625rem;
 }
@@ -119,8 +119,7 @@
    changes, to suit each header's title. Keep the two in step. */
 .sb-bible-reader-translation,
 .sb-bible-reader-mobile-header-translation {
-  display: inline-flex;
-  align-items: center;
+  display: inline-block;
   cursor: pointer;
   border: none;
   font-family: var(--text-bookTitle-font);
@@ -250,6 +249,27 @@
   font-size: 1rem;
   padding: 0.3125rem 0.625rem;
   border-radius: 0.9375rem;
+}
+
+/* A line box reserves room for descenders that an all-caps abbreviation never
+   uses, so it sits high in an evenly padded chip — badly so in Newsreader,
+   which leaves 0.265em below the baseline but only 0.065em above the caps.
+   Trimming the box to the caps centres the letters themselves, whichever font
+   the reader has chosen. The padding grows to hold the chip's size once the
+   box shrinks from the full 1em to the cap height. */
+@supports (text-box: trim-both cap alphabetic) {
+  .sb-bible-reader-translation,
+  .sb-bible-reader-mobile-header-translation {
+    text-box: trim-both cap alphabetic;
+  }
+
+  .sb-bible-reader-translation {
+    padding-block: 0.5625rem;
+  }
+
+  .sb-bible-reader-mobile-header-translation {
+    padding-block: 0.475rem;
+  }
 }
 
 .sb-bible-reader-mobile-header-play {
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
@@ -71,8 +71,19 @@
   border-bottom: 1px solid var(--sb-divider-color);
 }
 
-.sb-bible-reader-title {
+/* Chapter title and translation chip sit on one row so the chip stays beside
+   the chapter number instead of drifting over to the action cluster. */
+.sb-bible-reader-heading {
   flex: 1 1 auto;
+  min-width: 0;
+  display: flex;
+  align-items: flex-start;
+  flex-wrap: wrap;
+  gap: 0.625rem;
+}
+
+.sb-bible-reader-title {
+  flex: 0 1 auto;
   min-width: 0;
   font-family: var(--text-bookTitle-font);
   /* color: var(--text-bookTitle-color); */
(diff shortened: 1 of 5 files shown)


Blitzy:
> [!NOTE]

## Code Review — Make the reader header's translation a real button on both layouts

**Verdict: looks good, no blocking issues.** This is a clean, well-scoped change. The translation control becomes a real `<button>` on both layouts, keyboard-reachable and self-describing, and the desktop title click routes through `openBookSelector` so it always lands on the book list. Nothing here needs to change before merge — the notes below are low-severity and optional.

I ran the following locally against the PR head (`1f6d357`):

- `BibleReader.test.tsx` — **76/76 pass**
- `TabSlotReader.test.tsx` (integration, touched by this PR) — **31/31 pass**
- `check:ts:client` — clean (0 visible errors)
- ESLint + stylelint on the touched files — clean (the only ESLint warnings are pre-existing `translation-unused-keys` for unrelated keys; the new `change-translation` key is correctly detected as *used*)

### Tests are real, not always-green ✅

The task I care most about — that the three new tests actually exercise the fix rather than passing regardless — checks out. I swapped the base (`develop`, `bd8f3ab`) version of `BibleReader.tsx` back in while keeping the new test file, and all three failed as they should:

- _"opens the translation picker from the header's translation button"_ → **red** on base (`button.sb-bible-reader-translation` is null, because the desktop translation was a `<span>`).
- _"opens the translation picker from the mobile header's translation button"_ → **red** on base (same reason for the mobile chip).
- _"opens the selector on the book list when the title is clicked, even after the translation picker was left open"_ → **red** on base with `expected true to be false` — precisely the leftover-`selectingTranslation` bug the change fixes.

So each test is anchored to observable behavior (the rendered button, its text/`aria-label`, and the resulting selector state) and genuinely fails on the pre-fix code. Good regression coverage.

One thing worth calling out about how the tests are wired, for future readers rather than as a change request: the mobile test's `createMobileState(selectorState)` passing is load-bearing — `TabSlotReader` reads `state.selector` and forwards it as the `selectorState` prop, so without that wiring the button's `onClick` would hit `undefined`. The doc comment on `createMobileState` already explains this, which is nice.

---

### Low-severity notes

**No. 1 — The URL-splatter fallback isn't fully closed.**
`BibleReader.tsx` computes:

```ts
const translationLabel =
  translation.value?.shortName ?? translationId.value ?? "";
```

The PR description frames the old bug as the header printing the raw `translationId`, "which for a translation served from a non-default endpoint is a full URL that would have splattered across the header." That's largely fixed — normally `translation.value.shortName` wins. But the fallback chain still lands on `translationId.value` whenever `translation.value` is null (e.g. the brief window before the translation object has loaded) or a loaded translation lacks a `shortName`. In the exact non-default-endpoint case the description calls out, that fallback is the URL again.

- _Reproduction (hard, hence low):_ point the reader at a translation whose id is a URL and whose object has no `shortName` (or catch the pre-load render frame) — the pill shows the URL.
- _Suggested fix:_ if loaded translations are guaranteed to carry a `shortName`, this is only a transient-loading concern and fine to leave; otherwise consider a friendlier last-resort fallback (a derived short code, or eliding the URL) rather than the raw id.

**No. 2 — The adjacent book-title control is still not keyboard-reachable (accessibility).**
This PR carries the `area:accessibility` label and makes the *translation* a real button on both layouts — great. The neighboring control that opens the *book* picker, though, remains a clickable non-button: `<h2 {...flingSafeTapHandlers(openBookSelector)}>` on desktop and `<span onClick={openBookSelector}>` on mobile. Neither is in the tab order or announces as actionable. This is pre-existing and arguably outside this PR's scope, but given the accessibility framing it's the obvious next follow-up — a keyboard user still can't reach the book selector from the header. Worth a tracking issue if there isn't one.

**No. 3 (nit) — `stopPropagation` asymmetry between the two buttons.**
The mobile translation button calls `e.stopPropagation()`; the desktop one doesn't. Both behave correctly — on desktop the click harmlessly bubbles to the pane shell's `onClick={() => app.selectSlot(slot.id)}`, which just marks the slot active — so this isn't a bug. Only noting it so the next reader doesn't wonder whether the omission on desktop was intentional. A one-line comment, or matching them, would remove the question.

---

Nice attention to detail on the `text-box: trim-both` optical-centering work and the shared-rule-plus-per-layout-size structure that keeps the two pills from drifting. The async ordering in `openTranslationSelector` (await `setOpen`, *then* set `selectingTranslation = true`) is also correct and deliberate — `syncStateFromSlot()` resets that flag to `false` inside `setOpen`, so setting it afterward is the only ordering that sticks.

---

PR #1746, opened 2026-09-03
Add content-language meta tag matching og:locale

KallynGowdy:
## Summary

Adds a `content-language` HTTP-equivalent meta tag to server-rendered pages, with its value synchronized to the page's `og:locale` meta tag. This ensures search engines and browsers have explicit information about the page's language.

## Changes

- **`standalone/entry-ssr.tsx`**: Added `<meta httpEquiv="content-language" content={state.i18n.language.value} />` to the SSR'd HTML head, placed immediately after the description meta tag and before the og:locale tag.
- **`test/unit/standalone/entry-ssr.test.ts`**: Added test case verifying that the content-language meta tag is emitted and its value matches the og:locale property value.

## Implementation Details

The `content-language` meta tag uses the same `state.i18n.language.value` signal that drives the `og:locale` property, ensuring they stay in sync. The tag is placed early in the head for clarity and follows the existing pattern of other meta tags in the SSR template.

https://claude.ai/code/session_019Y55fmdycJPYi1qbXkXtmu

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html>
+<html lang="<!-- HTML_LANG -->">
   <head>
     <meta charset="UTF-8" />
     <meta
--- a/standalone/entry-ssr.tsx
+++ b/standalone/entry-ssr.tsx
@@ -487,6 +487,7 @@ export async function render(
         media="(prefers-color-scheme: dark)"
       />
       <meta name="description" content={state.app.description.value} />
+      <meta httpEquiv="content-language" content={state.i18n.language.value} />
       <meta property="og:locale" content={state.i18n.language.value} />
       <meta
         property="og:locale:alternate"
--- a/standalone/entry-ssr.tsx
+++ b/standalone/entry-ssr.tsx
@@ -58,6 +58,8 @@ export interface RenderOptions {
    *   CSS text should be injected, for the pre-hydration script that applies
    *   a returning visitor's saved theme before first paint.
    * - `<!--META-->` where any additional meta tags should be injected (optional).
+   * - `<!-- HTML_LANG -->` inside the root `<html lang="...">` attribute,
+   *   where the detected page language should be injected.
    *
    * The host server loads this from disk at startup and passes it to the render function on each request, allowing it to be customized or overridden per request if needed.
    * By default, it is just the contents of `index.html` in the project root.
@@ -66,6 +68,17 @@ export interface RenderOptions {
 }
 
 const escapeForScript = (json: string): string => json.replace(/</g, "\\u003c");
+
+/**
+ * Unlike the `<meta>`/`<link>` values above (rendered through Preact, which
+ * escapes attribute values automatically), `HTML_LANG` lands inside a
+ * `lang="..."` attribute via a raw string substitution into the static
+ * template — and the language it carries can trace back to an unvalidated
+ * `?lang=` query param (see `getUrlLanguage`). Escape it explicitly so a
+ * crafted value can't break out of the attribute.
+ */
+const escapeForHtmlAttribute = (value: string): string =>
+  value.replace(/&/g, "&amp;").replace(/"/g, "&quot;");
 
 /**
  * Excludes the full multi-translation catalog from what gets embedded in the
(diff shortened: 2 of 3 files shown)


Blitzy:
> [!NOTE]

## Code Review — Add content-language meta tag matching og:locale

**Overall:** Small, focused, and low-risk. The code does what it says: `content-language` is emitted from the same `state.i18n.language.value` signal as `og:locale`, so the two are guaranteed to stay in sync. The test is a genuine regression test — I reverted the one-line source change locally and confirmed the new test goes red (`expected '<!doctype html>…' to contain '<meta http-equiv="content-language" content="en"'`), then green again once restored. It is not a test that always passes.

My main reservation is not about correctness but about **whether this is the right mechanism for the stated goal**. Details below, by severity.

---

### Medium — The tag being added is obsolete in HTML5, and the more effective fix (`<html lang>`) is missing

The PR's stated goal is to give "search engines and browsers explicit information about the page's language." Two problems:

1. **`<meta http-equiv="content-language">` is non-conforming in the HTML Living Standard.** The spec's pragma-directives section explicitly documents this state as discouraged: *"Conformance checkers will include a warning if this pragma is used. Authors are encouraged to use the `lang` attribute instead."* The W3C Nu validator emits: *"Using the meta element to specify the document-wide default language is obsolete. Consider specifying the language on the root element instead."* Google has also stated it ignores this tag when determining page language (it uses on-page content and `hreflang`). So the tag is unlikely to actually advance the SEO goal, and it introduces a validator warning on every SSR'd page.

2. **The page's `<html>` element has no `lang` attribute at all — and that is the signal that actually matters.** The root template (`index.html:2`) is a bare `<html>`, and the SSR pipeline in `standalone/entry-ssr.tsx` never patches a `lang` onto it. The `lang` attribute is the standards-based, accessibility-critical signal: screen readers select pronunciation/voice from `<html lang>`, not from a `content-language` meta tag, and it is the conforming way to declare document language. Given the file already blocks on `state.i18n.ready` specifically so the HTML is rendered in the right language (see the comment at `entry-ssr.tsx:466-469`), serving that HTML with an unmarked `<html>` is the bigger gap.

**Reproduction:** Build/serve any SSR page (e.g. `/en/AAB/genesis/1`) and run it through the [W3C Nu validator](https://validator.w3.org/nu/). You'll get the "content-language … is obsolete" warning, and separately you can confirm `<html>` carries no `lang`.

**Suggested fix:** Set the `lang` attribute on the root element instead of (or in addition to) the meta tag. Since `index.html` is a static template filled by `replacePlaceholder`, the cleanest route is to add a substitution — e.g. make the template `<html lang="<!-- HTML_LANG -->">` and push `["<!-- HTML_LANG -->", state.i18n.language.value]` into the `substitutions` array next to `<!-- META -->`. That satisfies the SEO/browser goal the conforming way and fixes the accessibility gap at the same time. If you have a specific reason to keep `content-language` too (legacy crawler support), it's harmless to leave — but it shouldn't be the primary mechanism.

---

### Low — The test only exercises the default-language path, so it can't catch a "content-language pinned to default" regression

`test/unit/standalone/entry-ssr.test.ts` renders `/en/AAB/genesis/1`, where the detected language, the default language, and the emitted value are all `"en"`. The test asserts `content-language === og:locale`, which is good, but because every value in play is `"en"`, it would still pass if someone later changed `content-language` to read from `state.i18n.defaultLanguage` (a constant) while `og:locale` continued to track the detected language — the exact "they drifted apart for non-default locales" bug the test is meant to guard against.

**Suggested fix:** Add (or switch to) a non-default locale so the assertion proves they track the *detected* language, not just a shared constant — e.g. render a `/es/…` route (or drive detection via `Accept-Language`) and assert both `og:locale` and `content-language` equal `"es"`. That makes the synchronization guarantee real for the case that actually varies.

---

### Note (not a blocker, pre-existing) — og:locale value format

Both tags now emit short codes like `en`. The Open Graph spec technically expects `og:locale` in `language_TERRITORY` form (e.g. `en_US`). This is pre-existing behavior that this PR only mirrors, not something it introduces, so it's out of scope here — flagging only for awareness in case locale formatting is ever revisited.

---

Issue #1737, opened 2026-09-01
Add support for custom settings for extensions


KallynGowdy:
Currently, there is no standardized way to configure extensions. You can install them, but if you want to configure a setting in an extension, then the extension has to provide a way to manage that itself. This task is to add support for this, so that extensions can specify what settings they support so that a user (or customization) can configure them.

Take inspiration from [VSCode extension configuration](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration). Extensions should support a `settings` object where each key is a setting that can be configured by the user. For now, we will support three types:
- `string`
- `boolean`
- `number`

Additionally, the `string` type should support a `sensitive` boolean flag, which indicates that the value for the setting should be stored securely (i.e. configuring API keys).

KallynGowdy:
Implemented by #1747

PR #1747, opened 2026-09-03
Add support for custom settings for extensions

KallynGowdy:
## Summary

Implements #1737: extensions can now declare a `settings` map in `extension.json` (VSCode-`contributes.configuration`-style), and users — plus Customization authors — get a standard way to configure those values instead of every extension building its own.

- Extensions declare `settings: Record<string, { type: "string" | "boolean" | "number"; default?: ... }>` in `extension.json`. This ships alongside `dependencies`/`autoinstall` in the boot-time trimmed meta, since it's structural rather than per-locale text.
- A new `ExtensionSettingsManager` stores each viewer's own values per extension via CasualOS records (`publicRead`, matching every other manager's convention), keyed by the viewer's own user id.
- `CustomizationsManager` gained a new `extensionSettingDefaults` field (deliberately separate from its existing `extensionSettings` availability field, to avoid a naming collision) so a Customization can preset defaults for everyone using it.
- Resolution order for a setting's effective value: **viewer's own value → active Customization's default → the setting's own `default`**.
- A shared `ExtensionSettingsForm` component renders the typed fields (text/checkbox/number) and is reused by a new "Configure" action in Settings → Extensions and a "Configure defaults" action in the Customization editor's extensions list.
- The example extension (`seed-bible-refresh-example-extension`) now declares two sample settings (`greeting`, `showBanner`) so the feature has a real, testable extension to exercise end-to-end.

**Scope note:** the issue also asks for a `sensitive` flag on `string` settings (for secrets like API keys), stored securely. This codebase has no secure-storage primitive today — every existing record uses the `publicRead` marker. Building that properly deserves its own design pass, so this PR ships `string`/`boolean`/`number` only; `sensitive` is left for a follow-up.

## Test plan

- [x] `pnpm check:ts` — passes
- [x] `pnpm lint` — 0 errors (pre-existing warnings for other-locale translation gaps are unrelated to this change)
- [x] `pnpm test` — full suite passes (5615 tests), including new/extended coverage in `ExtensionSettingsManager.test.ts`, `extensionsModule.test.ts`, and `CustomizationsManager.test.ts`
- [x] Manual: install the example extension, open Settings → Extensions, use the new "Configure" action to set `greeting`/`showBanner`, reload and confirm persistence; then set a Customization default for the same extension and confirm it applies when no personal override is set

https://claude.ai/code/session_012oSVQPmvKLGPFTaKEn4vJj

---
_Generated by [Claude Code](https://claude.ai/code/session_012oSVQPmvKLGPFTaKEn4vJj)_


Blitzy:
> [!NOTE]

## Review of No. 1747 — Custom settings for extensions (implements No. 1737)

Overall this is a well-structured, well-documented change. The resolution order (viewer value → Customization default → setting default) is clearly modeled, the new `ExtensionSettingsManager` follows the app's record conventions, and the manager/module tests genuinely exercise real behavior rather than always passing (see the tests section below). The scope note about deferring `sensitive` is reasonable given there's no secure-storage primitive yet. The findings below are mostly about edge cases; none are blockers, but the first is worth addressing before merge.

Note: the PR targets the stacked branch `claude/seed-bible-customization-center-l6jfo7`, not `develop`, so this review is against that base.

---

### Medium

**M1 — Account switch can surface (and persist) the previous user's setting values.**

`ExtensionSettingsManager`'s load effect only clears `valuesByExtensionId` when the user becomes signed-out (`!userId`). On a **direct** account switch from user A to user B without an intervening signed-out state, the effect goes straight to `void load(userId)` and leaves A's in-memory values in place until B's `getData` round-trip resolves:

```ts
effect(() => {
  const userId = login.userId.value;
  if (!userId) { valuesByExtensionId.value = {}; loadedUserId.value = null; return; }
  if (loadedUserId.value === userId) return;
  void load(userId); // A's values remain in memory until this resolves
});
```

During that window two things can go wrong for user B:
1. `getValue` returns A's values in B's UI.
2. If B writes during the window, `persist()` merges into the still-stale `valuesByExtensionId.value` and records the merged blob — including A's values — to **B's** record, corrupting B's data.

This is exactly the case `HighlightsManager` guards against: its account-change effect clears the previous account's entries **synchronously** (`HighlightsManager.tsx:487`), and its writes target `entry.userId` rather than "whoever is signed in now." This new manager doesn't follow either half of that pattern.

Scope: the common logout→login path passes through `userId === null` first, which clears correctly, so the bug only bites on a direct A→B transition (e.g. multi-account switching). It's narrow, but it's a data-integrity issue rather than cosmetic.

*Reproduction (conceptual):* sign in as A, set an extension value, then switch directly to B (no signed-out step) on a slow connection; before B's load resolves, B's settings modal shows A's values, and a save by B writes A's values into B's record.

*Suggested fix:* clear `valuesByExtensionId` synchronously whenever `loadedUserId` no longer matches the current user (not only on sign-out), and/or capture the target user id in `persist`/`setValue` and skip the write if `login.userId.value` has changed since — mirroring the `entry.userId` approach in `HighlightsManager`.

---

### Low

**L1 — The number field can't accept decimals (and re-formats mid-edit).**

In `ExtensionSettingsForm.tsx`, the numeric input is controlled by `value={String(value)}` while `onInput` parses each keystroke with `Number(raw)` and calls `onChange(parsed)`:

```tsx
value={value === undefined ? "" : String(value)}
onInput={(event) => {
  const raw = event.currentTarget.value;
  if (definition.type === "number") {
    const parsed = Number(raw);
    if (raw === "" || Number.isNaN(parsed)) return;
    onChange(key, parsed);
  } ...
```

Because the displayed value is re-derived as `String(Number(raw))` on every keystroke, an in-progress decimal is destroyed: typing `1` then `.` yields `Number("1.") === 1`, `onChange(1)`, and the controlled re-render rewrites the field to `"1"`, dropping the `.` the user just typed. Result: users can't enter values like `1.5` for a `number` setting. Integer settings (the example extension's `maxEmbers`/`count`) are unaffected, which is why it isn't caught today.

*Reproduction:* declare a `number` setting, open its Configure form, try to type `1.5` — the decimal point won't "stick."

*Suggested fix:* keep a local string buffer for the raw input and only coerce to a number on blur / when it parses cleanly, or drop the controlled re-formatting so the field shows the raw text the user typed.

**L2 — Persist is optimistic with no rollback, and callers fire-and-forget the promise.**

`persist()` sets `valuesByExtensionId.value = next` *before* awaiting `os.recordData(...)`, and both call sites invoke it as `void extensionSettings.setValue(...)` / `void extensionSettings.clearValue(...)`. If the record write rejects (network/permission), the in-memory state stays "saved," the UI shows success, and the rejection is an unhandled promise rejection because `void` attaches no `.catch`. The value silently reverts on the next reload. Consider surfacing the failure (and/or rolling back the optimistic update) so a failed save isn't invisible.

**L3 — No component-level test for `ExtensionSettingsForm`.**

The manager and build-module logic are tested, but the form component — which is where the L1 number-parsing bug lives and which owns the checkbox/text/number rendering and the reset-button visibility — has no test. Given the project's emphasis on testing observable behavior, a small render test around the number/boolean field handling and the reset action would have caught L1 and would guard the resolution wiring (`getValue`/`onChange`/`hasOwnValue`).

---

### Informational

**I1 — `sensitive` deferral (agree, with a follow-up guardrail).** The PR correctly defers the `sensitive` flag the issue asks for, since values are stored with the `publicRead` marker (consistent with every other manager). Just flagging for the follow-up: `publicRead` means these values are readable by anyone who knows the record name (the user id), so the eventual `sensitive` work must not simply reuse this record/marker for secrets. No action needed here.

**I2 — Brief note:** in the Customization "Configure defaults" form, toggling a boolean off writes an explicit `false` default (distinct from "no default set"); clearing back to "no default" requires the reset button. That's a defensible design, just slightly non-obvious — not a defect.

---

### Tests assessment

The tests read as real-behavior tests, not tautologies:

- `ExtensionSettingsManager.test.ts` covers the full resolution precedence (own → Customization → extension default), signed-out no-ops, unknown-extension/key guards, persistence payload/marker shape, and reload-on-user-change. The corrupt-payload test is a genuine regression test — with the schema reverted to `z.any()` the corrupt blob would be stored and surfaced, so the test would fail as written.
- `extensionsModule.test.ts` asserts `settings` is both inlined into the trimmed meta and omitted (not emitted as `undefined`) when absent — both would fail if `trimMeta` were wrong.
- `CustomizationsManager.test.ts` verifies draft-only mutation (no `recordData` call), clear/no-op paths, and active-vs-unset resolution.

The one gap is L3 (no `ExtensionSettingsForm` component test), and there's no test asserting that a mismatched-type stored value is ignored by `getValue` (the type guard exists but is uncovered) — optional to add.

---

Nice work on the clear separation between per-viewer values and Customization defaults, and on keeping `settings` inline in the trimmed meta with a documented rationale. Addressing M1 (and ideally L1) would get this in good shape.

---

Blitzy:
@KallynGowdy regarding the `sensitive` flag for `string` property types, what do you think would be the best solution here? Clearly we dont want to store API keys or other sensitive codes and such in a `publicRead` record? Is there any reason that extension configurations should even be `publicRead`? Should we just make them private and rely on CasualOS private record security for protecting that kind of information?

KallynGowdy:
Yeah it's not actually that well thought out. The idea behind sensitive flags is that it would keep them server side and set something up so the server could proxy requests so the secrets never travel back to the client.

But halfway through writing it I realized that it would need more configuration which is why this PR doesn't handle sensitive values yet

Blitzy:
Update on what's landed since the note about `sensitive`. Everything below is pushed to this branch.

### Review follow-ups

**L1 — number fields now accept decimals.** The field re-rendered with the parsed number on every keystroke, so typing `1.0` on the way to `1.05` collapsed back to `1`. It now keeps what you typed while the field has focus and normalizes on blur.

**L2 — failed saves are no longer silent.** The records client reports a refused write by resolving with `success: false` rather than throwing, so the old code missed most real failures. Saves now check the result, and the Configure window shows "Couldn't save your settings." The change stays applied for the session and the next successful save stores it, matching how reading-plan draft autosave behaves. Writes are also serialized now, since a number field saves on every keystroke and an older save could otherwise land after a newer one.

**L3 — done.** `ExtensionSettingsForm` now has component tests covering the text, number and checkbox fields, the reset action's visibility and behaviour, the empty state, the override checkbox, and the default note. There are also tests for the extensions view covering the save-error message and the sign-in prompt.

**I2 — resolved by the override checkbox** described below: switching a toggle off used to store an explicit "off" that was indistinguishable from storing nothing.

### Decisions from the discussion

**Records stay `publicRead`**, consistent with every other record in the app. No change in this PR.

**`sensitive` is out of scope here** and gets its own issue, per Kal's note that it needs a server-side store plus a proxy so secrets never reach the client.

**Signed-out viewers are asked to log in.** Extension settings live in the viewer's own record, so Configure now shows a sign-in prompt instead of a form that silently discards changes. It switches to the form once you sign in, without reopening. Offline/local settings are a deliberate follow-up rather than part of this PR, and there's a TODO at that prompt.

**Setting constraints are tracked in #1835** (`minimum`/`maximum`/`multipleOf`, `enum`), since a setting can currently declare only a type and a default.

### Bug found while testing

The "Configure defaults" window read defaults from the *active* customization while writing them to the customization being *edited*. Editing any customization other than the one you're currently using — including one you just created — left the fields empty and the reset action hidden even though values were being saved. Fixed in `2b58141`.

### Changes to the settings UI

**The example extension actually uses its settings now.** `greeting` and `showBanner` were declared but never read, so changing them did nothing. Its panel now shows the greeting and the banner. It also declares `greetingSize` (number) and `subtitle` (string with no default), so all three supported types and the "no default" case can be exercised end to end.

**The Configure form was reworked for legibility:** each setting's description sits under its title (it used to sit below the field, at a larger size than the title it belonged to), and the reset action is inline with the title.

**The customization window is now "default overrides".** Each setting shows "Default value: …" (or "No default") plus an explicit Override checkbox. Unchecked leaves the extension's default in place and hides the field; checked stores a value and opens a field seeded from that default. Unchecking is the reset, so the separate reset action is gone there.

_Customization defaults override window for example extension:_
image

_User-level settings for example extension:_
image

Blitzy:
> [!NOTE]

## Review of No. 1747 (re-review after the No. 3 review's follow-ups landed)

This PR was just marked ready for review again after the `2b58141`-and-later commits addressed the earlier M1/L1/L2/L3 findings and the `sensitive`/`publicRead` discussion. I re-checked those fixes against the current head (`1c4dbf78`) and they hold up:

- **M1 (account-switch data leak)** — fixed: the load `effect` now clears `valuesByExtensionId`/`saveError` synchronously on any user-id change, not just on sign-out.
- **L1 (decimal input)** — fixed: `NumberSettingInput` now keeps a local `draft` string while focused and only clears it on blur, so `1.05` types cleanly.
- **L2 (silent save failures)** — fixed: `write()` checks `result.success` and no longer relies on a throw, and the Configure modal now surfaces `saveError`.
- **L3 (missing component tests)** — addressed with new `ExtensionSettingsForm` tests.

Found a few new issues while reviewing the rest of the diff (the CSS change, the schema, and the new save-error wiring). Nothing here blocks the earlier fixes from being correct — these are new gaps introduced by the same batch of changes.

---

### Medium

**No. 1 — `saveError` is shared across every extension's Configure modal, not scoped per extension.**

`ExtensionSettingsManager.tsx` keeps `saveError` as one manager-wide signal, set in `write()`:
```ts
if (loadedUserId === userId) {
  saveError.value = failed;
}
```
`SettingsPage.tsx` (~line 1272) renders that same flag inside *any* extension's Configure modal:
```tsx
{extensionSettings.saveError.value && (
  <p className="sb-settings-save-error" role="alert">
    {t("extension-settings-save-failed", ...)}
  </p>
)}
```
*Reproduction:* Open Extension A's Configure modal, force a save to fail (e.g. a network blip while editing a text field). Close the modal without it recovering, then open Extension B's Configure modal — B has never had a save attempt, but it immediately shows "Couldn't save your settings," because the flag isn't cleared by opening a different extension's modal, only by the next successful save (of any extension) or a sign-in/out.

*Suggested fix:* key `saveError` by `extensionId` (e.g. `Record<string, boolean>`), or at minimum clear it when `handleConfigureExtension` opens a modal for a different extension than the one that last failed.

**No. 2 — Removing shared padding from `.sb-settings-toggle-row` is a page-wide layout regression, not a scoped fix.**

The diff removes `padding: 0.375rem 0.25rem;` from `.sb-settings-toggle-row` in `SettingsPage.css`:
```diff
   justify-content: space-between;
   gap: 0.75rem;
-  padding: 0.375rem 0.25rem;
 }
```
That class isn't new or specific to extension settings — it's reused by ~10 pre-existing toggle rows in `SettingsPage.tsx` (show headings, show verse numbers, footnotes, highlights, dark mode, the toolbar-config toggle, etc.) plus the one new boolean-setting row in `ExtensionSettingsForm.tsx`. Removing the padding to fix spacing for the new boolean field also strips the vertical/horizontal inset from every one of those unrelated existing rows.

*Reproduction:* Open Settings and compare any of the pre-existing toggle rows (e.g. "Show verse numbers") on `develop` vs. this branch — they sit tighter against the section edges and each other on this branch.

*Suggested fix:* give the new boolean setting row its own class (or a modifier) for the tighter spacing instead of editing the shared `.sb-settings-toggle-row` rule.

---

### Low

**No. 3 — `ExtensionMetaSchema` wasn't updated to validate the new `settings` field.**

`script/lib/extension.ts`'s `ExtensionMetaSchema` (used by `upload()` to validate `extension.json` before it goes to the records server) still only types `id`, `translations`, `dependencies`, `autoinstall` and is `z.looseObject`, so `settings` passes through completely unchecked:
```ts
export const ExtensionMetaSchema = z.looseObject({
  id: z.string(),
  translations: z.object({ en: ExtensionTranslationSchema }).catchall(ExtensionTranslationSchema),
  dependencies: z.array(z.string()).optional(),
  autoinstall: z.boolean().optional(),
});
```
This file wasn't touched anywhere in the PR's diff, so it looks like an oversight rather than a deliberate scope cut.

*Reproduction:* Give an extension a malformed `settings` entry in `extension.json` (e.g. `{"type": "strnig", "default": "x"}`, or a `number` setting with a string default) and run the upload script — it passes schema validation silently; the mismatch only surfaces later as a broken/blank field in `ExtensionSettingsForm` at runtime, with no diagnostic pointing back at the bad `extension.json`.

*Suggested fix:* add a `settings` shape to `ExtensionMetaSchema` mirroring `ExtensionSettingDefinition` (discriminated on `type`, with `default` typed to match), so a bad `extension.json` fails at upload time instead of at render time.

**No. 4 — Text/number settings write on every keystroke with no debounce.**

`ExtensionSettingsForm`'s text and number fields call `onChange` → `ExtensionSettingsManager.setValue` on every `onInput`, and `setValue` always calls through to `persist`, which queues a real `os.recordData` write via `saveChain`. There's no debounce, unlike `CustomizationsManager`'s draft editing, which coalesces edits with a 5-second `scheduleAutoSave()`.

This isn't a correctness bug (writes are correctly serialized in-order by `saveChain`, per the L2 follow-up), just an efficiency gap: typing a 20-character `subtitle` value fires 20 sequential network writes instead of one after the viewer pauses.

*Suggested fix:* debounce `onChange` for the text/number cases (e.g. reuse the pattern from `CustomizationsManager.scheduleAutoSave`), keeping the boolean checkbox's immediate save as-is.

---

### Not flagged (checked, held up)

- The `publicRead` marker decision and deferring `sensitive` — already discussed and decided in the thread with Kal; no new concerns.
- The "Configure defaults reads active vs. writes editing customization" bug mentioned as fixed in `2b58141` — confirmed fixed, `getActiveExtensionSettingDefault` calls in the defaults form now consistently target the customization being edited.
- `NumberSettingInput`'s draft/blur handling and the account-switch effect's ordering — both correct under review.

None of the four issues above are blockers on their own — No. 1 and No. 2 are the ones worth fixing before merge since they're both user-visible regressions (a false error banner, a page-wide spacing shift); No. 3 and No. 4 are safe to follow up on.

---

Blitzy:
All four findings are fixed and pushed.

**No. 1 — error messages no longer appear on the wrong extension** (`20f6cc8`). If saving one extension's settings failed, every other extension's Configure window also said "Couldn't save your settings". Each extension now tracks its own failure. One thing to note: once any save succeeds, the message clears everywhere, because all extensions share a single record — so a save that lands also stores whatever the earlier failed one didn't.

**No. 2 — the spacing change no longer affects the rest of Settings** (`5867e6e`). The shared toggle row style is back to what it is on `develop`, so all the existing rows look as they did. The extension settings form removes that spacing only on its own rows, where the row around them already provides it.

**No. 3 — a bad `extension.json` now fails when it's uploaded** (`550969b`). Settings were never checked, so a misspelled type or a default that doesn't match its type slipped through and only showed up later as a blank field. They're checked now. Unknown fields are still kept rather than dropped, so settings can carry things this script doesn't know about yet, like the constraints in #1835. All 14 existing extension files still pass.

**No. 4 — typing no longer saves on every keystroke** (`7b22727`). Text and number changes wait until you stop typing for 700ms and then save once. Toggles and resets still save straight away, since each is a single deliberate change. 700ms rather than the 5 seconds the customization editor uses, because there's no way to save on closing the window yet, and a longer wait means more to lose if you close the tab. A waiting change is also saved before you switch accounts, and when you leave the page.

Each fix has tests, and I checked each one fails against the old code.

KallynGowdy:
This seems good enough for a first pass

PR #1750, opened 2026-09-03
Add audio timing support to Bible API and managers

KallynGowdy:
## Summary

This change adds support for per-verse audio timings across the Bible API and data managers. Audio timings are timestamps (in seconds) marking when each verse ends in a reader's audio file, enabling features like synchronized highlighting during playback.

## Key Changes

- **New API interfaces and method:**
  - `AudioTimings` — represents a single reader's per-verse timings for one chapter (fetched from `*.audioTimings.json` files)
  - `TranslationBookChapterAudioTimingsLinks` — maps reader names to timing file URLs in chapter responses
  - `CompleteTranslationChapterAudioTimings` — inlines timing values directly in complete-translation downloads (for offline use)
  - `FreeUseBibleAPI.getAudioTimings()` — fetches a reader's timing file from a given link

- **Updated chapter data structures:**
  - `TranslationBookChapter` now includes `thisChapterAudioTimings`, `nextChapterAudioTimings`, and `previousChapterAudioTimings` (links to timing files)
  - `CompleteTranslationChapter` now includes `thisChapterAudioTimings` (inlined timing values)

- **Manager support:**
  - `BibleDataManager.getAudioTimings()` resolves timing links against the correct translation endpoint
  - `OfflineTranslationsManager` returns empty timing links for offline chapters (timings are fetched live, not stored)

- **Test coverage:**
  - Unit tests for `FreeUseBibleAPI.getAudioTimings()` with endpoint override support
  - Unit tests for `BibleDataManager.getAudioTimings()` endpoint resolution
  - Mock data helper `makeAudioTimings()` for test fixtures
  - Updated existing test fixtures to include the new timing fields

## Implementation Details

Timing files are kept separate from chapter responses (linked rather than inlined) to keep chapter payloads small when timing data isn't needed. The complete-translation download is an exception — it inlines timings for all readers to remain self-contained for offline use, avoiding extra round trips per chapter per reader.

Offline translations never provide timing links, since timings are fetched live from the API rather than stored in downloads.

https://claude.ai/code/session_01NNTNnkmZkQ7xGgDb5e7Vdx

### DIFF
--- a/packages/seed-bible/seed-bible/managers/BibleDataManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/BibleDataManager.tsx
@@ -3,6 +3,7 @@ import { safeLocalStorage } from "../app/ssrEnv";
 import {
   FreeUseBibleAPI,
   type ApiRequestOptions,
+  type AudioTimings,
   type Translation,
   type TranslationBook,
   type TranslationBookChapter,
@@ -111,6 +112,20 @@ export interface BibleDataManager {
     chapter: TranslationBookChapter,
     options?: ApiRequestOptions
   ) => Promise<TranslationBookChapter | null>;
+
+  /**
+   * Loads a single reader's per-verse audio timings for a chapter.
+   *
+   * @param translationId The translation the chapter belongs to, used to
+   * resolve which API endpoint to read the link from.
+   * @param link A URL from that chapter's `thisChapterAudioTimings` (or
+   * `nextChapterAudioTimings`/`previousChapterAudioTimings`) map.
+   */
+  getAudioTimings: (
+    translationId: string,
+    link: string,
+    options?: ApiRequestOptions
+  ) => Promise<AudioTimings>;
 
   /**
    * Gets the API endpoint associated with a given translation. If the translation is not associated with a specific endpoint, it returns the default endpoint.
@@ -1173,6 +1188,15 @@ export function createBibleDataManager(
     return await api.getPreviousChapter(chapter, endpoint, options);
   };
 
+  const getAudioTimings = async (
+    translationId: string,
+    link: string,
+    options?: ApiRequestOptions
+  ): Promise<AudioTimings> => {
+    const endpoint = getEndpointForTranslation(translationId);
+    return await api.getAudioTimings(link, endpoint, options);
+  };
+
   const buildTranslationId = (translationId: string) => {
     const endpoint = getTranslationEndpointInfo(translationId);
     if (endpoint.isDefault) {
@@ -1262,6 +1286,7 @@ export function createBibleDataManager(
     getTranslationBookChapter,
     getNextChapter,
     getPreviousChapter,
+    getAudioTimings,
     getTranslationEndpointInfo,
     buildTranslationId,
     hydrateCachedCatalog,
--- a/packages/seed-bible/seed-bible/managers/FreeUseBibleAPI.tsx
+++ b/packages/seed-bible/seed-bible/managers/FreeUseBibleAPI.tsx
@@ -234,6 +234,13 @@ export interface TranslationBookChapter {
   thisChapterAudioLinks: TranslationBookChapterAudioLinks;
 
   /**
+   * Per-reader links to this chapter's audio-timing files. Fetch one with
+   * {@link FreeUseBibleAPI.getAudioTimings} to get that reader's per-verse
+   * timestamps.
+   */
+  thisChapterAudioTimings: TranslationBookChapterAudioTimingsLinks;
+
+  /**
    * The link to the next chapter.
    * Null if this is the last chapter in the translation.
    */
@@ -246,6 +253,12 @@ export interface TranslationBookChapter {
   nextChapterAudioLinks: TranslationBookChapterAudioLinks | null;
 
   /**
+   * Per-reader links to the next chapter's audio-timing files.
+   * Null if this is the last chapter in the translation.
+   */
+  nextChapterAudioTimings: TranslationBookChapterAudioTimingsLinks | null;
+
+  /**
    * The link to the previous chapter.
    * Null if this is the first chapter in the translation.
    */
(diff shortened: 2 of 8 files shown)


Blitzy:
> [!NOTE]

## Code review — Add audio timing support to Bible API and managers

**Verdict:** Solid, well-documented, purely additive plumbing. Types compile (`pnpm check:ts` — 0 visible errors) and the full suite passes (`pnpm vitest run FreeUseBibleAPI BibleDataManager OfflineTranslationsManager` — 686 tests green). The new tests exercise real URL/endpoint behavior rather than rubber-stamping (see the Tests section). I found one design inconsistency worth resolving before consumers are built on top of this, plus a few low-severity notes.

No linked issue was referenced in the PR description, so this review is against the PR's stated goals. (The `#1607` references in the touched test files belong to the unrelated audio play-button work, not this change.)

---

### Medium — Offline downloads throw away the inlined timings that `complete.json` exists to carry

The two rationales baked into this PR contradict each other:

- `CompleteTranslationChapter.thisChapterAudioTimings` inlines the actual timing **values** *"so it [the complete-translation download] can remain self-contained for offline use"* (interface doc + PR summary).
- `OfflineTranslationsManager` then returns an empty map for every offline chapter, because *"timings are fetched live from the API rather than stored in downloads."*

In the current code the second wins and the first is dead weight: a downloaded `complete.json` contains per-verse timings, but `toChapterEntries()` (`OfflineTranslationsManager.tsx` ~L419-437) persists only `{ numberOfVerses, thisChapterAudioLinks, chapter }` and drops `thisChapterAudioTimings`. `StoredChapter` (`OfflineTranslationStore.ts` ~L85-94) has no field to hold them anyway, and the read path (`OfflineTranslationsManager.tsx` ~L962) hardcodes `thisChapterAudioTimings: {}`. Since `BibleDataManager.getAudioTimings()` is network-only (it always calls `api.getAudioTimings`, never the offline store), a fully-offline reader can never obtain timings — even for a chapter whose timings were physically downloaded. So the feature the inlining was added to enable (offline synchronized highlighting) cannot work as built.

There's also a type-shape trap that makes this more than a one-line fix: `TranslationBookChapter.thisChapterAudioTimings` is typed `TranslationBookChapterAudioTimingsLinks` = `{ [reader]: string }` (URLs), whereas the inlined `complete.json` value is `CompleteTranslationChapterAudioTimings` = `{ [reader]: number[] }` (values). The downloaded timing arrays literally don't fit the field an offline `TranslationBookChapter` exposes, so you can't just forward them.

**Reproduction (once a consumer exists):**
1. Download a translation for offline use — the fetched `complete.json` includes `thisChapterAudioTimings` values per chapter.
2. Read a downloaded chapter through `OfflineTranslationsManager`/`BibleDataManager` while offline.
3. Observe `chapter.thisChapterAudioTimings === {}`, and any `getAudioTimings()` call fails (no network) — the downloaded timings are unreachable.

**Suggested resolution — pick one and make the code + docs agree:**
- **If offline timings are in scope:** add a values field to `StoredChapter` (e.g. `thisChapterAudioTimings?: CompleteTranslationChapterAudioTimings`), persist `entry.thisChapterAudioTimings` in `toChapterEntries`, and make `BibleDataManager.getAudioTimings()` (or a sibling accessor) consult the offline store before the network — the read must return the stored `number[]` values, since they can't ride in the URL-typed `thisChapterAudioTimings` field.
- **If offline timings are explicitly out of scope for now:** drop the "self-contained for offline use" justification from `CompleteTranslationChapterAudioTimings`'s doc comment and the PR summary so the inlining isn't described as serving an offline path it doesn't reach, and leave a `TODO`/note that offline timing support is deferred.

Either way is fine — the point is that the schema decision is being locked in here, and the two halves currently disagree.

---

### Low — Test coverage gaps around the paths most likely to bite

The added tests are genuinely useful (details below), but they only cover happy paths. Two behaviors that the change introduces go untested:

1. **Error path of `getAudioTimings()`.** `_getJson` throws on any non-2xx response, so `getAudioTimings` rejects rather than returning `null` (unlike `getNextChapter`/`getPreviousChapter`, which return `null`). Callers will need `try/catch`. A test asserting it rejects on a 404 would document this contract and guard against a future refactor silently swallowing errors.
2. **The intentional offline discard.** No test asserts that an offline/downloaded chapter comes back with `thisChapterAudioTimings === {}`. `OfflineTranslationsManager.test.ts:239` already asserts `thisChapterAudioLinks`; adding the timings assertion right there would pin the documented "empty offline" behavior so a future change to the storage schema can't regress it unnoticed.

### Low — `getAudioTimings` reject-vs-null contract is worth a doc line

Related to the above: the JSDoc on both `FreeUseBibleAPI.getAudioTimings` and `BibleDataManager.getAudioTimings` describes the happy path but not that a missing/failed file rejects. A one-line `@throws` note would save the eventual consumer a surprise, since the neighboring chapter fetchers resolve to `null` instead.

---

### Nit — Offline `next`/`previous` timing maps key off ref-existence, not download-state

`nextChapterAudioTimings: nextRef ? {} : null` / `previousChapterAudioTimings: previousRef ? {} : null` (L973, L985) derive from whether an adjacent chapter *exists*, whereas the sibling `nextChapterAudioLinks`/`previousChapterAudioLinks` derive from whether it's *stored* (`nextStored?.… ?? null`). This is not a bug — it actually matches the documented contract better (`null` iff first/last chapter) — but the divergence from the adjacent audio-links lines is easy to misread as a copy-paste slip. A one-line comment noting the deliberate difference would help the next reader.

---

### Tests — do they test real behavior? Yes.

- `FreeUseBibleAPI` tests: `fetchMock` returns the payload for any URL, so `expect(result).toEqual(payload)` alone would pass regardless — but the meaningful assertion is `toHaveBeenCalledWith(<exact url>, …)`, which genuinely verifies `_buildUrl`/endpoint threading and would fail if URL resolution broke. The override test correctly proves the `endpoint` argument wins over the API's default. Not vacuous.
- `BibleDataManager.test.ts` "resolves the link against the translation's endpoint" is the strongest one: it wires a real manager to an alt-endpoint translation and asserts the request goes to `ALT_ENDPOINT`. If `getEndpointForTranslation` resolution regressed to the default endpoint, the response map wouldn't match and the test would fail. Good real-world coverage.
- `makeAudioTimings` fixture and the `makeCompleteTranslation` timings (`[1.5, 3]` for a 2-verse chapter) are internally consistent (one timestamp per verse).

---

PR #1751, opened 2026-09-03
Highlight the verse being read in the audio reader extension

KallynGowdy:
Stacked on #1750 — wires the audio timings support added there into the actual "Listen" button.

## Summary
- `ext_audioReader` now fetches the currently-playing reader's per-verse timings (via `BibleDataManager.getAudioTimings`) as soon as playback starts.
- As the `<audio>` element's `timeupdate` fires, whichever verse is currently being read gets flashed with the "diminish" effect — the same effect `BibleReadingManager`'s `emphasizeVerses` already uses for cross-reference and search-result jumps, reused here rather than reimplemented so a verse-boundary crossing looks and behaves identically (including its own 3s auto-fade — a verse read for longer than that returns to normal brightness until the next one starts).
- Falls back to plain, unhighlighted playback when the chapter has no timing link for the picked reader — an older translation, or an offline-downloaded chapter, whose reconstructed data never carries a timings link.

## Test plan
- [x] `pnpm check:ts` — clean
- [x] `pnpm test` — all 5,483 tests pass
- [x] `pnpm lint` — no new errors
- [x] New tests: pure-function coverage for the verse/time math (`verseIndexForTime`, `chapterVerseNumbers`) plus an integration test that presses play, drives the audio element through two verses' worth of `timeupdate` events, and confirms the diminish decoration follows along — then navigates to a chapter with no timings and confirms nothing highlights. Verified this test fails against the pre-fix code.

https://claude.ai/code/session_01NNTNnkmZkQ7xGgDb5e7Vdx

---
_Generated by [Claude Code](https://claude.ai/code/session_01NNTNnkmZkQ7xGgDb5e7Vdx)_


KallynGowdy:
@Blitzy This should actually work now (assuming you have an updated translation downloaded)

PR #1752, opened 2026-09-03
Add update prompt for offline translations on data-saver devices

KallynGowdy:
## Summary

When a newer version of a downloaded translation is detected and the device has data-saver mode enabled (`navigator.connection.saveData`), the app now asks the user before downloading the update instead of doing it automatically. On devices without data-saver mode or without connection info, updates download immediately as before.

## Changes

- **New component `OfflineUpdatePrompt`**: Displays a modal offering to update a downloaded translation. Shares styling with the existing download prompt. Handles keyboard (Escape to dismiss), backdrop clicks, and error reporting via toast notifications.

- **New manager method `checkAndApplyUpdate`**: Decides whether to download an update immediately or show the prompt based on device connection info. Called automatically by `noteTranslationInUse` so updates are checked whenever the reader navigates to a translation, but also exposed as a public method for explicit calls.

- **New manager signal `updatePrompt`**: Holds the translation being offered as an update, or null when no offer is on screen. Paired with `dismissUpdatePrompt` to close the offer without downloading.

- **Update tracking in `checkAndApplyUpdate`**: Tracks which content hash the update prompt has already been shown for per translation (session-only, resets on reload) to avoid re-offering the same update every time the reader turns a page, while still allowing newer updates to be offered after an older one was dismissed.

- **Integration in main app**: `OfflineUpdatePrompt` is rendered alongside the existing `OfflineDownloadPrompt` in the app layout.

- **i18n strings**: Added English translations for the update prompt title, body, and button labels.

## Implementation Details

- The update prompt only appears when both a download prompt and the update prompt are not already on screen, preventing prompt stacking.
- A cancelled download (user stops it mid-transfer) reports no error, so no toast is shown — only actual failures trigger error feedback.
- The prompt respects the device's offline state; no update check happens when `navigator.onLine` is false.
- Comprehensive test coverage includes success/failure paths, keyboard/click interactions, and the automatic triggering via `noteTranslationInUse`.

https://claude.ai/code/session_0141GKUuVSLzNHxhEAYiV7J7


Blitzy:
> [!NOTE]

## Code Review — PR #1752: Update prompt for offline translations on data-saver devices

Overall this is a well-structured, well-commented change. The manager logic is clear, the doc comments explain the *why*, and the tests exercise real behavior through the real manager (mocking only at the API boundary), poll instead of sleeping, and would genuinely fail if the dedup logic were wrong. I ran both new test files locally — 40 tests pass. A few things are worth a second look before merge.

---

### 🟠 Medium

**No. 1 — The immediate auto-download path has no dedup or backoff, so a failing update re-downloads the whole translation on every page turn.**

On a device *without* data-saver, `checkAndApplyUpdate` calls `downloadTranslation` outright:

```ts
const connection = getNetworkConnection();
if (!connection || connection.saveData !== true) {
  await downloadTranslation(translationId);
  return;
}
```

`noteTranslationInUse` — which calls `checkAndApplyUpdate` — runs from an effect in `SeedBibleStateManager.tsx:1337` on *every chapter navigation*, not just the first. The `saveData`/prompt branch protects itself with `updatePromptShownForHash`, but this immediate branch has no equivalent guard. Its only stopping condition is that `updateAvailable` flips to `false`, which only happens on a *successful* download.

So when a download keeps failing (flaky/slow connection), `updateAvailable` stays `true` and each chapter turn kicks off a brand-new full-translation download attempt (a full Bible is ~7 MB per the `BYTES_PER_VERSE_ESTIMATE` note). It's throttled to one attempt per navigation (the in-flight `downloads.value.has(...)` check prevents overlap), but that's still repeated multi-MB fetches driven purely by page-turning, with no toast or progress to explain the data use.

*Reproduction:*
1. On a browser where `navigator.connection.saveData` is falsy/unset (desktop Chrome, or Safari/Firefox on any connection), download a translation.
2. Make the API report a newer `sha256` (so `updateAvailable` becomes `true`).
3. Cause `complete.json` to fail (throttle/offline mid-request, or a 500).
4. Read several chapters in that translation. Observe a fresh full-download attempt starting after each previous one fails.

*Suggested fix:* give the immediate path the same one-attempt-per-hash memory the prompt path already has — e.g. record the attempted `sha256` in a map (or reuse `updatePromptShownForHash`, renamed) before calling `downloadTranslation`, and clear it on a newer hash. That keeps a failed auto-update from retrying on every navigation while still picking up genuinely newer versions.

**No. 2 — Silent auto-download on non-data-saver devices is a new behavior; the description's "as before" isn't accurate.**

The PR summary says updates "download immediately as before" on non-data-saver devices. But before this PR nothing downloaded updates automatically — `checkForUpdates` only refreshes the list and flips the `updateAvailable` flag, and the sole update path was the manual "tap to update" button in `BibleSelector.tsx:1328`. This PR introduces the *first* automatic update download, and it happens with no toast and no visible progress (the success toast only fires on the prompt-accepted path in `OfflineUpdatePrompt`, not on the auto path).

That means any device that doesn't report `saveData` — which includes non-Chromium mobile browsers on cellular — will silently pull a multi-MB update in the background. That may well be intended, but it's a real product decision rather than a no-op, so it's worth an explicit confirmation and, ideally, a corrected description. If silent-on-cellular is a concern, note that `navigator.connection` also exposes `effectiveType`/`type`, which could gate the auto path further.

---

### 🟡 Low

**No. 3 — Two modals can stack: `offerDownloadPrompt` doesn't check `updatePrompt`.**

`checkAndApplyUpdate` guards against stacking with `if (downloadPrompt.value || updatePrompt.value) return;`, but the reverse isn't true — `offerDownloadPrompt` (line 1157) only checks `downloadPrompt.value`, not `updatePrompt.value`:

```ts
if (promptedThisSession || downloadPrompt.value) {
  return false;
}
```

So if an update offer is already on screen and a download offer becomes eligible, both overlays render at once. They use distinct title `id`s so there's no `aria-labelledby` clash, but two stacked modal dialogs is still a rough state. *Suggested fix:* add `|| updatePrompt.value` to the early-return in `offerDownloadPrompt`.

**No. 4 — A failed prompt-accepted update won't be re-offered for the rest of the session.**

`updatePromptShownForHash` is set at the moment the prompt is shown, before the user acts. If the user taps **Update** and the download fails, `updateAvailable` stays `true` but the hash is already recorded, so `checkAndApplyUpdate` won't re-offer the prompt for that hash again until a reload or a newer hash appears. The user still gets the "Couldn't update" toast and can fall back to the manual button, so this is minor — but if you'd prefer the auto-prompt to return after a failure, only record the hash on a *successful* download (or clear it in the failure branch).

**No. 5 — Narrow double-download race via the auto-trigger.**

Because both `checkAndApplyUpdate` and `downloadTranslation` `await ready` before `downloadTranslation` populates `downloads`, two `checkAndApplyUpdate` calls firing back-to-back before the first sets progress could both pass the `downloads.value.has(...)` guard. This race pre-exists in `downloadTranslation`, but the new auto-trigger makes it reachable without a user double-click. Very low likelihood (`ready` is resolved after startup, so the window is a microtask) — noting for completeness.

---

### Tests

The tests are solid and test real behavior, not tautologies:

- **Manager tests** drive the real `OfflineTranslationsManager` end-to-end through a real in-memory store, mocking only `fetch` — exactly the right boundary. The dedup cases are meaningful: *"does not re-offer the same update once shown"* fails if the `updatePromptShownForHash` guard is removed, and *"offers a newer update even after an older one was dismissed"* fails if the map is keyed by ID instead of hash. Good use of `waitForCondition` polling rather than fixed sleeps, and `navigator.connection` is cleaned up in `afterEach`.
- **Component tests** render the real `OfflineUpdatePrompt` with a manager stub (appropriate for a component test) and cover success, failure, user-cancelled (no toast), decline, backdrop, card-click, and Escape. The *"stays quiet when cancelled reports no error"* case correctly distinguishes a cancel from a failure.

Two small notes, neither blocking:
- The component stub's `errors` map is hardcoded to the `"BSB"` key, so it's coupled to the one fixture; fine as-is but worth knowing if the fixture changes.
- These are new-feature tests rather than regression tests, so there's no red-before-fix baseline to check — that's expected here. The gap in No. 1 (auto-download retry loop) has no test covering it; a test that fails a download and then re-notes the translation in use would pin that behavior down.

---

### Nits (optional)

- `OfflineUpdatePrompt.tsx` is ~90% identical markup to `OfflineDownloadPrompt.tsx` (overlay, handle, header, two buttons, Escape effect). Two copies is within the project's tolerance, but if a third variant ever appears this is the moment to extract a shared prompt shell.

None of the above are hard blockers, but No. 1 and No. 2 are worth resolving or explicitly signing off on before merge, since they change how much data the app spends on the user's behalf.

---

KallynGowdy:
@Blitzy this could use a re-review

PR #1753, opened 2026-09-03
Fix chapter verse counts

KallynGowdy:
Grabbed the chapter verse counts directly from the API instead of relying on the AI estimation. Additionally, it uses the last verse number, not the total number of verses since some chapters will skip verse numbers for one reason or another.


Blitzy:
> [!NOTE]

## Code review — Fix chapter verse counts

**Verdict: the fix is correct and the tests are meaningful.** I verified a sample of the changed New-Testament values against standard KJV/Protestant versification and they all check out (e.g. Matthew 17→27, 18→35, 23→39; Mark 7→37, 9→50, 11→33, 15→47; Luke 17→37, 23→56; John 5→47; Acts 8→40, 15→41, 24→27, 28→31; Romans 16→27). The old "AI-estimated" values were each short by exactly the traditional Textus-Receptus verses that critical editions relegate to footnotes (Matt 17:21, Mark 9:44/46, Acts 8:37, Rom 16:24, etc.). Deriving from the last *verse number* (not a verse *count*) is the right choice here: the table is only used as a permissive upper bound in `isVerseReferenceInBounds`, so picking the highest-numbered verse correctly lets references like "Romans 16:24" pass rather than being rejected.

**Tests are real, not always-green.** `scanVerseReferencesInText` is called without a translation book list, so it falls back to this static table for verse bounds — the new cases genuinely exercise the changed data. I confirmed this by reverting just the two touched rows (Luke 23 → 55, Mark 15 → 46) and re-running: both new cases fail (`endVerse` exceeds the old max), and pass again once restored. Good regression coverage.

Everything below is **low severity / polish** — none of it blocks merge.

---

### Low — semantics/naming now drift from the data

`BOOK_CHAPTER_VERSE_COUNTS`, `getChapterVerseCount()`, and its doc comment ("Verses in a 1-based chapter") now hold and return the **last verse number**, which differs from the verse *count* for any chapter that skips or combines verse numbers — exactly the case the PR description calls out. There's no bug (the only consumer treats it as a max), but the name and doc are now slightly misleading for a future reader.

*Suggested fix:* update the JSDoc on `getChapterVerseCount` to something like "Highest verse number in a 1-based chapter (an upper bound for scanning; not necessarily the verse count)", and consider whether a `getChapterLastVerseNumber` alias is worth it. Documentation-only.

### Low — generator rewrites the exported functions from a hardcoded template

`script/generate-book-chapter-verse-counts.ts` regenerates the *entire* file, including `getBookChapterCount`/`getChapterVerseCount` and the header, from an inline template string. Two consequences:

1. Any future edit to those two functions is silently reverted the next time someone runs `pnpm generate-book-chapter-verse-counts`.
2. The header/doc text is now duplicated in two places (the script template and the source file) and can drift.

*Suggested fix:* have the generator rewrite only the data object (e.g. splice between the `= {` and `};` markers, or between sentinel comments) and leave the hand-written functions/header untouched. If that's more than you want, at minimum add a comment in the source file noting that both functions are reproduced verbatim in the generator template so they stay in sync.

### Low — generator relies on a manual prettier pass

The script writes each array on a single line and then just prints "Run `pnpm exec prettier --write` on the file". If a contributor forgets, the committed file won't match `pnpm format` and could trip the pre-commit hook / lint in CI.

*Suggested fix:* run prettier programmatically at the end of the script (via its Node API or a `spawn` of the local binary) so the output is commit-ready in one step.

### Informational — the table is derived from one translation (AAB)

The result is correct only because AAB numbers the traditional TR verses, yielding the KJV maxima. A translation that omits those verses would produce *lower* bounds and start rejecting otherwise-valid references. That's a fine and deliberate choice for a permissive bound, but it's an implicit assumption worth a one-line note in the script header (why AAB specifically), and there's no CI check that the committed table still matches the API — regeneration is a manual step, so the table can silently drift over time. No action needed now; flagging for future maintainers.

---

### Reproduction (regression-test verification, for a human)

```bash
git checkout fix/book-chapter-verse-counts
# temporarily restore the two old maxima:
#   MRK chapter 15: 47 -> 46
#   LUK chapter 23: 56 -> 55
pnpm vitest run BibleDataManager.test.ts -t "scanVerseReferencesInText"
# => "should find Luke 23:50-56" and "should find Mark 15:42–47" FAIL
# restore the new values -> both PASS
```

Nice, tightly-scoped fix. My only real ask is the naming/doc note (No. 1) and the generator-clobbers-functions risk (No. 2); the rest is optional.

https://claude.ai/code/session_01Rq8E2UJzSdXYQCGFfVQdqY

---

KallynGowdy:
Thanks for the review — addressed the two blocking-adjacent items plus the informational note, all in 835ba0d1:

1. **Naming/doc drift** — `getChapterVerseCount`'s JSDoc now says it returns the highest verse number (an upper bound for scanning), not the verse count, and explains why those differ for chapters with skipped/combined verse numbers.
2. **Generator clobbering hand-written functions** — the script now reads the existing file and only splices the `BOOK_CHAPTER_VERSE_COUNTS` object back in via a regex-bounded replace, leaving the header comment and both functions on disk untouched. Verified by hand-editing `getChapterVerseCount` locally, re-running the splice logic, and confirming the edit survived.
3. **Manual prettier step** — the generator now formats its own output via prettier's Node API (`prettier.resolveConfig` + `prettier.format`) before writing, so there's no more "remember to run prettier --write" step.
4. **AAB translation choice** — added a one-line note in the script header on why AAB specifically (it numbers the traditional Textus Receptus verses, matching the KJV maxima).

Didn't add a `getChapterLastVerseNumber` alias since you flagged that as optional and the doc fix already removes the ambiguity. Also skipped the CI-drift-check suggestion since it was flagged informational/no-action-needed.

Ran `pnpm exec prettier --check`, `pnpm exec eslint`, and the `BibleDataManager`/`verseReferenceBounds` vitest suites (all pass) against the changed files; also exercised the generator's splice+format logic directly since I couldn't reach the live Bible API from this sandbox (403).

---

PR #1755, opened 2026-09-04
Replace diminish keyframe animation with an opacity transition

KallynGowdy:
The keyframe animation ran a fixed 3s timeline once triggered, so a
second diminish decoration arriving before the first expired had no
effect on the animation's own clock, causing the effect to end
early relative to the new navigation. A plain opacity transition
holds dimmed for as long as the container class is present, which
already lasts until the last matching decoration is removed, so
overlapping decorations now naturally extend the dim instead of
racing a stale timeline.

Claude-Session: https://claude.ai/code/session_019SqDuddYrYdX9r6Sku2B8g

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
@@ -541,6 +541,7 @@
   margin-right: var(--text-heading-margin-right);
   font-size: 1.125em;
   line-height: 1;
+  transition: opacity 0.3s ease-in-out;
 }
 
 .sb-line-break {
@@ -571,10 +572,10 @@
   transition:
     background-color 0.2s ease,
     background 0.3s ease,
-    color 0.2s ease;
+    color 0.2s ease,
+    opacity 0.3s ease-in-out;
 
   &:has(.sb-verse-decoration-diminish) {
-    animation: none;
     opacity: 1;
   }
 }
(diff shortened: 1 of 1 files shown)


Blitzy:
> [!NOTE]

## Code Review — Replace diminish keyframe animation with an opacity transition

**Verdict: Looks good. Approve with minor notes.** The change is correct, the specificity handling is preserved intact, and it fixes a real timing bug. No correctness issues found. The notes below are about confirming an intended UX change and one pre-existing accessibility gap.

### What the change does (and why it's right)

The old code dimmed non-emphasized verses with a fixed `@keyframes diminish-in-out 3s` animation. The decoration that turns dimming on, however, is removed by a **`removeAfterMs` timer** (`decorateVerses` in `BibleReadingManager.tsx:2386-2395`), and that timer is **not always 3s**:

- `BibleReadingManager.emphasizeVerses`, `SeedBibleStateManager`, `TodayManager` → `removeAfterMs: 3000`
- **`TabsManager.tsx:289` and `TabsManager.tsx:492` → `removeAfterMs: 5000`**

Because the animation ran on its own 3s clock and completed with no `animation-fill-mode`, verses reverted to full opacity at 3s regardless of how long the decoration actually lived. So for the 5000ms call sites the dim ended ~2s early, and for overlapping navigations a still-running animation from an earlier decoration finished on a stale clock (exactly the race the PR description calls out). Replacing the animation with a plain `opacity: 0.4` state + a `transition: opacity 0.3s` means the dim now fades in, **holds for as long as the container class is present**, and fades back out when the last matching decoration is removed — which is the correct, lifetime-driven behavior.

I confirmed the specificity reasoning in the comments still holds after the change:
- Dim rule `.sb-chapter-decoration-diminish .sb-verse:not(:has(.sb-verse-decoration-diminish))` (0,3,0) and the keep-bright rule `.sb-chapter-decoration-diminish .sb-verse:has(...)` (0,3,0) target disjoint verse sets via `:not`, so a verse with its own diminish decoration stays at `opacity: 1` and every other verse dims. Correct in both directions.
- The removed `@keyframes diminish-in-out` has no remaining references anywhere in `packages/` (verified by grep).

### Findings

**Low — confirm the intended dim-duration change for the 5000ms paths.** This is not a bug, but it is a visible behavior change worth an explicit sign-off: verses emphasized via the tab-restore path (`TabsManager.tsx:289`, `:492`) previously un-dimmed after ~3s and now stay dimmed for the full ~5s decoration lifetime. That is almost certainly the intended fix, but it does mean the on-screen dim is now 2s longer for those flows than it was on `develop`.
- _Reproduction:_ Trigger a tab creation/restore that highlights verses (the `removeAfterMs: 5000` path), and observe the surrounding verses now hold dim for ~5s instead of returning to full brightness at ~3s.

**Low — the effect is now a steady hold rather than a pulse.** The old keyframe faded `1 → 0.4 → 1` *within* the decoration's life (a brief pulse). The new behavior fades to 0.4 and holds flat until removal, then fades back. This matches the PR's stated goal; flagging only so the product/design side confirms a steady dim (rather than a pulse-and-release) is what's wanted.

**Nit — no `prefers-reduced-motion` guard on the diminish effect.** Neither the old animation nor the new transition is covered by a `prefers-reduced-motion: reduce` block, even though other effects in this same file are (`BibleReader.inline.css:518` for `.sb-highlight-ribbon`, `:755` for the retry spinner). This is **pre-existing**, not introduced here, but since the PR is already editing exactly these rules it's a cheap opportunity to add, e.g.:
```css
@media (prefers-reduced-motion: reduce) {
  .sb-verse,
  .sb-chapter-heading {
    transition: none;
  }
}
```
This keeps the dim state (helpful for focus) while dropping the animated fade for users who ask for reduced motion. Optional — not a blocker.

### Tests

No automated test accompanies this change, which is reasonable here: this is a pure CSS visual behavior (opacity transition timing) that jsdom does not compute, so a unit test could only assert the CSS text itself and would pass regardless of whether the effect works — the kind of always-green test worth avoiding. The decoration lifecycle that *drives* this CSS is already covered by the manager/component tests that assert `decorateVerses` is called with `className: "sb-verse-decoration-diminish"` / `containerClassName: "sb-chapter-decoration-diminish"` (e.g. `TabsManager.test.ts`, `SeedBibleStateManager.test.ts`, `DiscoverPane.test.tsx`), and those remain valid. The right validation for the visual behavior is a manual check on the alpha deploy — specifically exercising both a 3000ms path and a 5000ms path, and two overlapping emphasize calls, to confirm the dim now holds until the last decoration expires.

---

PR #1756, opened 2026-09-04
Feature/1602

### LINKED ISSUE #1602: Build a unified color picker component
## Problem

Seed Bible currently has at least five separate raw `<input type="color">` usages scattered across the app, each with its own hand-rolled wiring for state, debouncing, and commit behavior:

- `BibleReaderToolbar.tsx` (~L1931) — custom highlight color for verse selections, shared between desktop and mobile
- `SettingsPage.tsx` (~L1521) — custom color for the text-format palette
- `SettingsPage.tsx` (~L1700) — per-field theme color override
- `SettingsPage.tsx` (~L1757) — highlight background color override
- `SettingsPage.tsx` (~L1770) — highlight text/foreground color override

Because `<input type="color">` renders using the browser/OS's native picker, its behavior differs a lot by platform: desktop Chrome draws its own inline picker and fires a stream of `input` events while dragging, while mobile browsers (iOS/Android) show a native modal that only fires one `change` event when the user taps "Set"/"Done." Our code doesn't account for this difference consistently — in the verse toolbar, for example, both `input` and `change` are wired to the same debounced commit function, so a brief pause while still fine-tuning a color on desktop gets treated as final: it gets written to the custom-color palette and persisted to the backend, even though the user hasn't confirmed anything. Mobile only ever sees one real `change`, so it doesn't have this problem — not because we built it correctly, but because the OS picker happens to only report a confirmed value.

The result is an inconsistent, occasionally janky experience that differs by platform for reasons outside our control, plus five independent, slightly-divergent implementations of "let the user pick a color."

## Goal

Build one reusable, generic color picker component that:

- Works consistently across desktop and mobile, with the same interaction model on both, instead of relying on OS/browser-native picker quirks
- Has a clear, explicit "confirm" step, so a color is only committed (added to any saved palette, applied, persisted) once the user actually finalizes it — not on every intermediate drag value
- Can optionally support a live, non-committing preview while the user is still adjusting, if that's the interaction we want
- Is generic enough to drop into any component that needs color selection — verse highlight colors, theme customization, text-format palette — replacing all five existing ad-hoc usages
- Matches Seed Bible's visual design language on both platforms, rather than looking and behaving differently depending on the OS's native picker UI

## Open question

Should we build this from scratch or adopt an existing color-picker library (e.g., `react-colorful`, `@uiw/react-color`, or similar) and skin it to match our design system? Either is fine as long as it gives us full control over styling and interaction on both desktop and mobile — worth evaluating options once someone picks this up.

## References

- `packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx` — verse highlight custom color picker (~L1824-L1943)
- `packages/seed-bible/seed-bible/components/SettingsPage/SettingsPage.tsx` — theme and text-format custom color pickers (~L1518-L1775)

Kabi4:
Fixes #1602


Blitzy:
> [!NOTE]

## Code review — Feature/1602 (unified color picker)

This PR delivers the from-scratch `ColorPicker` component agreed on in #1602 (Kabi4 + Blitzy both preferred building our own over pulling in `react-colorful`), and replaces all five ad-hoc `<input type="color">` usages with it. The component is thoughtful — HSV square + hue slider, keyboard support, `touch-action: none` for touch dragging, a single-open-at-a-time guard, an explicit Confirm/Cancel step, and a portal host that is cleaned up on close. The CSS cleanup is thorough (no dangling `.sb-verse-toolbar-color-input` / `toHexInputValue` leftovers), and the ThemeManager hardening (brace-stripping + the un-nested `&` selector + the `--sb-background:` guard) is a nice defensive win with real regression tests behind it.

Below are the issues I found, most severe first.

---

### 🔴 High

**No. 1 — Unused `navigation` destructure will fail `pnpm lint` (likely red CI).**
In `BibleReaderToolbar.tsx` the `props.state` destructure gained `navigation` (~L613), but nothing reads the local — the one real use (~L652) goes through `props.state.navigation.initialUrl` directly. `eslint.config.mts` sets `@typescript-eslint/no-unused-vars` to `"error"` (only names matching `^_` are ignored), and that rule reports unused destructured object properties by default, so `pnpm lint` should error out. CI's `build (24.x)` was still queued at review time, so this may not have surfaced yet.

- _Reproduce:_ `pnpm lint` (or `npx eslint packages/seed-bible/seed-bible/components/BibleReaderToolbar/BibleReaderToolbar.tsx`).
- _Fix:_ drop `navigation,` from the destructure — the code already reads `props.state.navigation`.

---

### 🟠 Medium

**No. 2 — Applying a brand-new custom highlight color now takes two steps instead of one.**
Before, picking a custom color both saved it to the palette *and* highlighted the selected verses. Now `applyCustomColor` (the picker's `onChange`) only calls `settings.addCustomHighlightColor(color)` — the user then has to tap the freshly-added swatch to actually highlight. The inline comment explains the reasoning (confirming-and-applying would clear the selection and unmount the selector before the new swatch could render), and the flow does work (a test confirms the selection survives so the swatch is tappable). But it's a real interaction change that #1602 didn't ask for, so it's worth an explicit sign-off from the team.

Related: when the palette is empty the verse toolbar now shows **three dashed empty "Add custom color" slots plus the separate "+" button** — four affordances that all open the same picker. Intentional design is fine; just flagging the redundancy.

- _Reproduce:_ select a verse → open the custom picker → pick a color → Confirm. Observe the verses are not highlighted until you tap the new swatch.

**No. 3 — Theme / text-format color editing lost its live preview.**
The old `<input type="color">` fired `onInput` on every drag, so `theme.setCustomColor` / `onChange({ color })` applied the color live as you dragged — you saw the theme update in real time. The new `ColorPicker` only fires `onChange` on Confirm, and none of the SettingsPage call sites pass `onPreview`, so there's no visual feedback until you commit. Dropping the *per-drag persistence* is consistent with #1602's goal (that intermediate-commit behavior was the bug), but losing all preview for theme customization is a downgrade. The component already supports `onPreview` — consider wiring a non-committing preview for the theme fields (would need a preview-only path in ThemeManager that doesn't persist).

---

### 🟡 Low

**No. 4 — The custom-color ring-buffer pointer isn't persisted, so eviction order is wrong after a refresh.**
`SettingsManager`'s `nextCustomHighlightSlot` is an in-memory closure variable. The palette itself is persisted to profile config, but the pointer resets to `0` on reload. So with a full 3-color palette, the first color picked after a page refresh always overwrites slot 0 regardless of which entry is genuinely oldest, and the round-robin restarts from there. It only affects *which* of three custom colors is evicted after a reload — minor — but the new `SettingsManager.test.ts` / toolbar tests exercise the ring buffer only within a single session, so this gap isn't covered. Consider deriving the slot from persisted state, or noting the limitation.

- _Reproduce:_ fill all 3 custom slots, refresh the page, add a 4th color → it replaces slot 0 rather than the actual oldest.

**No. 5 — Silent semantics change in `addCustomHighlightColor`.**
It went from "move-to-front if present, evict oldest" (MRU) to "no-op if present, fixed insertion order, round-robin eviction." The new behavior keeps swatch positions stable (a reasonable motivation), but re-picking an existing custom color is now a no-op rather than promoting it. Confirm this is intended.

---

### ⚪ Nits

- The modal dialog sets `aria-modal="true"` and focuses the panel, but there's no focus trap — Tab can move focus out of the open picker. Minor a11y.
- `getPickerHost` / `removePickerHost` share one `#sb-color-picker-host` node across all instances. The full-viewport backdrop currently prevents two pickers' open states from overlapping, so the shared node is safe in practice — but it's a latent footgun worth a comment, since a caller that opens a second picker programmatically (bypassing the backdrop) could have one instance remove the host the other just portaled into.

---

### ✅ On the tests

The tests read as real-world behavior checks, not always-green filler:

- `ColorPicker.test.tsx` drives open → type/drag → Confirm/Cancel/Escape/backdrop and asserts on `onChange`/`onPreview`/`onCancel` call counts and DOM teardown, including a targeted regression for "confirm unmounts the picker" (the overlay-cleanup bug the `confirm()` comment describes).
- The rewritten `BibleReaderToolbar.test.tsx` cases assert the *new* two-step flow (Confirm adds to selector without highlighting/clearing; swatch tap highlights), the ring-buffer eviction, and that cancel/backdrop/drag leave the selection intact — all of which would fail against the old debounced implementation.
- `ThemeManager.test.ts` injects a `}`-bearing override and asserts the body rule can't be closed, and asserts the theme survives a custom-color save — both fail on pre-fix code.
- `SettingsManager.test.ts` pins the exact ring-buffer sequence.

The one coverage gap is No. 4 (post-refresh eviction order), since every ring-buffer test runs inside a single session.

---

Nothing here is a correctness blocker except **No. 1**, which should be fixed before merge to get CI green. **No. 2** and **No. 3** are UX calls worth a quick team decision. Nice work overall.

---

Blitzy (approved):

PR #1758, opened 2026-09-04
fix: collapsed sidebar tab list scrolls instead of overflowing

Blitzy:
The collapsed sidebar's tab list didn't scroll, so tiles below the fold were unreachable once you had enough tabs open, and now it scrolls like the expanded list does. Fixing that also required stopping the tiles and the "new tab" button from getting flex-squashed to fit, and compensating for a side-effect where enabling vertical scroll made the browser start clipping the tiles' rounded corners on the left/right.

_Before:_
image

_After:_
image

### DIFF
--- a/packages/seed-bible/seed-bible/components/SidebarSearch/SidebarSearch.css
+++ b/packages/seed-bible/seed-bible/components/SidebarSearch/SidebarSearch.css
@@ -127,6 +127,7 @@
   position: relative;
   width: 3.3125rem;
   height: 3.3125rem;
+  flex-shrink: 0;
   border-radius: 0.625rem;
   background: var(--sb-tab-background);
   color: var(--sb-tab-font-color);
--- a/packages/seed-bible/seed-bible/components/Tabs/Tabs.css
+++ b/packages/seed-bible/seed-bible/components/Tabs/Tabs.css
@@ -657,6 +657,15 @@
   flex-direction: column;
   align-items: center;
   gap: 0.5rem;
+  flex: 1 1 auto;
+  min-height: 0;
+  overflow-y: auto;
+  -ms-overflow-style: none; /* IE and Edge */
+  scrollbar-width: none;
+  /* The tiles are slightly wider than this list's padded content box, and
+     overflow-y: auto forces overflow-x to clip too — without this, that
+     clips the tiles' rounded left/right edges. */
+  margin-inline: -0.10625rem;
 }
 
 .sb-sidebar-settings-view {
@@ -1168,6 +1177,7 @@
   &.sb-collapsed-tab-add-button {
     width: 3.3125rem;
     height: 3.3125rem;
+    flex-shrink: 0;
   }
 }
 


Blitzy:
> [!NOTE]

## Summary

This is a clean, low-risk, CSS-only change (11 additions across 2 files) that makes the collapsed sidebar's tab rail scroll instead of overflowing. I traced the layout end-to-end and the approach is sound — I have **no correctness issues** to report, only a couple of low-severity UX/consistency observations worth a glance before merge.

**What the change does, and why each part is needed:**

- `.sb-sidebar-collapsed-tab-list` gets `flex: 1 1 auto; min-height: 0; overflow-y: auto`. The parent `.sb-tabs-sidebar` is already a `flex-direction: column` with `min-height: 0`, so this correctly lets the list claim the leftover height and scroll internally while the bottom actions row (`margin-top: auto`) stays pinned. The `min-height: 0` is the key that lets the flex child shrink below its content and actually trigger the overflow — good catch.
- `flex-shrink: 0` on the tiles (`.sb-collapsed-tab-tile`) and the add button (`.sb-collapsed-tab-add-button`) is genuinely required: both are column children of the now-shrinkable list, so without it they'd be squeezed rather than overflowing. Correct.
- The `margin-inline: -0.10625rem` compensation checks out arithmetically. The rail is `4.1rem` wide with `0.5rem` inline padding → a `3.1rem` content box; the negative margin widens the list to `3.1 + (2 × 0.10625) = 3.3125rem`, which is exactly the tile's width. Because the global reset in `app/styles/base.inline.css` sets `box-sizing: border-box`, `3.3125rem` is the tile's full border-box width, so the tiles fit precisely — no rounded-corner clipping and no phantom horizontal scrollbar from `overflow-y: auto` forcing `overflow-x: auto`. The comment explaining this is helpful; nicely reasoned.

## Issues by severity

### High / Critical
None.

### Low

**No. 1 — In collapsed mode the "New tab" (+) button now scrolls out of view with the list.** The `+` button is the last child *inside* the scroll container (`Tabs.tsx` ~line 2070), so once you have enough tabs to scroll, reaching "New tab" requires scrolling to the bottom. This differs from the expanded sidebar, where the `+` lives in the pinned header (`.sb-sidebar-tabs-header .sb-tab-add-button`) and is always reachable. This looks intentional given the "after" screenshot (the `+` sits at the bottom of the rail), so this is a heads-up rather than a defect — but if always-reachable "add" is desired, the button would need to sit as a sibling *after* the scroll list rather than inside it.

**No. 2 — Scrollbar hiding covers Firefox/old-Edge but not WebKit.** The new rules use `scrollbar-width: none` (Firefox) and `-ms-overflow-style: none` (legacy Edge/IE) but no `::-webkit-scrollbar { display: none }`. This exactly mirrors the sibling `.sb-sidebar-tab-list` (expanded list), so it's consistent with the existing pattern and harmless — modern Chrome/Safari now honor `scrollbar-width`. Flagging only for completeness; not a blocker. If you ever want to guarantee a hidden scrollbar on older WebKit, both lists would need the `::-webkit-scrollbar` rule (other components in the repo already use it, e.g. `ChatView.css`).

## Testing

No automated test accompanies this change, which is appropriate here: the behavior is pure layout/scroll, and the Vitest + jsdom setup has no layout engine (jsdom reports zero heights and never triggers overflow), so a unit test couldn't meaningfully assert scrollability without asserting on the presence of the CSS classes themselves — which just restates the stylesheet. The existing `TabsSidebarLayout.test.tsx` suite still passes unchanged (it asserts on rendered structure, not CSS). The screenshots in the PR description are the right form of verification for this class of change.

**Manual verification steps (for a human reviewer):**
1. Open the app and collapse the sidebar (the narrow rail).
2. Open enough tabs that the stacked tiles exceed the rail's height (roughly 8–10+ on a typical viewport).
3. _Before this fix:_ tiles below the fold are unreachable and the tiles/`+` get squashed. _After:_ the rail scrolls vertically, tiles keep their full `3.3125rem` size, rounded corners aren't clipped on the left/right edges, and no horizontal scrollbar appears.
4. Confirm the same in both light and dark themes, and re-check the docked-vs-collapsed transition in the 481–768px compact band (where the expanded sidebar floats but the collapsed rail stays in flow).

---

PR #1759, opened 2026-09-04
Skip re-fetching customizations on client by seeding SSR result

KallynGowdy:
## Summary

When a user opens a page with a `?customization=` share link, the server (SSR) fetches that customization record and renders the page with its theme and favicon. The client was then re-fetching the same record over `os.getData()`, duplicating the network round trip. This change has the server embed the resolved customization as JSON in the page, so the client's `CustomizationsManager` can skip the redundant fetch.

## Key Changes

- **CustomizationsManager**: Added `InitialCustomizationSeed` interface to represent a completed `?customization=` load (the locator and resolved record, or null if not found). Added `getInitialCustomizationSeed()` method that returns this seed only when a load actually completed (not when the SSR timeout backstop fired first, which would incorrectly tell the client "this doesn't exist"). The manager now accepts an optional `initialCustomizationSeed` parameter; if its locator matches the current page's `?customization=` param, the seed is applied synchronously without calling `os.getData()`.

- **entry-ssr.tsx**: After rendering settles, calls `getInitialCustomizationSeed()` and embeds the result as JSON in a `<script id="app-customization-seed">` tag. Also emits a `<link rel="icon">` pointing to the active customization's uploaded logo (if one exists), so the page's favicon reflects the customization's branding.

- **customizationSeed.ts** (new file): Client-side helper that reads the injected customization seed from the DOM and returns it for the app to pass to `createSeedBibleState`.

- **init.tsx**: Reads the injected customization seed and passes it to `createSeedBibleState`, which forwards it to `CustomizationsManager`.

- **SeedBibleStateManager**: Added `faviconUrl` signal that reflects the active customization's logo URL (or null), and accepts `initialCustomizationSeed` in options.

- **Service worker caching** (`sw.ts`, `swRouting.ts`): Updated the app shell cache strategy to partition by `?customization=` locator. Each customization link now gets its own cached entry (keyed by locator), separate from the plain shell and other customizations. This prevents a customization's branded HTML (theme, favicon) from leaking onto or being overwritten by the plain shell or a different customization. The cache key function now takes the request URL and extracts the `customization` query param to build the partition.

- **Tests**: Added comprehensive coverage for the new seed flow, including tests that verify the client doesn't re-fetch when a seed is provided, that mismatched seeds are ignored, that the SSR timeout backstop prevents seeding incomplete loads, and that favicon overrides work correctly.

- **SSR timeout safety**: The distinction between `initialCustomizationLoadSettled` (forced true by the timeout backstop) and `initialCustomizationLoadCompleted` (only true when the actual load finished) ensures that if SSR times out waiting for `os.getData()`, the client never receives a "not found" seed that would incorrectly rule out a customization that might still resolve. `getInitialCustomizationSeed()` returns null in this case, so the client makes its own attempt.

- **Locator matching**: The seed is only applied if its locator exactly matches the page's current `?customization=` param. A stale or mismatched seed (e.g., from a cached HTML page) is ignored, and the client fetches normally instead.

- **Cache partitioning**: The service worker now uses `getAppShellCacheKey(origin, requestUrl)` to compute a cache key that includes the customization locator (if present). This ensures each customization's branded shell is cached separately, preventing cross-contamination while still pooling all paths within the same customization under one entry.

https://claude.ai/code/session_01Ev1CgZC5QxTgWhd33DwSLE


Blitzy:
> [!NOTE]

## Summary

This is a clean, well-documented change. The core mechanism — have SSR embed the resolved `?customization=` load as JSON so the client's `CustomizationsManager` can apply it synchronously instead of repeating the `os.getData()` round trip — is sound, and the "completed vs. settled" distinction that keeps a timed-out SSR load from seeding a false "not found" is exactly the right subtlety to get right. The service-worker cache partitioning by customization locator is a reasonable extension of the existing single-shell model, and the tests cover the important seed/mismatch/timeout branches with real behavioral assertions (verifying `os.getData()` is *not* called, checking `activeCustomization`, exercising the fallback fetch) rather than trivially-passing checks.

No correctness bugs that break functionality. The findings below are all low severity — mostly test-coverage gaps and a reactivity gap around the new favicon feature. Nothing here needs to block merge, but No. 1 and No. 2 are worth a look.

**Context notes:** The PR body links no issue, and there is no prior human discussion on the PR (only the deployment bot comment), so this review is against the diff and stated intent alone. CI: the current build (`build (24.x)`) was still in progress at review time; the prior commit's build passed.

---

## Findings by severity

### Low — No. 1: The favicon override's "last icon wins" mechanism isn't actually exercised by the tests

`entry-ssr.tsx` emits the customization `<link rel="icon">` inside the `<!-- META -->` fragment, relying on it landing *after* `index.html`'s default `<link rel="icon" href="/standalone/img/favicon.ico" />` (line 15) so browsers pick the later one. That ordering is the whole mechanism, and it's browser-dependent — browsers don't uniformly honor "last declared wins" for icons; some select by `type`/`sizes`, and here the default is a `.ico` and the override a `.png`.

The tests use a stripped-down `TEMPLATE` (test file, around line 394) that contains **no default `<link rel="icon">`**. So:
- `"points the favicon at the linked customization's uploaded logo"` asserts the override link is *present*, but never that it follows / overrides a default.
- `"omits a favicon override..."` passes trivially because the template has no favicon at all — it would pass even if the override logic were broken in a way that also dropped a real default.

**Why it matters:** the tests can't catch a regression where the override renders but a real browser still shows the default icon, which is the actual failure mode users would see.

**Suggested fix:** include a representative default `<link rel="icon">` before `<!-- META -->` in the test `TEMPLATE`, then assert on ordering (the override link's index is greater than the default's). And a human should verify in an actual browser (Chrome + Firefox + Safari) that the branded favicon wins, since automated tests can't validate the browser's icon-selection heuristic.

**Repro for a human:** open a `?customization=` link whose customization has an uploaded logo, in Chrome/Firefox/Safari, and confirm the tab icon is the logo and not `favicon.ico`.

### Low — No. 2: `faviconUrl` is reactive but only ever read by SSR; the client never updates the tab icon

`faviconUrl` is added as a `computed` signal on `AppState`, but its only consumer is the SSR `<link rel="icon">` (grep confirms no client-side effect writes it to the DOM). Because it's a reactive signal, it reads as though the favicon tracks the active customization — but on the client it never does:

- **Editor preview:** open your own customization with an uploaded logo in the editor. `editingCustomization` → `activeCustomization` → `faviconUrl` all update, but the browser tab icon stays whatever the served HTML had. (Repro: edit a customization that has a logo; watch the tab icon — it won't change.)
- **Late client load (non-seed path):** on a slow connection where SSR timed out (no seed embedded), the client resolves the customization late; the favicon never catches up.

This may well be intended scope (SSR-only branding), but the reactive signal invites the assumption that it's live. Either add a small client effect that syncs `faviconUrl` to a managed `<link rel="icon">` in the DOM (the same way `document.title` is presumably kept in sync), or add a comment on `faviconUrl` stating it is SSR-only and intentionally not applied on the client.

### Low — No. 3: The seed path bypasses the schema validation that `loadByLocator` performs

In `loadByLocator`, a fetched record goes through `customizationSchema.safeParse(...)` and `narrowVariants(...)` before it's assigned to `linkedCustomization`. The seed branch (CustomizationsManager.tsx ~line 860) assigns `initialCustomizationSeed.customization` **directly**, and `readInjectedCustomizationSeed` only checks `typeof parsed === "object"` before casting to `InitialCustomizationSeed`.

This is safe **today** because the seed is produced by the server from an already-parsed/narrowed record and injected via `escapeForScript`, so it's effectively same-origin trusted data. The concern is the asymmetry: the two code paths that both end at `linkedCustomization.value = ...` now have different validation guarantees, so a future change to the schema (new required field, a variant shape change) would be enforced on the fetch path but silently skipped on the seed path. Consider running the seed's `customization` through the same `safeParse`/`narrowVariants` before applying it (falling back to a normal fetch if it fails to parse), so the two paths can't drift.

### Nit — No. 4: `apple-touch-icon` is not overridden

Only `rel="icon"` is overridden; `index.html`'s `<link rel="apple-touch-icon" href="/standalone/img/apple-touch-icon.png" />` (line 16) still points at the default. On iOS "Add to Home Screen", a customization-branded page will use the default touch icon rather than the customization logo. Likely out of scope, but flagging for awareness if home-screen branding matters.

---

## What looks good

- The `initialCustomizationLoadCompleted` vs. `initialCustomizationLoadSettled` split, and the regression test `"getInitialCustomizationSeed() stays null when the SSR timeout backstop fired..."` (uses fake timers, a never-resolving `getData`, and `import.meta.env.SSR`), correctly guards the "don't seed a false not-found" invariant — that's the highest-value test in the PR.
- `getAppShellCacheKey` tests assert the meaningful properties: same locator across different paths → one key; different locators → different keys; customization link ≠ plain shell.
- `escapeForScript(JSON.stringify(...))` is used consistently with the existing seed-data injection, so the `</script>` breakout concern is handled.
- `JSON.stringify(null)` → the client's `readInjectedCustomizationSeed` treating `null`/non-object as "absent" is a clean, well-tested no-seed path.

---

PR #1760, opened 2026-09-04
Support signed-out variant selections via login.localConfig

KallynGowdy:
## Summary

Extends `CustomizationVariantSelectionsManager` to persist theme variant choices for signed-out viewers to `login.localConfig` (which mirrors to `localStorage`), matching how other anonymous settings work. Previously, variant selections were silently discarded when signed out.

## Changes

- **Split selections storage**: Introduced `remoteSelections` (CasualOS record for signed-in users) and `localSelections` (computed from `login.localConfig` for signed-out users), with a `selections` computed signal that switches between them based on login state.

- **Updated `selectVariant()` behavior**: Now persists to `login.localConfig` when signed out, instead of being a no-op. Signed-in behavior unchanged — still writes to the CasualOS record.

- **Exported `parseStringRecord` helper**: Made the existing `SettingsManager` utility public so `CustomizationVariantSelectionsManager` can parse the stored record from `login.localConfig`.

- **Updated documentation**: Clarified that the manager now handles both signed-in and signed-out viewers, and that signed-out choices persist via the same device-local mechanism as other anonymous settings.

- **Added comprehensive tests**: Three new test cases verify signed-out behavior — that selections start empty, that `selectVariant()` persists to `login.localConfig`, that selections survive a simulated reload (via a fresh manager instance reading the same signal), and that multiple selections don't clobber each other.

## Implementation Details

The key insight is that `login.localConfig` is already mirrored to `localStorage` by `LoginManager` and adopted into a new account's profile on first login. By storing variant selections under the same `VARIANT_SELECTIONS_ADDRESS` key, signed-out choices automatically get the same persistence and adoption behavior as any other anonymous setting — no special handling needed.

The `selections` computed signal ensures the manager's public API remains unchanged: callers always read from `selections.value` and call `selectVariant()`, regardless of login state. The routing between remote and local storage is internal.

https://claude.ai/code/session_017mJQQd7TBf2Z8e3j6sFTQ4

### DIFF
--- a/packages/seed-bible/seed-bible/managers/CustomizationVariantSelectionsManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/CustomizationVariantSelectionsManager.tsx
@@ -2,6 +2,7 @@ import * as z from "zod/v4";
 import { computed, effect, signal, type ReadonlySignal } from "@preact/signals";
 import type { CasualOSManager } from "./OsManager";
 import type { LoginManager } from "./LoginManager";
+import { getProfileConfigValue } from "./ProfileConfigSync";
 import { parseStringRecord } from "./SettingsManager";
 
 export const VARIANT_SELECTIONS_ADDRESS = "customizationVariantSelections";
@@ -25,10 +26,18 @@ const variantSelectionsPayloadSchema = z.object({
  * Signed-in viewers get their choices synced to a CasualOS record (`os.getData`/
  * `recordData` under `VARIANT_SELECTIONS_ADDRESS`). Signed-out viewers get
  * the same device-local persistence every other anonymous setting gets —
- * `login.localConfig`, which `LoginManager` already mirrors to `localStorage`
- * and offers to adopt into a brand-new account's profile on first login —
- * keyed under that same address so it reads/writes just like any other
- * `SettingsManager` field.
+ * `login.localConfig`, which `LoginManager` already mirrors to `localStorage` —
+ * keyed under that same address.
+ *
+ * Unlike a real `SettingsManager` field, this manager's signed-in read comes
+ * from its own dedicated record, not from `profile.config` — so when
+ * `LoginManager`'s brand-new-account adoption copies a device's whole
+ * `localConfig` (including a selection made here) into `profile.config` and
+ * clears `localConfig`, that lands somewhere this manager doesn't normally
+ * read from. `load()` below has an explicit fallback for exactly that case:
+ * when this manager's own record comes back empty, it checks `profile.config`
+ * for an adopted selection and, if found, both applies it and writes it into
+ * the record so this fallback is only needed once per account.
  */
 export interface CustomizationVariantSelectionsManager {
   /** Locator -> chosen variant id, for the current viewer (signed-in profile or signed-out device). Empty when nothing chosen yet. */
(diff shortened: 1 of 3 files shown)


Blitzy:
> [!NOTE]

## Code review — Support signed-out variant selections via `login.localConfig`

Nice, focused change. The core goal — a signed-out viewer's variant pick now survives a page reload via `login.localConfig` → `localStorage` — is implemented cleanly, the public API stays unchanged, and there's no circular import from re-exporting `parseStringRecord`. The tests are mostly real-behavior tests and genuinely fail on the pre-fix no-op (verified below). A few things worth a look, none of them blocking.

I ran `pnpm vitest run CustomizationVariantSelectionsManager.test.ts` — all 8 pass.

---

### Medium

**No. 1 — The "adoption on first login" behavior the PR claims does not actually happen for these selections.**

The doc comment (lines 27-31) and the PR's Implementation Details both assert that signed-out choices get *"the same persistence **and adoption behavior** as any other anonymous setting — no special handling needed"*, because `LoginManager` "offers to adopt into a brand-new account's profile on first login."

That adoption works for real `SettingsManager` fields **only because their signed-in read source is `profile.config`** — the exact place adoption writes to (`getProfileConfigValue(profile, key)` in `SettingsManager.readSettings`, `SettingsManager.tsx:816-819`). This manager is different: its signed-in read is a **dedicated CasualOS record**, `os.getData(userId, VARIANT_SELECTIONS_ADDRESS)` (`CustomizationVariantSelectionsManager.tsx:54`), and the `selections` computed returns `remoteSelections` only when signed in (line 96-98) — it never falls back to `profile.config` or `localConfig`.

So the adoption path in `LoginManager.getUserProfile` (`LoginManager.tsx:550-566`) copies `localConfig` (now including `customizationVariantSelections`) into `profile.config` **and then clears `localConfig`** — depositing the selection in a location this manager never reads, while the record it *does* read is still empty (`data_not_found`).

Net effect for the user:
1. Signed out, open a `?customization=...` share link, pick a non-default variant → persists across reload. ✓
2. Sign up / first login on that device.
3. The customization reverts to its **default** variant, because the signed-in record is empty and `localConfig` was cleared during adoption. The pick also now sits orphaned in `profile.config.customizationVariantSelections`, read by nothing.

To be clear, this is **not a regression** versus `develop` (before this PR, signed-out picks weren't saved at all, so login lost them too) — but the PR text advertises adoption that isn't delivered, and a future maintainer reading that comment could reasonably rely on it. The `profile.config` schema is `z.record(z.string(), z.unknown())` (`LoginManager.tsx:191`), so the orphan key won't break profile validation — it's just dead data.

*Suggested fix (pick one):*
- Cheapest: correct the comment and PR description so they don't claim adoption, and state explicitly that a signed-out pick is not carried into the account on login.
- Or implement real migration: when the signed-in `load()` gets `data_not_found`, seed `remoteSelections` from `parseStringRecord(profile.config?.[VARIANT_SELECTIONS_ADDRESS])` (or from `localConfig` before it's cleared) and write it back to the record so it becomes durable.

---

### Low

**No. 2 — The "survives a simulated reload" test overstates what it verifies.**

`test/unit/.../CustomizationVariantSelectionsManager.test.ts:95-111` shares one in-memory `login.localConfig` signal between both manager instances and never exercises the real `writeLocalConfig` / `readLocalConfig` helpers (`LoginManager.tsx:306`, `520`, `537`) — i.e. the actual `localStorage` serialize/rehydrate round-trip a reload performs. It does prove the value lives in `login.localConfig` rather than in manager-instance memory (and it *does* catch the pre-fix no-op), so it's not vacuous — but it would still pass even if the real persistence layer were broken, which is more than the test name implies. Consider either round-tripping through the real read/write helpers, or renaming it to something like "reads its selections from `login.localConfig`, not instance memory."

**No. 3 — No test that a `selectVariant` write preserves unrelated `localConfig` keys.**

`selectVariant` (signed-out branch) does `{ ...login.localConfig.value, [VARIANT_SELECTIONS_ADDRESS]: {...} }` (lines 109-115), so a sibling key like `fontSize`/theme is preserved — good — but nothing asserts it. Since `localConfig` is a shared store for every anonymous setting, a one-line regression guard is worth it: seed `login.localConfig.value = { fontSize: "XL" }`, call `selectVariant`, assert `fontSize` still `"XL"`. The existing "doesn't clobber another's stored selection" test only covers merging *within* the address, not sibling keys.

---

### Reproduction for No. 1

No standalone runtime repro harness, but the flow above (signed-out pick on a customization share link → first login on the same device → variant reverts to default) exercises it against the deployed branch build. Worth confirming manually before deciding between the "fix the comment" and "implement migration" options.

---

Verified: `pnpm vitest run CustomizationVariantSelectionsManager.test.ts` → 8 passed. The two new persistence tests fail against the pre-fix (no-op) `selectVariant`, so they're testing real behavior, not passing vacuously.

---

PR #1761, opened 2026-09-04
Hide Customize settings nav item when user is signed out

KallynGowdy:
## Summary

The Customize navigation item in the Settings view is now only shown to signed-in users. When a user is signed out, the Customize option is hidden from the settings navigation menu.

## Changes

- **SettingsMainView component**: Added a computed signal `isLoggedIn` that checks whether `state.login.userId` is not null, and wrapped the Customize nav item in a conditional render that only displays it when the user is logged in.
- **Test coverage**: Added comprehensive unit tests for `SettingsMainView` that verify the Customize nav item is hidden when signed out and visible when signed in.

## Implementation Details

The change uses `useComputed()` to create a reactive boolean that tracks login state, ensuring the UI updates automatically if the user logs in or out while the settings view is open. The conditional rendering is placed at the JSX level using a ternary operator, keeping the logic simple and declarative.

https://claude.ai/code/session_01EiLGnUUTFNVVvLpeEYEssL

### DIFF
--- a/packages/seed-bible/seed-bible/components/SettingsPage/SettingsPage.tsx
+++ b/packages/seed-bible/seed-bible/components/SettingsPage/SettingsPage.tsx
@@ -2387,6 +2387,7 @@ function CustomizationsSettingsView(props: { state: SeedBibleState }) {
 function SettingsMainView(props: { state: SeedBibleState }) {
   const { state } = props;
   const { t, language, availableLanguages, setLanguage } = useI18n();
+  const isLoggedIn = useComputed(() => state.login.userId.value !== null);
   const isLanguageMenuOpen = useSignal(false);
   const languageSearchQuery = useSignal("");
   const languageTriggerRef = useRef<HTMLButtonElement | null>(null);
@@ -2525,22 +2526,24 @@ function SettingsMainView(props: { state: SeedBibleState }) {
               </span>
             </button>
           </li>
-          <li>
-            <button
-              className="sb-settings-nav-item"
-              onClick={() => onNavigate("customizations")}
-            >
-              <span className="sb-settings-nav-icon">
-                <MaterialIcon>palette</MaterialIcon>
-              </span>
-              <span className="sb-settings-nav-label">
-                {t("customize", { defaultValue: "Customize" })}
-              </span>
-              <span className="material-symbols-outlined rtl-mirror">
-                chevron_right
-              </span>
-            </button>
-          </li>
+          {isLoggedIn.value && (
+            <li>
+              <button
+                className="sb-settings-nav-item"
+                onClick={() => onNavigate("customizations")}
+              >
+                <span className="sb-settings-nav-icon">
+                  <MaterialIcon>palette</MaterialIcon>
+                </span>
+                <span className="sb-settings-nav-label">
+                  {t("customize", { defaultValue: "Customize" })}
+                </span>
+                <span className="material-symbols-outlined rtl-mirror">
+                  chevron_right
+                </span>
+              </button>
+            </li>
+          )}
           <li>
             <div className="sb-settings-field-row">
               <span className="sb-settings-field-label">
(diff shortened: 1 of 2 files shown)


Blitzy:
> [!NOTE]

## Code Review — Hide Customize settings nav item when signed out

**Overall:** This is a small, focused, and correct change. It hides the "Customize" nav item in the main settings list unless the user is signed in, using `useComputed(() => state.login.userId.value !== null)` and gating the `<li>` with `isLoggedIn.value && (…)`. The idiom matches the existing pattern already used a few hundred lines up in `AccountSettingsView` (`const isLoggedIn = useComputed(() => login.userId.value !== null)`), so it reads as consistent with the codebase. No linked issue is referenced by the PR, and the only PR comment so far is the automated deployment link.

I verified the reachability question: the `"customizations"` settings view is only ever entered through this one nav item (`onNavigate("customizations")` at `SettingsPage.tsx:2533`). There is no deep-link or share-link path that sets `requestedSettingsView` to `"customizations"` from a URL, so hiding the nav item genuinely makes the view unreachable for signed-out users — the fix is complete for its stated goal.

**On the tests:** I ran the new `SettingsMainView.test.tsx` and confirmed both cases pass. I also confirmed the "hides" case is a real regression test, not one that always passes: I temporarily reverted the guard (`isLoggedIn.value &&` → `true &&`) locally, re-ran, and the "hides the Customize nav item when the user is signed out" test failed as expected (`expected [ 'Account settings', …(5) ] to not include 'Customize'`), then passed again once restored. So the test actually exercises the behavior it claims to.

---

### Findings by severity

#### Low — the claimed reactivity is not tested
The PR description highlights that the item updates automatically "if the user logs in or out while the settings view is open." That reactive transition is the interesting part of using `useComputed` (vs. a one-time check), but neither test covers it — both tests render once against a fixed `userId`. The two tests would pass equally well against a non-reactive implementation.

*Suggested addition* — flip the signal in place and re-assert, which locks in the reactive behavior:

```tsx
it("reveals the Customize nav item when the user signs in while open", () => {
  const state = renderMain(null);
  expect(navLabels()).not.toContain("Customize");

  act(() => {
    state.login.userId.value = "user-1";
  });

  expect(navLabels()).toContain("Customize");
});
```

This is a coverage gap, not a defect — the implementation is reactive as written.

#### Low (defense-in-depth) — the view itself is not guarded, only the entry point
`SettingsPage` still renders `<CustomizationsSettingsView>` whenever `requestedSettingsView.value === "customizations"` (`SettingsPage.tsx:2756`), regardless of login state. Today the only way to set that value is the nav item you just hid, so this is not currently reachable while signed out. The one theoretical edge case is a user who is already on the Customize view and then signs out — the view would keep rendering with the nav item hidden. In practice signing out happens from Account settings, so the user would have already navigated away, making this unreachable through the UI. Worth a mention only as a note: if you'd prefer belt-and-suspenders, you could also early-return from `SettingsMainView`'s navigation (or from the `"customizations"` branch in `SettingsPage`) when signed out. Not required for this PR.

#### Nit — mock state uses `as unknown as SeedBibleState`
`createMockState` casts a hand-built partial through `as unknown as SeedBibleState`, which bypasses type checking. That's a common and pragmatic pattern for component tests, but it means the test won't catch a future change to the real `login`/`sidebar` shape — it will keep compiling against a stale mock. No change needed; just be aware the type cast is doing real work to silence the compiler here.

---

### Reproduction (for the behavior being fixed)
1. Open the app while signed out.
2. Open Settings → the main settings list.
3. **Before this change:** the "Customize" (palette) item appears even though customizations require an account.
4. **After this change:** the "Customize" item is absent when signed out and present after signing in.

### Verdict
Approve-worthy as-is. The change is minimal, correct, consistent with existing conventions, and backed by a genuine regression test. The only thing I'd actively suggest adding is the reactive-transition test above; the other two points are informational.

---

PR #1762, opened 2026-09-04
Persist reading history locally to survive tab closes

### LINKED ISSUE #1754: Reading history can silently drop recently-read chapters — no local store, nothing replays on reconnect
## What is happening

Chapters you have genuinely read sometimes never appear as read on the Today screen. It shows up most often as the *tail* of a reading session going missing: the chapters you read early in the evening are there the next morning, the last few are not.

## What is expected

Every chapter you spend real time in is recorded, and it survives closing the app, reloading, and reading with a poor or absent connection.

## Why it happens

Reading history is stored differently from every other piece of user data in the app, and the difference is the problem.

A reading event is `{userId, bookId, chapter, start, end}` (`managers/ReadingHistoryManager.tsx:13`), appended to a Yjs shared document — one document per user per calendar year, addressed as record `<userId>` / inst `reading_history` / branch `doc/<year>` (`managers/ReadingHistoryManager.tsx:54`). While a chapter is on screen, an effect writes every 5 seconds (`managers/SeedBibleStateManager.tsx:1339`), bumping the existing event's `end` if you read the same chapter within the last 30 minutes.

The first tick landing 5 seconds in rather than immediately is **deliberate** — skimming past a chapter should not count as having read it. Anything built on top of this needs to preserve that gate.

That document lives **only in memory** and only reaches disk by being synced to the server over a live websocket. There is no local copy and nothing replays a failed write.

Four checks, all negative:

- `managers/indexedDbUtils.ts` has exactly two consumers: `OfflineTranslationStore.ts` and `OfflineAnnotationStore.ts`.
- `online` event listeners exist in exactly two files: `AnnotationSyncManager.tsx:813` and `OfflineTranslationsManager.tsx:697`.
- No `y-indexeddb` dependency, and `getSharedDocument` (`managers/OsManager.tsx:232`) builds the document with `{recordName, inst, branch, markers}` and no persistence provider.
- The service worker is Workbox asset caching (`vite.config.ts:152`); it does not queue data writes.

To be precise about the boundary: Yjs is a CRDT, so the in-memory document does accumulate edits while the socket is down, and a reconnect *within the same page load* should carry them across. What is missing is durability across the page lifetime. Reload, close the tab, or let iOS freeze and discard the PWA, and anything not yet on the wire is gone.

A fresh load with no connection cannot even get started. `getSharedDocument` awaits the initial sync before returning, and the resulting promise is cached in the module-level `readingHistoryDocs` map (`managers/ReadingHistoryManager.tsx:40`). If that promise rejects — expired session key, auth failure, dropped connection — the **rejected** promise stays cached, so every save for the rest of that page load fails. Silently: the 5-second interval discards the returned promise, so a rejection surfaces as an unhandled rejection and nothing else. `clearReadingHistoryDocs()` exists but nothing calls it.

Worth noting that `login.userId` is computed from the *cached* session key in localStorage (`managers/LoginManager.tsx:325`), so the app can hold a healthy-looking user id while the key is actually expired — reads and writes both fail, quietly.

## What annotations already do

This is a gap, not a platform limitation. `OfflineAnnotationStore` keeps every note in IndexedDB with a `pendingOp` field (`"upsert" | "delete" | null`) and an `attempts` counter, and the index on `["owner", "pendingOp"]` *is* the outbound queue — rows with `pendingOp: null` drop out of the index on their own because IndexedDB cannot index null. `AnnotationSyncManager` drains that queue, separates retryable failures from permanent ones, and re-drains on the `online` event. Offline translations have their own equivalent.

Reading events are strictly easier than annotations: appends to a CRDT array merge cleanly and `end` is monotonic, so none of the read-then-write conflict machinery is needed.

## Sketch of a fix

1. An IndexedDB store of pending events keyed by `${userId}/${year}/${bookId}/${chapter}/${start}`, written by the same 5-second tick that currently writes to the Yjs document — so the skim gate is preserved for free.
2. A drain that replays pending rows into the year document once it is available — on load and on `online`.
3. Read from the union of the local store and the synced document, so today's reading appears immediately rather than after a round trip.
4. Stop caching rejected promises in `readingHistoryDocs`, and retry rather than poisoning the key for the rest of the page load.

Items 1 and 3 alone would likely cover most of the reported loss.

## Adjacent defects found while tracing this

Separable from the above; listing them here so they are not lost. Happy to split any into their own issue.

- [ ] **Background throttling under-counts `end`.** Browsers throttle `setInterval` to roughly once a minute in a hidden tab and stop it entirely in a frozen one, so reading with the screen locked records far less time than elapsed.
- [ ] **The heatmap silently drops anything under 60 seconds**, while the chapter grid does not — so the same day can show chapters in the grid and an uncoloured square in the timeline. Two filters: `minDurationSeconds: SEC_PER_MINUTE` (`components/TodayPane/useReadingHistoryTimeline.tsx:186`) and `totalTimeSpentReading > SEC_PER_MINUTE` (same file, line 265). The filtering is reasonable in spirit; the two surfaces disagreeing about it is the problem.
- [ ] **The "Last 48 hours" window is pinned at mount.** `initialOption` is `useMemo(..., [])` (`components/TodayPane/SocialSection.tsx:77`), so it means 48 hours before the Today screen first rendered, not before now, and goes stale in a long-lived session.
- [ ] **`start` and `end` are filtered inconsistently.** The fetch keeps events by `start` (`managers/ReadingHistoryManager.tsx:414`), then `getCommunityReading` re-filters by `end` (`managers/TodayReadingHistory.ts:129`). A session beginning just before the window opens is dropped even though nearly all of it falls inside.
- [ ] **Day buckets drift across DST.** `loadDailyReadingHistory` derives the day index by dividing by a fixed 86400 from local midnight (`managers/ReadingHistoryManager.tsx:656`), while the grid's day ranges come from real local midnights (`managers/ReadingHistoryTime.ts:30`). In a week containing a clock change the two disagree by an hour, putting late-night reading on the neighbouring square.
- [ ] **Dead code in `ReadingHistoryManager.tsx`:** `getTodayTimeSpan`, `getPastYearTimeSpan`, `getCurrentYearTimeSpan`, `getUserReadingHistorySummary`, `saveUserReadingHistory` and `clearReadingHistoryDocs` are referenced only by tests.

Blitzy:
Closes #1754

## Summary

Reading history events are now recorded to IndexedDB before being pushed to the server, so chapters read are not lost when the tab closes, the page reloads, or the browser freezes the PWA. Previously, events lived only in memory and synced over a live websocket — anything not yet sent when the connection dropped was gone.

## Key Changes

- **New `OfflineReadingHistoryStore`** (`OfflineReadingHistoryStore.ts`): IndexedDB-backed local storage for reading events. Each event is keyed by `userId/year/bookId/chapter/start` and carries a `pendingOp` flag. Rows marked `"append"` are the outbound queue; once synced to the year document, `pendingOp` is cleared to null (IndexedDB skips null-indexed rows, so the index naturally becomes the queue). Synced rows are kept rather than deleted so the Today screen can show today's reading immediately offline; `prune()` removes old synced rows to prevent unbounded growth.

- **New `ReadingHistorySyncManager`** (`ReadingHistorySyncManager.tsx`): Replays locally-recorded events into their year documents on sign-in, reconnect, or when the browser comes back online. Handles transient failures (no connection, expired session) by leaving rows queued for the next pass. No attempt cap — failures are connection issues, not content issues, so retrying indefinitely is safe and necessary to avoid losing genuine reading.

- **Updated `ReadingHistoryManager`**:
  - `saveReadingHistorySpan()` records to the local store first (if available), then pushes to the document. Failures to reach the document leave the row queued for replay. `saveReadingHistory()` delegates to it with `start === end`.
  - New `writeReadingEventsToDocument()` function that extends existing events rather than duplicating them — matching on user/book/chapter/start and only moving `end` forward. This is what both the initial push and the replay use, so the same event can be safely pushed twice out of order.
  - `getReadingHistoryEvents()` now merges rows from the store with the document, so reading shows up immediately even if the push hasn't landed yet.
  - Fixed a bug where a failed document fetch was cached, poisoning all later reads of that year for the rest of the page load.

- **Test coverage**: New test suites for the store and the sync manager, plus tests in `ReadingHistoryManager.test.ts` covering durability scenarios (failed push, retry, deduplication, offline reading, partial per-year failure, concurrent sync passes).

## Implementation Details

- Writes are recorded a *stretch* at a time (`saveReadingHistorySpan`), matching the two writers that exist after merging develop's #1738: the reader's five-second tick credits the stretch since the last one, and audio playback reports listening after the fact from how far its own clock advanced. The local store buckets a row into the UTC year of the stretch's end and joins a sitting judged from the stretch's start, both matching what the document write does, so the local copy and the server's copy always agree about which event a stretch belongs to.
- The store returns null when IndexedDB is unavailable (server-side rendering, private windows, sandboxed frames, or user-blocked storage). Callers fall back to writing straight to the document, preserving the pre-store behavior. A store that exists but refuses the write falls back the same way.
- Events are extended rather than replaced: if a push fails and a second tick lands before the retry, the local row grows to include both, and the replay sends the larger event. `end` only ever moves forward, in the store as well as the document, so out-of-order pushes can't shrink an event.
- A single database connection per page load (via `getSharedReadingHistoryStore()`), not one per caller, since all reading events belong in the same store.
- Synced rows are pruned after 400 days by default, keeping offline history readable without accumulating forever.

## Not included

The "Adjacent defects" checklist in #1754 is intentionally untouched (background throttling under-counting `end`, the heatmap/grid threshold mismatch, the pinned 48-hour window, `start`/`end` filtering inconsistency, DST day-bucket drift, dead code). They are separable from the durability fix and each is an independent behaviour change. Since this PR closes #1754, they need their own tracking issue so they aren't lost.

https://claude.ai/code/session_014s1HHW9VjJbMmd9R2q2fXx


Blitzy:
> [!NOTE]

## Summary

This PR implements the fix sketched in issue #1754: reading events are now recorded to a new IndexedDB store (`OfflineReadingHistoryStore`) before being pushed to the year document, and `ReadingHistorySyncManager` replays anything that didn't make it out on sign-in, reconnect, and the `online` event. I traced through the core algorithm in detail — the monotonic-`end` invariant, the identity matching used for dedup/extend (`readingEventIdentity`), the `markSynced` end-mismatch guard that protects against a row being extended while its push is in flight, and the per-year partial-failure handling in `runPass` — and did not find a functional correctness bug in any of it. The design is careful and the reasoning in the code comments matches what the code actually does.

The main gaps are in test coverage, which matters here because this PR is explicitly about durability under failure — the scenarios that are hardest to get right are exactly the ones that need the most test pressure.

## Findings

### Medium: `ReadingHistorySyncManager.test.ts` can't fail a single year while another succeeds

`runPass` (`ReadingHistorySyncManager.tsx:136-175`) is written to handle multiple pending years independently — one year's document write can fail while another succeeds, and only the failed year's rows should stay queued. That's the right behavior, but nothing in the test suite can actually exercise it: `recordingWriter()` (`ReadingHistorySyncManager.test.ts:29-49`) has a single global `failWith` flag that fails (or doesn't fail) every `writeEvents` call uniformly. Every multi-year test (e.g. "replays events recorded before this load, grouped by year", lines 117-127) only checks the all-succeed case.

**How to see the gap:** if `runPass` were refactored to collect all pending rows across years and call `store.markSynced` once after the loop (instead of per-year, as it does now), a partial failure would incorrectly mark the failed year's rows as synced too — silently dropping reading history exactly the way issue #1754 describes. No existing test would catch that regression, because none of them ever fail one year while succeeding another.

**Suggested fix:** let `recordingWriter()` take a per-year failure predicate (e.g. `failYear(year, error)`), then add a test with two years where one fails and one succeeds, asserting the failed year's rows stay pending and the successful year's don't.

### Medium: The grouped-write assertion doesn't check event content

`ReadingHistorySyncManager.test.ts:124` — `expect(writer.writes.map((w) => w.year).sort()).toEqual([2025, 2026])` — only checks that two years were written to, never what was actually sent for each year. A bug that pushed the wrong events into a year (e.g. all rows bundled into one year's write, or the wrong event's fields) would still pass this test as long as the year numbers came out right.

**Suggested fix:** also assert on the payload, e.g. `writer.writes.find((w) => w.year === 2025)?.events` matches the expected single event for that year.

### Medium: No test that concurrent `sync()` calls join the in-flight pass

`sync()`'s whole `running` promise mechanism (`ReadingHistorySyncManager.tsx:177-214`) exists so that calling `sync()` while a pass is already running joins it instead of starting a second pass — this matters because `recordSpanForCurrentUser` in `ReadingHistoryManager.tsx:1164-1166` calls `sync.sync()` opportunistically after every successful push. Nothing in the test suite calls `sync()` twice without awaiting the first call, so a regression that let two passes race (e.g. double-pushing the same batch, or a data race in `pendingRows`) would go unnoticed.

**Suggested fix:** call `sync.sync()` twice back-to-back without awaiting the first, then assert `writeEvents` was only invoked once for the pending batch.

### Low: Missing the sibling pattern's IndexedDB-unavailable test

`OfflineAnnotationStore.test.ts:66-71` has a one-line test asserting `createIndexedDbAnnotationStore()` returns `null` in this environment (jsdom has no `indexedDB`). `OfflineReadingHistoryStore.test.ts` never calls `createIndexedDbReadingHistoryStore` at all — the PR description explicitly points to `OfflineAnnotationStore` as the pattern being followed, and this one small assertion from that pattern didn't carry over.

**Suggested fix:** add the equivalent one-liner (`expect(createIndexedDbReadingHistoryStore()).toBeNull()`). Worth noting this doesn't fix the deeper issue that the *real* IndexedDB code path (schema/indices/transactions in `OfflineReadingHistoryStore.ts:300-497`) has no coverage from either test file — that's a pre-existing limitation of the test environment (no `fake-indexeddb` dependency), not something introduced by this PR, so I'd call it out-of-scope for this review but worth a follow-up if reading-history durability is important enough to justify adding that dependency.

### Low: `getSharedReadingHistoryStore()` / `resolveStore()` singleton is untested

Every test in `ReadingHistoryManager.test.ts` passes an explicit `store` option, so the actual behavior production code relies on — the lazy module-level singleton in `ReadingHistoryManager.tsx:63-90`, and the "caller passed nothing → use the shared store" branch in `resolveStore` — is never exercised.

### Low: Minor nits
- `OfflineReadingHistoryStore.test.ts` "ignores an event that is no longer stored" (~line 230) only asserts the promise resolves, not that the rest of the store's state is untouched.
- The exact boundary where a gap equals `joinThresholdSeconds` (as opposed to `threshold + 1`) isn't tested in `extendOrCreateReadingRow`.
- `saveReadingHistorySpan`'s comment "No local store: the document is the only place the span can go" (`ReadingHistoryManager.tsx:387`) is a little misleading — that branch also runs when a store exists but `recordReadingLocally` failed, not only when there's no store at all.

### Informational: issue linkage
This PR doesn't include a "Closes #1754" (or similar) in its description, so merging it won't auto-close the tracking issue. Also worth confirming with whoever triages it whether the "Adjacent defects" checklist in #1754 (background throttling under-counting, heatmap/grid threshold mismatch, the pinned 48-hour window, start/end filtering inconsistency, DST day-bucket drift, dead code) should be split into their own follow-up issue now, since this PR intentionally doesn't touch any of them — which is reasonable, they're genuinely separable, but they'll be easy to lose track of if nothing tracks them once this issue closes.

---

Blitzy:
> [!NOTE]

## Summary

Re-review of this PR after the fixes for the two earlier review comments landed, followed by a full manual test pass against a 13-check plan covering the offline and reconnect paths.

**Outcome: nothing blocking.** All findings from the previous reviews are resolved, and the manual pass turned up three further bugs — all in the coming-back-from-offline path, all now fixed on the branch. CI is green and the branch is up to date with `develop`.

The part worth the team's attention is the manual findings. They were real data loss, they survived two rounds of code review, and none of them could have been caught by the test suite as it stood.

## What the earlier reviews found

The Sept 4 comment raised five test-coverage gaps; the Sept 9 comment raised a pruning bug, a misreported error, a lost error message, and sequential year replays. All are addressed, and I re-checked each one exercises what it claims rather than just passing.

Two further findings from this round, now fixed:

- **A timed-out document was abandoned but never closed** (`643d622cd`). The 10-second deadline lived above `getSharedDocument`, so giving up freed the caller while the document underneath stayed connected and watching its branch. With the cache correctly dropping failures so they retry, that left another connected document behind roughly every 10 seconds of reading. The deadline now belongs to the layer that owns the document, so the existing unsubscribe runs.
- **A row left queued after a successful push went unnoticed** (`a6c5faae6`), plus a zero-padded sort key, an orphaned doc comment, and a disposed manager that kept working.

CI was also failing on this PR for a reason that was not a failing test: all 5925 tests passed while the run exited 1 on a single unhandled rejection, from a test that let a promise reject before attaching its assertion (`da654707f`).

## What manual testing found that review did not

All three are fixed in `75da619b8`.

**No. 1 — a write into a disconnected document was recorded as delivered.** Writing an event is a local CRDT edit, so it succeeds with the network gone, and a year document cached before the connection dropped answers instantly afterwards. The push was counted as delivered on that basis and the row's `pendingOp` cleared — so reading recorded offline was lost when the page went away, with nothing left to replay it. That is precisely the failure this PR exists to prevent, surviving inside the fix for it.

**No. 2 — the document's own status arrives too late to be trusted alone.** The first fix for No. 1 read the document's sync status, which was correct in principle but still reproduced in the browser. `connectionState` is driven by `socket.onclose`, and pulling a connection out from under a socket leaves it half-open with no close event until a send times out — minutes later. The document went on reporting the last thing it had heard. The fix requires `navigator.onLine` *and* the document's status to agree; each is blind to what the other catches.

**No. 3 — the replay gate could wedge permanently.** `sync()` gated on a signal carried only by the `online`/`offline` window events, so one missed `online` shut it for the rest of the page load. It now reads the network at the moment it is asked.

### Why the suite could not have caught these

The offline tests mocked `getSharedDocument` to **reject**. There was no fake for "resolves fine but the socket underneath is dead" — in the tests' world a resolved document was a working one, so the state simply could not be expressed. Two tests also dispatched an `offline` event while leaving `navigator.onLine` at `true`, a combination no browser produces.

Both gaps are closed: the fake document now has a status channel, and connection changes in tests set the event and `navigator.onLine` together. Each of the three fixes has a regression test verified to fail against the pre-fix code.

## Verification

- `pnpm test` — 257 files, 6126 tests, exit 0
- `pnpm check:ts` — clean
- `pnpm lint` — 0 errors
- CI on this PR — `build (24.x)` and `deploy (24.x)` passing
- 13 of 13 manual checks passing, including offline recording, replay on reconnect, deduplication, sitting boundaries, and sign-out retention on a shared device

Sign-out behaviour was checked specifically for the shared-device case: every read of the local store is scoped to the signed-in user, so a second account on the same device neither sees nor replays the first account's queued reading.

---
_Generated by [Claude Code](https://claude.com/claude-code)_

Blitzy:
> [!NOTE]

## Summary

This PR has already been through three thorough review rounds (Sept 4: five test-coverage gaps; Sept 9: a pruning bug, a misreported error, and sequential-replay/error-visibility issues; Sept 14 manual-testing round: three genuine data-loss bugs found by hand-testing that no automated test could have caught). Rather than re-summarize those, I did an independent pass to verify them rather than take them on faith.

I read the current `OfflineReadingHistoryStore.ts`, `ReadingHistorySyncManager.tsx`, and the connection-tracking additions in `ReadingHistoryManager.tsx`/`OsManager.tsx` line by line and traced the core algorithm — the monotonic `end` invariant, identity-based dedup/extend, the per-year partial-failure handling in `runPass`, and the dual `navigator.onLine` + document-status "synced" gate — against the fix commits' own descriptions. For the newest and highest-stakes fix (`75da619b`, "only clear a reading row once it has actually reached the server"), I went further: I reverted the three source files to the commit *before* that fix while keeping the current tests, and reran the suite. 7 tests failed against the pre-fix code, in exactly the scenarios the fix describes (an offline write miscounted as delivered, a missed `online` event wedging the sync gate shut permanently). That confirms these are real regression tests, not ones that happen to pass either way.

I also independently ran the full suite for the four changed test files (114 tests, all pass on the current head), `pnpm check:ts` (0 visible errors), and confirmed CI on this exact head commit (`7f8b0bdd`) — `build (24.x)` and `deploy (24.x)` — is green. The CHANGELOG has an accurate `TBD` entry for this fix, and the PR description now says "Closes #1754" (an earlier round had flagged that it didn't).

**Outcome: nothing blocking.** I did not find a new correctness or data-loss issue beyond what the last three rounds already found and fixed.

## Findings

Nothing severe enough to block merging. One informational item worth carrying forward:

### Informational: the year-boundary quirk noted on Sept 9 is still present, by design
`extendOrCreateReadingRow` / `writeReadingEventsToDocument` bucket an event by the UTC year of its *end*, not its *start*, so a continuous read spanning a New Year's Eve midnight opens a second event in the next year's document instead of extending the first. This is pre-existing behavior, unchanged by this PR, and in the same family as the DST day-bucket drift that issue #1754's "Adjacent defects" list already tracks as separable and out of scope for this fix. Not a finding against this PR — just confirming it's still true on the latest commit, in case it needs to be carried into whichever follow-up issue ends up tracking those adjacent defects (the PR description says that tracking issue hasn't been created yet).

## Test quality

Matches the bar in this repo's CLAUDE.md: mocks sit at the real boundary (`OsManager`/`getSharedDocument`, `navigator.onLine`, window events), not sibling managers; the in-memory store used in tests shares the real IndexedDB store's semantics rather than stubbing behavior; and the offline/reconnect tests assert on observable state (`store.listPending`, `pendingCount`, what was actually written to which year) rather than internals. I didn't just take that on faith for the most safety-critical fix — see the revert-and-rerun above.

## Not re-litigated

Everything the Sept 4, Sept 9, and Sept 14 review comments raised has a corresponding commit addressing it, and I confirmed the tests attached to those commits genuinely exercise the fix rather than merely existing. I'm not repeating that list here since it's already on the PR's timeline.

---

Issue #1659, opened 2026-08-19
Finish the saves migration by dropping the legacy bookmarks record


Blitzy:
**Depends on #1657. Gated on telemetry — do not start on a schedule.**

## Currently

#1657 moved saves from the `bookmarks` record address to `saves` using an expand/contract migration, and deliberately stopped at the expand half: new clients write `saves`, fall back to reading the legacy `bookmarks` record when `saves` is absent, and never delete the legacy record. Both records coexist by design, so a rollback or a stale client can't lose anything.

This issue is the contract half. It matters for the reason the rename happened at all: while both records exist, anyone inspecting a user's data still finds a `bookmarks` blob full of saves, and the next person to build something bookmark-shaped can still mistake it for bookmark data. The code is already unambiguous; this makes the data unambiguous too.

## Preconditions

**This is gated on the telemetry added in #1657, not on a date.** Before starting:

- The **migration ran** event has flatlined near zero — new migrations have stopped because essentially everyone has been migrated.
- The **stale writer detected** event is at zero over a sustained window. If it isn't, stop and recover those users from their legacy record first; the whole point of keeping it was to make that possible.

If either check fails, this issue waits. Nothing about it is urgent.

## Changes

1. Drop the legacy fallback read and the copy-forward write from `SavesManager`, along with the `LEGACY_SAVES_ADDRESS` constant. `saves` becomes the only address the manager knows about.
2. Erase the legacy record. `os.eraseData(userId, "bookmarks")` already exists (`OsManager.tsx:333`) and takes the same calling pattern as `recordData`, so no new plumbing is needed.
3. Remove the migration-specific tests from `SavesManager`'s suite, keeping the ones that cover normal load/persist behavior.

**On the erase step:** only ever erase after confirming the user has a non-empty `saves` record. A user who somehow reaches this code with an empty `saves` and a populated legacy record is exactly the person the erase would hurt — leave theirs alone and let the telemetry surface it.

Two ways to do the erase, and they're worth a moment's thought:

- **Client-side, on load.** Simple, no new infrastructure, and self-limiting — it only touches users who actually open the app. But it also means users who never return keep their legacy record forever, so the cleanup is never actually complete.
- **A one-off sweep.** Actually finishes the job, but needs a script with access to enumerate user records, which is a bigger lift and a riskier blast radius.

[...]

Issue #1658, opened 2026-08-19
Redesign bookmarks as named, colored chapter markers


Blitzy:
## Currently

Bookmarks are categorized, archival, and can point at either a chapter or a verse range — see #1657, which moves all of that behavior to the new Saves system. That leaves "bookmark" free to mean what people actually expect it to mean: a temporary marker for where you are, that you move as you go.

This task is to build that. It should land **after** #1657, so bookmarks start from an empty slate rather than being reshaped in place.

There is prior design exploration for the move-an-existing-bookmark flow in #1267, with a Figma linked in its comments. Worth starting from that rather than cold — and #1267 is probably superseded by this issue.

## What a bookmark is now

- **No categories.** Bookmarks are flat.
- **Book + chapter only.** No more verse-level bookmarks.
- **Named, with a color.** You can call one "Reading plan" and another "Sermon prep".
- **Multiple bookmarks can sit on the same book + chapter.** They're purpose-built markers, so "My bookmark" and "Reading plan bookmark" can both live on Genesis 10 if that's where each of them needs to be.
- **Soft cap of 5, account-wide.** Not per translation — a bookmark carries its translation with it, so the cap counts bookmarks, not chapters or translations. This is the design pressure that keeps bookmarks from becoming an archive: they're markers you move, not a collection you grow. Saves is where things accumulate.

## How bookmarking works

**Tapping the bookmark button always opens the bookmark modal.** No exceptions, no first-run special case.

The modal lists your bookmarks. Pick one and hit save, and that bookmark moves to the current chapter. Your first bookmark is preselected, so the common case — someone with a single bookmark — is two taps: bookmark button, save. You can create an additional bookmark from inside the modal, and each row has a remove button that deletes that bookmark.

**When you have no bookmarks yet, the modal opens with a premade one already in the list** — named "My bookmark", default color, not yet pointing anywhere. Hit save and it lands on the current chapter. That's the same two taps and the same modal as every subsequent bookmarking action.

This is the part worth protecting during implementation: the premade bookmark is not a special first-run mode, it's the ordinary "create a bookmark" form with its name and color prefilled. First bookmark and fifth bookmark are one code path. If the implementation ends up with a branch for "user has no bookmarks," something has gone sideways.

### 1. The reader's bookmark button

`ReaderBookmarkButton` in `BibleReader.tsx`, in both the desktop reader header and the mobile header. It always opens the modal, and its icon reflects what is on the current chapter:

[...]

PR #1763, opened 2026-09-04
Repurpose the bookmark system into Saves

### LINKED ISSUE #1657: Repurpose the current bookmark system into a Saves system
Part 1 of 3: this issue (Saves), #1658 (bookmark redesign), #1659 (migration cleanup).

## Currently

What we call a "bookmark" today is really an archival save. `BookmarksManager` (`packages/seed-bible/seed-bible/managers/BookmarksManager.tsx`) already stores:

- a location (translation + book + chapter, optionally narrowed to a single verse or an inclusive `[start, end]` verse range),
- membership in one or more named categories (folders), defaulting to "My Bookmarks",
- a `createdAt` timestamp,

persisted per-user to the CasualOS data record at address `bookmarks` as `{ bookmarks, categories }`.

Users reach it from four places: the chapter bookmark button in the reader header (`ReaderBookmarkButton` in `BibleReader.tsx`), the verse selection toolbar in `BibleReaderToolbar.tsx`, the per-row bookmark button in the sidebar tab list, and the tab kebab menu (both in `Tabs.tsx`). All four open the same folder picker (`openBookmarkCategoryModal` / `BookmarkCategoryPickerContent`, `Tabs.tsx`).

In other words: the categorized, archival system we want for Saves already exists — it's just called "bookmarks". This task is to rename it to what it actually is, and add the one capability it's missing.

## What Saves are

- **Categorized.** Pick one or more existing categories, or create a new one.
- **For reference material** you'll come back to for study, research, or preaching.
- **Archival — the passage a save points at never changes.** Once you save Romans 8:28–30, that save means Romans 8:28–30 forever. If you want to reference something else, you make a new save. This is the core contrast with the redesigned bookmarks (#1658), whose entire job is to be picked up and repointed at wherever you've moved to.
  - To be clear, this is about the *reference*, not the filing. Moving a save between categories, or putting it in several at once, is expected and stays editable at any time — that's what the folder picker's edit mode and the "Move to folder" kebab action already do today.
- **Verse selections _and_ whole chapters.** Saving a whole chapter in one action is the new capability; today you can only save a verse selection or a chapter via the bookmark button.

## Changes

**1. Rename the domain.** `BookmarksManager` → `SavesManager`, `Bookmark` → `Save`, `bookmarkCategory*` → `saveCategory*`, and so on, along with the CSS class prefixes (`sb-bookmark-picker-*` → `sb-save-picker-*`) and the `state.bookmarks` key on `SeedBibleState`. The default category "My Bookmarks" becomes "My Saves".

**2. New save button in the quick actions bar** at the top right of the reader, sitting next to the bookmark button (the `sb-bible-reader-actions` cluster on desktop, and the mobile header cluster). Saves the whole current chapter, and opens the same save modal as a verse selection does — pick one or more existing categories, or create a new one.

Use a `stacks`-style Material icon, **not** a floppy disk — it fits the "collect into categories" idea far better.

**3. The verse toolbar's "Bookmark" action becomes "Save".** Same modal, same behavior; the label, icon, and the underlying manager calls change.

**4. The per-row button in the sidebar tab list stays, as "Save chapter"** — functionally identical to the new quick-actions save button, just reachable from the tab list. Same for the tab kebab menu entry. Note that today this button *toggles* (it removes an existing chapter bookmark on second press); as a save action it should always open the save modal for that chapter instead.

**5. Saves get no indicator on tab rows.** The dot that currently marks any chapter containing a bookmark (`isChapterBookmarked`) is not meaningful for saves, which accumulate — that indicator belongs to bookmarks and is picked up in #1658. For the same reason the per-row save button becomes a plain action button rather than one with a filled/pressed "already saved" state.

**6. Saves come off the Today screen entirely.** Remove `BookmarksSection` from `TodayContent.tsx` and the `showBookmarks` flag from `useTodayContent.tsx`, along with the category-grouping pieces (`BookmarksCategory.tsx`, `useBookmarksCategory.tsx`, the `CategorizedBookmarks` type). Bookmarks take that slot in #1658. Because the section only renders when there is at least one bookmark, nothing renders an empty state in the meantime — the strip is simply absent until #1658 lands.

Worth keeping rather than deleting outright: the per-translation book-name resolution and the overflow measurement in `useBookmarksSection.tsx`. The flat bookmarks strip in #1658 needs both, so it's cheaper to strip the category grouping out of that hook than to delete it and write it again.

**7. New sidebar management panel for saves.** This is largely the existing `BookmarksSection` in `Tabs.tsx` — collapsible category headers with rename and delete, individual rows with a kebab menu for editing which categories a save belongs to and for removing it — carried over as the saves panel, including its mobile full-screen variant (`sb-bookmarks-mobile-screen`) and the bottom-tab entry that opens it.

## Migration

Existing data maps to Saves one-to-one — categories, verse ranges, and all — because Saves _is_ the current bookmark system. What changes is where it lives: the record address moves from `bookmarks` to `saves`, and the payload keys move with it (`{ bookmarks, categories }` → `{ saves, categories }`).

The address rename is worth doing rather than leaving a comment on a misleading name. The risk of *keeping* `bookmarks` as the saves address isn't just that a reader has to spend a beat working out that it means saves — it's that someone building the new bookmarks feature wires it to `bookmarks` and writes over the saves record. Data-layer names also outlive code names by years. We'd rather pay a bounded one-time migration cost than carry that hazard indefinitely.

This is a standard **expand/contract** migration, which means it's a two-issue arc: this issue expands (both records coexist), and #1659 contracts (legacy record goes away). Filing both now so the second half doesn't get forgotten — see the note at the end of this section on why it matters.

### Phase 1 (this issue)

- `SavesManager` reads `saves` first. If that record exists, use it and **ignore the legacy address entirely.**
- Only when there is no `saves` record does it read the legacy `bookmarks` record, run it through the existing `normalizePayload`, and write it forward to `saves`.
- Name the legacy constant so it explains itself at the point of use — e.g. `LEGACY_SAVES_ADDRESS = "bookmarks"` with a one-line comment saying it's the pre-rename address, read-only, migrated forward to `saves`. Someone grepping "bookmarks" inside `SavesManager` should land on an explanation, not a puzzle.

Three invariants make this safe. All three are easy to get wrong:

1. **Guard on "`saves` is absent."** Never merge the two, never overwrite an existing `saves` record. This is what makes the migration idempotent and safe to evaluate on every load.
2. **The migration is a pure copy-forward of the legacy payload — never a merge with in-memory state.** `loadBookmarks` runs from a signal effect on login and on user change, so two tabs can migrate the same user concurrently. That's harmless *only* because both compute identical output from identical input. If the migration folded in anything from memory (a save the user just made in this tab), the two tabs would write different payloads and one would clobber the other.
3. **Write `saves` with the same `publicRead` marker** the bookmarks record uses today (`BookmarksManager.tsx:492`), or the record's visibility changes as an invisible side effect of the migration.

And one hard rule for #1658: **the new bookmarks system must use a distinct address (`readingBookmarks`) and must never read `bookmarks` or `saves`.** If it did, a user's entire save history would surface as bookmarks and blow straight past the cap of 5.

### What we're deliberately not building

**Never delete the legacy `bookmarks` record in this phase.** It costs nothing to keep and it's the difference between a rollback being a non-event and being data loss. Keep the fallback read path for several releases too — someone who hasn't opened the app in six months should still migrate on their next visit.

The one case this design doesn't cover: a user with an already-loaded tab running pre-migration code, who keeps using it after another device has migrated. That tab's writes go to `bookmarks` and the new client never looks there. Two mitigations look attractive and neither survives scrutiny:

- **Dual-writing to both addresses** protects a stale *reader* — an old tab keeps seeing new saves. It does nothing for a stale *writer*, which is the case we care about, because the new client reads `saves` and never consults the legacy record. Cost without the coverage.
- **Merging the legacy record on every load** does address stale writers, but it can't tell "a stale client added this" apart from "the new client already deleted this" — both look like *present in legacy, absent from saves*. So it resurrects deleted saves. Fixing that properly means permanent deletion tombstones, which is a lot of machinery for a transient problem.

The exposure is genuinely narrow: the PWA is configured `registerType: "autoUpdate"` with `skipWaiting()` + `clientsClaim()` (`standalone/sw.ts:263-267`, `vite.config.ts:137`), so there's no stale app shell surviving across sessions — the window is one already-open tab and it heals on the next load. And because we never delete the legacy record, anything written there is **invisible, not destroyed** — still recoverable by hand if it ever comes up. That turns data loss into data latency, which is the whole point of keeping it.

### Instrument it

`managers/Utils.tsx` already has a PostHog `capture` helper. Two events turn #1659 into a data-driven decision instead of a guess:

- **Migration ran** — gives the rate of remaining un-migrated users, so we know when it's safe to contract.
- **Stale writer detected** — phase 1 is reading the legacy record anyway, so it can cheaply notice the one case we chose not to handle: `saves` exists *and* the legacy record contains ids that aren't in it. If that stays at zero for a few weeks, contract with confidence. If it doesn't, the legacy record is still there to recover from.

This is the highest value per line of code in the whole migration.

### Why #1659 isn't optional

Until the legacy record is gone, anyone inspecting a user's data still finds a `bookmarks` blob full of saves — the exact confusion this rename exists to kill. Phase 1 makes the *code* clear and leaves the *data* ambiguous. The contract phase is the payoff, not housekeeping.

### Optional cleanup

The per-item field is named `category` but holds either one name or a list. If we want it to be `categories`, the migration write is the cheapest moment to do it — one normalize step, no extra write, no extra risk.

## Notes

- This is knowingly short-term UI. Once we have a real account-level "Your Content" page covering bookmarks, saves, and highlights in one place, both this sidebar panel and the bookmark one probably go away. We're accepting the duplicate work so saves can ship without waiting on Your Content.
- Only `en.json` gets new/renamed translation keys — the other 23 locale files are left for the translators. Affected keys today: `add-bookmark`, `remove-bookmark`, `bookmark-verses`, `bookmark-options`, `bookmark-folder-empty`, `edit-bookmark`, `add-bookmark-modal`, `move-bookmark`, `remove-bookmark-from-all-folders`, `new-bookmark-folder`, `bookmarks`, `show-bookmarks`, `hide-bookmarks`.
- `test/unit/seed-bible/managers/BookmarksManager.test.ts` (24 tests) covers the existing behavior and should carry over to `SavesManager` largely intact — the category, multi-category, recategorization, and legacy-payload cases are all still relevant. The migration needs four cases of its own:
  - a legacy-only record migrates forward to `saves`;
  - an existing `saves` record wins and the legacy record is ignored, even when legacy is non-empty;
  - a second load after migration neither rewrites nor duplicates anything;
  - the legacy record is never erased.

  The new whole-chapter-save entry point needs coverage too.

Blitzy:
Closes #1657. Part 1 of 3 — #1658 (bookmark redesign) and #1659 (migration cleanup) follow.

### What this is

What we called a "bookmark" was already an archival save: a categorized, permanent reference to a passage. This renames it to what it is and adds the one capability it was missing — saving a whole chapter in one action.

**Before:** the chapter bookmark button toggled a bookmark on and off. Bookmarks appeared on the Today screen grouped by folder, and a tab row showed a filled icon when its chapter was bookmarked.

**After:** a save button files the current chapter into one or more folders. Saves accumulate and never move — the passage a save points at is fixed for its lifetime. Saves are off the Today screen; #1658 puts bookmarks in that slot.

### Changes

**Domain rename.** `BookmarksManager` → `SavesManager`, `Bookmark` → `Save`, `sb-bookmark-*` → `sb-save-*`, `state.bookmarks` → `state.saves`. The default folder "My Bookmarks" becomes "My Saves".

**New save button** in the reader's quick actions, on desktop and mobile. Saves the whole chapter through the same folder picker a verse selection uses.

**The verse toolbar's "Bookmark" action is now "Save."** Same modal, same behaviour.

**The sidebar tab row keeps its per-row button as "Save chapter,"** and no longer toggles — a save is never removed by pressing the button that created it. Removal happens in the saves panel.

**Saves came off the Today screen.** `BookmarksSection` is unwired and stripped of its category grouping, but kept: #1658 needs its per-translation book-name resolution and overflow measurement, and rewriting those would be wasted work. Its prop is now a local `BookmarkStripItem` so it depends on neither manager.

**Icon.** Saves use a dual-state star — outlined when a chapter isn't filed, filled when it is.

### Migration (expand phase)

Saves move from record address `bookmarks` to `saves`. On load both addresses are read in parallel; `saves` wins outright when present, and only when it's absent is the legacy record copied forward.

The three invariants from the issue hold:

1. The migration runs only when `saves` is absent — never a merge, never an overwrite.
2. The copy-forward is a pure function of the legacy payload, so two tabs racing it write identical bytes.
3. The `publicRead` marker is preserved.

The legacy record is never erased, so a rollback is a non-event.

Two PostHog events make #1659 a data decision rather than a guess: `saves_migrated_from_legacy_bookmarks` and `saves_legacy_record_diverged`. The second is an upper bound, not an exact count — a save deleted after migrating also reads as "in legacy, absent from saves" — which is documented at the call site.

### Judgment calls worth a reviewer's eye

**"My Bookmarks" is renamed to "My Saves" during the copy-forward.** Without this, a migrating user opens the panel to find their full "My Bookmarks" folder beside an empty "My Saves" that normalization always adds. Applied only on the legacy path, so it stays idempotent. A user who already had a folder named "My Saves" gets the two merged rather than duplicated — there's a test for it.

**Dead manager API removed.** `isChapterBookmarked` had no UI caller at all, so item 5's "dot indicator" doesn't exist in this tree — item 5 reduced to dropping the pressed state from the tab row button. The three `toggleBookmark*` helpers had no production callers either, and toggling is what item 4 removes.

**A pre-existing load race is fixed.** A save made while the initial load was in flight was overwritten when the load applied. That was survivable before; with the migration it would silently skip the copy-forward. Mutators now await the in-flight load. The regression test was verified red without the fix.

**The star's filled state partly reverses item 5,** which had dropped the "already saved" state from save buttons. Requested during review.

**Pressing a filled star opens the picker in edit mode.** `addSave` ignores a location it already holds, so add mode on a saved chapter let you change folders and hit Save to no effect. All four entry points now route through `openSaveModalForLocation`, and `openSaveCategoryModal` is module-private so a new caller can't regress it.

**The other 23 locale files keep their old keys.** Per CLAUDE.md only `en.json` is touched; translators own the rest. This costs about 13 unused-key lint *warnings* per locale until translations land.

### Also here

- The reader's bookmark button is hidden behind `SHOW_BOOKMARK_BUTTON = false`. Its only behaviour was a "coming soon" toast. The placeholder stays in the tree so #1658 flips one line rather than rebuilding the cluster.
- The mobile tab drawer's "Add new tab" button was clipped by the fixed bottom toolbar. Pre-existing on `develop`, unrelated to Saves, fixed here since it surfaced during testing.
- A dev-only `console.log` of the saves list in `SavesManager`, gated on `DEV` + non-test + browser.

### Testing

35 files changed, +3,010 / −2,682. All 5,741 tests pass; `pnpm check:ts` reports 0 visible errors; `pnpm lint` 0 errors.

`SavesManager.test.ts` carries the original 24 cases forward and adds migration coverage: legacy-only copies forward, an existing `saves` record wins even against a non-empty legacy one, a second load neither rewrites nor duplicates, the legacy record is never erased, divergence reporting fires and stays quiet correctly, a failed copy-forward keeps saves in memory, and the folder-name collision merges. Both behavioural fixes (non-toggling row button, edit mode on a saved chapter) have regression tests confirmed red against the pre-fix code.

Manually verified in the browser.


Blitzy:
> [!NOTE]

## Code review — Repurpose the bookmark system into Saves (#1763)

Reviewed against `develop` (head `2a0043d8`), read alongside #1657 and its three invariants. This is a careful, well-documented PR: the domain rename is thorough, the migration is a genuine pure copy-forward, and the tests are the real thing — they assert observable behaviour, use microtask flushing rather than fixed sleeps, and the two behavioural regression tests genuinely fail on the pre-fix path (add-mode-on-a-saved-chapter would open `save-category-…` instead of `save-edit-…`; the load-race test loses the in-flight save). I ran `SavesManager.test.ts` locally: 29 passing.

One finding is a real data-loss hazard; the rest are minor. Listed by severity.

---

### High

**No. 1 — A failed `saves` read (not just a missing one) re-runs the migration and overwrites the live record with stale legacy data.**

`loadSaves` decides whether to migrate by asking "did the `saves` record parse?" — but `parseRecord` collapses three different outcomes into one `null`: *record genuinely absent*, *read failed transiently* (network blip, `server_error`, an auth/token edge, rate-limit), and *record present but malformed*. Only the first should trigger the copy-forward.

`SavesManager.tsx:273-287` (`parseRecord`) returns `null` for any `result.success === false`, and `os.getData` (`OsManager.tsx:309-316`) passes the client's `{ success:false, errorCode }` straight through — so `errorCode: "server_error"` and `errorCode: "data_not_found"` are indistinguishable to the caller. Then at `SavesManager.tsx:512-542`, a `null` current + a readable legacy record runs the migration and **writes migrated legacy data to the `saves` address.**

Why this is more than theoretical: the two reads are independent RPCs fired together (`SavesManager.tsx:484-487`), so one can fail while the other succeeds — normal distributed-systems behaviour. And because the legacy record is *deliberately never erased* until #1659, **every migrated user carries a readable legacy record on every load for the entire migration window.** So the exposure is: an already-migrated user whose `saves` read hiccups while the `bookmarks` read succeeds silently has their canonical record overwritten with their pre-migration snapshot — any save made *after* migrating (which only ever lived at `saves`) is gone, and `saves_legacy_record_diverged` won't fire because `saves` now equals legacy. It defeats invariant No. 1 from #1657, which is really "guard on `saves` *absent*," not "guard on the `saves` read *not succeeding*."

**Reproduction** (confirmed locally with a throwaway test):

1. Mock `os.getData` so the `saves` address returns `{ success: false, errorCode: "server_error" }` and the `bookmarks` address returns a legacy record.
2. `createSavesManager(os, login)` for a logged-in user; flush microtasks.
3. Observe `recordData` is called once, to `"saves"`, with the migrated legacy payload — the migration ran even though the `saves` record was never confirmed absent. In production that write lands on top of the user's real, still-present `saves` record.

**Suggested fix:** distinguish "genuinely absent" from "couldn't read." Only treat `saves` as eligible for migration when the read *succeeded* and there is no record — e.g. `success === true` with empty data, or `success === false && errorCode === "data_not_found"`. On any other unsuccessful `saves` result, abort `loadSaves` without calling `apply`/migrating (leave in-memory state untouched; it retries on the next load) rather than falling through to the legacy path. Concretely, have `parseRecord` (or `loadSaves`) return a three-way result — `found` / `absent` / `error` — and gate the copy-forward on `absent` alone. A regression test mirroring the repro above (saves read errors, legacy present → `recordData` must **not** be called) would lock this in; the existing "prefers an existing saves record" test only covers the success-with-data case, not the failed-read case.

---

### Low

**No. 2 — Verse-toolbar save button still advertises a toggle state it no longer has.**

`BibleReaderToolbar.tsx:2886` keeps `aria-pressed={isSelectionSaved}`, but this button now always opens the folder picker (`openSaveModalForLocation`, line 2878) — it never toggles. The equivalent reader star and the tab-row save button both correctly dropped `aria-pressed` for exactly this reason, so a screen reader will announce a pressed/unpressed toggle here that doesn't match the other two entry points. The saved state is already conveyed by `aria-label={saveLabel}` and the filled star. Suggest removing `aria-pressed` from this button for consistency. (The `aria-pressed={isSavesFilterActive}` usages in `Tabs.tsx` are correct — those are the genuine show/hide-saves filter toggle.)

**No. 3 — Divergence event can become steady background noise.** `reportLegacyDivergence` (`SavesManager.tsx:463-478`) fires on every load where legacy holds an id absent from `saves`. As the PR notes, a save *deleted after* migrating reads identically to a stale-writer orphan, so any migrated user who ever deletes a save will emit `saves_legacy_record_diverged` on **every subsequent load** until #1659. It's documented as an upper bound and the intent is to confirm it sits at zero — just flagging that for a user who deletes saves it won't sit at zero, so #1659's read of this metric should expect a non-zero floor from deletions rather than treat any signal as a stale writer. No code change needed.

**No. 4 — Dev-only `console.log` of the full saves list.** `SavesManager.tsx:833-841` logs on every change. It's correctly gated (`DEV && MODE !== "test" && window`) and disclosed in the PR, so it won't reach production or spam the test run — noting it only so it's a conscious keep.

**No. 5 — `BookmarksSection` is now orphaned.** Nothing imports `BookmarksSection` / `BookmarkStripItem` anymore (`TodayPane.tsx` and `SeedBibleStateManager.tsx` dropped the wiring). The PR intentionally preserves it for #1658, which is a reasonable call given its book-name resolution and overflow measurement — just confirming the team would rather carry unreferenced (but still type-checked) code across one PR than restore it from history when #1658 lands.

---

### Things I checked that are correct

- **No toggling on the tab-row button** — `handleSaveAction` (`Tabs.tsx:1086-1092`) unconditionally opens the modal; the old remove-on-second-press branch is gone.
- **Edit mode on a saved chapter** — all four filing entry points route through the exported `openSaveModalForLocation` (`Tabs.tsx:1024`), which picks edit vs. add via `getSaveForLocation`; `openSaveCategoryModal` is module-private. The one direct call to it (`Tabs.tsx:1951`) is the saves-panel "Edit save" action, which legitimately already knows the `saveId`.
- **Migration idempotency** — the copy-forward is a pure function of the legacy payload (no `Date.now`, no new ids, no in-memory fold-in), so two tabs racing it write identical bytes; the "My Bookmarks" → "My Saves" rename is applied only on the legacy path and dedupes against an existing "My Saves". `publicRead` is preserved. All well covered by tests, and the load-race fix (`whenLoaded`) is sound.
- **i18n** — `en.json` only; new keys present, removed keys no longer referenced, every `t()` carries a `defaultValue`.

Nice work overall — No. 1 is the one I'd want addressed before merge, since it turns a routine transient read failure into silent, canonical-record data loss for the whole migrated population.

---

Blitzy:
> [!NOTE]

## Pre-merge review — reviewed at head `ca6ef2bc`

A full independent pass, read against [#1657](https://github.com/HelloAOLab/seed-bible/issues/1657) and its three invariants rather than against the two earlier reviews on this PR.

**Short answer: this is safe to merge.** The migration is genuinely correct — the three-way `RecordRead` type is the right fix at the right layer, the copy-forward is a real pure function, and the tests are the honest kind. I found nothing that risks a user's stored data on merge.

I did find two things worth fixing first, neither of which is a data-loss hazard today:

- **No. 1** feeds a wrong number into the one metric [#1659](https://github.com/HelloAOLab/seed-bible/issues/1659) uses to decide when deleting legacy records is safe — so the cost lands later, in the phase that *is* destructive.
- **No. 2** shows the user a save that was never stored.

> [!IMPORTANT]
> **Status: Nos. 1, 2 and 3 have since been fixed on this branch** (`7ee3c51d`, `52a79dac`), at Ryan's request after this review was written. Each fix carries a regression test that was confirmed red against the pre-fix code. Nos. 4 to 8 are still open and are all nits or notes. Details are under each finding below; the analysis is kept as the record of why the changes were made.

Everything below No. 3 is a nit or a note.

---

### Medium

**No. 1 — The migration reports success even when the copy-forward write failed, inflating the metric #1659 depends on.**

`writeSaves` (`SavesManager.tsx:490-505`) awaits `os.recordData(...)` and never looks at what came back. The records client signals API-level failure by *resolving* `{ success: false, errorCode }`, not by throwing — `RecordDataFailure` covers `not_logged_in`, `not_authorized`, `data_too_large`, `record_not_found` and `server_error` (`patterns/pattern-typings/AuxLibraryDefinitions.d.ts:4251-4262`). So the `try`/`catch` around the migration write (`SavesManager.tsx:625-636`) only catches network-level throws, and on any of those resolved failures:

```ts
await writeSaves(userId, normalized.saves, normalized.categories);   // silently did nothing
captureEvent("saves_migrated_from_legacy_bookmarks", { ... });       // fires anyway
```

Why this matters more than a swallowed error normally would: per #1657, this event exists to give "the rate of remaining un-migrated users, so we know when it's safe to contract." Over-counting successes biases that reading in the one direction that hurts — it makes the population look more migrated than it is, and #1659 then erases the `bookmarks` record for users whose saves only ever lived there. `saves_legacy_record_diverged` won't cover for it either: for those users `saves` is absent, so there is nothing to diverge from.

**Reproduction:** take the existing test `keeps the migrated saves in memory when the copy-forward write fails` (`SavesManager.test.ts:638`) and swap the throw for the resolved-failure shape the client actually returns:

```ts
recordDataMock.mockResolvedValue({
  success: false,
  errorCode: "data_too_large",
  errorMessage: "Data too large",
} as never);
```

Its `expect(captureMock).not.toHaveBeenCalledWith("saves_migrated_from_legacy_bookmarks", …)` assertion flips to failing — the event fires on a write that never landed. (The test passes today only because `mockRejectedValue` exercises the one path that *is* handled.)

**Suggested fix** — make `writeSaves` treat a resolved failure like the failure it is, so the existing catch covers both paths:

```ts
const result = await os.recordData(userId, STORAGE_ADDRESS, payload, {
  marker: "publicRead",
});
if (result && result.success === false) {
  throw new Error(`Failed to write saves: ${result.errorCode}`);
}
```

Then add the resolved-failure variant alongside the existing rejection test.

One caveat so this lands proportionately: **not checking `recordData`'s result is a pre-existing, codebase-wide pattern** — the old `BookmarksManager` ignored it (`BookmarksManager.tsx:492`) and `HighlightsManager` still does (`HighlightsManager.tsx:548`). I'm not asking you to fix it everywhere. I'm asking for it at the one call site whose output feeds a destructive decision in the next phase.

> [!NOTE]
> **Fixed in `52a79dac`** with the guard above, kept scoped to this one call site. The `result &&` check turned out to be load-bearing rather than defensive: several existing tests mock `recordData` as resolving `undefined`, so a bare property access would have broken them. Regression test `does not report a migration the records server refused`, confirmed failing on the pre-fix code with `expected "vi.fn()" to not be called with arguments`.

---

**No. 2 — A save made while the record is unreadable is shown as saved but never stored, for the rest of the session.**

The new abort branch (`SavesManager.tsx:595-611`) is right: on a read error it changes nothing and leaves the user unloaded. But `loadedUserId` then stays `null`, so `persist` (`SavesManager.tsx:642-662`) correctly refuses every write from then on — while each mutator has *already* committed its optimistic change to the `saves` signal before calling it. In `addSave` (`SavesManager.tsx:748-756`):

```ts
saves.value = nextSaves;          // star fills, row appears in the panel
...
await persist(nextSaves, nextCategories);   // console.warn, drops the change
```

So the star fills, the save shows up in the saves panel, and it silently vanishes on the next reload. Nothing retries either: the load effect only re-runs when `login.userId` or `loadedUserId` changes, so one blip on *either* read — including the legacy read, which for an already-migrated user is only there for telemetry — degrades the whole session behind a `console.warn`. The saves panel shows an empty list, which is indistinguishable from genuinely having none.

**Reproduction:** the PR's own test already sets this up. In `does not write a save made after a failed load` (`SavesManager.test.ts:417`), add one line after the `addSave`:

```ts
await manager.addSave("BSB", "JHN", 3);

expect(recordDataMock).not.toHaveBeenCalled();   // passes today
expect(manager.saves.value).toEqual([]);         // fails — one phantom save in memory
```

**Suggested fix**, in rough order of value:

1. Have `whenLoaded()` retry the load when nothing has loaded for the current user, so a transient failure heals on the user's next action instead of persisting until refresh.
2. Have the mutators bail *before* mutating when the record still hasn't loaded — check `loadedUserId.value !== login.userId.value` right after `await whenLoaded()` and return. That's what turns "your save didn't save" into a no-op rather than a lie.
3. At minimum, surface it (`state.app.toast`) so the user knows the save didn't stick.

To be fair to the change: this is **strictly better than `develop`**, where a failed read applied an empty list and the next save overwrote the real record. This isn't a regression — it's the last soft edge on an otherwise solid fix.

> [!NOTE]
> **Fixed in `52a79dac`** with options 1 and 2 together. `whenLoaded()` became `ensureLoaded()`, which waits on the in-flight load (keeping the original race fix intact), retries it once when nothing has loaded for the current user, and returns whether the record is actually loaded; all seven mutators now bail before touching state via `if (!(await ensureLoaded())) return;`. The retry reuses a load another caller already started rather than firing a third read.
>
> Two regression tests, both confirmed red pre-fix: the existing `does not write a save made after a failed load` gained the phantom-state assertions (pre-fix: `expected [ { …(6) } ] to deeply equal []`), and `retries the failed load when the user next saves something` covers the recovery path.
>
> Option 3, the toast, was deliberately left out. With the retry in place the common transient case heals silently and correctly, so the remaining gap is narrow: a read that fails twice in a row makes the save a silent no-op. Surfacing that needs either a new constructor dependency on `SavesManager` (`BibleToolsManager` has precedent) or a signal plus wiring in `SeedBibleStateManager` (as `LoginManager.sessionEnded` does), and both widen this PR. Worth a follow-up if you want it.
>
> One behavior difference worth flagging in the diff: `ensureLoaded` also returns `false` when logged out, so mutators now bail before mutating in that case where they previously mutated and let `persist` drop it. The logged-out effect already clears the list, so nothing user-visible changes.

---

### Low

**No. 3 — `CHANGELOG.md` isn't updated.** The repo keeps a `## TBD` section with per-PR entries and there's a `changelog-update` skill for exactly this. This PR renames a whole user-visible feature, adds save-a-whole-chapter, takes saves off the Today screen, and hides the reader's bookmark button — larger than several entries already sitting in v1.7.1. It's also the single easiest thing to lose once merged, and the changelog currently says "Bookmarks" throughout with no note that it became Saves. Suggest: **Changed** for the rename and the Today removal, **Added** for the chapter save button, **Fixed** for the mobile "Add new tab" clipping.

> [!NOTE]
> **Fixed in `7ee3c51d`** — seven bullets across all four subheadings. The three intra-branch correction commits (`41de4ad2`, `f6974255`, `ca6ef2bc`) are deliberately not documented: they repair code introduced earlier on this same branch, so relative to v1.7.1 there is no user-visible before/after to describe. For the same reason the non-toggling save button is filed under Changed rather than Fixed, since against `develop` the old behavior was to *remove* the save.

**No. 4 — A test comment now describes the opposite of the shipped behavior.** `SavesMobile.test.tsx:219`:

> Filled is an indicator, not a pressed toggle — pressing still files another copy, so aria-pressed would mislead.

Pressing a filled star opens *edit* mode; it never files another copy. The test 40 lines above (`edits the existing save when the chapter is already saved`) asserts exactly that. The real reason to drop `aria-pressed` is that the button isn't a toggle at all. Worth correcting so nobody reads it as intent and "restores" the add path later.

**No. 5 — One regression test doesn't exercise the path it claims.** `refiles the existing save rather than dropping the change` (`SavesMobile.test.tsx:184`) calls `state.saves.setSaveCategories(...)` directly instead of going through the picker, so it passes against the pre-fix code too — it re-tests the manager, which `SavesManager.test.ts:960` already covers. The neighbouring `save-edit-${saveId}` assertion is the one actually carrying the fix. Per CLAUDE.md ("a regression test that passes either way isn't testing anything"), either drive it through the modal content or drop the regression framing.

**No. 6 — A malformed `saves` record is now a permanent dead end.** `readRecord` classifies a parse failure as `error`, which is the right call (don't overwrite what you can't read) — but one bad entry fails the whole array, so such a user loads nothing, never migrates, and can't save, indefinitely, behind a console warning. Old code reset them to empty, which was worse in a different way. Not asking for a change here; noting that per-entry tolerance (drop invalid items, keep the rest) is a cheap future improvement, and that this failure mode is now silent-and-sticky rather than silent-and-destructive.

**No. 7 — Two conscious keeps, confirmed fine.** The dev-only `console.log` (`SavesManager.tsx:938-947`) is correctly gated on `DEV` + non-test + `window`. The mobile "Add new tab" fix (`Tabs.css:1475`, `margin: 0px 1rem 4rem`) does work — I traced it, because a bare `4rem` next to a fixed toolbar looked like it would miss the safe-area inset: the drawer already contributes `2rem + env(safe-area-inset-bottom)` (`Tabs.css:1292`), so total clearance is ~6rem plus the inset against a toolbar that occupies ~86px, and home-indicator devices are covered. A one-line comment saying where `4rem` comes from would save the next reader that trace.

**No. 8 — Item 5 of #1657 is now partly reversed by design.** The issue says saves get "a plain action button rather than one with a filled/pressed 'already saved' state"; the shipped star fills. The PR explains this was requested during review and I agree with the outcome — but #1657 still reads the other way, and #1658 picks up the tab-row indicator work. Worth editing the issue or noting it in #1658 so the next PR doesn't "restore" the issue's version.

---

### Verified correct (the things I'd want checked before merging this one)

- **`data_not_found` is the right sentinel, and a brand-new user can't be misread as an error.** This is the load-bearing assumption of the whole migration, so: `GetDataFailure` is `ServerError | "data_not_found" | "not_authorized" | NotSupportedError` (`AuxLibraryDefinitions.d.ts:4294-4300`) — `record_not_found` isn't in the union, so a first-time user with no record at all can't land in the `error` branch. `LoginManager.getUserProfile` uses exactly this code for "new user, no profile record yet" (`LoginManager.tsx:544`), and `AnnotationSyncManager` does the same three-way split (`AnnotationSyncManager.tsx:298-312`). The three-way read is idiomatic here, not novel. Brand-new users reach `apply({ saves: [], … })` and can save immediately.
- **Migration purity holds.** No clock, no new ids, no in-memory fold-in; `legacyHighWaterMark` is a pure reduce over the payload. Two tabs racing it write identical bytes. `publicRead` is preserved, the legacy record is never erased or written (asserted directly at `SavesManager.test.ts:326`).
- **The watermark is threaded correctly** — written on migration, read back on every load, re-emitted on every subsequent write (`keeps the migration marker on later writes`). The divergence logic correctly separates a post-migration deletion from a stale pre-rename writer, and both directions are tested.
- **Nothing on `develop` collides with the rename.** The branch is 10 commits behind; those commits touch `BibleSelector`, `SettingsPage`, `CustomizationVariantSelectionsManager` and `SettingsManager`, and add **zero** new `bookmarks` references — so renaming `state.bookmarks` won't break at merge. Worth updating the branch anyway so CI runs once against the merged state.
- **The CSS rename is complete.** I cross-checked every `sb-*save*` / `sb-*bookmark*` class used in TSX against every CSS file. Five have no rule (`sb-save-category-header-expanded`, `sb-saves-mobile-header-add`, `sb-saves-mobile-header-close`, `sb-sidebar-tabs-header-saves-button`, `sb-verse-toolbar-save-trigger`) — all five were equally styleless under their old names on `develop`; they're modifiers sitting beside styled base classes. Every `-active` variant that *was* styled carried over, and dropping the `.material-symbols-outlined` sizing rule is correct now that the trigger renders an SVG star.
- **i18n is clean.** None of the 13 removed keys is still referenced anywhere, no new key is unused, every `t()` carries a `defaultValue`, and only `en.json` was touched. The four keys used-but-absent in the touched files (`discovery`, `discovery-coming-soon`, `social`, `failed-to-load-playlist`) are all pre-existing on `develop`. The PR also incidentally removed a duplicated `edit-bookmark` / `remove-bookmark-from-all-folders` pair — `en.json` carries ~41 pre-existing duplicate keys, so that's a small net improvement, not a new problem.
- **No extension API break.** Nothing in `packages/*-extension` or any live package reads `state.bookmarks`; the only hits are in `obsolete/`, on unrelated CasualOS bot tags.
- **`aria-pressed` is gone from all three save entry points**, and the two genuine `aria-pressed={isSavesFilterActive}` toggles in `Tabs.tsx` correctly keep it. The row button's visibility (`isSelected && !tab.sharedSession`) is byte-identical to `develop`.
- **The CI claims are verified, not just asserted.** `build (24.x)` runs `pnpm lint`, `pnpm check:ts` and `pnpm test:coverage` (`.github/workflows/ci.yml`) and is green on `ca6ef2bc`.
- **Test quality is real.** No `.only`, `.skip` or `todo`; no fixed-duration sleeps (microtask flush throughout, plus a properly gated released-promise for the in-flight race test); mocks stop at the `OsManager` boundary; the `getData` mock answers `data_not_found` for unlisted addresses the way the server does. The one assertion-free test I found (`opens the book selector on an ordinary tap`) is byte-identical to `develop` and not this PR's doing.

Overall: the migration is the part that would have been expensive to get wrong, and it's right. Nos. 1 and 2 were each a few lines and are now fixed on the branch, along with the changelog; what's left (Nos. 4 to 8) is a stale test comment, a redundant test, and notes. Verified after the fixes: 5752 tests pass across 240 files, `pnpm check:ts` reports 0 visible errors, `pnpm lint` 0 errors, Prettier clean.


---

Blitzy:
@KallynGowdy This is ready for your review.

I'm pretty happy with where this is at, I just want to have your eyes on it before merging in.

 Here is a video overview: https://www.youtube.com/watch?v=xc2HcuzF1io

KallynGowdy:
@Blitzy Could we fix the merge conflicts and update with develop? Also does it rename the "Bookmarks" section on the user's profile to "Saves"?

Blitzy:
Profile section needs to be updated to say Saves now that the profile screen is merged in. That should be updated as well with this PR before merge.

PR #1764, opened 2026-09-04
Strip existing query params from customization share links

KallynGowdy:
getShareLink() built on linkToQuery(), which starts from the current
page URL and only adds/overwrites the given params — so the share link
carried whatever the current page happened to have (language,
translation, book, chapter, etc.) in addition to the customization
param. Add linkToBareQuery(), which clears the URL's search params
before applying the given ones, and use it for the share link so it
only ever contains ?customization=....

Claude-Session: https://claude.ai/code/session_01UUCbSXQxUXYKanfuTE2dSH

### DIFF
--- a/packages/seed-bible/seed-bible/managers/CustomizationsManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/CustomizationsManager.tsx
@@ -1503,7 +1503,7 @@ export function createCustomizationsManager(
 
   const getShareLink = (customization: SeedBibleCustomization): string => {
     const recordName = login.userId.value ?? "";
-    return navigation.linkToQuery({
+    return navigation.linkToBareQuery({
       customization: buildCustomizationLocator(recordName, customization.id),
     });
   };
--- a/packages/seed-bible/seed-bible/managers/CustomizationsManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/CustomizationsManager.tsx
@@ -1503,7 +1503,7 @@ export function createCustomizationsManager(
 
   const getShareLink = (customization: SeedBibleCustomization): string => {
     const recordName = login.userId.value ?? "";
-    return navigation.linkToBareQuery({
+    return navigation.linkToBareRoot({
       customization: buildCustomizationLocator(recordName, customization.id),
     });
   };
--- a/packages/seed-bible/seed-bible/managers/NavigationManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/NavigationManager.tsx
@@ -438,6 +438,23 @@ export function createNavigationManager(
     return url.toString();
   };
 
+  /**
+   * Like `linkToQuery`, but drops every existing query parameter first —
+   * for links that should carry only the parameters passed in (e.g. a
+   * customization share link), not whatever the current page happens to
+   * have in its URL (language, translation, book, chapter, ...).
+   */
+  const linkToBareQuery = (query: Record<string, string | null>) => {
+    const url = new URL(currentUrl.value);
+    url.search = "";
+    for (const [key, value] of Object.entries(query)) {
+      if (value !== null) {
+        url.searchParams.set(key, value);
+      }
+    }
+    return url.toString();
+  };
+
   return {
     currentUrl: computed(() => currentUrl.value),
     initialUrl,
@@ -451,6 +468,7 @@ export function createNavigationManager(
     updatePathAndQueryParams,
     syncSignalsToUrl,
     linkToQuery,
+    linkToBareQuery,
     dispose,
   };
 }
--- a/packages/seed-bible/seed-bible/managers/NavigationManager.tsx
+++ b/packages/seed-bible/seed-bible/managers/NavigationManager.tsx
@@ -439,13 +439,18 @@ export function createNavigationManager(
   };
 
   /**
-   * Like `linkToQuery`, but drops every existing query parameter first —
-   * for links that should carry only the parameters passed in (e.g. a
+   * Like `linkToQuery`, but drops the current URL entirely first, keeping
+   * only the origin and the deployment root (`basePath`, or "/") — for
+   * links that should carry only the parameters passed in (e.g. a
    * customization share link), not whatever the current page happens to
-   * have in its URL (language, translation, book, chapter, ...).
+   * have. The reading position (language, translation, book, chapter) lives
+   * in the URL *path* now, not the query string — see `ReadingUrlPath.ts` —
+   * so a link that must not leak the sharer's current reading position has
+   * to drop the path too, not just the query.
    */
-  const linkToBareQuery = (query: Record<string, string | null>) => {
+  const linkToBareRoot = (query: Record<string, string | null>) => {
     const url = new URL(currentUrl.value);
+    url.pathname = basePath || "/";
     url.search = "";
     for (const [key, value] of Object.entries(query)) {
       if (value !== null) {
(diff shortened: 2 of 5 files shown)


Blitzy:
> [!NOTE]

## Summary

The PR adds `linkToBareQuery()` (clears all query params, then sets the given ones) and points `getShareLink()` at it, so a customization share link no longer inherits query params like `?language=`/`?translation=`/`?book=`/`?chapter=` from the current page. The change is small, well-commented, and the added tests pass (`117 passed` across the two touched test files).

The mechanism is correct for what it does. My concern is that **it only solves half the stated problem**, and the regression test is shaped in a way that hides the other half.

---

### High-level finding: the share link still carries the sharer's reading position — via the URL *path*, not the query

The PR's stated goal (from the description) is that the share link should carry *only* `?customization=...` and drop "whatever the current page happened to have (language, translation, book, chapter, etc.)".

But those values don't live in the query string anymore — they were **moved into the URL path**. See `managers/ReadingUrlPath.ts`:

- `buildReadingPath()` produces `/{lang}/{translationId}/{bookSlug}/{chapter}` (line 155).
- The comment block above `buildReadingUrl()` (lines 158–179) is explicit: `book`, `chapter`, `translation`, `translationId`, `lang` are **legacy** query params that "no longer win" because "the position moved into the path", and it deletes them precisely because the path is authoritative.

`linkToBareQuery()` clears `url.search` but leaves `url.pathname` untouched:

```ts
const linkToBareQuery = (query: Record<string, string | null>) => {
  const url = new URL(currentUrl.value);
  url.search = "";            // clears query only
  // ... pathname (e.g. /en/BSB/genesis/1) is preserved
};
```

So in the normal case — a user who is *reading* when they open the share button — the current URL is something like `http://host/en/BSB/genesis/1`, and the produced share link is:

```
http://host/en/BSB/genesis/1?customization=user-1.abc
```

The recipient still lands on the sharer's exact book, chapter, translation, and language. That is the very thing the PR set out to avoid; it's just encoded in the path instead of the query.

**Severity:** Medium. The customization itself loads correctly (the `?customization=` param is set), so nothing is broken — but the PR's actual motivation (a "bare" link that doesn't leak the sharer's current position) is only met when the sharer happens to be at the site root. Whenever they're reading, it isn't.

1. Navigate to a reading position so the URL path is `/{lang}/{translation}/{book}/{chapter}` (e.g. `/en/BSB/genesis/1`).
2. Open a customization and click Share (`CustomizationEditPane` or `SettingsPage`).
3. Inspect the copied link — the `/en/BSB/genesis/1` path is still there.

I confirmed the mechanism directly:
```
path-based  : http://localhost/en/BSB/genesis/1?customization=user-1.abc   // path survives
legacy/root : http://localhost/?customization=user-1.abc                    // only bare at root
```

**Suggested fix:** reset the path as well as the query for the share link. Because `linkToBareQuery` is named for *query*, the cleanest options are either to clear the path inside `getShareLink`, or to have the helper also drop the path. For example, in `getShareLink`:

```ts
const url = new URL(navigation.currentUrl.value);
url.search = "";
url.pathname = navigation.basePath || "/";   // keep the deployment prefix, drop the reading position
url.searchParams.set("customization", buildCustomizationLocator(recordName, customization.id));
return url.toString();
```

(Keeping `basePath` matters so a branch deployment like `/b/some-branch` still resolves.) If you'd rather keep it in the navigation layer, consider a small option flag or a dedicated helper so the intent — "origin + basePath + only these params" — is explicit.

---

### Test-quality finding: the regression test uses a URL shape the app no longer produces, which masks the gap above

`test/unit/seed-bible/managers/CustomizationsManager.test.ts` (the new case at line 1330):

```ts
const nav = createNavigationManager({
  initialHref:
    "http://localhost/?language=en&translation=BSB&book=GEN&chapter=1&foo=bar",
});
// ...
expect(link).toBe(`http://localhost/?customization=user-1.${created.id}`);
```

This does fail on the pre-fix code (old `getShareLink` used `linkToQuery`, which would keep those params), so it is a genuine regression test for the query-stripping mechanism — that part is good.

The problem is the *scenario*. It encodes the reading position as **query params on the root path**, which is the legacy representation the codebase itself has migrated away from (`ReadingUrlPath.ts` builds the position into the path and strips those very query keys). No real reading URL in the current app looks like this. Because the test starts at the root path (`/`), the path-leak in finding No. 1 never comes into play, so the test passes while the real-world behavior it's meant to guarantee ("keeping only customization") does not hold.

A test that reflects how the app actually builds URLs would set a path-based `initialHref` — and, importantly, that test **fails against this PR's code today**, which is exactly why it's worth adding:

```ts
it("getShareLink() drops the sharer's reading position (path) too, keeping only customization", async () => {
  const nav = createNavigationManager({
    initialHref: "http://localhost/en/BSB/genesis/1?foo=bar",
  });
  const { manager } = createManager(nav);
  const created = await manager.create();

  const link = manager.getShareLink(created);

  // Currently produces http://localhost/en/BSB/genesis/1?customization=...
  expect(link).toBe(`http://localhost/?customization=user-1.${created.id}`);
});
```

I'd suggest adding this (it will go green once finding No. 1 is fixed) and either keeping or replacing the legacy-shaped case. The `NavigationManager` unit tests for `linkToBareQuery` are fine as-is — they legitimately test the helper's documented "clears query params" contract; the issue is specifically that `getShareLink`'s contract should also cover the path.

---

### Minor / consideration: `linkToBareQuery` drops *every* query param

Clearing all query params also removes any that a recipient might genuinely need to render the shared page (e.g. `useFreeBibleAPI`, read in `FreeUseBibleAPI.tsx:1269`). For an end-user customization share link this is almost certainly the desired behavior, so I'm flagging it only as something to be deliberate about rather than a defect — if there's ever a deployment-level query flag that must survive sharing, this helper would strip it.

**Severity:** Low / informational.

---

### What looks good

- The helper is small, pure, and doesn't mutate `currentUrl` (nicely covered by the "does not mutate currentUrl" test).
- Null-value handling is consistent with the "bare" intent (a null key simply isn't set).
- The `NavigationManager` mock additions across the `I18nManager` tests keep the `NavigationManager` type satisfied — good housekeeping.

Overall: the direction is right and the code is clean; the main ask is to also clear the path so the link is truly bare, and to add a path-based test that would have caught the gap.

---

KallynGowdy:
Addressed in d71c03e9:

- **Main finding (path leak):** `linkToBareQuery()` is replaced with `linkToBareRoot()`, which resets `url.pathname` to the deployment root (`basePath`, or `"/"`) in addition to clearing the query string. `getShareLink()` now calls it, so a share link generated while reading (e.g. from `/en/BSB/genesis/1`) no longer carries the reading position in the path either — it comes out as `origin[+basePath]?customization=...`.
- **Test-quality finding:** added the suggested path-based regression test (`getShareLink() drops the sharer's reading position from the URL path too...`), which fails on the pre-fix code, plus one confirming a deployment's `basePath` prefix is preserved. Kept the existing legacy-query-param case, retitled to make clear it's covering the legacy shape.
- Added direct unit tests for `linkToBareRoot()` on `NavigationManager` (root reset, `basePath` handling, multiple params, null-omission, no-mutation), replacing the old `linkToBareQuery()` tests since that helper had no remaining callers once `getShareLink()` moved to `linkToBareRoot()`.
- Minor/informational note (stripping *all* query params) — left as-is per the review's own assessment that this is the correct behavior for a customization share link.

Full suite (5,746 tests), `check:ts`, and `lint` all pass.

---

PR #1771, opened 2026-09-08
Feature/1749

### LINKED ISSUE #1749: Use an "X" icon instead of trash can icon for hero images for playlists and reading plans
Currently, when you are editing a reading plan or playlist, the user is shown a trash can icon on the hero image:

Image

This implies that the image could be completely deleted simply by clicking that icon. Instead, it just removes the link to the image from the reading plan/playlist.

Instead of a trash can icon, we should use an "X" icon so that it implies removing/clearing the image instead of deleting it.

Kabi4:
Fixes: #1749


Blitzy:
> [!NOTE]

## Code Review — PR #1771 (Fixes #1749)

### Summary

The linked issue asks for one thing: the hero-image control on playlists and reading plans shows a **trash-can icon**, which wrongly implies the image will be permanently deleted, when in fact it only clears the link. The PR swaps that for an **"X" (close) icon** and relabels it.

The change does exactly that, and does it cleanly:

- Icon changed from `delete` → `close` (the Material Symbols "X").
- `aria-label` changed from "Delete" → "Remove cover image" (via the existing `remove-hero-image` key), which matches the new intent for screen-reader users too.
- CSS class renamed `sb-hero-field-delete` → `sb-hero-field-clear` consistently — I grepped the whole repo and found **no stale references** to the old class name.
- Because `HeroImageField` is a single shared component used by both `CreatePlaylistForm` and `ReadingPlanEditor`, this one change correctly covers **both** the playlist and reading-plan cases the issue names.
- The `remove-hero-image` translation key already exists in `en.json` (line 326, added with the gallery module), so nothing is missing there.

**Verdict: Approve.** No blocking issues found. The notes below are all optional/minor.

---

### Critical / High

None.

### Medium

None.

### Low / Optional

**No. 1 — Button stays in the top-left corner (`left: 8px`).**
The clear button is positioned top-left (unchanged from before). A "close/X" affordance is conventionally placed top-**right**, so an X in the top-left reads a little more unusual than a trash can did. This is pre-existing and outside the strict scope of #1749, so it's fine to leave — just flagging it in case the team wants the icon and its position to match the usual convention. If you do want to move it, only `left: 8px` → `right: 8px` in `.sb-hero-field-clear` (HeroImage.css:109) is needed.

**No. 2 — The same trash-icon-for-remove pattern exists elsewhere (out of scope).**
`ReadingPlanEditor.tsx:518` uses `<MaterialIcon>delete</MaterialIcon>` with `aria-label="Remove session"`, and `CreatePlaylistForm.tsx:391` uses a `delete` icon for removing an item. These aren't hero images, so they're correctly out of scope here — but if the goal is broader "remove vs. delete" consistency, they might be worth a follow-up issue.

---

### Tests

The tests are updated to assert **real, user-visible behavior**, not implementation details, and they would genuinely fail on the pre-fix code:

- `HeroImageField.test.tsx` and `CreatePlaylistForm.test.tsx` now assert `aria-label === "Remove cover image"` and `textContent` **contains `"close"`**. Because `MaterialIcon` renders its child as the span's text (`<span class="material-symbols-outlined">close</span>`, icons.tsx:1353), that text-content check is a real guard on the icon: on the old code it was `delete`, so the assertion flips red. The `aria-label` check likewise flips red (old value was "Delete"). So this is a legitimate regression test, not one that always passes.
- The negative assertion — no button whose `textContent === "Remove cover image"` — correctly confirms the label is screen-reader-only and not duplicated as visible text.
- `CreatePlaylistForm.test.tsx` still verifies the click clears the value (`updateEditingPlaylistMetadata({ heroImageUrl: null })` and `editingPlaylist.value?.heroImageUrl` is null), so the behavior behind the button is covered, not just its appearance.

One transparency note: I reviewed the tests statically (the environment had no installed dependencies, so I did not execute the Vitest suite). Based on the mock (`mockTranslate` returns each call site's `defaultValue`) and the `MaterialIcon` implementation, the assertions line up with the component output. CI's `build (24.x)` job — which runs the suite — was still in progress at review time; worth a glance once it finishes.

### Reproduction of the original bug (for reference)

1. On `develop`, open the playlist or reading-plan editor and set a cover image.
2. Observe a **trash-can** icon over the image; a screen reader announces it as "Delete".
3. This PR's build shows an **X** icon announced as "Remove cover image", clarifying that the action clears the link rather than deleting the underlying image.

https://claude.ai/code/session_01StQACq4Khw4m2GwQofdkzg

---

Blitzy (approved):

PR #1772, opened 2026-09-08
Feature/1730

### LINKED ISSUE #1730: Referencing chapters in annotations causes the verse number that the annotation is for to be applied to the chapter link
Currently, when you reference another chapter in an annotation without a verse number, it will incorrectly add the verse number of the selected verse to the chapter link.

### Steps to reproduce

1. Select a verse to annotate it (like John 3:9)
2. Click "Note" in the verse toolbar
3. Edit the note and add a reference to another chapter, like "Jonah 2"
4. Save the note
5. View the note
6. See that the link includes `?verse=9` when it shouldn't have a `verse` query parameter at all.

### What is expected

When the user references another chapter without a verse, then they are referencing the entire chapter. The verse that the annotation was for has nothing to do with this.

Kabi4:
Fixes: #1730


Blitzy:
> [!NOTE]

## Summary

This PR fixes #1730: a chapter-only scripture reference in an annotation (e.g. typing "Jonah 2" in a note attached to John 3:9) was inheriting the page's `?verse=9`, producing a link to `.../jonah/2?verse=9` instead of `.../jonah/2`.

**Verdict: looks good, no blocking issues found.** The fix is correct, minimal, and well-scoped, and the tests are genuine regression tests.

### Why the fix is in the right place

The root cause is that `buildReadingUrl` deliberately preserves unrelated query params — including `?verse=` — so that the Share buttons can keep the reader's selected verses on a shared link. Only the caller knows whether a given reference actually *has* a verse, so `getVerseReferenceLinkHref` is the correct layer to strip it. The added `else` branch (`next.searchParams.delete("verse")`) handles both code paths: the canonical reading-path branch and the `legacyVerseReferenceUrl` fallback (which copies the whole current URL, verse and all).

I confirmed there is no parallel bug elsewhere: the only other verse-link producer, `openShareModal`'s builder in `BibleToolsManager.tsx`, clears `current.search = ""` before calling `buildReadingUrl`, so it never carries a stale verse.

### Tests

I verified the tests exercise real behavior and are not trivially green. Reverting just the `else` branch locally turns **12 of the new assertions red** (8 in `verseReferenceLink.test.ts`, 4 in `VerseReferenceMark.test.ts`); restoring the fix makes all 22 pass. The `VerseReferenceMark` cases are especially good — they drive the actual TipTap editor end-to-end (type "Jonah 2", edit "John 3:16" down to "John 3") and assert on the rendered anchor `href`, matching the exact reproduction steps in the issue. Prettier is clean on all three files.

## Issues by severity

None blocking. Two optional observations:

- **Low / observation — chapter-range references link only to the start chapter.** A reference like "Jonah 2-3" (`chapter: 2, endChapter: 3`) links to `.../jonah/2`; `endChapter` is not represented in the URL. This is pre-existing behavior (the URL scheme has no way to express a chapter span), not introduced here, and the new "does not copy the page's verse onto a chapter range" test correctly documents it. Flagging only so it's a conscious decision rather than an oversight — no change requested.

- **Trivial — `if (ref.verse)` treats verse 0 as "no verse".** A hypothetical `verse: 0` reference would fall into the delete branch. Scanned references start at verse 1, so this is not reachable in practice; not worth a code change.

## Manual verification steps (for a human)

1. Open a chapter and select a verse to annotate, e.g. John 3:9 → the URL becomes `.../en/AAB/john/3?verse=9`.
2. Click **Note**, type a chapter-only reference such as `Jonah 2`, and save.
3. Confirm the rendered link points at `.../jonah/2` with **no** `verse` query param.
4. In the same note type `John 3:16` and confirm that link *does* keep `?verse=16` — the fix must not strip verses from references that actually name one.

---

Blitzy:
Verified that this fix is working for the reported verse issue.

This does bring up something else though, currently it will leave every other query parameter untouched (pre-exisiting behavior). Even the new tests utilize `sessionId` as an example of one that gets preserved. That means if I'm in a shared session and save a note with a verse range, its going to leave that shared session hardcoded in the verse link later on, that seems not ideal. 

I feel for at least annotations we should strip `sessionId` as well? I'm not sure if there are others that we should strip or if we should take the approach of a whitelist of query parameters to preserve and otherwise strip the rest? What do you think @KallynGowdy?

Blitzy (dismissed):

KallynGowdy:
Yes I would agree with that. It should probably just strip all query parameters except the `verse` query param

Blitzy:
@Kabi4 can you look at implementing it the way Kal described? We really just want to avoid this bug quietly returning in the future with another unrelated query parameter getting baked into the verse references.

Blitzy (approved):

PR #1773, opened 2026-09-08
Feature/1770

### LINKED ISSUE #1770: Support dot-notation for Bible references
Correctly, we only support colon-notation for Bible references, like "Gen 1:1". Some people also use dots, like this: "Gen 1.1" or even this: "Gen.1.1".

We should support all three ways. In particular, we should make sure this is supported in these areas:
- The playlist editor (scripture items)
- the reading plan editor
- Annotations
- Shared session chat
- AI chat

Kabi4:
Fixes: #1770


Blitzy:
> [!NOTE]

## Code review — dot-notation for Bible references (#1770)

**Short version:** This is a clean, well-structured change. It pulls the reference-parsing syntax into one place (`verseReferenceSyntax.ts`), and it genuinely closes the gap the issue was about — the playlist / reading-plan path (`computeSuggestions` and `parseVerseReferenceCandidates`) previously accepted only `Gen 1:1`, and now accepts `Gen 1.1`, `Gen.1.1`, etc. All five areas named in the issue are covered (playlist editor, reading-plan editor, annotations, shared chat, AI chat). The new tests are real regression tests — I reverted the fix locally and confirmed they go red on the old code, so they're not the "always green" kind.

There is one real problem, though: to centralize the syntax, the same pattern is now shared between two things that should **not** behave identically — parsing a string that is *entirely* a reference (typed input), and *finding references embedded in prose* (chat/editor/annotations). Broadening the shared pattern quietly broadened the prose scanner too, and that introduces false positives.

---

### Medium — colon is now a valid book→chapter *joiner*, creating false verse links in prose

`BOOK_CHAPTER_JOIN_PATTERN` (`verseReferenceSyntax.ts:16`) is `[\s.:]+`, so a colon can now sit between the book name and the chapter number. That is fine for **typed** input, but this same constant is spliced into the **prose scanner** `scanVerseReferencesInText` (`BibleDataManager.tsx:414`) and the annotation parser `parseVerseReference` (`BibleDataManager.tsx:328`). Those run over free text in AI chat, shared-session chat, the TipTap editor, and annotation anchors.

Before this PR the prose joiner was `[\s.]+` (space or period, no colon), so text like `Mark: 3 things` did **not** match. Now it does — `Mark: 3` gets linkified as Mark chapter 3. This hits ordinary, header-style prose that people actually write in a Bible-study chat: `Mark: 3 things stood out`, `James: 5 steps`, `Luke: 2 reasons`.

Note the issue only asks for `Gen 1:1`, `Gen 1.1`, and `Gen.1.1` — none of those use a colon as the book→chapter joiner, so this broadening isn't required by the feature.

**Reproduction (no colon needed in the reference itself):**
```ts
import { scanVerseReferencesInText, parseVerseReference } from ".../managers/BibleDataManager";

scanVerseReferencesInText("Mark: 3 things to remember");
// => [{ ref: { book: "MRK", chapter: 3 }, start: 0, end: 7 }]   ❌ (was [] before this PR)

scanVerseReferencesInText("James: 5 steps here");
// => [{ ref: { book: "JAS", chapter: 5 }, ... }]                ❌

parseVerseReference("Mark: 3 things");
// => { book: "MRK", chapter: 3 }                                ❌ (annotation anchor mis-resolves)
```
In the deployed branch (https://alpha.seedbible.org/b/feature_1770) the same text typed into chat will render `Mark: 3` as a clickable verse link.

**Suggested fix:** keep the colon joiner for the typed/whole-string parser, but give the prose scanner (and, arguably, the annotation parser) a joiner without the colon. For example, a second constant `PROSE_BOOK_CHAPTER_JOIN_PATTERN = "[\\s.]+"` used at `BibleDataManager.tsx:328` and `:414`, leaving `BOOK_CHAPTER_JOIN_PATTERN` for `TYPED_REFERENCE`. (The colon is still available as the chapter–verse separator via `REFERENCE_NUMBERS_PATTERN`, so `Mark 3:16` keeps working everywhere.)

---

### Low — whitespace around the range dash also widens the prose scanner

`REFERENCE_NUMBERS_PATTERN` (`verseReferenceSyntax.ts:23`) uses `\s*[-–—]\s*` for the range. That's the right behavior for typed input (`parseVerseReferenceCandidates` already allowed surrounding spaces), but the old **prose** range had no surrounding `\s*`, so a spaced dash was not treated as a range. Now it is:

```ts
scanVerseReferencesInText("Read Luke 1 - 2 reasons why");
// => [{ ref: { book: "LUK", chapter: 1, endChapter: 2 }, ... }]  ❌ (was Luke 1 only, no range)
```
So `Luke 1 - 2 reasons` becomes a Luke 1–2 chapter-range link. Same root cause as above (one pattern serving both the strict typed parser and the loose prose scanner). If you split the join pattern as suggested, consider whether the prose range should also drop the surrounding `\s*`.

---

### Low — test gap: the new syntax has no *negative* prose tests

The added tests only assert the new forms are accepted. There are no tests asserting the scanner *doesn't* over-match prose, which is exactly where the two findings above slipped through (the full suite stays green with them present). Whatever you decide the intended behavior is, please add cases to `BibleDataManager.test.ts` `scanVerseReferencesInText()` pinning it down — e.g. `"Mark: 3 things"` and `"Luke 1 - 2 reasons"` should resolve to whatever you intend (my recommendation: no match / chapter-only), so a future refactor can't silently re-broaden them.

---

### Nits (non-blocking)

- The two prose regexes in `BibleDataManager.tsx` (lines 328 and 414) still duplicate the leading book-name sub-pattern verbatim. Since you're already centralizing, that book portion could move into `verseReferenceSyntax.ts` too — though I'd hold off until the join-pattern split above is settled, to avoid over-abstracting.
- Commit message typo: "Aded Chapter Reference…" → "Added".

---

Nice cleanup overall — the `buildTail` / `splitTypedVerseReference` split reads well and the `null` handling for the ambiguous `John 1-2:3` case is correct and tested. The only real ask is to stop the prose scanner from inheriting the looser typed-input syntax.

---

Blitzy:
## Follow-up after review: four commits added

**Short version:** the review above is stale on two of its three findings — `7d4ad8c0` fixed them about four minutes after it posted. Its remaining finding is now addressed, along with one regression it missed. Verified by running both `develop`'s and this branch's parsers over the same inputs rather than by reading the regexes.

### What the branch actually adds over `develop`

`develop`'s prose scanner already used `[\s.]+` between book and chapter and `[:.]` between chapter and verse, so **chat, annotations and the TipTap editor already understood dot notation**. `Gen 1.1`, `Gen.1.1`, `Gen.1:1` and `Gen.1` produce identical output on both branches.

The real gap this PR closes is the playlist / reading-plan entry field (`computeSuggestions` → `ScriptureItemInput`), which accepted only `Gen 1:1`:

| Typed into a playlist / reading-plan entry | develop | this branch |
|---|---|---|
| `Gen 1.1`, `Gen.1.1`, `Gen.1:1`, `Gen.1` | no suggestions | ✅ |
| `Gen. 1:1`, `Phil. 2.3`, `Song of Solomon 1.1` | no suggestions | ✅ |
| `Gen 1.1-3`, `Jude.2` | no suggestions | ✅ |

Worth knowing: `parseVerseReferenceCandidates` / `parseSingleVerseReference` have **no production caller anywhere in the repo** — only their own definitions, doc comments and tests. So `computeSuggestions` is the whole user-visible payoff.

### Review findings

**Colon in prose (Medium) — already fixed before the review landed.** `7d4ad8c0` split the joiner into a typed and a prose variant and added the negative tests. `Mark: 3 things` → no match, identical on `develop` and here.

**No negative prose tests (Low) — not accurate.** `develop` already had 25 negative assertions in `BibleDataManager.test.ts` (out-of-bounds refs, `the Song of Moses 2:1`, the Spanish `Esdras 3` cases), and `7d4ad8c0` added the three cases the review asked for plus the annotation path. 25 → 30.

**Spaced dash in prose (Low) — real, now fixed** in `dc6c33d4`. Prose requires the tight form; `Luke 1-2` is a range, `Read Luke 1 - 2 reasons why` links only Luke 1. This also removed a backwards-range false positive (`Mark 4 - 3 things stood out` → Mark 4–3). Typed entries still accept `Gen 1 - 3`, matching what `develop`'s playlist parser did.

### One regression the review missed

`9d213a7a`. Typing a bare `1`, `2` or `3` into a playlist / reading-plan entry returned **no suggestions**, where `develop` listed 1 John, 1 Corinthians, 2 John, 3 John. `splitTypedVerseReference` rejected any book portion without letters — right for `1.1` (which would read as book `1`, chapter 1), wrong for a bare `1`, which is a numbered book mid-typing. The check now applies only once a chapter has been split off. It was silent: the full suite passed with it present, and no test covered a bare numeric book prefix.

### Colon removed from the typed path too

`6c60ba01`, per @Blitzy's call — `Gen:1` isn't in #1770's spec, which asks only for `Gen 1:1`, `Gen 1.1` and `Gen.1.1`. Both joiner constants are now identical, so `PROSE_BOOK_CHAPTER_JOIN_PATTERN` collapses back into `BOOK_CHAPTER_JOIN_PATTERN` and all five surfaces share one syntax again. The colon remains the chapter–verse separator, so `Gen 1:1` and `Mark 3:16` are untouched.

Removing it from the pattern wasn't sufficient on its own: with `Gen: 1` the joiner matches the *space*, leaving `Gen:` as the book query, which `getBookId` resolves anyway because it strips trailing punctuation so the `Gen.` abbreviation works. Since no book name contains a colon, one left in the book portion is now rejected outright.

### Verification

- Every new negative test was checked red-then-green: source reverted → fails, restored → passes.
- Full suite **5590 passing** (was 5583, +7 tests). `pnpm check:ts` 0 errors, `pnpm lint` 0 errors.
- Re-ran the full behavioral battery across all four parsers against `develop`. Prose scanning is now byte-identical to `develop` except for `John 1-2:3`, which `develop` linked as John 1–2 with the `:3` silently dropped and which now links nothing.

### Changelog

Added, and deliberately scoped to what differs from `develop`. The spaced-dash and colon commits reverted this branch's own intermediate broadening, so they announce nothing to users — `develop` already behaved that way on every surface a user can reach. That leaves one Added line for the dot-notation feature and one Fixed line for `John 1-2:3`.

### Left alone — pre-existing on `develop`, out of scope here

- `Mark. 4` and `Mark .4` match in prose on both branches. `Mark. 4` is worth keeping (`Matt. 5:3` is standard style); `Mark .4` matches only because the joiner is a character class. The genuinely bad case, `I read Mark. 4 people were there.`, can't be fixed by the joiner — it needs to know "Mark" isn't an abbreviation.
- Reversed ranges: `Mark 4-3` parses as chapter 4 through chapter 3 on both branches. Probably worth its own ticket.

---

PR #1775, opened 2026-09-08
feat: Show where other people are reading in a shared session (#1692)

### LINKED ISSUE #1692: Display a visual indicator of scroll positions for other users in a shared session
Currently, the only way to know where another user is in a chapter is to watch for highlights or to have some other context for where they are and what they are doing. We have a design for showing user avatars and where they are inside a chapter, but it isn't implemented yet. This task is to do that.

Design: https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=9417-4626&t=QmKrn1LuFmNLnKHZ-4

Implementation notes:

- The avatars are only shown when the user is in a shared session.
- Your avatar is never shown. Only show avatars for other participants.
- We display user presence by showing a user's avatar combined with a range (start verse, end verse). This means that we have to capture what verses are in view for a user and then render their avatar plus a little range indicator.

Image

mazenemad:
Until now the only clue to where a peer was in a chapter was watching their highlights appear. Each other participant now gets a thin bar down the start edge of the reader, in their session colour, spanning the verses they can see, with their avatar on it.

fixes #1692

- The reader reports the verses on screen through an IntersectionObserver (so a scroll doesn't measure every verse on every frame) into a new `visibleVerseRange` on the reading state.
- SessionsManager publishes that range alongside the chapter it already broadcast. A scroll gets a longer debounce than a navigation: the range changes continuously, and peers only need where the reader came to rest, so a whole scroll gesture leaves one entry in a document that never shrinks.
- A position written without a range (or with a back-to-front one) still reports its chapter, so entries from before this existed keep working.
- Only other participants are drawn, only in the chapter on screen, and no gutter appears outside a session.

### DIFF
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
@@ -498,8 +498,13 @@
   font-size: 1em;
 
   /* Covers dimming and un-dimming via `.sb-chapter-content-stale` (see below).
-     Short enough to read as immediate rather than as an animation. */
-  transition: opacity 0.1s ease;
+     Short enough to read as immediate rather than as an animation. The indent
+     is here too: it opens for the session presence gutter and widens as people
+     gather on the same verses (see `.sb-chapter-content-presence`), and the
+     text should move with them rather than reflow in a jump. */
+  transition:
+    opacity 0.1s ease,
+    padding-inline-start 0.35s ease;
 
   /* Positioning context + own stacking context so the highlight ribbon layer
      can sit behind the scripture text (z-index: -1) without slipping behind the
@@ -707,7 +712,14 @@
  * this chapter, so a solo reader keeps the full text width.
  */
 .sb-chapter-content.sb-chapter-content-presence {
-  --sb-presence-gutter-width: 1.5rem;
+  --sb-presence-lane-width: 1.5rem;
+
+  /* One lane per bar that has to stand beside another (`--sb-presence-lanes`
+     comes from the measured markers); a session reading in different places
+     needs only the one. */
+  --sb-presence-gutter-width: calc(
+    var(--sb-presence-lanes, 1) * var(--sb-presence-lane-width)
+  );
 
   padding-inline-start: calc(var(--sb-presence-gutter-width) + 0.25rem);
 }
@@ -721,13 +733,47 @@
 }
 
 /* Placed against measured line boxes: from the top of the first verse the
-   participant can see to the bottom of the last. */
+   participant can see to the bottom of the last.
+
+   A peer's whole scroll arrives as a single update once they come to rest, so
+   without a transition the bar teleports down the page. Gliding to the new
+   verses reads as somebody moving instead. A marker that has just appeared has
+   no previous position to travel from and none is invented: it fades in where
+   its owner is (`@starting-style` below), and the same fade covers the remount
+   on a chapter change, which the key in BibleReader.tsx forces so a bar can't
+   glide across a page whose text was replaced under it. */
 .sb-presence-marker {
   position: absolute;
-  inset-inline: 0;
+  inset-inline-start: calc(
+    var(--sb-presence-marker-lane, 0) * var(--sb-presence-lane-width)
+  );
+  width: var(--sb-presence-lane-width);
   display: flex;
   align-items: center;
   justify-content: center;
+  opacity: 1;
+  transition:
+    top 0.35s ease,
+    height 0.35s ease,
+    inset-inline-start 0.35s ease,
+    opacity 0.25s ease;
+}
+
+@starting-style {
+  .sb-presence-marker {
+    opacity: 0;
+  }
+}
+
+@media (prefers-reduced-motion: reduce) {
+  .sb-presence-marker {
+    transition: none;
+  }
+
+  /* Keep the stale-content fade, drop the gutter's reflow. */
+  .sb-chapter-content {
+    transition: opacity 0.1s ease;
+  }
 }
 
 .sb-presence-range {
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
@@ -698,6 +698,60 @@
   cursor: pointer;
 }
 
+/* -------------------------------------------------------- session presence */
+
+/*
+ * Where the other people in a shared session are reading (#1692). Each one
+ * gets a thin bar down the start edge of the reader spanning the verses they
+ * can see, with their avatar on it. Only drawn when there is somebody else in
+ * this chapter, so a solo reader keeps the full text width.
+ */
+.sb-chapter-content.sb-chapter-content-presence {
+  --sb-presence-gutter-width: 1.5rem;
+
+  padding-inline-start: calc(var(--sb-presence-gutter-width) + 0.25rem);
+}
+
+.sb-presence-gutter {
+  position: absolute;
+  inset-block: 0;
+  inset-inline-start: 0;
+  width: var(--sb-presence-gutter-width);
+  pointer-events: none;
+}
+
+/* Placed against measured line boxes: from the top of the first verse the
+   participant can see to the bottom of the last. */
+.sb-presence-marker {
+  position: absolute;
+  inset-inline: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.sb-presence-range {
+  position: absolute;
+  inset-block: 0;
+  inset-inline-start: 50%;
+  width: 2px;
+  border-radius: 1px;
+  transform: translateX(-50%);
+  /* Softer than the avatar so several overlapping ranges stay readable. */
+  opacity: 0.55;
+}
+
+.sb-presence-avatar {
+  position: relative;
+  display: flex;
+  font-size: 0.625rem;
+}
+
+.sb-presence-avatar .sb-tab-user-icon {
+  width: 1.25rem;
+  height: 1.25rem;
+}
+
 .sb-verse-number .material-symbols-outlined {
   font-size: 1em;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### ✨ Added
 
+- See where everyone else is reading in a shared session. Each other participant gets a slim coloured bar beside the text spanning the verses on their screen, with their avatar on it, so you can tell at a glance whether they are with you or still a few verses back. Before this, the only clue was watching their highlights appear. ([#1692](https://github.com/HelloAOLab/seed-bible/issues/1692))
+
 ### 🔧 Changed
 
 ### 🐛 Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### ✨ Added
 
-- See where everyone else is reading in a shared session. Each other participant gets a slim coloured bar beside the text spanning the verses on their screen, with their avatar on it, so you can tell at a glance whether they are with you or still a few verses back. Before this, the only clue was watching their highlights appear. ([#1692](https://github.com/HelloAOLab/seed-bible/issues/1692))
+- See where everyone else is reading in a shared session. Each other participant gets a slim coloured bar beside the text spanning the verses on their screen, with their avatar on it, so you can tell at a glance whether they are with you or still a few verses back. The bars glide along as people move, and two people on the same verses stand side by side rather than one hiding the other. Before this, the only clue was watching their highlights appear. ([#1692](https://github.com/HelloAOLab/seed-bible/issues/1692))
 
 ### 🔧 Changed
 
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
@@ -17,12 +17,7 @@ import {
   useLayoutEffect,
   useState,
 } from "preact/compat";
-import {
-  computed,
-  useComputed,
-  type ReadonlySignal,
-  type Signal,
-} from "@preact/signals";
+import { computed, type ReadonlySignal, type Signal } from "@preact/signals";
 import {
   adjacentInlineRect,
   buildRibbonPath,
@@ -1344,6 +1339,13 @@ interface PresenceMarker {
   /** Offset from the top of the chapter content box, in px. */
   top: number;
   height: number;
+  /**
+   * Which column of the gutter this bar sits in. People reading the same
+   * verses would otherwise be drawn on top of each other, hiding everyone but
+   * whoever was painted last, so bars whose spans overlap are dealt out into
+   * side-by-side lanes. Zero whenever nobody overlaps.
+   */
+  lane: number;
 }
 
 interface ChapterContentProps {
@@ -1670,6 +1672,10 @@ function ChapterContent(props: ChapterContentProps) {
   // Signature of the last markers written to state, so the measure -> setState
   // -> re-render -> measure cycle settles instead of looping.
   const presenceSignatureRef = useRef("");
+  const presenceLaneCount = presenceMarkers.reduce(
+    (widest, marker) => Math.max(widest, marker.lane + 1),
+    1
+  );
 
   // A participant's bar spans from the top of the first verse they can see to
   // the bottom of the last, measured from the live line boxes so it follows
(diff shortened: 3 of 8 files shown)


Blitzy:
> [!NOTE]

## Review of No. 1775 — "Show where other people are reading in a shared session" (fixes #1692)

Nicely built and well-commented feature. The data flow (reader → `visibleVerseRange` → `readingPositionsMap` → peers → measured bars) is clean, the backward-compatibility handling (chapter-only entries, half-written/back-to-front ranges) is thoughtful, and the tests genuinely exercise real rendering and measurement rather than passing vacuously. I pulled the branch, verified the code against the surrounding source, and ran the new suites — **all 28 presence/range tests pass** (`test/unit/.../BibleReader.test.tsx` and `test/unit/.../SessionsManager.test.ts`).

Below are the issues I found, by severity. Nothing here is blocking; the headline item is a latency regression worth a look.

---

### Medium

**M1 — The 150 ms "navigation" fast path is effectively unreachable; every navigation now publishes at the 500 ms scroll cadence.**

`BibleReader.inline` intent (and the PR description) is that a chapter change gets the short `PUBLISH_DEBOUNCE_MS` (150 ms) and a scroll gets the longer `RANGE_PUBLISH_DEBOUNCE_MS` (500 ms). But `isNavigation` is recomputed on every run of the `stopBroadcastLocalPosition` effect, and that effect also subscribes to `visibleVerseRange`. A navigation is *always* immediately followed by verse-range changes:

1. Book/chapter change → effect run: `isNavigation = true` → 150 ms timer armed, `lastBroadcastBookId/Chapter` updated to the new chapter.
2. The old chapter's `IntersectionObserver` tears down and calls `onVisibleVersesChange(null)` → `visibleVerseRange` becomes `null` → effect run: `isNavigation = false` (last-broadcast already equals the new chapter) → the 150 ms timer is cleared and a **500 ms** one is armed.
3. The new chapter's observer measures and reports its range → effect run: `isNavigation = false` → 500 ms again.

Net result: because the 150 ms timer is cleared and replaced before it can fire, peers see your chapter change ~500 ms after you make it instead of ~150 ms. On `develop` the single `setTimeout(…, PUBLISH_DEBOUNCE_MS)` published navigations at 150 ms, so this is a ~3x latency regression for the presence/participant-position indicators (both the new bars and `SeedBiblePresenceProvider`'s chapter display).

There is also a narrow secondary effect under heavy load: if a re-render is delayed past 150 ms, the navigation timer can fire while `visibleVerseRange` still holds the *previous* chapter's range, briefly publishing `{ newChapter, oldFirstVerse, oldLastVerse }`. It self-corrects on the next measurement, so this is minor, but it stems from the same coupling.

*Reproduction:* In a two-client shared session, have client A jump between chapters while a stopwatch runs; observe on client B that A's marker/position updates ~half a second later. (The regression is easiest to confirm by unit test — assert that a pure book/chapter change with no range set publishes after `PUBLISH_DEBOUNCE_MS`, not `RANGE_PUBLISH_DEBOUNCE_MS`.)

*Suggested fix:* Decouple the two triggers so a navigation isn't downgraded by the range change that always accompanies it. For example, track whether *any* contributing change since the last actual broadcast was a navigation (a `pendingIsNavigation` flag OR'd across runs and reset inside the timer callback), and pick the shorter window when it is set. Alternatively, drive the range publish on its own debounce independent of the book/chapter effect. Also consider updating `lastBroadcastBookId/Chapter` inside the timer callback (when the broadcast actually happens) rather than on every effect run — the current names imply "last broadcast" but they track "last seen".

---

### Low

**L1 — Long continuous scrolls publish nothing until the reader pauses.**

`RANGE_PUBLISH_DEBOUNCE_MS` is a trailing-only debounce with no max-wait. If a user scrolls slowly and continuously (the visible range changing at least every 500 ms), the entry is never rewritten mid-gesture, so peers see the marker frozen at the pre-scroll position for the whole scroll. This is the documented tradeoff ("where the reader came to rest"), and it's a reasonable default, but a throttle with a max-wait (e.g. leading+trailing, or "publish at least every N ms while changing") would keep peers roughly in sync during a long read. Worth a comment acknowledging the behavior if it's intentional.

**L2 — Reported visible range is approximate and can over-report.**

The `IntersectionObserver` uses default options: viewport root and `threshold: 0`. So a verse counts as "visible" when even 1px enters the viewport, and verses occluded behind a bottom toolbar/chrome (but still within the viewport rectangle) are counted as visible. Peers' bars may therefore extend a verse or two beyond what the reader can actually read. For a presence hint this is probably fine, but if precision matters, consider a small negative `rootMargin` at the bottom (and/or a modest threshold) to exclude sliver-visible/occluded verses.

**L3 — Overlapping avatars when multiple peers share a verse span.**

When several participants are reading the same verses, their markers get the same `top`/`height` and their avatars fully overlap in the 1.5rem gutter (the CSS comment softens the *range bars* for this case, but the avatars themselves aren't de-collided). Fine for the common 1–2 peer case; may look messy in a larger session. The Figma in #1692 may specify stacking/offset behavior worth matching.

---

### Nits

**N1 — Confusing dead binding in the effect.**

```ts
const rangeChanged = readingState.visibleVerseRange.value;
void rangeChanged;
```

The name reads like a boolean but it's the range value, and it differs from the sibling `void readingState.bookId.value;` lines. Collapse to `void readingState.visibleVerseRange.value;` for consistency.

**N2 — Test debounce waits use fixed sleeps.**

`flushRangePublishDebounce` (700 ms) and the existing `flushPublishDebounce` (250 ms) are hardcoded real-time waits. This matches the file's existing convention, so it's not new, but `CLAUDE.md` asks to prefer polling / fake timers over fixed sleeps. `vi.useFakeTimers()` would make the two new range tests deterministic and drop ~1.4s of wall time. Optional.

---

### On the tests (the routine's specific ask)

I checked whether the tests could pass with the feature broken, and they hold up:

- **"draws a bar spanning the verses another participant can see"** asserts concrete measured geometry (`top: 40px`, `height: 54px`) computed from stubbed line boxes, plus the range color and avatar title — a broken measurement would fail it. Because it relies on the real component rendering `.sb-verse[data-verse-number]` elements for `querySelector` to find, it exercises the actual render path, not a stub.
- **"publishes one entry for a scroll that passes several verses"** genuinely verifies debounce collapse (5 synchronous range changes → exactly one `set`) — it would fail if the debounce were removed.
- The negative tests (own position, different chapter, no range, outside session) assert absence, but they're anchored by the positive test's presence assertions on the same class names, so a class-name typo would surface.
- **"reports the verses on screen…"** drives a stubbed `IntersectionObserver` and asserts the range both widens and *narrows* as verses cross the edge — real behavior, not always-green.

One gap worth closing given M1: there's no test asserting the *navigation* debounce window (150 ms) versus the *scroll* window (500 ms). Adding one (a book/chapter change with no range set should publish at `PUBLISH_DEBOUNCE_MS`) would both document the intended distinction and catch the M1 regression.

---

KallynGowdy:
Taking a first look at this, it seems like highlights and the viewing indicators get in each other's way sometimes:

### Device 1
image

### Device 2

image

Notice how the highlights on Device 2 aren't lined up right:

image

Also notice how even though both devices are viewing verses 1-19, it says that they're actually viewing verses 11-19:

image

The range only seems to get fixed once you scroll down past the highlights.

KallynGowdy:
@mazenemad This is a big improvement. I've got a couple more things:

1. User avatars should be displayed at the top of the indicator range and then be sticky so they follow as you scroll. In this example, I can see the avatars of of the red and purple users, but not the blue one (because it is below where I am):
image

This means that when two users are overlapping, then their avatar list should collapse like how we do it for the mobile session list:
image


2. When ranges overlap, they have a bunch of margin that pushes the user's scripture over to the side more. Instead, all the user avatar lines should share the same area so that the space that they take is constant. Compare the two examples below:
image
image

In the first, the blue, red, and purple lines take up a lot more horizontal space, which pushes the scripture over. Instead, they should all be placed in the same "gutter" with minimal spacing between them. It should be more like this:
image

If there are more lines than will fit in the gutter, we should just ignore the overlapping areas for those lines. We should still include them in the avatar list.

3. When a range extends off the screen above or below where the viewport is, there should be tiny arrow indicating that it extends further. This is so the user has a hint that the other user is viewing more than the other user (either more above, more below, or both)

e.g. 

image

Blitzy:
I really like the way this feature is shaping up. I agree with all of @KallynGowdy's notes in the comment above.

## My findings

### No. 1 — Gutter will push the text to the right when a user enters your page that didn't previously have users

This is supporting @KallynGowdy's No. 2 in the comment above. If a chapter doesnt have any other readers, and then another user enters the same book+chapter as you your text will shift to the right because the gutter for the avatar presence visual is made. 

https://github.com/user-attachments/assets/d3d83924-e9b3-41e5-90af-0abf24364218

This text shift also occurs when the gutter grows to accomodate another user and their line.

https://github.com/user-attachments/assets/145c7b01-28e2-4dbd-a3b8-152d407ef19b

Ideally the gutter is always present and the size is fixed in shared sessions so that it never shifts text while reading it.

_We really do need a fixed gutter size and should collapse lines when a bunch are overlapping :laugh:_
image


### No. 2 — If a user switches (or backgrounds) the tab, their presence on the page should also disappear

This one is pretty simple, if a user is literally not present on the page then the presence avatar and line should not be shown for them. If they flip to a new tab or the tab is otherwise backgrounded, their presence line should disappear. 

Keep in mind this is different from their actual presence in the session itself, this shouldn't necessarily remove that they are connected to the session but instead remove the avatar + line that shows where they are actively looking at on the page.

https://github.com/user-attachments/assets/df50d67b-a876-4531-bbb5-64ea5c46cd01

Blitzy:
> [!NOTE]

## Review of #1775 — "Show where other people are reading in a shared session" (fixes #1692)

This was a single-pass review done without the Agent tool (no multi-agent fan-out, no subagent verify pass), so treat findings as leads to double-check rather than a fully cross-verified report.

I re-read the whole diff against `develop`, plus the linked issue #1692 and the existing conversation on this PR (Blitzy's two reviews and @KallynGowdy's two rounds of visual feedback). The latest commit (`ad47854e`, today) looks like it lands fixes for essentially everything raised so far: the gutter is now a fixed width and open for the whole session (no more text-shifting when someone arrives/leaves — Blitzy's No. 1), presence disappears when the tab is backgrounded (Blitzy's No. 2), avatars are now sticky/pinned with edge arrows and collapse into a "+N" stack (Kallyn's Nos. 1–3), and the "highlighted verses drop out of the reported range" bug is fixed with a real regression test. The debounce fix for Blitzy's M1 (chapter changes were sliding onto the 500ms scroll cadence) also checks out — I traced the `pendingIsPrompt` logic through several effect re-runs and it correctly keeps the short window armed until the actual broadcast fires.

**On tests**: I don't think these are tautological. The `SessionsManager.test.ts` additions use a real in-memory `Map`-backed mock (`createMockSharedMap`) rather than a stub that just echoes back whatever was passed in, so `expect(...set).toHaveBeenCalledWith(...)` is asserting on values the code actually computed, not on the mock agreeing with itself. The `BibleReader.test.tsx` presence tests derive their expected pixel values independently from the stubbed line-box geometry (e.g. "verse 1 at 40, verse 2 at 70 → bar height 54px") rather than just checking that *something* rendered, and the highlighting-replaces-the-DOM-element test genuinely fails without the `sync()`/`syncVersesRef` mechanism (I confirmed the reasoning by tracing what `observed` would contain without that code path — the replacement element would never get observed).

A few things I'd still flag:

### 1. A participant's presence marker can go stale (wrong name/color shown)

`packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx`, `measurePresence` (~line 2125):

```js
const signature = next
  .map((m) => `${m.connectionId}:${Math.round(m.top)}:${Math.round(m.height)}:${m.lane}:${m.imageUrl ?? ""}`)
  .join("|");
if (signature === presenceSignatureRef.current) return;
presenceSignatureRef.current = signature;
setPresenceMarkers(next);
```

The dedup signature that guards `setPresenceMarkers` only covers `connectionId`, `top`, `height`, `lane`, and `imageUrl` — it leaves out `displayName` and `visual` (color), even though both are carried on `PresenceMarker` and rendered (`marker.visual.color` as the bar's background, `marker.displayName` as the avatar's tooltip title). If a peer renames themselves mid-session (the app does support editing your display name live, via `ProfileConfigSync`/`LoginManager`) while their scroll position happens not to change, `measurePresence` computes a `next` array whose signature is identical to the previous one, so `setPresenceMarkers` is skipped and the old, stale `displayName`/color keep rendering — the reader keeps seeing the old name/color until that participant's bar geometry changes for some unrelated reason. Suggest folding `displayName` and `visual.color` into the signature (or just comparing objects field-by-field instead of a signature string).

### 2. Self-reported "visible verse range" doesn't account for the floating mobile header, but the gutter's own viewport measurement does

`packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx`: the `IntersectionObserver` that drives `onVisibleVersesChange` (~line 2018) uses the default root/threshold, so a verse that's geometrically inside the viewport but visually hidden behind `.sb-bible-reader-mobile-header` (which is `position: absolute; top: 0` and floats over the top of the scroller, per `BibleReader.inline.css`) is still reported as "visible" and gets published to peers as part of `visibleVerseRange`. Compare that with `measurePresenceViewport` (~line 1552), used to place *other* participants' bars/arrows/pinned avatars, which explicitly subtracts the header's height (`top = Math.max(top, header.getBoundingClientRect().bottom)`) plus the bottom chrome inset. So the same app has two different definitions of "on screen" — one for what you tell your peers you're reading, one for what you draw of your peers — and a reader whose top verse is currently tucked behind the mobile header will report a `firstVerse` that's actually invisible to them. Minor in practice (off by however many verses fit under the header), but it's the kind of mismatch that will look like a real bug in testing on a phone ("why does it say Mary's on verse 3 when she can't even see verse 3 yet").

### 3. Efficiency: `presence` is rebuilt (and a fresh array reference produced) on every `BibleReader` render

`packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx`, in the `BibleReader` component body (~line 2350 area, the `const presence: ParticipantPresence[] = []; ...` block): this loop over `sharedSession.connectedUsers.value` runs unconditionally in the render body, and produces a brand-new array every time `BibleReader` re-renders for *any* reason (a verse selection, a highlight, scroll position bookkeeping, etc.), not just when session state actually changes. That new array reference is then handed to `ChapterContent` as the `presence` prop, so `ChapterContent`'s own reflow-measuring layout effect (`measureRibbons(); measurePresence(); syncVersesRef.current();`) also runs unconditionally every render regardless of prop identity — meaning this doesn't currently cause *extra* runs of that effect, but it does mean the O(participants) presence-list build runs far more often than it needs to. Given the PR intentionally split `PresenceGutter` out specifically to avoid extra rendering cost on scroll, it seems worth applying the same care here — e.g. wrapping the list construction in a `useComputed`/memo keyed on the actual position/user data instead of rebuilding on every render.

### 4. Reuse: `findScrollContainer` duplicates existing "walk up for the nearest scrollable ancestor" logic

`packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx`, `findScrollContainer` (new in this PR) re-implements the same "walk `parentElement` checking computed `overflowY`" pattern that already exists in `packages/seed-bible/seed-bible/components/BibleSelector/BibleSelector.tsx` (~line 1023–1033, which also checks `overflowY === "overlay"` and `scrollHeight > clientHeight`). Not identical, but close enough that it's worth checking whether `BibleSelector`'s version (or a shared extraction of it) could be reused/extended instead of a third, slightly different copy appearing the next time this is needed.

---

Nice feature overall — the debounce/backward-compat/tab-visibility handling in `SessionsManager` in particular is careful and well-tested. None of the above feels blocking; #1 and #2 are the ones I'd actually want a decision on before merge, #3 and #4 are just cleanup.


---

packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
```diff
@@ -705,6 +713,180 @@
...
+  --sb-presence-gutter-width: 1.75rem;
+  /* Mirrored by PRESENCE_AVATAR_PX / PRESENCE_STACK_STEP_PX in BibleReader.tsx,
+     which places the avatar stacks against measured verse positions. */
+  --sb-presence-avatar-size: 20px;
+  --sb-presence-stack-step: 12px;
+  /* One bar's width plus the gap to the next. Three fit, centred. */
+  --sb-presence-lane-pitch: 6px;
+
+  padding-inline-start: calc(var(--sb-presence-gutter-width) + 0.25rem);
+}
```
KallynGowdy: @mazenemad All the added CSS doesn't need to be in the `BibleReader.inline.css` file. It should just be in a regular `BibleReader.css` file so that it doesn't get inlined in the HTML on first load

KallynGowdy (changes requested):

Blitzy (approved):
This is looking really good. I am excited to get this in.

I went through the 4 review points. None of them block the merge:

1. **Stale name/color:** the color can't go stale because it comes from the connection ID. Only the hover tooltip name can lag after a rename. It's a one-line fix if we want it, but low impact.
2. **Mobile header overlap:** at most a line or two off. The header hides on scroll, so a proper fix isn't cheap. Not worth it.
3. **Presence list rebuilt each render:** skipping `useComputed` here is deliberate (see the comment). Memoizing it would bring back the "nobody's here" bug. The cost is tiny.
4. **Duplicate scroll-container lookup:** not a real duplicate. BibleSelector's loop does a different job.

packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
Blitzy: fixed

PR #1776, opened 2026-09-08
feat: Show note markers beside the text on mobile (#1691)

### LINKED ISSUE #1691: Implement side display for notes on mobile
Currently, the only visual indicator that a user has a custom note for a verse on mobile is a box drawn around the verse number. On desktop, they can open the discover panel to see their notes while reading, but this isn't possible on mobile.

Follow this design: https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=9452-436&t=QmKrn1LuFmNLnKHZ-4

In particular:

- Note icons should be rendered to the side of the Bible Reader.
- They should only be shown when a user has a note.
- The icons should be displayed inline with where the verse number is.
- Clicking on a note icon shows the note(s) for the verse by selecting the verse and automatically expanding the verse toolbar and scrolling to the note.
- If a note spans multiple verses, then an icon should only be shown on the first verse for the note.

mazenemad:
Desktop can read its notes in the Discover panel while reading; mobile could not, so the only sign a verse had a note was the box around its verse number.

fixes #1691

Mobile now draws a marker in a gutter at the end edge of the reader for every verse a note begins at. Each marker is placed against the measured first line box of its verse, so it lines up with the verse number however the text wraps. Tapping one does what tapping the boxed number does: selects the verse, expands the verse toolbar and scrolls to the note.

A note spanning several verses marks only the verse it starts at, and a verse with several notes gets one marker. The gutter (and its extra inline padding) only appears when the chapter has notes, so an unannotated chapter keeps its full text width.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### ✨ Added
 
+- See your notes at a glance while reading on a phone. Every verse a note starts on now gets a small note icon beside the text, lined up with its verse number; tapping it selects the verse and opens the note. Previously the only sign a verse had a note was a box around its number, and there was no way to read the note without hunting for it. ([#1691](https://github.com/HelloAOLab/seed-bible/issues/1691))
+
 ### 🔧 Changed
 
 ### 🐛 Fixed
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.inline.css
@@ -698,6 +698,57 @@
   cursor: pointer;
 }
 
+/* ------------------------------------------------------------ note gutter */
+
+/*
+ * Mobile note markers (#1691). Desktop reads its notes in the Discover panel
+ * beside the text; mobile has no such panel, so the markers move into a gutter
+ * at the end edge of the reader. The gutter only appears when the chapter
+ * actually has notes, so an unannotated chapter keeps its full text width.
+ */
+.sb-chapter-content.sb-chapter-content-noted {
+  --sb-note-gutter-width: 1.75rem;
+
+  padding-inline-end: calc(var(--sb-note-gutter-width) + 0.25rem);
+}
+
+/* With a marker in the gutter, the inline one (shown when verse numbers are
+   turned off) would be the same note flagged twice on the same line. */
+.sb-chapter-content-noted .sb-verse-annotation-icon {
+  display: none;
+}
+
+.sb-note-gutter {
+  position: absolute;
+  inset-block: 0;
+  inset-inline-end: 0;
+  width: var(--sb-note-gutter-width);
+  pointer-events: none;
+}
+
+/*
+ * Placed against a measured line box: `top` and `height` come from the first
+ * line rect of the verse the note starts at, which is the line carrying the
+ * verse number — so the icon lines up with it however the text wraps.
+ */
+.sb-note-gutter-marker {
+  position: absolute;
+  inset-inline: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: 0;
+  border: none;
+  background: none;
+  color: var(--sb-primary-color);
+  cursor: pointer;
+  pointer-events: auto;
+}
+
+.sb-note-gutter-marker .material-symbols-outlined {
+  font-size: 1.125rem;
+}
+
 .sb-verse-number .material-symbols-outlined {
   font-size: 1em;
 }
--- a/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
+++ b/packages/seed-bible/seed-bible/components/BibleReader/BibleReader.tsx
@@ -1310,6 +1310,15 @@ const CHAPTER_SKELETON_PARAGRAPHS = [
   ["95%", "96%", "98%", "89%", "68%"],
 ] as const;
 
+/** A note marker in the mobile gutter, placed against a measured line box. */
+interface NoteMarker {
+  verseNumber: number;
+  /** Offset from the top of the chapter content box, in px. */
+  top: number;
+  /** Height of the verse's first line box, so the icon centres on it. */
+  height: number;
+}
+
 interface ChapterContentProps {
   chapterData: Signal<TranslationBookChapter | null>;
   chapterDataPromise: Promise<void>;
@@ -1337,6 +1346,11 @@ interface ChapterContentProps {
    * the chapter they navigated to arrives.
    */
   isStale?: boolean;
+  /**
+   * Mobile has no Discover panel to read notes alongside the text, so the
+   * note markers move out into a gutter beside the scripture there (#1691).
+   */
+  isMobile?: boolean;
 }
 
 function ChapterContent(props: ChapterContentProps) {
@@ -1354,8 +1368,10 @@ function ChapterContent(props: ChapterContentProps) {
     justConvertedSelectionRef,
     scriptureElements,
     onAnnotationVerseClick,
+    isMobile = false,
   } = props;
 
+  const { t } = useI18n();
   const currentChapter = chapterData.value;
   const chapterAnnotations =
     currentChapter && annotations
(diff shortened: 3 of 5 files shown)


Blitzy:
> [!NOTE]

## Summary

This PR adds a mobile note gutter for issue [#1691](https://github.com/HelloAOLab/seed-bible/issues/1691): every verse a note starts on gets a `sticky_note_2` marker in a gutter at the end edge of the reader, measured against the verse's first line box so it lines up with the verse number, and tapping it selects the verse and scrolls to the note.

I checked the change against the issue's acceptance criteria and it meets all of them:

- Icons render to the side of the reader ✔
- Only shown when the chapter has notes ✔ (`showNoteGutter = isMobile && noteVerseNumbers.length > 0`)
- Placed inline with the verse number ✔ (measured first-line rect)
- Tapping selects the verse, expands the toolbar and scrolls to the note ✔ (routes through the same `onAnnotationVerseClick` the boxed number uses)
- A multi-verse note marks only the first verse ✔ (`Math.min(...verses)` + dedup)

**Verdict:** solid, ready to merge after considering the low-severity notes below. `pnpm check:ts`, `pnpm lint` (0 errors), and the full `BibleReader.test.tsx` suite (82 tests) all pass locally on the head commit.

### Test quality

The 6 new tests exercise real behavior, not mocks agreeing with mocks: they stub only the layout boundary jsdom can't provide (`getClientRects`, since jsdom does no layout) and then assert on rendered output — marker count, `style.top`, and the click side effects (`selectVerse`, `pendingAnnotationScrollVerse`). I sanity-checked that they aren't vacuous by mutating the dedup logic from `Math.min` to `Math.max`; the "marks only the first verse of a note that spans several" test then fails, which is what you want. The span test stubs a line box for the second verse too, so an implementation that marked every spanned verse would produce a second marker and be caught. No always-green tests here.

---

## Findings

No High or Medium severity issues. Everything below is Low / design-confirmation.

### Low

**No. 1 — Two indicators on the same verse when verse numbers are shown.**
The CSS hides the inline `.sb-verse-annotation-icon`, but that icon only renders when verse numbers are turned *off* (see `renderVerseNumberOrIcon`). When verse numbers are **on**, an annotated verse still shows the boxed, clickable number (`.sb-verse-number-annotated`) *and* now also the gutter marker — two indicators, both focusable, both firing the same `onAnnotationVerseClick`.
- *Repro:* on a mobile viewport with verse numbers enabled, open a chapter with a note. The verse number is boxed **and** a marker appears in the gutter.
- This may be exactly what the Figma intends (box on the number + tappable icon at the side). Worth a quick confirm against the design at [node-id 9452-436]. If the intent is a single affordance, consider dropping the box on mobile when the gutter is active, or making the boxed number non-interactive there to avoid the duplicate tab stop.

**No. 2 — Reading / tab order puts the gutter buttons before the scripture.**
The `.sb-note-gutter` block is rendered before `renderChapterContent(...)`, so a keyboard or screen-reader user traverses every note button first, then reaches the chapter text. Combined with No. 1, an annotated verse can be two tab stops to the same action. Not blocking, but consider `tabindex`/DOM-order or grouping so the markers don't front-load the chapter.

**No. 3 — Marker centers on the whole first-line box, while the verse number is a raised superscript.**
`.sb-note-gutter-marker` gets `top`/`height` from the line rect and flex-centers the icon, so it sits at the vertical middle of the line. The verse number is a `<sup>` (`top: -0.4em`), so it rides near the top of that line. The icon can therefore sit a few px below the number rather than truly level with it. Minor and likely fine at reading sizes, but flagging since the issue specifically says "inline with where the verse number is."

**No. 4 — Stale `ResizeObserver` closure (harmless, mirrors existing ribbon code).**
The `ResizeObserver` is created once (`[]` deps) and closes over the first render's `measureNoteMarkers`, so its captured `noteVerseNumbers`/`showNoteGutter` can be stale after the note set changes. In practice this self-heals: the unconditional `useLayoutEffect(() => { measureRibbons(); measureNoteMarkers(); })` re-measures with a fresh closure on every render *before paint*, so any stale write is corrected before the browser paints — no visible flicker, just an occasional redundant render. It also exactly mirrors the pre-existing `measureRibbons()` usage right beside it. Calling it out only for completeness; if you ever want to tidy both, hold the measure functions in a ref the observer reads. No action needed for this PR.

---

## Suggested follow-up tests (optional)

The current tests cover the core behavior well. If you want to close the remaining gaps:

- Assert the inline `.sb-verse-annotation-icon` is hidden when the gutter is active (the CSS rule at `BibleReader.inline.css:717` currently has no test).
- Assert the marker's `style.height` matches the measured line box (only `top` is asserted today), so a regression in vertical sizing is caught.
- A test for a note whose start verse is missing from the rendered chapter (no line rect) — confirm it produces no marker rather than throwing.

---

KallynGowdy:
@mazenemad It seems to place notes in the center of the verse instead of near the start of the verse when the verse contains poem lines. Here's an example with Genesis 1:27:

image

Blitzy:
I think my main problem is that the addition of the gutter for the note icons leaves the reader text feeling unbalanced now that it is uncentered.

image

I think if we go in this direction the left side should maintain identical gutter size of the right size to keep reader text balanced in the center of the screen. Keeping in mind that we are also looking to add user presence via #1775 

I also believe that No 2 from the code review is a real problem for accessibility. However, testing it out myself brings to light a number of keyboard navigation issues that far pre-date this PR. I dont consider it something work fixing with this PR but we do need to do a follow up and fix some keyboard navigation issues with the reader.

_Recording of keyboard tabbing navigation on `develop` just for reference:_

https://github.com/user-attachments/assets/d75f54c4-f499-4078-92f2-58b83b445163

Blitzy:
> [!NOTE]

## Summary

This PR closes issue #1691: on mobile, every verse a note starts on now gets a `sticky_note_2` marker in a gutter beside the reader, measured against the verse's actual first line so it lines up with the verse number even when the text wraps (including poetry lines). Tapping a marker selects the verse and scrolls to the note, same as tapping the boxed verse number does today.

I re-read the code at the current head (`57bc04a`), re-checked the two prior review threads on this PR, and reproduced/verified the specific claims below directly against the diff and the test file rather than taking them at face value.

**What's already been resolved since the last review:**
- KallynGowdy's Sept 11 report that a note on a poetry verse (e.g. Genesis 1:27) placed the marker halfway down the verse instead of beside the verse number — fixed in commit `e3ec64b` (Sept 15), which switches the measurement to `collectLineRects` (the same helper the highlight ribbons already use) instead of the verse element's own bounding rect. There's a dedicated regression test for this ("marks a poetry verse level with its first line, not its middle") that stubs multi-line poetry rects and asserts `top`/`height` come from the first line, not the whole block.
- The Sept 8 automated review's accessibility note (tab order puts the gutter buttons before the scripture text) was explicitly discussed and deferred in Blitzy's Sept 21 comment — pre-existing keyboard-navigation problems on `develop` were confirmed out of scope for this PR, with a follow-up planned. Not re-flagging it here since that's already a team decision, not an open question.

**What's still open and, as far as I can tell from the commit history, unaddressed by any code change since it was raised:**

### Medium — Visual balance concern from Blitzy's Sept 21 comment has no code change behind it

Blitzy's Sept 21 comment says the reader text now feels uncentered/unbalanced with the gutter added, and asks that the non-marker side reserve the same width as the marker side. Looking at the CSS, that symmetric reservation is already there — `.sb-chapter-content-noted` applies `padding-inline` (both logical sides) to `var(--sb-note-gutter-width) + 0.25rem`, and only the `.sb-note-gutter` element itself (absolutely positioned, `inset-inline-end: 0`) holds the markers. So the text column's *width* is already centered.

The only two feature commits are the initial one (Sept 8) and the poetry fix (Sept 15); everything after Sept 21 is a `develop` merge, not a content change. So this symmetric padding was present when Blitzy left that comment, and nothing has changed in response to it — meaning either (a) the comment is about something CSS alone doesn't fix (the markers themselves are only ever on one side, so the page still *reads* asymmetric even though the text box is centered), or (b) it was addressed by a design decision I don't have visibility into. Since there's no formal review submitted on this PR yet (`get_reviews` returns empty — Blitzy's and the earlier bot review are both PR comments, not a GitHub "review"), I'd treat this as an open item worth an explicit yes/no before merge rather than assuming it's settled: either Blitzy confirms the symmetric-padding approach addresses the concern, or there's a follow-up needed on the marker side itself (e.g., a barely-visible tick mark or matching reserved-but-empty affordance on the start edge) to make the asymmetry read as intentional rather than accidental.

### Low — `verseByNumber`'s type guard is looser than the existing `isVerseEntry`, though not currently reachable

`BibleReader.tsx` (around line 1722) builds `verseByNumber` with an inline check:

```ts
if (entry && typeof entry === "object" && entry.type === "verse" && typeof entry.number === "number") {
  verseByNumber.set(entry.number, entry as ChapterVerse);
}
```

This duplicates the existing `isVerseEntry` guard defined earlier in the same file (line 1063), which additionally requires `Array.isArray(entry.content)`. CLAUDE.md's own duplication guidance calls for reusing a close-match helper rather than writing a parallel version, and here the parallel version is strictly weaker.

I traced whether this divergence is actually reachable: a malformed content entry (`type: "verse"`, non-array `content`) would fail `isVerseEntry`, so `renderChapterContent` would never render a `.sb-verse[data-verse-number]` element for it. `measureNoteMarkers` only creates a gutter marker when it finds that exact DOM element (`content.querySelector('.sb-verse[data-verse-number="${verseNumber}"]')`) — no element, no marker, `continue`. Since `verseByNumber` is only ever read via `verseByNumber.get(marker.verseNumber)` inside a marker's own click handler, a verse that `isVerseEntry` would reject can never actually produce a marker to click, so this doesn't crash anything in practice today.

It's still worth tightening for maintainability — the moment either guard changes independently (e.g., `isVerseEntry` grows another check for some future data shape), they'll silently drift again. Suggested fix: replace the inline loop with `if (isVerseEntry(entry)) verseByNumber.set(entry.number, entry);` to reuse the existing helper instead of re-deriving it.

## Tests

The 6 new tests in the `"note gutter"` describe block assert on real rendered output (marker count, `style.top`/`style.height`, click side effects on `selectVerse` and `pendingAnnotationScrollVerse`) rather than internals, and stub only the layout boundary jsdom genuinely can't provide (`getClientRects`), consistent with CLAUDE.md's test guidance. The poetry-specific test is a proper regression test for the Sept 11 bug: it stubs a multi-line poetry verse and would fail if the marker were centered on the whole verse block instead of its first line — I confirmed this by reading `e3ec64b`'s diff, where the same commit that fixes the measurement also introduces this test, which is the shape a real regression test should take (test and fix land together, and the test's assertions directly encode the bug that was reported).

One gap, not blocking: there's no test for a note whose start verse isn't present in the rendered chapter at all (missing `verseEl`) confirming it produces no marker rather than throwing. `measureNoteMarkers`'s `if (!verseEl) continue;` handles this correctly today, but it's currently unverified by a test — worth adding if you're touching this code again.

## Verdict

No High/Critical issues. CI is green (`build (24.x)` passing on the current head) and the change is scoped correctly to mobile only, with the gutter itself gated on the chapter actually having notes. The one thing I'd want resolved before merge is the Medium item above — not because the code is wrong, but because it's an explicit, still-open reviewer objection with no confirmed resolution in the thread, and this PR currently has zero formal review approvals.

---

PR #1778, opened 2026-09-08
feat: let a Customization set its own default Bible translation

KallynGowdy:
Adds an optional defaultTranslationId to SeedBibleCustomization (with an
editor field to pick it), reusing the same override concept
BrandingConfig.defaultTranslationId already provides for a whole
deployment. TabsManager applies it to the selected tab once the
active Customization is known — a share link resolves it over the
network, so unlike branding it can't be read synchronously at tab
creation — deferring to an explicit deep link or a signed-in reader's
own saved translation when either is present.

Claude-Session: https://claude.ai/code/session_01TyQ1hZYDy2LtY1fjKVk8KD


Blitzy:
> [!NOTE]

## Review: `feat: let a Customization set its own default Bible translation`

Thanks for this — the implementation is clean and the intent is well documented in the code comments. The feature mirrors `BrandingConfig.defaultTranslationId` sensibly, and the async application via an effect (rather than at tab-creation time) is the right call given a `?customization=...` link only resolves over the network. No linked issue is referenced on the PR, so this review is against the stated goal in the description.

The precedence design itself looks correct: an explicit deep link wins, then a signed-in reader's own saved translation, then the customization default, then Seed Bible's per-language default. My findings are mostly about **test coverage of that precedence**, plus a couple of minor UI/behavior notes. Nothing here is a blocker.

---

### 🟠 Medium

**No. 1 — The "profile wins over customization default" test passes even when the precedence logic is deleted (it doesn't actually test what it claims).**

`test/unit/seed-bible/managers/TabsManager.test.ts` → *"prefers a signed-in reader's own saved translation over the active customization's default"*.

The test sets the customization default to `NIV` at construction, then sets the profile (`AAB`) **after** the initial load. Because the profile-restore effect always runs last and unconditionally wins, the final state is `AAB` regardless of whether the customization effect's profile check exists. In other words, the test can't fail for the reason it's named for — the scenario it describes (a signed-in reader whose translation must beat the customization default) is never actually exercised, because the guard that enforces that (`if (savedTranslationId) return;` in the customization effect in `TabsManager.tsx`) is never the thing under test.

**Reproduction (confirmed locally):**
1. In `TabsManager.tsx`, in the `activeCustomizationDefaultTranslationId` effect, delete the block that reads the saved profile translation and returns early:
   ```ts
   const savedTranslationId = getProfileConfigValue(
     login.profile.value,
     PROFILE_TRANSLATION_ID
   );
   if (typeof savedTranslationId === "string" && savedTranslationId) {
     return;
   }
   ```
2. Run `pnpm vitest run TabsManager.test.ts -t "prefers a signed-in reader"`.
3. It still passes — the deleted guard is real logic, but no test turns red.

The gap is the ordering that actually depends on the guard: **the profile is already loaded when the customization default resolves** (the realistic case of a signed-in user opening a `?customization=...` link). A test for that fails without the guard and passes with it. Suggested addition:

```ts
it("does not let a customization default override an already-loaded profile translation", async () => {
  setWebResponses(createExampleManagerResponseMap());
  const activeCustomizationDefaultTranslationId = signal<string | undefined>(
    undefined
  );

  const { tabs: manager, login } = createTabsManager({
    activeCustomizationDefaultTranslationId,
  });
  const firstTab = manager.tabs.value[0]!;
  await waitForInitialLoad(firstTab.readingState);

  // Signed-in reader whose saved translation is NIV, loaded BEFORE the
  // customization default resolves.
  login.userId.value = "user-1";
  login.profile.value = { name: "", config: { translationId: "NIV" } };
  await waitFor(() => firstTab.readingState.translationId.value === "NIV");

  // Customization default (AAB, different) now resolves over the network.
  activeCustomizationDefaultTranslationId.value = "AAB";
  await new Promise((resolve) => setTimeout(resolve, 20));
  await waitForInitialLoad(firstTab.readingState);

  expect(firstTab.readingState.translationId.value).toBe("NIV");
});
```

I verified this addition **fails** on a build with the guard removed (`expected 'AAB' to be 'NIV'`) and **passes** with the guard in place — so it genuinely protects the precedence the PR is trying to guarantee.

---

### 🟡 Low

**No. 2 — The editor `<select>` silently shows "Seed Bible's default" when the saved translation isn't in `availableTranslations`.**

In `CustomizationEditMainView`, the options come from `bibleData.availableTranslations.value`. If `record.defaultTranslationId` holds an id that isn't in that list — a translation later removed from the catalog, or simply the brief window before `availableTranslations` has populated — the `<select value={record.defaultTranslationId ?? ""}>` matches no option and the browser falls back to displaying the first option, "Seed Bible's default." The stored value is not lost (nothing is written unless the user actively changes the field), but the editor misrepresents the current setting.

This is an edge case and non-destructive, so low priority. If you want to close it, render the saved id as an extra disabled option when it's absent from the list (e.g. `{savedId && !sortedTranslations.some(t => t.id === savedId) && <option value={savedId}>{savedId} (unavailable)</option>}`), so the user can see that a value is set even when the translation is gone.

**No. 3 — Deactivating a customization (or clearing its default) does not revert the reader off that translation.**

The effect early-returns when `activeCustomizationDefaultTranslationId` becomes `undefined`, so a previewer who applied customization A's `NIV` default and then clears the customization stays on `NIV` rather than returning to Seed Bible's default. Switching directly from customization A (`NIV`) to B (`ESV`) *does* re-apply correctly; only the transition to "no customization / no default" doesn't revert. This matches how `branding.defaultTranslationId` behaves (apply-forward, never revert), so it may well be intended — flagging it only so the choice is deliberate. No change needed if that's the accepted behavior.

---

### 🔵 Nit

**No. 4 — The `updateEditingDefaultTranslationId()` no-op test asserts against `undefined === undefined`.**

*"...no-ops when there is no open draft"* asserts `manager.customizations.value[0]?.defaultTranslationId` equals `created.defaultTranslationId`, both of which are `undefined`, so the assertion holds trivially. It does still confirm the persisted list wasn't mutated, but it would be a touch stronger to also assert `manager.editingCustomization.value` is `null` (proving the no-op path was the one taken), matching the other no-open-draft tests.

---

### ✅ Things that look good

- The `appliedCustomizationDefaultTranslationId` guard correctly stops an unrelated draft edit (e.g. a rename) from re-applying the default and clobbering a translation the previewer picked manually — and the "renaming doesn't clobber" reasoning is captured in the comment.
- An invalid/removed `defaultTranslationId` degrades gracefully: `applySavedTranslation` → `getTranslationBooks` catches the failure and bails, leaving the reader on the existing translation rather than erroring.
- Schema round-trips correctly — the `load()` test confirms a persisted record carrying `defaultTranslationId` parses back through Zod.
- The two positive-path TabsManager tests (No `?translation=` share-link case, and the explicit-URL-wins case) are real regression tests and do fail if their respective guards break.
- Only `en.json` was touched for i18n, per the repo convention.

### Test run
`pnpm vitest run TabsManager.test.ts` → 67 passed. `pnpm vitest run CustomizationsManager.test.ts` → 96 passed. (Both on this PR's head.)

https://claude.ai/code/session_018mNUFdfXbhu2ksuzV7qaSH

---

KallynGowdy:
Thanks for the thorough review — addressed the findings:

- **No. 1 (Medium)**: Replaced *"prefers a signed-in reader's own saved translation over the active customization's default"* with your suggested test. Verified locally that it fails (`expected 'AAB' to be 'NIV'`) with the profile guard removed from the customization effect in `TabsManager.tsx`, and passes with it restored.
- **No. 2 (Low)**: `CustomizationEditMainView` now renders the saved `defaultTranslationId` as a disabled `{{id}} (unavailable)` option when it's missing from `availableTranslations`, so the `<select>` reflects the actual saved value instead of silently falling back to "Seed Bible's default".
- **No. 3 (Low)**: Left as-is — intentionally matches `branding.defaultTranslationId`'s apply-forward-only behavior, as you flagged.
- **No. 4 (Nit)**: The no-open-draft test for `updateEditingDefaultTranslationId()` now asserts `editingCustomization.value` is `null`, matching the sibling no-op tests.

Full suite (5735 tests), typecheck, and lint all pass on the updated head.

https://claude.ai/code/session_01TyQ1hZYDy2LtY1fjKVk8KD

---

Blitzy:
> [!NOTE]

Reviewed at head `320b4c86`. This PR has no linked issue, so I checked it against its stated goal (letting a Customization set its own default Bible translation). Blitzy's earlier review on this PR raised four findings (the test-precedence guard, the unavailable-translation `<select>` display, the apply-forward-only behavior, and a no-op test's trivial assertion) — all four are already fixed or explicitly accepted as intended in the follow-up commits, so I didn't re-raise those and focused on new ground. Two new issues below, both are real bugs I traced through the current code, and neither has a test that would catch it.

## Medium — The translation picker's "sort my language to the top" logic runs after the list is already cut down to one page

**File:** `packages/seed-bible/seed-bible/components/CustomizationEditPane/CustomizationEditPane.tsx` (~line 872-912)

**What's happening:** The comment right above this code says the point is that a viewer's own UI-language translations should sort to the top of the picker by default. But the order of operations is backwards. `filterTranslationGroups()` (in `managers/translationGrouping.ts`) sorts and then cuts the list down to `limit.value` (50 by default) *before* it hands the result back. Only after that does the component re-sort — but it's re-sorting `filtered.groups`, which is already the trimmed 50-item page, not the full catalog. If the viewer's language group didn't make it into that first 50, there's nothing left to promote — it's just gone from the picker, not merely buried further down. `translationGrouping.ts`'s own docstring notes "complete" mode has "a little over a hundred" language groups, so a language that alphabetizes into the second half is a common case, not an edge case.

**How to reproduce:** As a viewer whose UI language is one that sorts late alphabetically among Bible-translation languages (e.g. Korean `kor`, Chinese `zho`, Thai `tha`), with no translation currently selected (or a selected translation in an earlier-alphabet language), open the customization editor's translation picker in "complete" view mode. Expected: your own language's group appears first. Actual: it's missing entirely from the first page — you'd have to click "Load more" repeatedly to reach it, and even then it won't jump to the top since the re-sort already ran on the earlier page.

**Suggested fix:** Pass `priorityLanguages` into `filterTranslationGroups` and have it sort by priority *before* slicing to `limit`, rather than re-sorting the already-sliced output in the component. Worth adding a regression test with a priority language deliberately placed outside the default page to catch this — there's currently no test file for `CustomizationEditPane` at all.

## Low/Medium — Switching back to a customization after picking a translation manually doesn't re-apply its default

**File:** `packages/seed-bible/seed-bible/managers/TabsManager.tsx` (~line 1131-1163)

**What's happening:** `appliedCustomizationDefaultTranslationId` remembers the last translation id this effect applied, and skips re-applying it if `activeCustomizationDefaultTranslationId.value` comes back around to the same value. The comment explains this guard exists so that an unrelated draft edit (e.g. renaming the customization) doesn't clobber a translation the previewer picked manually in between — that part is fine. The gap: when `targetTranslationId` goes falsy (the customization becomes inactive), the code returns early at the `if (!targetTranslationId)` check without resetting the guard. So the guard also survives a full deactivate-then-reactivate cycle, which the comment doesn't say it's meant to cover.

**How to reproduce, in the customization editor's live preview:**
1. Activate customization A, whose `defaultTranslationId` is `NIV` — preview switches to NIV, guard is now set to `NIV`.
2. Manually pick a different translation in the reader, e.g. ESV.
3. Switch the preview away from A (to another customization, or none).
4. Switch back to A.

Expected: re-activating A re-applies its default (NIV), since the previewer explicitly chose to preview A again. Actual: the reader stays on ESV — the guard still thinks `NIV` was "already applied" from step 1 and skips.

**Suggested fix:** Reset `appliedCustomizationDefaultTranslationId = null` on the early return at the `if (!targetTranslationId)` check, so the guard only survives identity churn of the *same still-active* customization (renames, etc.), not a deactivate/reactivate cycle. `TabsManager.test.ts` has three tests around this effect but none exercise deactivate-then-reactivate — worth adding one that fails on current code and passes after the fix.


---

KallynGowdy:
Thanks for the second pass — both findings were real bugs, fixed:

- **Medium (picker sort-before-slice)**: `filterTranslationGroups()` (`managers/translationGrouping.ts`) now takes an optional `priorityLanguages` list and applies it in the same sort that runs *before* `limit` is sliced off, instead of the component re-sorting the already-paged result afterward. `CustomizationEditPane.tsx` now passes the UI language's Bible-language codes straight into that option rather than doing its own post-hoc re-sort. Added a regression test in `translationGrouping.test.ts` that places a priority language deliberately past a small `limit` — confirmed it fails on the pre-fix code (the language is missing from the page entirely, not just buried) and passes with the fix.
- **Low/Medium (guard survives deactivate/reactivate)**: `TabsManager.tsx`'s customization-default effect now resets `appliedCustomizationDefaultTranslationId` to `null` when `activeCustomizationDefaultTranslationId` goes falsy (the customization deactivates), not just when its default id changes. Added a regression test in `TabsManager.test.ts` that activates a customization, manually picks a different translation, deactivates, then reactivates the same customization — confirmed it times out waiting for the re-applied default on the pre-fix code and passes with the fix.

Full suite, typecheck, and lint all pass on the updated head (`33a15e35`).

https://claude.ai/code/session_01TyQ1hZYDy2LtY1fjKVk8KD

---

Blitzy (approved):

PR #1779, opened 2026-09-08
Add first-time tutorial for creating a customization

KallynGowdy:
Shows a guided walkthrough the first time a user creates a customization:
its name and logo fields, then a deep-linked tour into the new theme's
editor covering base theme, brand/surface/text/verse-selection colors,
fonts, and highlight colors.

Extends TutorialManager.startContextual() to accept an explicit steps
array (still tracked under the featureId's "seen" flag), since this
tour's theme-editor steps need per-instance state (the variant id) that
isn't available in the static CONTEXTUAL_TUTORIALS registry.

Claude-Session: https://claude.ai/code/session_01CxCrvi2CnGJbWnEV1FvM9v


Blitzy:
> [!NOTE]

## Summary

This PR adds a first-run guided tour that fires the first time a user creates a customization: it walks through the name and logo fields, then deep-links into the new theme's editor to explain each section (base theme, brand/surface/text/selection colors, fonts, highlights). The mechanism is a new optional `steps` override on `TutorialManager.startContextual()`, so a contextual tour can carry per-instance state (the variant id) that the static `CONTEXTUAL_TUTORIALS` registry can't.

Overall this is a clean, well-commented change. The core plumbing is sound: `startContextual`'s override is backward-compatible, the "seen" flag is still tracked per feature, the overlay's 150ms polling loop means the view-switch → DOM-render delay when entering the theme editor resolves gracefully (the spotlight just locks on a beat later), and the 22 new `en.json` keys exactly match the 22 keys referenced by the steps (only `en.json` touched, per repo convention). No blocking correctness bugs were found. The findings below are mostly robustness, coverage, and one thing worth verifying on mobile.

The findings are ordered by severity.

---

### Medium

**No. 1 — Verify the tour is actually visible on mobile (missing `elevated`).**
None of the new steps set `elevated: true`. The `.sb-tour-overlay` sits at `z-index: 1200` (`Tutorial.css`), and its raised variant `.sb-tour-overlay-raised` at `10001`. The only other contextual tour that fires from within the settings surface — `CONTEXTUAL_TUTORIALS["mobile-settings"]` — deliberately sets `elevated: true` precisely because the mobile settings sheet renders at `z-index: 9998/9999` (`MobileSettingsSheet.css`), which would otherwise hide the overlay.

This new tour is launched from `CustomizationsSettingsView.handleCreate`, i.e. from inside the settings UI, and then deep-links into an `openPane({ placement: "side" })` pane. On desktop this is fine — side panes top out around `z-index: 50` (`PaneLayout.css`), well below `1200`. On mobile I could not confirm statically whether the customization pane (or a still-open settings sheet behind it) sits above `1200`; if it does, the spotlight and popover will render behind it and the tour will appear broken or blank.

*Reproduction (to confirm):* On a ≤480px viewport, sign in, open Settings → Customize, tap "Create". Watch whether the coachmark spotlight/popover appear on top of the name field, or are hidden behind the pane/sheet.
*Suggested fix:* If the tour is intended to run on mobile, set `elevated: true` on the steps (or gate the tour to non-mobile). If it renders fine already, no change needed — this is a "please verify," not a confirmed defect.

**No. 2 — The four color-group steps are hardcoded and can silently drift from `CUSTOMIZATION_COLOR_GROUPS`.**
`colorGroupSteps` writes out one step per group as string literals for `brand`, `surfaces`, `text`, `selection`. Today those exactly match `CUSTOMIZATION_COLOR_GROUPS` (`CustomizationsManager.tsx:398`), and the code comment explains the literals are required so the i18n lint rule can statically read the keys — that reasoning is legitimate. The risk is drift: if someone later adds a fifth group or renames a group `id`, the tour will silently skip the new section (or point `data-tutorial="theme-group-<newid>"` at nothing) with no failing test to catch it.
*Suggested fix:* Add a unit test asserting the tour's `theme-group-*` targets cover exactly `CUSTOMIZATION_COLOR_GROUPS.map((g) => g.id)`. That keeps the required static literals but turns a future silent omission into a red test.

---

### Low

**No. 3 — Test isolation: the module-level `customizationEditView` signal is never reset between tests.**
In `CustomizationEditPane.test.tsx`, calling a step's `onEnter`/`onLeave` mutates the module-scoped `customizationEditView` signal, which persists across tests in the file. There's no `beforeEach` resetting it. The suite currently passes, but partly by ordering luck — each test happens to seed `editingVariantId` such that the guard resolves correctly regardless of the leaked view value. This is fragile; a reordering or a new test inserted between them could flip a result.
*Suggested fix:* Reset `customizationEditView.value = "edit"` in a `beforeEach` (exporting it or a small reset helper), so each test starts from a known view.

**No. 4 — Coverage gap: nothing tests that the tour is actually wired into creation, or that exit restores the main view.**
The builder is well-covered, but two behaviors that a user actually experiences aren't asserted:
- That `handleCreate` in `SettingsPage.tsx` calls `startContextual("customization-created", …)` after `create()`. A refactor could drop that line and every test would still pass.
- The "restore the main editor on exit" behavior. `it("defines onLeave for every theme-editor step …")` only checks `onLeave` is *defined*, not that invoking it sets the view back to `"edit"`. Since `finish()` relies on the last step's `onLeave` running to snap the pane back (see No. 5), a test that calls `onLeave()` and asserts `customizationEditView.value === "edit"` would protect the real contract.

**No. 5 — Nit / UX: finishing the tour snaps the user out of the theme editor.**
When the user clicks "Done" (or "Skip") on the final "Highlight colors" step, `finish()` sets `running=false`, and the onEnter/onLeave effect runs the last step's `leaveVariantEditor`, resetting `customizationEditView` to `"edit"`. So the moment the tour ends, the theme editor the user was just shown closes and they're bounced back to the customization overview. This may be intended (returns them to the top level), but leaving them inside the editor they just learned about could be friendlier. Worth a conscious decision either way.

**No. 6 — Info: live preview is a no-op during this tour.**
`openCustomizationVariant` calls `selectActiveVariant(variantId)`, but `selectActiveVariant` returns early when there's no `activeCustomizationLocator` (`CustomizationsManager.tsx:1511`). A just-created customization generally isn't the active one, so during this tour the theme editor won't live-preview edits. This matches the pre-existing behavior the original `openVariant` comment described, and the tour doesn't make edits, so it's informational — no action needed.

---

### Tests — real-world behavior assessment

Per the request to scrutinize the tests: the two `TutorialManager` tests are genuine regression tests, not always-green. "runs the steps passed in…" would fail on the pre-fix code because `startContextual("a-brand-new-feature")` with an unregistered feature id would no-op (never set `running=true`). "still tracks the seen flag…" exercises the real replay-suppression path. The `buildCustomizationTutorialSteps` tests also bite: the "does not re-select … already inside the same group" test genuinely fails if the `enterVariantEditor` idempotency guard is removed (it would call `selectActiveVariant` again), and the "deep-links…" test fails if entry stops selecting. So the guard logic is meaningfully covered. The gaps are the wiring/exit-restore behaviors in No. 4 and the isolation smell in No. 3, not false-green assertions.

https://claude.ai/code/session_01X6k8CS9qJjDPhxymDn5cGj

---

PR #1789, opened 2026-09-09
Next line + Ssave through keyboard in annotation

### LINKED ISSUE #1774: Add keyboard shortcuts: Cmd/Ctrl+Enter to save annotation, Enter to insert new line
Currently, there's no keyboard shortcut to save an annotation while typing. Adding a save shortcut plus proper newline behavior would improve the annotation workflow for users who type multi-line annotations.

Proposed Behavior
Cmd+Enter (Mac) / Ctrl+Enter (Windows/Linux): Save the current annotation while the text field is focused, without needing to click a separate Save button.
Enter (alone): Insert a new line within the annotation text field (standard textarea behavior), rather than submitting/saving or doing nothing.
Motivation

Annotators need to write multi-line notes/annotations without accidentally triggering a save or having Enter do nothing. Meanwhile, a quick modifier-based shortcut (Cmd/Ctrl+Enter) lets them save without reaching for the mouse.

Steps to Reproduce (current behavior)
Open an item for annotation.
Click into the annotation text field and start typing.
Press Enter.
Observe current behavior — (e.g., nothing happens / field submits / focus is lost — please confirm actual behavior).
Expected Behavior
Pressing Enter while typing inserts a new line in the text field, just like a normal textarea.
Pressing Cmd+Enter (Mac) or Ctrl+Enter (Windows/Linux) saves the annotation.
Save action should give some visual confirmation (e.g., brief toast, checkmark, or button state change).
Additional Notes
Ensure OS detection correctly maps Cmd (Mac) vs Ctrl (Windows/Linux) to the same save action.
Confirm the text field is a <textarea> (or equivalent) that supports multi-line input by default — if it's currently a single-line <input>, that may need to change first.

Kabi4:
Fixes: #1774


Blitzy:
> [!NOTE]

## Code Review — PR No. 1789 (Fixes No. 1774)

**Summary:** This is a clean, well-scoped change that does exactly what issue No. 1774 asks for, and in the way the maintainers settled on in the issue thread (KallynGowdy: on desktop, plain `Enter` should insert a newline like mobile, and a modifier shortcut saves). `Enter` keeps TipTap's default newline behavior, `Cmd/Ctrl+Enter` saves, a toast confirms the save, and the Save button advertises the shortcut via `title` + `aria-keyshortcuts`. The wiring is threaded through cleanly (`onModEnter` is optional, so the other TipTap consumer — `TextItemInput` — is unaffected), `pnpm check:ts` is clean, and all 14 tests pass.

The tests are genuinely testing real behavior, not just passing by construction:
- `TipTapEditor.test.tsx` drives the **real** TipTap editor with real `keydown` events and asserts on the actual document (`Enter` → `<p>Hello</p><p></p>`, `Mod+Enter` → handler fires and the document is unchanged). I confirmed this test fails if the `Mod-Enter` handler is neutered, so it is not vacuous.
- `CreateAnnotationForm.test.tsx` mocks TipTap only at its lazy-loaded module boundary (consistent with the existing `TextItemInput.test.tsx` approach) and asserts observable behavior: toast on success, no toast + error message on failure, no save while empty, and the platform-specific shortcut label.

Findings below, most severe first.

---

### 🔴 Blocking — Merge conflict with `develop`

GitHub reports this PR as `mergeable_state: dirty`. Merging `develop` into the branch conflicts in two files:
- **`CHANGELOG.md`** — both this branch and a merged PR (No. 1773) added a new first bullet under `### ✨ Added`, so they collide.
- **`packages/seed-bible/seed-bible/i18n/en.json`** — also changed on both sides (this one likely auto-resolves, since the repo configures an `i18n-json` git merge driver, but `CHANGELOG.md` will not).

**Reproduction:**
```
git fetch origin develop feature/1774
git checkout feature/1774
git merge origin/develop      # conflicts in CHANGELOG.md (and en.json)
```

**Suggested fix:** Merge (or rebase onto) latest `develop` and resolve `CHANGELOG.md` by keeping both bullets under `### ✨ Added`. While you're there, see the CHANGELOG nit below.

---

### 🟡 Low — Keyboard save can fire twice before the in-flight guard engages

`doSave` guards against re-entry with `if (saving || editorEmpty) return`, but `setSaving(true)` runs only *after* the `await sanitize(...)`:

```ts
const doSave = async () => {
  if (saving || editorEmpty) return;   // reads React state (stale between rapid presses)
  const editor = editorRef.current;
  if (!editor || editor.isEmpty) return;
  const html = await sanitize(editor.getHTML());  // <-- async gap
  ...
  setSaving(true);
  ...
};
```

The Save **button** is protected by `disabled={saving || editorEmpty}`, but the new keyboard path bypasses the disabled attribute. Two `Mod+Enter` events dispatched inside that `await` window (a fast double-tap, or auto-repeat from a held keystroke) both pass the `saving` check — because the state update from the first hasn't re-rendered yet — and both call `annotations.saveEditingAnnotation()`. The manager's save is an upsert so data corruption is unlikely, but it can double-fire the network call and the "Annotation saved" toast.

**Reproduction (manual):** open an annotation, type, then press `Cmd/Ctrl+Enter` twice in quick succession (or hold it) — depending on timing, `saveEditingAnnotation` runs twice.

**Suggested fix:** gate on a ref that flips synchronously before the first `await`, e.g.
```ts
const savingRef = useRef(false);
const doSave = async () => {
  if (savingRef.current || editorEmpty) return;
  const editor = editorRef.current;
  if (!editor || editor.isEmpty) return;
  savingRef.current = true;
  try { ... } finally { savingRef.current = false; }
};
```
A regression test could assert `saveEditingAnnotation` is called once when `onModEnter` is invoked twice synchronously.

---

### 🟡 Low — `navigator.platform` is deprecated (affects the label only)

`isApplePlatform()` uses `navigator.platform`, which is deprecated. This only drives the cosmetic hint (`title` and `aria-keyshortcuts`) — the actual shortcut is handled by TipTap's `Mod` binding, which does its own Mac detection, so the *functional* behavior is unaffected. Two small consequences worth noting:
- The label and the real binding are derived from two independent platform checks, so they could theoretically disagree (in practice both key off "Mac", so they stay consistent).
- On iPadOS, `navigator.platform` reports `"MacIntel"`, so the hint shows `⌘Enter` there too — which happens to match TipTap's behavior, so it's fine, just worth being aware of.

**Suggested fix (optional):** prefer `navigator.userAgentData?.platform` with a `navigator.platform` fallback, or leave as-is and treat this as a known minor limitation. No existing shared helper exists in the codebase for this, so a local helper is reasonable.

---

### 🔵 Nit — CHANGELOG entry doesn't cite the PR

Every other entry in `CHANGELOG.md` ends with a PR reference link, e.g. `([#1773](https://github.com/HelloAOLab/seed-bible/pull/1773))`. The new bullet omits it:
```
- Save an annotation with Cmd+Enter on Mac or Ctrl+Enter on Windows and Linux while typing, and keep Enter inserting a new line.
```
**Suggested fix:** append `([#1789](https://github.com/HelloAOLab/seed-bible/pull/1789))` (fold this into the conflict resolution above).

---

### 🔵 Nit — Empty-editor keyboard test could false-pass if the prop were dropped

In `CreateAnnotationForm.test.tsx`, the "does not save while the editor is empty" test calls `latestOnModEnter?.()` with optional chaining but never asserts the handler was actually wired. If `onModEnter` ever stopped being passed, this call would become a silent no-op and the test would still pass. The sibling "Cmd/Ctrl+Enter saves…" test already guards this with `expect(latestOnModEnter).toBeTypeOf("function")` — adding the same assertion here would make the empty-case test robust to a regression in the wiring.

---

**Overall:** aside from the merge conflict (which blocks merge), the remaining items are low-severity. Nice work on the real-editor keyboard test in particular.

---

KallynGowdy (approved):

PR #1790, opened 2026-09-09
Feature/1265

### LINKED ISSUE #1265: AI Chat should use current translation in the active tab and be able to recommend or switch translations
AI in the chat should use the active translation of the most recent tab, let a user pick a default translation (which would override) or offer translation options for the user to pick from and change it for their active tab.

Kabi4:
Fixes: #1265


Kabi4:
Please merge this after #https://github.com/HelloAOLab/seed-bible/pull/1789/

So commit don't overlap

Blitzy:
> [!NOTE]

## Code review — PR #1790 (Feature/1265: AI Chat uses the active translation)

Thanks @Kabi4 — this is a thorough piece of work: pure, well-factored resolver functions, a genuinely nice test for the modal/chat-dismiss race, and the annotation keyboard-shortcut tests exercise real ProseMirror key handling rather than mocks. Tests and `pnpm check:ts` both pass locally.

That said, there is one **high-severity correctness bug** that defeats the feature's main goal for a large class of translations, and the unit tests hide it because they don't use real translation-id formats. Details below, most severe first.

---

### 🔴 High — Translation mapping keys off `id`, but real translation IDs are often `lang_code` (e.g. `eng_kjv`), so supported translations are wrongly treated as "not available"

`resolveApologistBible` / `mapSeedTranslationToApologist` (`packages/apologist-extension/ext_Apologist/main/apologistBible.ts:57-124`) look the requested value up in `SEED_TO_APOLOGIST_BIBLE`, whose keys are bare short-codes (`BSB`, `KJV`, `KJAV`, `ESV`, `NASB1995`, …). But the value passed in is `readingState.translationId.value`, which is the raw `Translation.id` from the HelloAO Free Use API — and those IDs are frequently in `lang_code` form: `eng_kjv`, `eng_esv`, `eng_kjav`, `eng_nasb95`, `eng_web`, `hin_cvb`, etc.

**Before/after in plain terms:** a user reading the KJV should have the AI quote KJV (Apologist supports `kjv`). Instead, because the id is `eng_kjv` and no map key matches, the code decides KJV "isn't available", silently downgrades the AI to **BSB**, and pops the "Translation not available for AI" modal — even though KJV is fully supported. The same happens for ESV, NASB, WEB and essentially every eBible-sourced translation whose id carries a language/source prefix. So for a large share of the catalog the feature does the opposite of what #1265 asks.

This is not a hypothetical about the id format — the repo already documents it. `packages/compare-extension/ext_Compare/compareState.ts:188-197` deliberately matches on `shortName` **"rather than `id`, since `id`'s format varies by source … but `shortName` is the stable, recognizable abbreviation,"** and `script/lib/sitemap.ts:196` notes IDs are *"usually simple (`BSB`, `eng_kjv`)"* — i.e. both forms occur. This PR takes the fragile path the rest of the codebase avoids.

**Reproduction** (drop into `test/unit/apologist-extension/`, run `pnpm vitest run`):

```ts
import { resolveApologistBible } from "@packages/apologist-extension/ext_Apologist/main/apologistBible";
// catalog entry for KJV as the real API returns it: id "eng_kjv", shortName "KJV"
const catalog = [
  { id: "eng_kjv", language: "eng", shortName: "KJV", name: "KJV", englishName: "KJV",
    website: "", licenseUrl: "", textDirection: "ltr", availableFormats: ["json"],
    listOfBooksApiLink: "", numberOfBooks: 66, totalNumberOfChapters: 1189, totalNumberOfVerses: 31102 },
];
resolveApologistBible({ seedTranslationId: "eng_kjv", catalog });
// EXPECTED: { bible: "kjv", usedFallback: false, reason: "mapped" }
// ACTUAL:   { bible: "bsb", usedFallback: true,  reason: "default-english" }  → false "not available" warning, AI downgraded to BSB
```

I ran exactly this and it fails as shown.

**Suggested fix:** resolve against `shortName` (and/or strip a `xxx_` language prefix from the id), the way `compareState.ts` does, instead of matching the raw id:
- Pass the `Translation` object (or its `shortName`) into `mapSeedTranslationToApologist`, and key `SEED_TO_APOLOGIST_BIBLE` on the normalized `shortName`.
- As a belt-and-suspenders, also try the id with any `^[a-z]{2,3}_` prefix removed (`eng_kjv` → `kjv`) before giving up.
- Note that real short names don't all match the current keys either — the catalog uses `NASB95` (not `NASB1995`) and `KJAV`, so the table needs to be aligned to the actual `shortName` values, not guessed.

---

### 🟠 Medium — The unit tests pass but don't test the real-world id format, so they mask the bug above

`test/unit/apologist-extension/apologistBible.test.ts` only feeds bare-code ids (`BSB`, `KJAV`, `AAB`) that happen to match the hardcoded keys, so "uses a direct mapping without fallback" is green. The only realistic-format id used, `hin_cvb`, is exercised solely on the negative/fallback path. Had the "direct mapping" test used `eng_bsb`/`eng_kjv` (the form the live API returns), it would have failed and surfaced the defect. Per the repo's testing guidance ("test real-world behavior… a regression test that passes either way isn't testing anything"), please add cases that use `eng_*` ids paired with their `shortName`, asserting `reason: "mapped"` / `usedFallback: false` for translations Apologist supports.

### 🟠 Medium — `APOLOGIST_SUPPORTED_BIBLES` is a hardcoded, unverified allow-list; the inverse failure sends an unsupported code with no fallback

`apologistBible.ts:34-51` hardcodes a broad set including several copyrighted translations (`esv`, `niv`, `nkjv`, `nlt`, `csb`, `nasb`, `nasb1995`, `lsb`, …). The code's own comment admits *"Agent-specific catalogs can differ; keep this list aligned with what Seed's Apologist agent actually exposes."* If Apologist's agent does **not** actually support one of these, `resolveApologistBible` treats it as a valid direct mapping, sends it as `metadata.bible`, and neither falls back nor warns — the opposite failure to the one above, and silent. The linked issue explicitly anticipated this ("Apologist only supports some of the translations we support … we should have fallback logic"). Recommend verifying the list against the real Apologist agent (ideally fetching its supported set rather than hardcoding), and handling the case where the API rejects a `bible` code.

### 🟡 Low — Response language is collapsed to the primary subtag, losing Traditional/Simplified and regional distinctions

Both `resolveApologistLanguage` (`apologistBible.ts:304-312`) and the Bonfire path (`bonfire.tsx`) reduce the UI locale to its first subtag (`zh-TW` → `zh`, `pt-BR` → `pt`), yet the injected instruction claims a *"BCP-47 language code."* For Chinese and Portuguese users this drops a meaningful distinction (Traditional vs Simplified; Brazilian vs European). Where the value goes into a free-text instruction (Bonfire's `custom_instructions`), consider passing the full tag; only reduce for `metadata.language` if Apologist requires a bare code.

### 🟡 Low — CHANGELOG omits the headline feature

`CHANGELOG.md` only lists the annotation Cmd/Ctrl+Enter shortcut. The AI-translation feature — the actual subject of #1265 — isn't mentioned. Please add an "Added" entry for it.

### 🟡 Low — Unrelated annotation changes are bundled into this PR

The Cmd/Ctrl+Enter annotation save, the `savingRef` double-save guard, and the save toast (`CreateAnnotationForm.tsx`, `TipTapEditor.tsx`, plus their tests) are unrelated to #1265. They look solid and are well-tested, but bundling them makes this PR harder to review and revert independently. Consider splitting them out. (No change requested to the annotation code itself — it reads well.)

### 🟡 Low — `ChatsManager` persistence test covers only the logged-out path

`ChatsManager.test.ts`'s new test drives the `localConfig` (not-logged-in) branch of `setAiBibleTranslationId`. The logged-in path — where `getProfileConfigValue(profile.config)` should take precedence over `localConfig` — is untested. Worth a case with a populated `profile.config` to lock in that precedence.

### ⚙️ Informational — PR is currently in conflict with `develop`

GitHub reports the branch as un-mergeable (`mergeable_state: dirty`); it'll need a merge/rebase of `develop` before it can land.

---

**Nice touches worth keeping:** the `.sb-footnote-modal-overlay` guard in `FloatingReaderPanels.tsx` correctly matches how `ModalHost` renders managed modals, and `pauseChatWhileModalOpen` with its once-only settle guard is a clean solution to the modal/outside-dismiss race — the test that flaps the signal after settle is a good regression guard.

The high-severity item is the blocker: as written, users on KJV/ESV/NASB/etc. get the wrong Bible sent to the AI plus a misleading warning. Fixing the id→shortName matching (and adding tests in the real id format) should resolve it.

---

PR #1791, opened 2026-09-09
Feature/house of the lord bidirectional communication

### LINKED ISSUE #1698: House of the Lord: Camera framing is hard-coded for all pieces — each piece should define its own
## Summary

When a piece is selected, the camera moves to a fixed position that is identical for all 19 pieces. Because the pieces vary in size by roughly 30×, one framing cannot suit them all: small furniture is framed reasonably, while large pieces are either cut off or fill the entire viewport.

## Current behaviour

`patterns/house-of-the-lord/house-of-the-lord/infrastructure/adapters/pieces/PieceHighlightAdapter.tsx:138`

```ts
os.focusOn(bot, {
  duration: 1,
  easing,
  rotation: { x: 1.01229, y: 0.5 },
  zoom: 40,
});
```

Every piece receives this same distance and angle. Piece scales, for comparison:

| Piece | `scale` |
|---|---|
| fence, ground | 30 |
| grey / brown / red / purple curtain | 10.5 – 11.7 |
| walls, bars, rings | 9.35 – 9.5 |
| altar of sacrifice | 4.55 |
| front / inner curtain, front / inner pillars | 3.03 – 3.06 |
| table of showbread, ark, menorah, incense altar | 1.1 – 1.5 |
| bronze laver | (unset, defaults to 1) |

## Observed in walkthrough

- **Fence** (EXO 27:9) — framed at ground level from a tabletop distance, the perimeter sits outside the viewport. The pane reads as an empty scene; it takes moving the camera to realise the fence is present and simply out of frame.
- **Red curtain** (EXO 26:14) and **brown curtain** (EXO 26:7) — fill the viewport edge to edge, so the piece's shape and its relationship to the rest of the tabernacle are not visible.

There is a knock-on effect: ao.bot ignores camera drags while the pointer is over a mesh. When a curtain fills the viewport there is no empty area left to drag from, so the camera cannot be moved without first scrolling to an edge. Correct framing removes this.

## Proposed change

Add a per-piece camera config, following the convention already used three times in this pattern — `positionsMap`, `hitboxMap` and `layersMap` each pair a per-piece config map with a Provider class under `infrastructure/config/`.

```ts
// infrastructure/config/camera/cameraMap.tsx
export interface CameraFraming {
  zoom: number;
  rotation?: { x: number; y: number };
}

export type CameraMap = {
  [E in ExperienceKey]: {
    [K in ExperienceKeyMap[E]]?: CameraFraming;
  };
};
```

Carry `rotation` as well as `zoom`. The fence needs a higher, wider angle rather than only a greater distance, and the curtains likely want a different approach vector than the tabletop furniture. Keep the current values as the fallback when a piece has no entry, so unconfigured pieces behave exactly as they do now.

## Files

- **New:** `infrastructure/config/camera/cameraMap.tsx`, `infrastructure/config/camera/CameraConfigProvider.tsx`
- `infrastructure/adapters/pieces/PieceHighlightAdapter.tsx` — accept the provider as a constructor dependency; read framing per piece in `highlightPiece`
- `infrastructure/di/bootstrap.tsx` — instantiate and inject the provider

## Acceptance criteria

- [x] Each of the 15 menu-reachable pieces is fully in frame and legible when opened from the verse toolbar
- [x] Fence is visible in its entirety, recognizable as a perimeter
- [x] No piece fills the viewport to the point where camera drag has no empty region to start from
- [x] Pieces with no camera entry fall back to the current values

## Non-goals

Does not address the layer-occlusion behaviour that also affects fence — that is a separate design question for the team review session.

### LINKED ISSUE #1699: House of the Lord: Hitboxes should support multiple volumes per piece
## Summary

Each piece has exactly one hitbox, and a hitbox is a solid volume rather than a shell. For pieces that enclose other pieces — walls, curtains — that single volume also encloses everything inside the tent, so clicks aimed at the interior furniture are absorbed by the enclosing piece and the furniture cannot be selected.

## Current behaviour

Every piece carries `pointableDefault: false`, so the meshes themselves are never click targets. Selection happens entirely through hitboxes: invisible (`color: "clear"`) bots spawned once per piece at setup, parented to the piece via `transformer`.

`infrastructure/config/hitboxes/hitboxMap.tsx` — walls:

```ts
[TABERNACLE_PIECE_KEYS.WALLS]: {
  position: { x: 0, y: 0, z: -0.5 },
  scaleX: 1, scaleY: 0.3, scaleZ: 0.32,
},
```

The hitbox is a CasualOS cube spanning the walls' full footprint. Since the walls enclose the tent, that cube also contains the menorah, table of showbread, incense altar and ark. A ray cast from the camera toward any of them intersects the walls hitbox first.

The same applies to all four curtains, and to bars and rings.

## Observed in walkthrough

With the walls visible, clicking any interior item selects the walls instead. The only way to reach the furniture is to select an outer piece first so that its layer hides and exposes what is beneath — an outside-in traversal that works but is undiscoverable.

Separately: **fence and ground have no hitbox entries at all**. Combined with `pointableDefault: false` on every piece, this means neither can be selected in the scene under any circumstances.

## Proposed change

Allow a piece to define an array of hitbox volumes rather than a single one. Walls would define four volumes, one per wall, instead of one box enclosing the interior. The piece stays selectable from any side, and the tent interior becomes directly clickable through the gaps between them — removing the forced outside-in traversal without removing the option to work that way.

The click plumbing already keys off the piece rather than the box, so this needs no adapter change:

```ts
// HitboxLifecycleAdapter.spawn — creates one bot per call, tagged with the piece
os.addBotListener(hitboxBot, "onClick", () => {
  this.#eventManager.emit("OnHitboxClicked", hitboxBot.tags.pieceKey);
});
```

N volumes sharing a `pieceKey` all resolve to the same selection.

**Type change:**

```ts
// before
export type HitboxMap = {
  [E in ExperienceKey]: { [K in ExperienceKeyMap[E]]?: HitboxData };
};
getHitboxData<E>(experienceKey: E, pieceKey: ExperienceKeyMap[E]): HitboxData | null;

// after
export type HitboxMap = {
  [E in ExperienceKey]: { [K in ExperienceKeyMap[E]]?: HitboxData[] };
};
getHitboxData<E>(experienceKey: E, pieceKey: ExperienceKeyMap[E]): HitboxData[];
```

Returning an empty array rather than `null` lets the service drop its `if (!data) continue` guard in favour of a plain inner loop:

```ts
for (const piece of pieces) {
  for (const data of this.#hitboxProviderPort.getHitboxData(experience, piece.key)) {
    hitboxes.push(this.#hitboxSpawnerPort.spawn({ data, piece }));
  }
}
```

## Files

| File | Change |
|---|---|
| `infrastructure/config/hitboxes/hitboxMap.tsx` | Value type → `HitboxData[]`; migrate 17 existing entries as one-element arrays |
| `infrastructure/config/hitboxes/HitboxConfigProvider.tsx` | `getHitboxData` returns `HitboxData[]`, empty array when absent |
| `application/ports/out/hitboxLifecycle.tsx` | `HitboxProviderPort` signature follows |
| `application/services/HitboxLifecycleService.tsx` | Inner loop, one spawn per volume |
| `infrastructure/adapters/pieces/HitboxLifecycleAdapter.tsx` | No change |
| `test/unit/house-of-the-lord/pattern/application/services/HitboxLifecycleService.test.ts` | Mocks return arrays; add a multi-volume case |

## Scope of new definitions

Genuinely multi-volume: **walls**, **inner curtain**, **front curtain**, **brown curtain**, **red curtain**, **grey curtain**, **purple curtain**.

Stay single-volume: altar of sacrifice, ark, bronze laver, front pillars, inner pillars, incense altar, menorah, table of showbread, bars, rings.

New entry needed: **fence** — currently has none, and as a perimeter it is a natural composite. **Ground** also has none; whether it should be selectable is a separate question.

## Acceptance criteria

- [x] With walls and curtains visible, clicking the menorah, table of showbread, incense altar or ark selects that piece
- [x] Walls and each curtain remain selectable by clicking the piece itself from outside
- [x] Fence is selectable in the scene
- [x] The 10 pieces that remain single-volume behave exactly as before
- [x] `HitboxLifecycleService.test.ts` covers a piece with several volumes, asserting one spawn per volume and a single resolved selection

## Non-goals

Does not change the layer-occlusion behaviour that hides pieces above the selection — that stays as-is and is a separate design question.

### LINKED ISSUE #1700: Add a discover provider for the House of the Lord extension
Currently, the only way to discover the House of the Lord extension and its features is to select related verses and click the Tabernacle verse toolbar item. In addition to this, we should also have the House of the Lord extension register a discover provider that returns results for the chapters that have Tabernacle content. This would allow users to use the compact discovery panel and the discover panel to discover the Tabernacle interactive exhibit.

soukigabriel:
Fixes #1699
Fixes #1698
Fixes #1700


Blitzy:
> [!NOTE]

## Code Review — House of the Lord bidirectional communication

Thanks for this — it's a large, well-organized refactor (ports/adapters split, a domain event bus, a nav menu, and a two-way bridge between the reader host and the cross-origin pattern iframe). The layering is clean and the new message-narrowing helpers (`ToExperienceKey`/`ToPieceKeyOf`) are a nice touch for untrusted input. Most of my findings are about **test coverage of the new behavior** and one **cleanup bug**. There's no linked issue and the PR description is empty, so I reviewed against the code and commit history alone — if there's a spec for the bridge, linking it would help future reviewers.

---

### High

**No. 1 — `clearExperience` is registered as an unbound listener, so cleanup throws instead of running**

`infrastructure/di/bootstrap.tsx` (~lines 398–407):

```ts
os.addBotListener(entrypointBot, "onInstLeave", experienceService.clearExperience);
os.addBotListener(entrypointBot, "onDestroy",   experienceService.clearExperience);
```

`clearExperience` is a normal (non-arrow) async method that reads private fields (`this.#experience`, `this.#piecesSequencePort`, …). Passing it as a bare reference detaches it from its instance, so when the event fires `this` is not the service and the first private-field access throws a `TypeError`. Every other listener in this same file wraps the call in an arrow (e.g. `os.addBotListener(entrypointBot, "onGridClick", () => { environmentInteractionController.handleGridClick(); })`), which is why this pair stands out as an oversight.

**Effect:** on inst-leave / destroy the teardown never completes — the clear sequence doesn't run and pieces/hitboxes aren't torn down.

**Reproduction:** open the experience, then leave the inst or trigger destroy; observe a `TypeError` in the console and that `displayClearSequence`/`clearPieces` never run.

**Suggested fix:**
```ts
os.addBotListener(entrypointBot, "onInstLeave", () => experienceService.clearExperience());
os.addBotListener(entrypointBot, "onDestroy",   () => experienceService.clearExperience());
```
(or bind the method / make it an arrow property).

---

### Medium

**No. 2 — The headline "experience change" logic is untested and no test would fail if it were deleted**

`ExperienceService.tryDisplayExperience` gained real branching for switching experiences: the `#queuedExperience` guard, `await this.clearExperience()`, `tryAbortCurrentDropSequence()`, and `displayClearSequence()`. But every case in `ExperienceService.test.ts` calls `tryDisplayExperience(testKey)` with the **same single key**. `displayClearSequence` and `tryAbortCurrentDropSequence` are declared as mocks (lines 23–24) but never asserted, and the "already displaying" test now short-circuits on `experience === this.#experience` before any of the new code runs. In other words, the branches this PR is named after are never executed by the suite — deleting the queue/abort/clear code would keep the tests green.

Compounding this: only one experience exists (`EXPERIENCE_KEYS = { TABERNACLE }`), and `ScriptureInteractionController.handlePieceFocusRequest` only calls `tryDisplayExperience` in its `else` (different-experience) branch. So the switch path is also **unreachable in production today**. That's fine as forward-looking scaffolding, but it should be tested as such.

**Suggested fix:** add a test-only second experience fixture (or narrow the mock types) and cover: (a) switching A→B aborts A's drop, runs the clear sequence, then displays B; (b) a rapid A→B→C sequence lands on C, not B; (c) `OnExperienceChanged` is emitted with the right payloads. Assert on `displayClearSequence`/`tryAbortCurrentDropSequence` call order.

**No. 3 — New pure domain functions have no unit tests**

`domain/functions/keys.tsx` (`ToExperienceKey`, `ToPieceKeyOf`) and `domain/functions/verseRanges.tsx` (`ToVerseRanges`, `FormatVerseRange`) are new, pure, and easy to test — and `keys.tsx` validates values that arrive over `postMessage` from the reader, so its rejection paths matter. Yet `test/unit/.../domain/functions/` still contains only `colors.test.ts`, even though an earlier commit here was "Added unit tests for all functions and services at domain and application layers."

`ToVerseRanges` in particular encodes an implicit contract — it only merges a reference into the previous range when it's the same book/chapter and exactly `last.end + 1`. Unsorted, duplicated, or descending input silently produces fragmented or wrong ranges. A couple of tests would pin that contract (and document whether callers must pre-sort).

**Suggested tests:** `ToExperienceKey`/`ToPieceKeyOf` accept valid keys and reject non-strings, unknown strings, and keys from another experience; `ToVerseRanges` merges consecutive verses, keeps a gap as two ranges, and doesn't merge across chapter/book boundaries; `FormatVerseRange` renders single vs. range (`Exodus 25:1` vs `Exodus 25:1–5`).

**No. 4 — The new bridge services and controllers have no unit tests**

`NavMenuStateService`, `PieceFocusService`, `PieceHighlightService`, `ScriptureNavigationService`, and the controllers (`NavMenuController`, `ScriptureInteractionController`, `SeedBibleController`) are the actual glue of the bidirectional feature, and there's no `infrastructure` test dir at all. `ScriptureInteractionController.handlePieceFocusRequest` is a good, cheap target: it decides same-experience-focus vs. switch-experience and has two `console.warn` rejection paths — exactly the branching that's easy to break later. A small test with a mocked `experienceServicePort`/`scriptureInteractionPort` would lock in that routing.

---

### Low / Minor

**No. 5 — Dead no-op path: `handleExperienceShowRequest`**

`ScriptureInteractionService.handleExperienceShowRequest(): void {}` is empty, and `ScriptureInteractionController.handleExperienceShowRequest` forwards to it — but nothing is wired to call the controller method (there's no message type for it in `MESSAGE_TO_EVENT_MAP`). Either wire it up or drop it so it doesn't read as an implemented feature.

**No. 6 — Latent concurrency corruption when a second experience is eventually added (ties to No. 2)**

In `ExperienceService`, `#isDisplayingExperience` is only reset in `#displayExperience`'s `finally`, while `clearExperience` runs independently. Under rapid switching, an aborted display's `finally` can flip `#isDisplayingExperience` to `false` while a newer display is mid-flight, and a subsequently-queued call can then start a second concurrent `#displayExperience`. Not reachable with a single experience, but worth hardening (e.g. a single in-flight promise/generation guard) before a second experience ships. A test from No. 2 would surface it.

**No. 7 — Duplicated `@ts-expect-error os.sendEmbedMessage`**

`HostNotifierAdapter` and `ScriptureNavigationAdapter` each carry the same private `#send` with an identical `@ts-expect-error` about `sendEmbedMessage` living on `os` at runtime. Per the repo's "extract on the third copy" convention this is fine for now at two copies, but a single tiny typed `sendEmbedMessage` wrapper would keep the suppression in one place if a third sender appears.

---

Nothing here blocks the direction of the PR — No. 1 is a real bug worth fixing before merge, and the rest is about giving the new bridge/switch logic tests that actually exercise it (right now the suite would stay green even if the new branches were removed). Happy to dig deeper into any specific area.

---

KallynGowdy:
@soukigabriel this is really cool! I really like the menu and the ability to jump to verses. One thing I had to note: when I click on a verse from the menu, I notice that it highlights the verses instead of using the "diminish" decoration.

image

image

Ideally it should use the diminish decoration so it doesn't interfere with any highlights that the user may already have for those verses.

Blitzy:
> [!NOTE]

## Code Review

Reviewed the substantive logic changes under `patterns/house-of-the-lord/` and `packages/house-of-the-lord/` (services, adapters, controllers, DI wiring, presentation hooks) plus the deletions under `obsolete/`. This is a large PR (469 files, ~4800/~11500 lines), so the generated data-map files (hitboxMap, pieceCatalogMap, referenceMap) and the `.bot.aux` file were skimmed rather than read line-by-line.

### 1. High — Non-English translations deleted from `extension.json`

**File:** `packages/house-of-the-lord/extension.json` (around line 33)

The diff removes the translation blocks for every locale except `"en"` (roughly 76 locale blocks: `af`, `am`, `ar`, ... `zh`, `zu`), rather than only adding/updating the `en` entry as the project convention calls for.

**Repro:** Set the app locale to any previously-supported non-English locale (e.g. `ar`, `zh`, `af`) and open the House of the Lord discover card or the verse-toolbar tool. The title, description, and all 19 piece labels now fall back to the English `defaultValue` for every non-English user, because the corresponding locale blocks no longer exist anywhere in the PR.

**Suggested fix:** Restore the deleted locale blocks (or confirm with the localization team whether this was an intentional temporary removal pending re-translation). CLAUDE.md's "only update en.json" convention is about not hand-editing other locale files' translations — it isn't license to delete them outright.

### 2. Medium — Stale pane title/icon when reusing an open exhibit pane for a different experience

**File:** `packages/house-of-the-lord/bootstrap.tsx` (around line 93)

`openExhibit()` keeps a single `portalRef` shared across all experiences. When a pane is already open (`portalRef` set) and `openExhibit()` is called for a *different* experience, the code only sends a `highlight-piece` message — it never calls `context.panes.openPane` again, which is what sets the pane's title/icon from `meta.title`/`meta.icon`.

**Repro:** Open the Tabernacle experience (pane opens, title shows "Tabernacle"). While that pane is still open, trigger `openExhibit("solomon-temple", key)` — reachable through the same verse-toolbar/discover loop this PR adds (`for (const experience of Object.values(EXPERIENCE_KEYS))`) once Solomon's Temple has verse data. The iframe is asked to display Solomon's Temple content, but the pane still shows the "Tabernacle" title/icon.

**Suggested fix:** Track which experience the currently-open pane belongs to, and call `context.panes.openPane` again (updating title/icon) whenever the requested experience differs from the one the open pane is showing — not just when no pane is open at all.

### 3. Low — Unhelpful error message in `PieceAdapter.getCurrentState()`

**File:** `patterns/house-of-the-lord/house-of-the-lord/infrastructure/adapters/pieces/PieceAdapter.tsx` (around line 39)

The newly-added `getCurrentState()` reuses the same truncated error string as the pre-existing `setPosition()`: `"PieceAdapter: bot not found at "` with nothing appended — no piece key, bot id, or method name.

**Repro:** If `PiecesProvider` returns a `Piece` whose bot no longer exists (e.g. despawned mid-interaction) when `PieceInteractionService.handlePieceSelection` calls `pieceAdapterPort.getCurrentState(piece)`, the thrown error gives no way to tell which piece/bot triggered it, making it hard to diagnose from logs alone.

**Suggested fix:** Include the piece key/id and the calling method name in the thrown error, e.g. `` `PieceAdapter.getCurrentState: bot not found for piece "${piece.key}"` ``.

---

**On tests:** I did not find test coverage added in this PR for the `openExhibit()` pane-reuse path (finding No. 2) or for `PieceAdapter.getCurrentState()`'s not-found path (finding No. 3) — both would be good candidates for regression tests per the project's testing conventions (assert on observable behavior, cover error paths, and confirm the test fails on the pre-fix code).


---

PR #1792, opened 2026-09-09
Chat Translation Support

### LINKED ISSUE #1265: AI Chat should use current translation in the active tab and be able to recommend or switch translations

Kabi4:
Fixes: #1265


Blitzy:
> [!NOTE]

## Chat Translation Support — review of PR #1792 (Fixes #1265)

Thanks for this — it's a well-structured feature. The Apologist bible-mapping module is cleanly factored (single source of truth for the canonical list, reject-and-retry in one place), the reject-fallback logic is tested against real HTTP shapes, and the `.sb-footnote-modal-overlay` guard in `FloatingChatPanel` correctly matches the class `ModalHost` actually renders. Below are the issues I found, most severe first.

---

### Medium

**1. Clearing the AI-default translation silently fails when the device-local config still holds a value.**

`readAiBibleTranslationId` in `ChatsManager.tsx` resolves the preference as:

```ts
const value = fromProfile ?? fromLocal;
```

Because it uses `??`, a profile value of **explicit `null`** (which is what "clear the override" writes when signed in) is treated the same as "absent," so it falls through to the device-local value. `setAiBibleTranslationId(null)` only writes `null` to `profile.config` — it never clears `login.localConfig` — so if `localConfig.aiBibleTranslationId` still holds an old id, the override snaps right back.

This is not just a reload issue: the `effect(() => { aiBibleTranslationId.value = readAiBibleTranslationId(loginManager); })` re-runs the moment `setAiBibleTranslationId` updates the profile, so the in-memory signal is immediately overwritten and the "Follow active tab instead" button appears to do nothing (or switches to the stale translation).

Realistic path to a populated `localConfig` while signed in: pin an AI default **while signed out** (writes to `localConfig`), then sign in (LoginManager adopts local config into `profile.config`), then click "Follow active tab instead."

Reproduction (I confirmed this against the real `createChatsManager`, matching the mock harness in `ChatsManager.test.ts`):

```ts
userId.value = "user-1";
localConfig.value = { aiBibleTranslationId: "eng_web" };
profile.value = { name: "A", config: { aiBibleTranslationId: "eng_esv" } };

const chats = createChatsManager(login, mockI18nManager);
expect(chats.aiBibleTranslationId.value).toBe("eng_esv"); // ok
chats.setAiBibleTranslationId(null);
expect(chats.aiBibleTranslationId.value).toBeNull();       // FAILS -> "eng_web"
```

Result: `AssertionError: expected 'eng_web' to be null`.

Suggested fix — make "signed in, profile loaded" authoritative so an explicit `null` wins, e.g.:

```ts
function readAiBibleTranslationId(login: LoginManager): string | null {
  const profileConfig = login.profile.value?.config as
    | Record<string, unknown>
    | undefined;
  // When signed in with a loaded profile, the profile is authoritative —
  // an explicit clear must not fall back to a stale device-local value.
  if (login.userId.value && profileConfig) {
    const value = profileConfig[PROFILE_AI_BIBLE_TRANSLATION_ID];
    return typeof value === "string" && value.length > 0 ? value : null;
  }
  const fromLocal = login.localConfig?.value?.[PROFILE_AI_BIBLE_TRANSLATION_ID];
  return typeof fromLocal === "string" && fromLocal.length > 0 ? fromLocal : null;
}
```

Either way, please add a regression test covering "signed in + stale localConfig + clear" — the two current tests don't exercise it (the signed-out test happens to pass only because clearing also nulls `localConfig`, and the signed-in test never clears).

---

### Low

**2. The "prefer a same-language translation before English BSB" fallback is largely inert for non-English readers, and its passing test relies on contrived data.**

`findSameLanguageSupportedCode` filters the Seed catalog by Seed `language`, but the codes it can match (`APOLOGIST_CANONICAL_BIBLES`) are all English translations. A Spanish translation only maps when its `shortName`/`id` happens to lowercase onto an English code — at which point you'd be sending an English translation (e.g. `niv`) while labeling it "same language." The test at `apologistBible.test.ts:102` only passes because `spa_nvi` is given `shortName: "NIV"`, which collides with the English NIV code; a realistically-named Spanish translation would return `null` and fall to `bsb`.

This isn't a crash, but the feature does less than the name implies. Consider either (a) verifying the matched Apologist code is actually in the requested language before preferring it, or (b) dropping the same-language step and documenting that non-English falls straight to BSB, so the test reflects real catalog data rather than a shortName collision.

**3. The new `AiBibleTranslationControl` UI has no component-level test.**

`ChatView.test.tsx` only extends the shared mock state; there's no test that renders the control and exercises opening the menu, "Use current as AI default," "Follow active tab instead," or "Change active tab translation" (the picker wiring through `state.selector.setOpen`). The persistence logic is covered in `ChatsManager.test.ts`, but the component wiring — including the `!followingTab` conditional that hides "Follow active tab instead" — is untested. A small render + click test would guard against regressions in the menu behavior.

---

### Nits

**4. Duplicated test blocks in `apologistBible.test.ts`.** `describe("isLikelyUnsupportedApologistBibleError", ...)` appears twice (lines ~269 and ~374) and `describe("postApologistChatCompletion", ...)` appears twice (lines ~295 and ~397), asserting nearly the same things. Worth collapsing so the intent is clear and the file doesn't drift.

**5. `MAX_COMPLETION_TURNS` exhaustion yields nothing.** In `init.tsx`, if the model keeps emitting tool calls (or never final content) for 25 turns, `generateResponse` falls out of the loop and returns without yielding any `text`. The user would see tool-call chips but no answer. A trailing fallback message (or a logged warning) would make that failure mode visible rather than silent.

**6. `aria-expanded={isMenuOpen.value}`** renders the string `"true"`/`"false"`, which is what ARIA wants — just confirming it's intentional; no change needed.

---

### Tests — real-world behavior check

Per the review focus on tests that could "always succeed": the Apologist mapping/retry tests assert against real `Response`/status/body shapes and the streaming SSE default-retry, so they're meaningful. The `FloatingReaderPanels` test manufactures a `.sb-footnote-modal-overlay` node rather than opening a real modal, but since that class matches what `ModalHost` renders, it's a valid regression (it would fail on the pre-guard handler). The main gap is issue No. 1's uncovered clear-while-signed-in path and issue No. 3's missing component test — both are places where the current tests pass without exercising the risky behavior.

---

KallynGowdy:
@Kabi4 I have some notes:

1. We don't need special controls for the translation selection. It should always follow the user's tab by default.
image

2. We should automatically fallback to a supported translation listed in the Apologist fusion docs instead of potentially sending an invalid translation. This means that we also don't need the ability to display warning messages or need feature flags to gate them.

Essentially, what we're looking for is that it supports KVJ, BSB, etc. for now. It doesn't need to support every possible translation, it just needs to have a mapping from our translations to the ones that are supported by Apologist and fallback to the BSB after that.

packages/apologist-extension/ext_Apologist/main/apologistBible.tsx
```diff
@@ -0,0 +1,291 @@
...
+);
+
+/**
+ * Seed shortNames / bare ids that do **not** lower-case onto a canonical code.
+ * Everything else (KJV→kjv, BSB→bsb, ESV→esv, …) is handled by lowercasing
+ * against {@link APOLOGIST_CANONICAL_BIBLES} — do not duplicate those here.
+ */
+const APOLOGIST_BIBLE_ALIASES: Record<string, ApologistCanonicalBible> = {
+  WEB: "webu",
+  KJAV: "kjv",
```
KallynGowdy, line 52: @Kabi4  We should also have mappings for `KJVA` and `KJVCP` to `kjv`

Blitzy:
@KallynGowdy before this goes any further I wanted to get your thoughts on direction, because as it stands this PR only covers part of what #1265 asks for.

The issue asks for three things:

1. AI chat uses the active tab's translation
2. A user can pick a default translation for the AI that overrides the tab
3. The AI can offer translation options that the user picks from to change their active tab

This PR does No. 1. No. 2 was in an earlier commit but got pulled out during review, and No. 3 was never built. The PR says `Fixes: #1265` though, so merging it as-is would close the issue with No. 3 still missing. I think No. 3 is really the heart of the issue, it's what lets someone reading the BSB ask in French "is there a French Bible I can read?" and switch to it in one tap.

### Either way — a `searchTranslations` tool

Regardless of which option we go with below, I think we should add a `searchTranslations` core tool that lets an agent look up what translations we actually have by language or name. Right now agents have no idea what's in our catalog so they guess, and they'll happily recommend an NIV we don't have. An agent that can call the tool can give the user real options. Agents without tool support (like Bonfire) wouldn't get this, but we could cover them by including the translations available in the user's language right in the prompt (French only has 4: JND, LSG, NCL, OST).

I see two ways we could build the actual picking part.

### Option 1 — Choice cards through agent tools

Add a `suggestTranslations` tool so when an agent suggests translations, a card of buttons shows up in the chat and tapping one switches the active tab while keeping you on the same chapter. The card could be a generic "choices" message so we can reuse it later for things like suggested passages.

The upside is the agent can only ever offer translations we actually have. The downside is it only works for agents that support tool calling (Apologist does, Bonfire doesn't), and it needs a new chat message type which older clients will silently hide in shared chats.

### Option 2 — Translation reference links

Detect translation codes in the message text the same way we already detect verse references, so "LSG" becomes a link that switches your active tab in-place. This one is nice because it doesn't care what agent is answering, tools or not, and it even works when a person types it in a shared chat.

The catch is the edge cases are real. I had Claude look through the 1,260 translations:

- `KJV` in our catalog is actually a **Thai** translation (`tha_kjv`), the English KJV is `KJAV`. So without an alias table, "KJV" would switch an English reader to Thai.
- 50 codes are shared by more than one translation. `WBT` alone is used by ~400, `IRV` by 12 different Indian language Bibles.
- Some codes are just words: `WEB`, `MAY`, `OLD`, `NEW`, `THE`, `MAN`, `AND`. Every code is uppercase so case-sensitive matching covers most of this.
- NIV, ESV and NLT aren't in the catalog, so when an agent recommends those they'd just stay plain text. Though for many we could provide aliases and map them to translations we do have that are part of their families. `searchTranslations` helps a lot here since agents that use it will name codes we actually have.
- `selectTranslation` jumps to the first book, so we'd need to keep the current chapter and have a fallback for NT-only translations.

Honestly I'm torn. I like Option 2 more because it's agnostic to the agent, but those edge cases will get us if it isn't built carefully and with proper test cases for each of them. Option 1 is safer but only helps agents with tools.

A few questions:

- Are you on board with adding `searchTranslations` either way?
- Which direction would you rather go for the picking part, or should we go with both?
- Do we want to build No. 3 in this PR, or do this as a follow up and keep this PR as is and just be sure not to close #1265

Blitzy (dismissed):
<removed comment>

KallynGowdy:
@Blitzy I agree we should add a `searchTranslations` tool. As far as the options, I think number 1 is probably the best for now. While number 2 could work, the support that Apologist has for various translations is strange because it is configurable per-agent and there is no easy way to search them. It could theoretically be using Free Use Bible API translation IDs on its own, but there's no real way for us to know at this point, so it is probably safter to just have a specific way for the AI to answer directly from our data instead of Apologist's.

Thinking through use-cases, we're unfortunately not going to be able to support seamless conversation with every translation that the Free Use Bible API supports unless we're able to pass FUBAPI IDs through. So this feature is mostly about having a natural language way to query the available translations and give the user options to select from for now

Blitzy:
@KallynGowdy Agreed that this `searchTranslations` for now is really about natural language translation searching / switching.

@Kabi4 Let's add the `searchTranslations` agent tool along with Option 1 (choice cards) as described above in this PR.

Issue #1120, opened 2026-04-08
Port the References extension to the refactor


KallynGowdy:
This task is to port the References extension to the Seed Bible refactor.

PR #1794, opened 2026-09-10
remade the reference extension

Bored-Wizard:
task #1120

### DIFF
--- a/packages/references-extension/extension.json
+++ b/packages/references-extension/extension.json
@@ -1,0 +1,30 @@
+{
+  "id": "ext_references",
+  "translations": {
+    "en": {
+      "title": "References",
+      "description": "An extension that shows references in bible verses.",
+      "loading-references": "Loading verse text",
+      "text-unavailable": "Verse text unavailable.",
+      "no-references": "No cross references for this verse.",
+      "show-more": "Show more references",
+      "open-in-new-pane": "Show this verse's own references",
+      "loading-list": "Loading references",
+      "show-verse-references": "Verse {{verse}} references",
+      "references-unavailable": "Could not load the references for this verse.",
+      "download-title": "Save cross-references for offline use?",
+      "download-body": "Keep every cross-reference on your device so they still open when you have no connection. The download carries on in the background.",
+      "download-body-with-translation": "Keep every cross-reference on your device, along with {{name}}, so they still open when you have no connection. The download carries on in the background.",
+      "download-size": "About {{references}} of cross-references.",
+      "download-size-with-translation": "About {{references}} of cross-references, plus {{translation}} for {{name}}.",
+      "download-confirm": "Download",
+      "download-not-now": "Not now",
+      "download-started": "Downloading in the background…",
+      "download-done": "Cross-references are now available offline",
+      "download-done-with-translation": "Cross-references and {{name}} are now available offline",
+      "download-partial": "Saved {{downloaded}} of {{total}} chapters. {{failed}} couldn't be downloaded.",
+      "download-translation-failed": "Saved the cross-references, but {{name}} couldn't be downloaded.",
+      "download-failed": "Couldn't download the cross-references. Check your connection and try again."
+    }
+  }
+}
--- a/packages/references-extension/index.ts
+++ b/packages/references-extension/index.ts
@@ -1,0 +1,1 @@
+export { default } from "./references/manager/init";
--- a/packages/references-extension/package.json
+++ b/packages/references-extension/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "@seed-bible/references",
+  "version": "1.0.0",
+  "description": "",
+  "type": "module",
+  "exports": {
+    ".": {
+      "import": "./index.ts"
+    }
+  },
+  "dependencies": {
+    "seed-bible": "workspace:*"
+  }
+}
--- a/packages/references-extension/references/manager/askToDownload.css
+++ b/packages/references-extension/references/manager/askToDownload.css
@@ -1,0 +1,84 @@
+.sb-references-download {
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  color: var(--sb-font-color);
+  font-family: var(--sb-font-family);
+}
+
+.sb-references-download-header {
+  display: flex;
+  align-items: center;
+  gap: 0.875rem;
+}
+
+.sb-references-download-icon {
+  width: 2.75rem;
+  height: 2.75rem;
+  border-radius: 0.875rem;
+  background: var(--sb-secondary-color);
+  color: var(--sb-primary-color);
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex: 0 0 auto;
+}
+
+.sb-references-download-icon .material-symbols-outlined {
+  font-size: 1.5rem;
+}
+
+.sb-references-download-text {
+  display: flex;
+  flex-direction: column;
+  gap: 0.375rem;
+}
+
+.sb-references-download-body {
+  margin: 0;
+  line-height: 1.6;
+  font-size: 0.95em;
+  opacity: 0.7;
+}
+
+/* The size estimate sits under the explanation as a quieter second line. */
+.sb-references-download-size {
+  margin: 0;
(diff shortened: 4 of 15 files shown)


Blitzy:
> [!NOTE]

## Code review — References extension port (#1120)

This PR ports the References (cross-references) extension onto the refactor. I reviewed the whole diff (14 files, +2545), read the linked issue #1120, ran the new test suites, ran ESLint on the new package, and checked the app APIs the extension consumes.

**What I verified locally**
- `pnpm vitest run test/unit/references-extension` → **44 tests pass** (2 files).
- `eslint packages/references-extension` → **0 errors**, 13 CSS baseline warnings (see nit No. 3).
- Every app API the extension imports (`OfflineTranslationsManager`, `formatBytes`, `estimateTranslationSizeBytes`, `registerVerseToolbarTool`, `PanePlacement`, `indexedDbUtils`, `getTranslationBookChapter`) exists with the expected shape.

Overall this is solid, well-structured work with unusually thoughtful tests. The findings below are mostly about graceful degradation and a couple of things to confirm before merge. Nothing is a hard blocker except your own judgment on No. 1.

---

### Medium

**No. 1 — One unresolvable reference blanks the *entire* text list**

`createRefsWithText` (`utils.tsx:183`) fetches each reference's verse text with `Promise.all`. `Promise.all` rejects as soon as *any* one fetch rejects, and `getTranslationBookChapter` **does** reject on a missing book/chapter — it delegates to `FreeUseBibleAPI._getJson`, which throws on any non-2xx (`FreeUseBibleAPI.tsx:1373`, `Failed request ... Status: 404`). In `referenceApp.tsx:232` that rejection sets `failed.value = true`, and the render then shows the "Verse text unavailable" note for **every** row (`referenceApp.tsx:391`), not just the one that failed.

Why this matters: cross-references routinely point across books. If the current translation omits a referenced book (partial/NT-only translations, or a single 404), or if any *one* of the up-to-10 concurrent fetches hits a transient network error, the reader sees an entire pane of "Verse text unavailable" even though most references loaded fine.

This failure mode is currently untested — every `createRefsWithText` test supplies all the chapters it needs, so a rejecting fetch is never exercised (which is exactly why the bug slipped through).

*Reproduction (unit-level):*
```ts
const dm = dataManagerWith({ "GEN.1": [{ type: "verse", number: 1, content: ["ok"] }] });
// second ref points at a chapter the manager doesn't have -> it throws
await createRefsWithText({
  references: [
    ref({ book: "GEN", chapter: 1, verse: 1, score: 9 }),
    ref({ book: "EXO", chapter: 1, verse: 1, score: 8 }), // throws
  ],
  limit: 5, dataManager: dm, translationId: "WEB",
});
// Actual: rejects entirely. Expected: returns GEN text, EXO row -> text "".
```
*In-app:* open the References pane on a verse whose top references include a book missing from the current translation (or throttle so one chapter 404s) → whole list shows "unavailable".

*Suggested fix* — degrade per row instead of all-or-nothing:
```ts
const refsWithText = await Promise.all(
  limitedReferences.map(async (ref) => {
    try {
      const verseText = await dataManager.getTranslationBookChapter(
        translationId, ref.book, ref.chapter
      );
      // ...existing extraction...
      return { ...ref, text: verseTexts.filter(Boolean).join(" ") };
    } catch (error) {
      console.warn("Could not load text for reference", ref, error);
      return { ...ref, text: "" };
    }
  })
);
```
An empty string already renders as the "unavailable" note per-row (`referenceApp.tsx:385-390`), so this makes each row fail independently. Add a regression test with a mixed success/throw data manager.

---

### Low

**No. 2 — A degraded `200` response is cached permanently as empty**

`GetReferences` (`utils.tsx:121-138`) writes whatever it gets to IndexedDB whenever the request resolves, using `referenceReq.data?.chapter?.content ?? []`. If the endpoint ever returns a `200` whose body is missing `chapter.content` (an error page served as 200, a truncated payload), the chapter is cached as `references: []` and the user never sees references for it until the DB version is bumped. In a full `downloadReferences` run such a chapter is also counted as a successful `downloaded`, hiding the gap. Consider only caching when `chapter.content` was actually present (fetch failures already throw and correctly skip the cache).

**No. 3 — The offline-download offer is spent the instant it's shown, once per device, forever**

`offerReferenceDownload` (`askToDownload.tsx:325-326`) calls `rememberOffered()` *before* the user interacts, and `hasBeenOffered()` gates on the key's mere presence. So closing the tab without answering, clicking "Not now", or a download that fails all leave the user never offered again. That may be intentional, but there's currently no path to trigger the download later (no toolbar/settings entry I could find). Worth confirming this is the desired UX, and consider re-offering when a download fails, or exposing a manual "save for offline" action.

**No. 4 — References without a `score` are silently dropped**

`selectTopReferences` (`utils.tsx:168-171`) filters out any reference with `score === undefined`. A test documents this as intended, but the consequence is that a verse whose references are all unscored renders as "No cross references for this verse." Please confirm the real `open-cross-ref` dataset always scores its references; if not, unscored entries vanish silently.

---

### Nits / things to confirm

**No. 5 — Tests fully mock `axios`, so the real API shape is never validated.** Both suites replace `axios` entirely and assert against a `{ chapter: { content: [{ verse, references }] } }` payload. That's internally consistent with the code, but nothing confirms the live `open-cross-ref` endpoint actually returns that shape — if it differs, `GetReferences` silently returns `[]` (the `?? []`) and every test still passes. I couldn't reach `vmfnri.helloao.org` from the review sandbox (network policy) to verify. Since this is a *port*, the shape is presumably carried over from the old extension, but a one-line comment documenting the confirmed real response shape (or one non-mocked smoke test) would lock it in.

**No. 6 — Clickable `<span>` for the reference label** (`referenceApp.tsx:367-372`) isn't keyboard-focusable or activatable. The adjacent "open in new" is a real `<button>`; consider making the label one too (or adding `role`/`tabindex`/key handling) for keyboard and screen-reader users.

**No. 7 — New CSS uses container queries** (`@container`, `container-type` in `referenceApp.css`) which don't appear elsewhere in the app's CSS and produce the baseline ESLint warnings. Fine if your browser targets cover them — just flagging that it's a new dependency for this codebase. The `color-mix` warnings are consistent with existing app CSS and can be ignored.

---

### On the tests (the part I was asked to scrutinize)

These are genuinely good and do **not** look like they always-pass:
- They assert observable behavior — rendered text, toast wording and order, download counts, and even the concurrency **peak** (`expect(peak).toBe(20)` / `toBe(3)`) — not internals.
- Error paths are covered: network failure, partial download, translation-download failure, blocked storage / SSR (no IndexedDB), and "size unknown".
- Async is polled via `waitForCondition`, not fixed sleeps — matches the repo convention.
- The "shipped English == the code's `defaultValue`" tests (`askToDownload.test.tsx:276-319`) are a nice guard against fallback wording drifting from the bundle.

The one real gap is the untested rejection path behind finding No. 1: no `createRefsWithText` test drives a fetch that throws, which is precisely where the all-or-nothing bug lives. Adding that case would both catch the bug and prevent regressions.

---

KallynGowdy:
@Bored-Wizard this is a good first start. I'd like to see this be integrated with the discover panel though. For example, when the user clicks a verse we could show cross references (just the references, not the verses) in the compact discover panel or in the mobile verse toolbar.

Maybe we should have @amir2iqbal447 draw something up for this?

Bored-Wizard:
> @Bored-Wizard this is a good first start. I'd like to see this be integrated with the discover panel though. For example, when the user clicks a verse we could show cross references (just the references, not the verses) in the compact discover panel or in the mobile verse toolbar.
> 
> Maybe we should have @amir2iqbal447 draw something up for this?

That sounds like a really good idea @KallynGowdy , will pick this PR again when @amir2iqbal447 have a design ready for it. Also while you are at it can you also make new design for the reference panel and modal amir?

KallynGowdy:
@Bored-Wizard Please use this design:
image

https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=10056-1871&t=ExdFAeNnBCedWaiR-4

For now, it should only show when one or more verses are selected, and the reference "pills" should be grouped by the selected verse number.

PR #1802, opened 2026-09-11
feat: Chrome-style extension manager: icon, enable/disable toggle, details pane

VictorUvarov:
## Summary

Phase 1 of #1737, a Chrome-style extension manager for Seed Bible's extension settings: an enable/disable switch that is separate from remove, and a details pane showing version, source and homepage. 

- `extension.json` gains optional `icon`, `version` and `homepage` fields, threaded through the existing build-time schema. No migration needed.
- A new `disabledExtensionIds` mechanism (localStorage only) turns an installed extension off without uninstalling it. The installed-extensions ledger and its cross-device sync are untouched.
- Each extension row is now two rows inside one card. The top row keeps the original layout (state icon, name, description, remove or install button) and a footer row carries "Enabled" with its switch.
- Clicking anywhere on the card opens the details pane: version, source (built in, or loaded from a URL), homepage link, the same switch, and uninstall.
- Every breadcrumb segment is clickable, not just the back arrow. From an extension's details pane you can jump to Extensions or straight to the settings root.

## What changed

- The details pane's switch and uninstall now depend on install state. They used to render unconditionally, so opening details on a never-installed extension from the Available tab and flipping the switch wrote a persisted "phantom installed" entry. Fixed, with a regression test.
- A disabled extension is re-enabled when another extension force-installs it as a dependency. Before: disabling `seed-bible-utils` and then installing `bible-stack`, which depends on it, left the UI showing it as off while it was running. Now the disabled flag is cleared when the dependency loads.
- A signals anti-pattern (writing to a signal mid-render) was simplified away.

## Screenshot / Video(s)

These predate the row redesign, so they still show the icon avatar and the chevron button.

Screenshot 2026-09-12 at 5 44 30 PM
Screenshot 2026-09-12 at 5 44 16 PM
Screenshot 2026-09-12 at 5 44 40 PM


KallynGowdy:
Duplicate of #1747

Issue #1769, opened 2026-09-07
Add support for web browser Text to speech for translations which don't have audio versions 


KallynGowdy:
Currently, only the BSB and AAB have audio. All other translations don't support audio playback. Web browsers support rudimentary text to speech for a variety of languages already, so we could use that to support audio playback for Bible translations which don't have audio.

See this: https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API/Using_the_Web_Speech_API

PR #1803, opened 2026-09-12
added text to speech functionality for supported languages using web …

Bored-Wizard:
…speech api

task #1769

### DIFF
--- a/packages/audio-reader-extension/ext_audioReader/host/init.tsx
+++ b/packages/audio-reader-extension/ext_audioReader/host/init.tsx
@@ -1,10 +1,15 @@
 import { computed, effect, signal } from "@preact/signals";
+import { debounce } from "es-toolkit";
 import { registerExtension, type SeedBibleState } from "seed-bible";
-import type {
-  BibleReadingState,
-  ChapterVerse,
-  QuickToolContext,
-  TranslationBookChapter,
+import { LANG_META } from "seed-bible/i18n";
+import {
+  bibleLanguageToUiLocale,
+  extractContentText,
+  type BibleReadingState,
+  type ChapterVerse,
+  type QuickToolContext,
+  type SpeechVerse,
+  type TranslationBookChapter,
 } from "seed-bible/managers";
 
 /** Drives the icon swap between play and pause. Shared across the tool. */
@@ -22,12 +27,31 @@ const isPlaying = signal(false);
  */
 const VERSE_HIGHLIGHT_LEAD_IN_SECONDS = 0.3;
 
+/**
+ * How long the Listen control waits out a burst of presses before acting, so a
+ * double-press toggles once rather than playing and immediately stopping.
+ */
+const TOGGLE_DEBOUNCE_MS = 300;
+
 /** Lazily-created shared audio element and the URL currently loaded into it. */
 let audioEl: HTMLAudioElement | null = null;
 let currentUrl: string | null = null;
 
 /** The chapter whose narration is loaded into that element, if any. */
 let currentChapter: ListeningTarget | null = null;
+
+/**
+ * The reader whose verses the speech synthesiser is currently lighting up, and
+ * the decoration doing the lighting. Kept separate from `verseTrack` because
+ * spoken playback needs no timing data at all — an utterance's `start` event is
+ * the verse boundary — so the two paths share nothing but `decorateVerses`.
+ */
+let speechHighlight: {
+  readingState: BibleReadingState;
+  bookId: string;
+  chapterNumber: number;
+  decorationId: string | null;
+} | null = null;
 
 /**
  * Set for as long as the extension is installed. The recorder writes through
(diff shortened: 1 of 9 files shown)


Blitzy:
> [!NOTE]

## Code review — Text-to-speech for translations without audio (#1769)

Thanks for this — it's a well-thought-out feature. The core is a new `TextToSpeechManager` that speaks a chapter one utterance per verse (so an utterance's `start` event doubles as the verse boundary for highlighting), plus a genuinely careful language-matching layer in the audio-reader extension that maps the Bible API's ISO 639-3 codes to the BCP-47 tags voices are labelled with. The comments explain the *why* throughout, and the Web Speech API's known rough edges (Chrome's ~15s cutoff, `cancel()` not being a guarantee) are handled deliberately.

I found one real correctness gap, a few cleanups, and one place where the tests could be tightened. Nothing here is blocking on its own — the Medium item is the one worth fixing before merge.

---

### Medium

**1. A synthesiser error partway through a chapter abandons the UI but not the audio.**

In `TextToSpeechManager.speak`, every utterance's `onerror` calls `reset()`, which flips `isSpeaking` to false and clears `currentVerse` — but it does **not** `cancel()` the rest of the queue, and it does not abandon the run (the run token is left unchanged). So when an error hits a verse that isn't the last one:

- the browser carries on to the next queued utterance and keeps reading the chapter aloud;
- because the run token still matches, later `onstart` handlers keep firing, so `currentVerse` keeps advancing and the reader's verse highlight keeps moving — while the toolbar button now shows **Play**;
- the final utterance's `onend` still fires `onFinished`, reporting a clean finish for a chapter that errored.

This is the same "reads itself aloud with nothing tracking it" state the manager works hard to prevent for the double-press case — it's just reachable through a different door.

_Reproduction (via the manager's own test fakes):_
```ts
manager.speak(
  [{ number: 1, text: "one" }, { number: 2, text: "two" }, { number: 3, text: "three" }],
  { lang: "en", onFinished }
);
speech.queued[0].onstart();   // currentVerse = 1, isSpeaking = true
speech.queued[0].onerror();   // reset(): isSpeaking = false, currentVerse = null — queue NOT cleared
speech.queued[1].onstart();   // currentVerse advances to 2 again, though isSpeaking is false
speech.queued[2].onend();     // onFinished fires, as if nothing went wrong
```
I ran exactly this against the branch: `currentVerse` came back as `2`, `isSpeaking` as `false`, and `onFinished` was called once.

_Suggested fix_ — treat an error as ending the whole run, mirroring `stop()`'s "abandon the token, then silence the queue" order so the cascade of `canceled` errors it triggers is ignored by the guard:
```ts
utterance.onerror = () => {
  if (token !== runToken) return;
  runToken++;      // abandon this run so the follow-on error/end events are ignored
  reset();
  speech.cancel(); // clear the verses still queued behind the failed one
};
```

---

### Low

**2. Debug `console.log`s left in `TextToSpeechManager`.**

`managers/TextToSpeechManager.tsx:239` and `:253` log on every `stop()` and every `speak()`:
```ts
console.log("Stopped speech run with token:", runToken);
console.log("Starting new speech run with token:", token);
```
These fire on every play/stop during normal use and will spam the console in production. Worth removing (there's no project logger these are routing through).

**3. The manager never tears down its global listeners, and a bfcache restore can strand the button.**

`createTextToSpeechManager` adds `speech.addEventListener("voiceschanged", …)` and `window.addEventListener("pagehide", …)` but exposes no teardown, so they're never removed. In practice the manager is an app-lifetime singleton, so this isn't a live leak — but it also means the `pagehide` handler only `cancel()`s the engine without resetting state. If the page is restored from the back/forward cache (bfcache) mid-chapter, speech is cancelled but `isSpeaking`/`currentVerse` stay as they were, leaving the button stuck on **Pause** with nothing playing. Resetting state alongside the `cancel()` on `pagehide` would close that edge.

---

### Nit

**4. The 300ms debounce now delays every Listen press, including recorded audio.**

Wrapping the whole `onSelect` in `debounce(…, TOGGLE_DEBOUNCE_MS)` is a clean way to collapse a double-press, but it also adds a 300ms delay to a single, deliberate press — including the recorded-audio play/pause path, which previously responded instantly. Probably an acceptable trade for the double-press fix, but flagging in case the latency on the common single-press case matters for the recorded-audio experience.

---

### Tests

Overall the test coverage is strong and tests real behavior rather than rubber-stamping the implementation — I confirmed the suite is green (64 passing across the four files). Some specifics I checked:

- `TextToSpeechManager.test.ts` drives the fake synthesiser through real scenarios (token guards, the double-press "start anyway" case, the "engine ignores `cancel()`" enforcement loop with fake timers, the give-up deadline). These are not tautological — e.g. the debounce test in `speechVerseHighlightSync.test.tsx` genuinely fails if the debounce is removed (two presses would become play-then-stop).
- The `#1769` language-mapping tests in `init.test.tsx` exercise the CLDR/`LANG_META` edge cases directly (Hausa→`ha`, Indonesian `ind`→`id`, Malay `zlm`→`ms`, the same-script fallback allowing Marathi→Hindi but blocking Gujarati→Hindi). Good, meaningful coverage.

One gap, tied to Finding No. 1: `TextToSpeechManager.test.ts`'s _"resets when an utterance fails"_ test only asserts the immediate `reset()` (it errors verse 1 of 2 and stops there). It never checks that the run is actually *abandoned*, so the behavior in Finding No. 1 slips through green. Extending it to fire the next utterance's `onstart` after the error and assert `currentVerse` stays `null` (and that `onFinished` never fires) would catch the bug — and would fail on the current code, which is what makes it a real regression test.

---

_Reviewed at head `263ad79`. I did not run the full type-check or lint; findings are from reading the diff and running the affected test files (which transpile without type-checking)._

---

KallynGowdy:
@Bored-Wizard one thing to note is that it doesn't currently seem to handle the 15 second cutoff on Chrome right at the moment. It still cut off for me halfway through reading a verse.

Bored-Wizard:
> @Bored-Wizard one thing to note is that it doesn't currently seem to handle the 15 second cutoff on Chrome right at the moment. It still cut off for me halfway through reading a verse.

the cut-off after 15 secs is not happening for me when I was testing it and when I test it again. It seems it is a known bug in speechSynthesis so I will add a guard for it so it doesn't happen in other devices.

Bored-Wizard:
> @Bored-Wizard one thing to note is that it doesn't currently seem to handle the 15 second cutoff on Chrome right at the moment. It still cut off for me halfway through reading a verse.

tried it o my phone and yeah it's happening there lol

PR #1804, opened 2026-09-13
feat: Let openPane callers set a pane's starting size

VictorUvarov:
## Summary

- `openPane` takes an optional `size: { width, height }`, so a caller can open a pane larger than the fixed 480x320 default. Depending on the extension, embedded content such as an extension's iframe might be unreadable at that size.
- The size applies only when the pane is first created, so reopening a pane by `id` keeps whatever size the user dragged it to.

## Test plan

- [ ] Open a pane without a size, confirm it still opens at 480x320
- [ ] Open a pane with a size, resize it by dragging, reopen it by the same id and confirm the dragged size survives

## Screenshots / Video(s)
Screenshot 2026-09-12 at 5 29 24 PM


KallynGowdy:
@Blitzy this looks good to me

Blitzy (approved):

PR #1805, opened 2026-09-13
feat: Add a System theme that follows the device color scheme

VictorUvarov:
## Summary

- Added a **System** theme to Settings → Display & Theme, next to Light and Dark

## Test plan

- [ ] Flip the OS between light and dark with the app open — theme follows without a reload
- [ ] Reload while System is selected — no flash of the wrong theme
- [ ] Pick Light or Dark explicitly — the OS setting no longer affects the app
- [ ] Theme card names stay aligned as selection moves between cards
- [ ] Line-height icons still render in both Settings and the mobile settings sheet

## Screenshots / Video(s)
system-settings

### DIFF
--- a/index.html
+++ b/index.html
@@ -73,6 +73,11 @@
           if (!raw) return;
           var themeId = JSON.parse(raw).themeId;
           if (!themeId) return;
+          if (themeId === "system") {
+            themeId = window.matchMedia("(prefers-color-scheme: dark)").matches
+              ? "dark"
+              : "light";
+          }
           var presetsEl = document.getElementById("sb-theme-presets");
           if (!presetsEl || !presetsEl.textContent) return;
           var css = JSON.parse(presetsEl.textContent)[themeId];
--- a/packages/seed-bible/seed-bible/components/MobileSettingsSheet/MobileSettingsSheet.tsx
+++ b/packages/seed-bible/seed-bible/components/MobileSettingsSheet/MobileSettingsSheet.tsx
@@ -7,35 +7,9 @@ import {
   type UISize,
 } from "../../managers/SettingsManager";
 import { useI18n } from "../../i18n/I18nManager";
-import { SettingsIcon } from "../icons";
+import { ScriptureLineHeightIcon, SettingsIcon } from "../icons";
 
 const FONT_SIZE_OPTIONS = ["XS", "S", "M", "L", "XL", "XXL"] as const;
-
-function ScriptureLineHeightIcon({ index }: { index: number }) {
-  const gap = 3.5 + index * 1.5;
-  const startY = 1;
-  return (
-    <svg width="20" height="14" viewBox="0 0 20 14" fill="none">
-      <rect x="0" y={startY} width="20" height="2" rx="1" fill="currentColor" />
-      <rect
-        x="0"
-        y={startY + gap}
-        width="20"
-        height="2"
-        rx="1"
-        fill="currentColor"
-      />
-      <rect
-        x="0"
-        y={startY + 2 * gap}
-        width="20"
-        height="2"
-        rx="1"
-        fill="currentColor"
-      />
-    </svg>
-  );
-}
 
 interface MobileSettingsSheetProps {
   state: SeedBibleState;
--- a/packages/seed-bible/seed-bible/components/SettingsPage/SettingsPage.css
+++ b/packages/seed-bible/seed-bible/components/SettingsPage/SettingsPage.css
@@ -821,6 +821,21 @@
     transform 0.1s ease;
 }
 
+.sb-theme-ready-preview-system {
+  padding: 0;
+  gap: 0;
+  align-items: stretch;
+  overflow: hidden;
+}
+
+.sb-theme-ready-system-half {
+  flex: 1;
+  display: flex;
+  align-items: flex-end;
+  justify-content: center;
+  padding-bottom: 0.625rem;
+}
+
 .sb-theme-ready-card:hover .sb-theme-ready-preview {
   border-color: color-mix(in srgb, var(--sb-font-color), transparent 70%);
 }
@@ -853,6 +868,10 @@
 .sb-theme-ready-check {
   font-size: 1rem;
   color: var(--sb-primary-color);
+}
+
+.sb-theme-ready-check-hidden {
+  visibility: hidden;
 }
 
 /* Scripture quick-settings (font size + line height) */
--- a/packages/seed-bible/seed-bible/components/SettingsPage/SettingsPage.tsx
+++ b/packages/seed-bible/seed-bible/components/SettingsPage/SettingsPage.tsx
@@ -1,5 +1,6 @@
 import "./SettingsPage.css";
 import { useComputed, useSignal } from "@preact/signals";
+import { ScriptureLineHeightIcon } from "../icons";
 import type { SeedBibleState } from "../../managers/SeedBibleStateManager";
 import {
   TEXT_FONT_OPTIONS,
@@ -16,7 +17,10 @@ import {
   type UISize,
 } from "../../managers/SettingsManager";
 import {
+  DARK_THEME,
   DEFAULT_HIGHLIGHT_IDS,
+  LIGHT_THEME,
+  SYSTEM_THEME_ID,
   THEME_COLOR_GROUPS,
   type ThemeColorKey,
 } from "../../managers/ThemeManager";
@@ -502,32 +506,6 @@ function AccountSettingsView(props: { state: SeedBibleState }) {
   );
 }
 
-function ScriptureLineHeightIcon({ index }: { index: number }) {
-  const gap = 3.5 + index * 1.5;
-  const startY = 1;
-  return (
-    <svg width="20" height="14" viewBox="0 0 20 14" fill="none">
-      <rect x="0" y={startY} width="20" height="2" rx="1" fill="currentColor" />
-      <rect
-        x="0"
-        y={startY + gap}
-        width="20"
-        height="2"
-        rx="1"
-        fill="currentColor"
-      />
-      <rect
-        x="0"
-        y={startY + 2 * gap}
-        width="20"
-        height="2"
-        rx="1"
-        fill="currentColor"
-      />
-    </svg>
-  );
-}
-
 /**
  * Built-in theme names are authored in English on the theme object, so they'd
  * otherwise render untranslated. Spelled out as separate `t()` calls (rather
@@ -540,13 +518,43 @@ export function localizedThemeName(
   t: I18nHook["t"],
   theme: { id: string; name: string }
 ): string {
-  if (theme.id === "light") {
+  if (theme.id === LIGHT_THEME.id) {
     return t("theme-light", { defaultValue: theme.name });
   }
-  if (theme.id === "dark") {
+  if (theme.id === DARK_THEME.id) {
     return t("theme-dark", { defaultValue: theme.name });
   }
+  if (theme.id === SYSTEM_THEME_ID) {
+    return t("theme-system", { defaultValue: theme.name });
+  }
   return theme.name;
+}
+
+/**
+ * A theme card's footer: the theme name, plus a check mark when it's the
+ * selected theme. The check is hidden rather than unmounted when unselected, so
+ * the name sits in the same place on every card.
+ */
+function ThemeCardLabel(props: { name: string; isSelected: boolean }) {
+  const { t } = useI18n();
+  return (
+    <div className="sb-theme-ready-label">
+      <span>{props.name}</span>
+      <span
+        className={`material-symbols-outlined sb-theme-ready-check${
+          props.isSelected ? "" : " sb-theme-ready-check-hidden"
+        }`}
+        aria-hidden={!props.isSelected}
+        aria-label={
+          props.isSelected
+            ? t("selected", { defaultValue: "Selected" })
+            : undefined
+        }
+      >
+        check_circle
+      </span>
+    </div>
+  );
 }
 
 function ThemesGallerySection(props: { state: SeedBibleState }) {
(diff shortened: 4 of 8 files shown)


KallynGowdy:
One thing we need to test here is that the system theme selection doesn't cause issues for Customizations, and that it is still available when a customization has a dark and light theme.

VictorUvarov:
Fixed: `system` can never be stored as a variant's base theme, a customization with a light and a dark variant now gets its own System card that follows the device, and the text-color reset moved onto `setTheme` so an OS flip no longer wipes saved colors.

KallynGowdy:
@VictorUvarov This looks good and I can confirm that the functionality works. I have two things that I think would make this even better:

1. Set the `system` theme to be the default (if possible). Currently, it is `light` which won't match the user's choice in a lot of cases. This also means that the system theme should be the first in the list.
2. Wrap the themes into a 2 column grid instead of scrolling:

image

Blitzy:
> [!NOTE]

## Summary

Nice work overall. Picking System and flipping the OS while the app is open works, System support for Customizations is in place, and moving the text-color reset onto `setTheme` is the right call. There is one blocking bug, though: since System became the default (commit 2e1a7e9), **on a dark-mode device the page stays Light after it loads** for anyone whose theme should be Dark. That includes existing users who explicitly picked Dark, so it's a regression and not only a gap in the new feature. CI is also failing on a type error in the new test file.

| No. | Severity | Issue |
| --- | --- | --- |
| 1 | 🔴 High | On dark-mode devices the page renders Light after load (first visits, saved System, and saved Dark) |
| 2 | 🔴 High | CI fails `check:ts`: a new test fixture is missing `extensionSettingDefaults` |
| 3 | 🟡 Medium | Custom text colors carry over a light→dark flip and can become unreadable |
| 4 | 🟡 Medium | Following the device ignores a Customization's default variant |
| 5 | 🔵 Low | Tests check the in-memory theme, not what's painted on screen |
| 6 | ⚪ Nit | `en.json` reshuffled by the merge driver; one comment is inaccurate |

---

### No. 1 🔴 On a dark-mode device, the page stays Light after loading

**What happens:** The server can't see the device, so it always renders the Light theme. The browser app then reads `matchMedia` right at startup (`watchSystemColorScheme`, `ThemeManager.tsx:971`), before Preact hydrates. On a dark device that means its very first render is already Dark, and two things then keep the server's Light CSS on screen:

1. **The in-tree `<style>body{…}</style>`** that `ExternalResourceDependencies` renders (`app/main.tsx:110`) uses `dangerouslySetInnerHTML`. Preact never applies that during hydration. After that, it only rewrites the tag when the CSS *changes* from what the client first rendered. The client rendered Dark first and stays Dark, so the Light CSS from the server is never replaced. This tag comes after the head tag in the document, so it wins.
2. **The `#sb-theme-styles` head tag** skips its first write (the `skipWrite` guard) and only writes again when the theme changes. That change never comes, because the theme was already Dark at startup. On a first visit the inline script in `index.html` also exits early when nothing is saved (`if (!raw) return;` / `if (!themeId) return;`, lines 80–82), so nothing corrects this tag either.

The result is that the app *thinks* it's Dark (Settings shows System as selected, and `basePresetTheme` is `dark`), but the page is painted Light until the OS flips or the user picks a theme.

**Who's affected, on a dark-mode device:**

| Visitor | develop | This PR |
| --- | --- | --- |
| First visit, nothing saved | Light (was the default) | Light, even though System is selected ❌ |
| Saved **System** | n/a | Light ❌ |
| Saved **Dark** | Dark (after load) | **Light ❌ (regression)** |
| Saved Light | Light | Light ✅ |
| `?customization=` link with light + dark variants | the default variant | the Light variant ❌ |

Saved Dark worked on develop because the startup default was Light, which matched the server. When `hydrateLocalConfig()` then swapped in Dark, the CSS *changed* and Preact re-rendered the tag. Now the startup default already resolves to Dark, so no change happens and nothing triggers a re-render.

I confirmed this with a scratch SSR + hydrate test (same setup as `hydration.test.tsx`) that stubs `matchMedia` to dark. In the three ❌ rows, both style tags ended on `--sb-background: #f8fafc` (Light). The control cases were correct: saved Light on a dark device, and saved Dark on a light device. Setting `DEFAULT_SETTINGS.themeId` back to `"light"` makes saved Dark and saved System render correctly again, so making System the default is what exposes this. It was probably missed because the feature was confirmed working before the default changed, and the bug only appears on a reload on a dark device.

**Steps to reproduce:**

1. Set your OS to dark mode (or DevTools → Rendering → "Emulate CSS media feature prefers-color-scheme" → dark).
2. Open the app in a fresh private window. The page is Light, but Settings → Display & Theme shows **System** as selected.
3. Pick **Dark** and reload. The page comes back Light. On develop it comes back Dark.
4. Pick **System** and reload. The page is Light again.
5. While it's stuck on Light, flip the OS to light and back to dark. Now the page goes Dark, because that's the first real *change*.

**Suggested fix:** Use the same "seed, then correct after mount" pattern the app already uses for `localConfig`. Start `prefersDarkScheme` as `false` to match the server, and read the real device setting in the post-mount effect next to `hydrateLocalConfig()`. The client's first render then matches the server's HTML, and the switch to Dark after mount is a real change, so both style tags repaint. I prototyped this and every ❌ row above then paints Dark.

<details><summary>Sketch</summary>

```ts
// ThemeManager.tsx
function watchSystemColorScheme(
  prefersDarkScheme: Signal<boolean>
): () => void {
  if (
    typeof window === "undefined" ||
    typeof window.matchMedia !== "function"
  ) {
    return () => {};
  }
  const query = window.matchMedia("(prefers-color-scheme: dark)");
  query.addEventListener("change", (event) => {
    prefersDarkScheme.value = event.matches;
  });
  // Seeded `false` to match SSR (the server can't see the device); the real
  // value is applied post-mount, the same way `hydrateLocalConfig` works.
  return () => {
    prefersDarkScheme.value = query.matches;
  };
}

// in createTheme(), and exposed on ThemeManager:
const hydrateSystemColorScheme = watchSystemColorScheme(prefersDarkScheme);
```

```ts
// app/main.tsx (line 201)
useEffect(() => {
  state.login.hydrateLocalConfig();
  state.theme.hydrateSystemColorScheme();
}, []);
```

`createTestSeedBibleState.ts:292` should call it as well. The ThemeManager tests that expect the device value right after `createTheme()` will need to call it first.

</details>

Also update the inline script in `index.html` so that a missing saved theme counts as System, since that's now the default. That makes the head tag and the Android status-bar color correct before the bundle loads on a first visit:

```js
var raw = localStorage.getItem("sb-profile-config-local");
var themeId = (raw && JSON.parse(raw).themeId) || "system";
```

> A separate issue that this PR didn't cause: even with both fixes, dark-theme users will see a brief Light flash before the app mounts. The in-tree `<style>` from point 1 above is server-rendered Light, and it overrides the head tag that the inline script corrects. The same thing already happens on develop for users who saved Dark, so it's fine to leave for a follow-up. It does mean the test-plan item "Reload while System is selected — no flash of the wrong theme" won't fully hold on a dark device.

---

### No. 2 🔴 CI is failing on a type error in the new test

`check:ts` fails, so the test suite never ran in CI:

```
test/unit/seed-bible/components/CustomizationVariantGallery.test.tsx:45:3 - error TS2741: Property 'extensionSettingDefaults' is missing in type '{ ... }' but required in type 'SeedBibleCustomization'.
```

`extensionSettingDefaults` came in from develop with #1747 after this test was written. **Fix:** add `extensionSettingDefaults: {},` to `makeCustomization()`.

---

### No. 3 🟡 Custom text colors can become unreadable after a device flip

Custom verse, heading and book-title colors are applied as inline overrides on `body`, which beat both themes. Under System, a color picked while the device was light stays in force after the device turns dark. The PR's own test shows this: `systemThemeTextColors.test.ts:11` sets verse text to `#101010` and expects it to survive a flip to Dark, whose background is `#0a0a0a`. That's near-black text on a near-black background.

Keeping the colors was a deliberate choice (commit 1999036, so that a flip doesn't silently wipe saved colors), and wiping them would be worse. But changing the default widens who this affects. Existing users who never picked a theme but did customize text colors are moved onto System automatically. On a dark device they could open the app to unreadable text with no idea why.

**Steps to reproduce:**

1. With the device in light mode, go to Settings → Display & Theme → System.
2. Set the Verse text color to something dark, e.g. near-black.
3. Switch the device to dark mode. Verse text is now close to invisible.

**Options (this is a product call):**

- **Keep one set of text colors per scheme** (light and dark), so each keeps its own. This is the cleanest, but it's a bigger change and could be a follow-up issue.
- **Under System, skip a custom color that has poor contrast** with the current background and use the theme's color instead. It's a small change, but the user's color silently won't apply in one of the two schemes.
- **Accept it for now** and track it as a follow-up.

A related minor point: picking System while on Light, on a light device, wipes custom text colors (`setTheme` resets them whenever the theme id changes, `ThemeManager.tsx:1384`) even though nothing on screen changes. Comparing the *resolved* theme before and after would avoid that.

---

### No. 4 🟡 Following the device ignores the Customization's default variant

`bySystemScheme` (`CustomizationsManager.tsx:1164`) takes the *first* variant whose base theme matches the device. If a Customization has more than one variant on the same base, the author's chosen default can be skipped. The System card's preview (`CustomizationSystemCard`, `SettingsPage.tsx:738`) makes the same first-match choice.

**Example:** the variants are "Paper" (Light), "Sepia" (Light, **set as the default**) and "Night" (Dark). A viewer on System with a light device who hasn't picked a variant sees "Paper" instead of the default "Sepia".

**Suggested fix:** prefer the default variant when it matches the device's scheme, and share one helper between the manager and the card:

```ts
export function variantForScheme(
  customization: SeedBibleCustomization,
  presetId: string
): CustomizationThemeVariant | undefined {
  const matches = customization.variants.filter(
    (v) => v.baseTheme === presetId
  );
  return (
    matches.find((v) => v.id === customization.defaultVariantId) ?? matches[0]
  );
}
```

**A question for the team:** System is now everyone's default, so every viewer who hasn't picked a variant follows their device. That means a Customization's default variant only matters to viewers who explicitly chose Light or Dark for the whole app. For example, a group that sets a Dark variant as its default will show Light to most viewers on light devices. Is that intended? The earlier request was that System be *available* for Customizations that have light and dark variants, which isn't quite the same thing.

---

### No. 5 🔵 Tests check the in-memory theme, not what's painted on screen

- The new System tests in `ThemeManager.test.ts` assert on `basePresetTheme` / `currentTheme`, not on what actually gets painted, and the painted output is where No. 1 lives. A hydration test catches it. Something like this should work; my scratch version of it failed on the current code and passed with the prototype fix:

<details><summary>Test sketch for test/unit/standalone/hydration.test.tsx</summary>

```tsx
it.each([
  ["a first-time visitor", null],
  ["a visitor who saved System", "system"],
  ["a visitor who saved Dark", "dark"],
])(
  "paints the dark theme for %s on a dark device",
  async (_label, savedThemeId) => {
    const { container } = await installSsrDocument();
    stubColorScheme(true);
    if (savedThemeId) {
      localStorage.setItem(
        "sb-profile-config-local",
        JSON.stringify({ themeId: savedThemeId })
      );
    }
    const { config, state } = await createClientState();

    // Runs MainBody's post-mount effects (hydrateLocalConfig and the new
    // color-scheme read).
    await act(async () => {
      hydrate(<Main initialState={state} config={config} />, container);
    });

    const inTreeCss = Array.from(container.querySelectorAll("style"))
      .map((s) => s.innerHTML)
      .find((css) => css.includes("--sb-background"));
    expect(inTreeCss).toContain(
      `--sb-background: ${DARK_THEME.variables.background};`
    );
  }
);
```

</details>

- Every System test calls `setTheme("system")` while the theme is *already* System, because System is now the default. That makes the call do nothing, so `ThemeManager.test.ts` and `systemThemeTextColors.test.ts` can't catch a broken "pick System" path. To check, I made `setTheme` reject `"system"`. Only two CustomizationsManager tests failed, and only because their mocked settings happen to start on Light. These tests should start on Light or Dark and then switch to System, which is what a real user does.
- `CustomizationVariantGallery.test.tsx` replaces `state.customizations` with hand-set signals. For example, "hides System when the customization only has one color scheme" passes `canFollowSystemScheme: false` in directly, so it would still pass if the real `canFollowSystemScheme` logic were broken. CLAUDE.md says to mock only at the CasualOS boundary. Build the state with `createTestSeedBibleState` and a real Customization instead, then assert on the rendered cards, and on which card is selected after a click rather than on `toHaveBeenCalledWith`.

---

### No. 6 ⚪ Nits

- **`en.json` reshuffle:** the file shows +156/−122 lines, but only one key was actually added (`theme-system`). Everything else is keys moved around by the i18n merge driver (`script/mergeI18nJson.ts`) during the develop merges. For example, `theme-light` / `theme-dark` moved from the Themes block to the middle of the file. It works, but it hides the real change and will cause merge churn for other i18n PRs. Suggest taking develop's `en.json` and re-adding `"theme-system": "System"` right after `"theme-dark"`.
- **Inaccurate comment:** the updated comment in `ThemeManager.test.ts` ("the server renders the system-resolved default into the tag") suggests the server can resolve System. It can't; it always renders Light. That's the root of No. 1, so it's worth wording precisely.

---

**What looks good:** `system` can never be saved as a variant's base theme. Moving the text-color reset from an effect onto `setTheme` also fixes an existing develop bug, where a returning Dark user's saved text colors were wiped on every load, and the returning-visitor test covers that. `prefersDarkScheme` is scoped to each manager and safe on the server. Unknown theme ids, including `"system"` read by an older cached build, fall back to Light cleanly. Arrow-key navigation still works in the new 2-column grid.

Blitzy:
@VictorUvarov No 1 is is definitely an issue and is consistent with my own testing. Cold load in an incognito will respect the device preference (dark mode in my testing) but after a refresh it gets stuck in light mode.

@KallynGowdy do we want to have an explicit system mode or would we rather do a device preference check at load when the user has no preference and pick a light or dark theme and from that point the control is up to the user to change? I ask because No 3 and 4 are as a result of making system the default where we never had one before. I can see keeping it but this is gonna need a lot more testing around customization and user colors.

KallynGowdy:
@Blitzy I actually think we should remove user custom colors now that we have customizations. The custom colors that users can set were always about being able to customize and preview things, and we now have a much better way to do that with customizations (since customizations support custom colors on a per-theme basis - unlike user custom colors). Then we'd be able to keep a `system` theme preference value that would allow a user to use their device preference across multiple devices without forcing either `light` or `dark`.

As far as number 4, I'd say that we should give customization authors the option to set a light and dark theme default separately. This would then let them set the same theme as the default for both kinds, and give them the most flexibility.

Blitzy:
@KallynGowdy I actually love the idea of getting rid of user defined colors. Especially since we great expanded the number of available built-in colors recently this great simplified a number of things when it comes to maintain readability while losing mostly nothing now that we have the customizations framework.

Just to be clear the removal of user colors should not be part of this PR but a follow up one instead.

I also agree that defaults for both light and dark themes make perfect sense for customizations.

I also think this could be a follow up PR.

Blitzy:
@VictorUvarov I think the only merge blocking issues then are No 1 and 2 from the agent code review.

VictorUvarov:
@Blitzy  No. 1 and No. 2 are fixed

Blitzy (approved):
Working great now. Tested logged in, anonymous, with and without customizations that had both light and dark themes and customizations with only light or dark themes.

PR #1817, opened 2026-09-15
Feature/1765

### LINKED ISSUE #1765: When launched as a PWA, the Seed Bible always displays a white status bar on Android
See at the top of the screen that the status bar appears completely white and is unreadable.

Kabi4:
Fixes: #1765


Blitzy:
> [!NOTE]

## Code Review — PR #1817 (Feature/1765)

**What this PR does, in plain terms:** It fixes issue #1765, where the installed Android PWA showed a white, unreadable status bar even in dark theme. The old code shipped two server-rendered `<meta name="theme-color">` tags keyed off `prefers-color-scheme` (the phone's OS light/dark setting), so the bar followed the *OS*, not the *in-app* theme, and defaulted to white. The PR:

- **Removes** those `prefers-color-scheme` meta tags from the SSR output (`standalone/entry-ssr.tsx`).
- **Adds** a single static `<meta name="theme-color" id="sb-theme-color" content="#f8fafc">` to `index.html`, then keeps its `content` in sync with the active theme's `--sb-background` from three places that all converge: the pre-hydration inline script (for the cached/saved theme), and `ThemeManager` (on every theme write, plus the first-run `skipWrite` branch for cached documents).
- **Aligns** the manifest's `theme_color`/`background_color` with `LIGHT_THEME.background` (`#FFFFFF` → `#f8fafc`) so the install splash/first paint matches.

**Overall assessment:** This is a clean, well-scoped fix with unusually thorough inline comments explaining the *why*. I found no correctness bugs. The tests are real regression tests, not always-green ones (I verified this — see below). The findings below are all low severity / polish.

### Verification I ran

- `pnpm vitest run test/unit/seed-bible/managers/ThemeManager.test.ts` → **36 passed**.
- `pnpm vitest run test/unit/standalone/entry-ssr.test.ts` → **87 passed**.
- **Regression check (the important one):** I temporarily reverted the two production call sites (`applyBrowserThemeColorFromCss` in the `skipWrite` branch and in the normal theme-write branch) and re-ran the suite. All three new `theme-color` tests failed:
  - `sets the theme-color meta to the active theme's background`
  - `does not clobber a dark #sb-theme-styles tag with the light default on boot`
  - `updates theme-color when the app background color is customized`

  So these tests genuinely exercise the fix rather than passing regardless. The SSR test (`does not inject prefers-color-scheme theme-color tags…`) is likewise a true guard: on the old code it would match the removed `media=` tags and fail. Tests mock only at the real boundary (`SettingsManager` over a fake `login`) and assert on observable DOM state, in line with the repo's testing guidance.

---

### Findings by severity

#### 🟠 Blocker to merge (not a code defect)

**No. 1 — The PR currently has a merge conflict.** GitHub reports the branch as not mergeable (`mergeable_state: dirty`) against `develop`. It will need `develop` merged in and the conflict resolved before it can land. No reproduction needed — it's visible in the PR's merge box.

#### 🟡 Low

**No. 2 — CHANGELOG entry doesn't reference the issue.** The new line under `### 🐛 Fixed` reads:
> Fix the installed Android app's status bar staying white when Seed Bible is in dark theme…

Every neighboring entry links its issue/PR (e.g. `([#1552](…))`), but this one has no `([#1765](…))` link. Suggested fix: append `([#1765](https://github.com/HelloAOLab/seed-bible/issues/1765))` for consistency and traceability.

**No. 3 — Test isolation in the `ThemeManager storage` describe block.** These tests share one jsdom `document` and there's no `beforeEach`/`afterEach` that resets the `theme-color` meta or the `#sb-theme-styles` tag for the block as a whole; individual tests clean up ad hoc at their own start (some do, some don't — e.g. `does not clobber…` never clears prior `theme-color` metas). I confirmed the tests still fail correctly on unfixed code today, so this isn't hiding anything *now*, but relying on run order contradicts the "keep tests independent: no dependence on run order" rule in CLAUDE.md and could let a future reorder mask a regression. Suggested fix: add a block-level `afterEach` that removes any `meta[name="theme-color"]` and `#sb-theme-styles` so each test starts from a known-empty head.

#### ⚪ Nit / informational

**No. 4 — Custom background values reach `theme-color` unvalidated.** `parseThemeBackgroundColor` copies whatever `--sb-background` holds into the meta, and custom color overrides are only checked for being non-empty (see `filterValidColorOverrides`), not for being a valid CSS color. In practice this is harmless — a browser silently ignores an invalid `theme-color` and falls back to the manifest color — and it mirrors how these values already flow into CSS, so no change is required. Noting only for completeness.

---

### On the fix's real-world behavior

The three converging update points are the right shape for this problem, and the comments capture the tricky cases well (the `skipWrite` first-run path that must *not* clobber the dark CSS the inline script painted, and the "cached document whose script predates this update" scenario handled by iterating and stripping `media` from every existing tag). One thing worth a manual on-device pass before shipping (as issue #1765's own testing notes call out, this can't be reproduced on a branch preview — only a `main`/production install): after switching to dark theme, fully closing the app, and relaunching from the home-screen icon, confirm the status bar background and its clock/battery/signal icons are legible in both themes. The unit tests lock in the meta *value*; only a real install confirms Android honors it for the status bar.

---

KallynGowdy:
This is an improvement, but it doesn't fix the issue for Android Chrome.

It looks like this has been fixed and is currently available in Chrome Beta: https://issues.chromium.org/issues/554055703
We just have to wait for it to be released to stable. I've tested Chrome stable and beta and can confirm that the beta fixes the issue with this PR

PR #1824, opened 2026-09-16
chore: Migrate build tooling to Vite+ and TypeScript 7

VictorUvarov:
## Context

Replaces this repo's lint, format, and type-check tooling — ESLint, Prettier, tsc-silent, husky, and a leftover Jest setup — with Vite+ (oxlint, oxfmt, `vp hooks`) and TypeScript 7's native compiler, without changing what any check catches.

## Why

- `check:ts` runs in about 3 seconds instead of 14, since TypeScript 7's native compiler replaces both `tsc` and the `tsc-silent` wrapper that slowed it down further.
- `tsc-silent` was suppressing 6 real type errors. This PR fixes them instead of hiding them.
- One config block in `vite.config.ts` replaces five separate tools and their configs (ESLint, Prettier, tsc-silent, husky, pretty-quick), which is less to keep in sync with `develop` going forward.
- Verified no regression: same lint rule coverage and warning counts, same formatting output (only ~11 files had trivial reflow differences against the old Prettier output), same test suite green, same build output.

## What changed

- Replaced ESLint and Prettier with oxlint and oxfmt (same formatting rules, same lint coverage; custom i18n and hydration lint rules ported to oxlint's JS plugin API)
- Replaced tsc-silent with TypeScript 7's native compiler for `check:ts`
- Replaced husky + pretty-quick with `vp hooks` / `vp staged` for pre-commit checks
- Moved the i18n JSON checks (unused and incomplete translation keys) from ESLint rules into a standalone script
- Removed the unused Jest/Babel setup (Vitest has been the actual test runner)

## Test

- [ ] `pnpm test` passes
- [ ] `pnpm lint` passes with the same warning counts as before
- [ ] `pnpm check:ts` passes
- [ ] `pnpm build` completes (client, SSR, server, patterns)
- [ ] `pnpm dev` starts and serves the app


Blitzy:
Talked with @KallynGowdy about this PR a bit. We are going to hold off on changing the build tooling until later when we are in a more stable position and things arnt moving so quickly. Nothing is fundamentally broken right now with the pipeline so we arnt going to try and change it at the moment.

PR #1825, opened 2026-09-16
Feature/1766

### LINKED ISSUE #1766: Preserve scroll position in navigation history
Currently, if you scroll halfway down a chapter, navigate to another chapter (via the Bible Selector), and then use the browser back button to go back to the previous chapter, then it will always take you to the top of the last chapter you were at. Instead, it should preserve your previous scroll position so that it takes you back to exactly where you were. 

### Steps to reproduce 

1. Go to Psalm 119
2. Scroll down the chapter a bit
3. Open the Bible Selector and go somewhere else.
4. Press the web browser's back button.
5. See that it takes you back to Psalm 119 but at the top of the chapter instead of where you were.

Kabi4:
Fixes: #1766

This support not only last chapter but multiple chapter stacked over another pr [PR](https://github.com/HelloAOLab/seed-bible/pull/1727) please merge that before looking into this @KallynGowdy @Blitzy


Blitzy:
> [!NOTE]

## Code Review — PR #1825 (Fixes #1766: preserve scroll position in navigation history)

**Short version:** The core of the scroll-restore feature is well thought out, the comments are excellent, and the tests are real regression tests (they fail on the pre-fix code). The main things to sort out before merge are structural: the PR currently can't merge (conflicts with `develop`), and it carries a large, unrelated session-toast rework that belongs to the stacked PR. There are also a couple of behavior questions worth confirming against what issue #1766 actually asked for.

---

### 🔴 Blocking / merge-readiness

**No. 1 — The PR is not mergeable and has no passing CI.**
GitHub reports `mergeable_state: "dirty"` (i.e. it conflicts with `develop`), and the combined commit status is `pending` with zero checks recorded on the head commit `08a47b0`. As the author notes in the description, this is stacked on [#1727](https://github.com/HelloAOLab/seed-bible/pull/1727) and that should land first.
*Suggested fix:* Once #1727 merges, merge `develop` back into `feature/1766`, resolve conflicts, and confirm the test + lint + type-check jobs go green before requesting human review.

**No. 2 — The diff bundles the session-toast rework, which is out of scope for #1766.**
Roughly half the changed lines (`SeedBibleStateManager.tsx`, the four new `session-*` i18n keys, and the `SeedBibleStateManager.test.ts` cases) are the "host disconnected / you lost connection" toast work, which is unrelated to scroll-position restore and appears to come from the stacked branch. This makes the PR harder to review in isolation and couples two features in one history.
*Suggested fix:* Rebase onto `develop` after #1727 merges so this PR's diff shrinks to only the scroll-position changes (`TabsLayout.tsx`, `BibleReadingManager.tsx`, `NavigationManager.tsx`, `TabsManager.tsx` + their tests). If the session changes are genuinely meant to ship here, please say so explicitly so they can be reviewed on their merits. I've reviewed those files lightly below but treated them as belonging to #1727.

---

### 🟡 Behavior / design questions

**No. 3 — Restore now fires on *every* return to a chapter, not just the browser Back button.**
Issue #1766 and the discussion on it framed this as browser-history restore ("it relies on the browser history tracking … only apply on a single session on a single device"). The implementation adds a second, independent mechanism: an in-memory `chapterScrollByKey` map (`BibleReadingManager.tsx`) that restores your last offset whenever you re-enter a chapter — including a *fresh forward* navigation from the Bible Selector or Next/Previous, not only Back. The CHANGELOG documents this as intended, but it's broader than the issue asked for.

Concrete consequence to confirm is desired:
1. Read a chapter and scroll to the middle.
2. Navigate away, then deliberately re-open that same chapter from the Bible Selector (a fresh visit, expecting the top).
3. You land in the middle where you last were, not at the heading.

This is defensible ("take me back where I was"), but it's a product call. Please confirm with @KallynGowdy that selector/Next-Previous restore is wanted, or scope the restore to history-driven navigation only if not.

**No. 4 — Please verify restore on the longest chapters (Psalm 119 — the issue's own example).**
Restoration applies the offset immediately (against a possibly-shorter placeholder) and then re-applies once via a single `requestAnimationFrame` when the matching chapter text arrives (`TabSlotReader` in `TabsLayout.tsx`). For a very long, progressively-rendered chapter, one rAF may fire before final layout height is settled, so `scrollTop = offset` could still be clamped short. The unit test mocks rAF to run synchronously, so it can't catch this.
*Repro to check by hand (matches the issue steps):* Go to **Psalm 119**, scroll deep into the chapter, open the Bible Selector, go elsewhere, then press browser Back. Confirm you land exactly where you were and not partway/short. If it clamps, a second rAF or a "re-apply until `scrollHeight` stops growing" guard would make it robust.

---

### 🟢 Minor / low severity

**No. 5 — The `fromHistory > 0` guard discards a legitimately-stamped `0`.**
In `applyPosition` (`BibleReadingManager.tsx`), a Back navigation only honors the stamped history offset when `fromHistory > 0`; a stamped `0` falls through to the in-memory map instead. Because the map is updated on push-navigations but the popstate (Back/Forward) path doesn't re-stamp the entry it leaves, a scroll → Forward → Back sequence can restore a stale non-zero offset where a native browser would restore the top. Very edge-casey, but the intent reads as "use history when it has a value," which is `fromHistory !== undefined`, not `> 0`.
*Suggested fix:* Distinguish "no stamp" from "stamped 0" — e.g. treat `typeof fromHistory === "number"` as authoritative for history-driven navigations, and reserve the map fallback for `undefined`.

**No. 6 — CHANGELOG entry has no PR/issue reference.**
The new "Restore your place in a chapter…" line under 🐛 Fixed omits the `([#1825](…))` link that neighboring entries carry. Small consistency nit.

**No. 7 — Orphaned `session-host-reconnecting` key in non-English locales.**
`session-host-reconnecting` was removed from `en.json` but still exists in ~75 other locale files. That's consistent with the project convention (only edit `en.json`; leave the rest to translators), so no action needed here — just flagging it so it isn't a surprise. (Belongs to #1727 scope.)

---

### ✅ Tests

The scroll-position tests are genuine and test observable behavior, not internals — and they do fail on the pre-fix code:

- `BibleReadingManager.test.ts` → "chapter scroll memory" (restore after next/previous, after a selector round-trip, after visiting >3 chapters, and *not* when scrolling to a verse) assert the `scrollPosition` signal value; without the map they'd read `0` and fail.
- `NavigationManager.test.ts` → stamps the current entry and starts the next push at `0`, and doesn't stamp after dispose — real history-state assertions.
- `TabsManager.test.ts` → drives an actual `window.history.back()` and asserts the offset comes back (240), plus the corrected `replaceSpy` count (3 → 4) for the new origin-stamp.
- `TabSlotReader.test.tsx` → "re-applies the saved scroll offset when the matching chapter text arrives" asserts `scrollTop` is written twice (`[320, 320]`); removing the content-match re-apply block drops it to `[320]` and the test goes red — a proper regression guard. Only caveat is the synchronous rAF mock (see No. 4).

The session-toast tests (`SeedBibleStateManager.test.ts`) also look like solid, scenario-driven regression tests for the "remaining host device" and "airplane rejoin" bugs, but I've treated them as #1727's to verify.

---

_Reproduction steps above are best-effort from static review; the runtime clamping in No. 4 in particular is worth a hands-on check on a real long chapter._

---

Blitzy:
@KallynGowdy What do you think of preserving the scroll when re-entering chapters not via browser history navigation? For example this PR will restore history when using the next/prev chapter buttons. Personally I don't like it, I feel like scroll preservation should be limited to browser history navigation, and we should otherwise start at the top of the chapter. On desktop especially this feels like a bug to restore scroll position when using next/prev chapter buttons because the book+chapter are at the top of the reader (as opposed to mobile having it in the navigation bar alongside the chevron buttons) so you can flip through chapters and just see body text change and not actually get a good sense of what book+chapter you are on.

KallynGowdy:
I agree. It should only preserve the scroll position when using the browser back/forward buttons. The previous/next chapter buttons don't carry the idea of "take me back where I was", but the browser buttons do. That is why we want to preserve the scroll position for the browser buttons but not the regular ones.

Blitzy:
## Follow-up: restore is now browser-history only, plus two bugs found while testing it

**Short version:** @KallynGowdy's call is implemented — scroll position is restored only for the browser's back and forward buttons, and every other way into a chapter starts at the heading. Scoping it that way exposed two further bugs, one of them latent in the original implementation and hidden by the code this change removed. Both are fixed. The full manual test plan passes by hand, and CI is green on `4140dfc`.

Three commits since that decision:

| | |
|---|---|
| `f751e82` | Restore scroll position only for browser history navigation |
| `3d9dfb2` | Keep the current history entry's scroll offset up to date |
| `4140dfc` | Read the restored scroll offset after the history entry is current |

---

### 1. Restore is limited to browser back/forward (`f751e82`)

The in-memory per-chapter map is gone. The browser history entry is now the only place a scroll offset is recorded, which makes the rule short enough to state in full:

1. A verse link wins — it goes to the verse, ignoring any recorded offset.
2. Otherwise, if you arrived via back/forward, go to the offset recorded on the entry you landed on.
3. Otherwise, the chapter heading.

So Next/Previous, the Bible Selector, and a translation switch all start at the top now.

Two things fell out of this that are worth calling out separately, because both were reported as defects:

- **Switching translation on the same chapter used to leave you mid-chapter.** Only a book/chapter change counted as a position change, so a translation switch neither reset nor recorded the offset. It now counts like any other position change and starts at the heading.
- **Going back across a translation switch returned the wrong translation's place.** The switch never recorded translation A's offset before moving on. It does now.

This also resolves review finding **No. 5**: an offset of `0` is honoured as "the top of the chapter" rather than falling through to an older value, which is what a native back button does. That turned out to matter far more than it looked — see No. 3 below.

### 2. Forward stopped working (`3d9dfb2`)

Removing the map broke forward, and the map had been quietly covering for it.

The offset was only ever written onto a history entry immediately *before* a push. A back or forward never re-stamps the entry it leaves. So: scroll down John 4, press Back to John 3, press Forward — John 4 came back at the **top**, because its entry still held the zero it was pushed with.

The fix writes the offset onto the current entry as the reader scrolls, debounced so one scroll gesture costs one `replaceState`. The push-time stamp is kept, because it covers the case where you scroll and navigate away inside the debounce window.

### 3. Restore was running one navigation behind (`4140dfc`)

This is the interesting one, and it was latent in the original implementation rather than introduced here.

**Symptom:** pressing Back landed at the top of the chapter, and pressing Forward then scrolled to the offset Back should have used.

**Cause:** Chrome's Navigation API fires its `navigate` event *before* the browser swaps the history entry in, and `NavigationManager` published the new URL from there. The URL→state effects run synchronously off that publish, so when they asked for the scroll offset they read `window.history.state` while it still described **the entry being left**. Confirmed from inside the handler: during a Back to John 3, `location.pathname` still read `/en/AAB/john/4`.

The destination's own offset is not readable from that event either — `destination.getState()` returns the Navigation API's state, which `pushState` never writes, so it comes back `undefined`.

**Fix:** back/forward is left to the `popstate` listener, which already existed for browsers without the Navigation API and fires once the entry is current, where both the location and its state read correctly. Other same-origin navigations are published from `navigate` exactly as before.

**Why nobody saw it earlier:** the per-chapter map was masking it. A stale offset of `0` failed the old `fromHistory > 0` check and fell through to the map, which happened to hold the right number. Make history authoritative — which finding No. 5 correctly asked for — and the stale read becomes visible.

### A testing gap worth flagging

jsdom has no Navigation API, so the entire suite took the `popstate` path and never touched the broken code. **6,362 tests passed while back/forward restore was broken in every Chromium browser.** It was only found by driving the real app in a browser.

Anything that depends on the ordering between a navigation event and the history entry becoming current cannot be trusted to the unit suite as it stands. I have not audited whether other URL→state reads go through that same pre-commit window; only the scroll-offset read was fixed here.

### Tests

- `BibleReadingManager.test.ts` — the old "chapter scroll memory" block is replaced by "scroll position on navigation" (6 cases): next/previous and the selector start at the heading, a history-supplied offset restores, a stamped `0` is honoured, a translation switch on the same chapter starts at the heading, and a linked verse still wins.
- `NavigationManager.test.ts` — two cases that stub `window.navigation`: a traverse must not publish the URL early, and a non-traverse same-origin navigation still does.
- `TabsManager.test.ts` — forward returns the reader to where they were.
- `SeedBibleStateManager.test.ts` — the hand-rolled shared-session reading-state mock gained the `scrollPosition` signal it now stands in for.

Every new test was verified red against the pre-fix code before being kept.

### Manual testing

A 24-case plan was worked through by hand against the branch and all cases pass. That includes the Psalm 119 deep-scroll case from review finding **No. 4** — the concern that a single `requestAnimationFrame` might apply the offset before the longest chapter in the Bible has settled its height. It lands correctly.

Known limits, confirmed rather than fixed: what is recorded is a pixel distance, not a verse, so changing the reading text size or resizing the window substantially between leaving and returning lands you at a different verse.

### Review findings

- **No. 1** (not mergeable, no CI) — `develop` is merged in and CI is green on `4140dfc`.
- **No. 2** (unrelated session-toast work in the diff) — resolved by #1727 landing; this PR's diff is now the scroll-position change only.
- **No. 3** (restore fires on every return, not just Back) — this is the change above.
- **No. 4** (verify on Psalm 119) — verified by hand.
- **No. 5** (`fromHistory > 0` discards a stamped `0`) — fixed, and it turned out to be the thread that led to No. 3 above.
- **No. 6** (CHANGELOG entry missing its PR link) — added, and the entry is rewritten to describe the history-only behaviour.

The PR description still says #1727 should land first. Leaving it as-is deliberately — it is accurate history for how the branch was built.

---

Blitzy:
> [!NOTE]

## Follow-up review — PR No. 1825 at `4140dfc`

This re-reviews the branch after the three follow-up commits (`f751e82`, `3d9dfb2`, `4140dfc`) that resolved the earlier review thread and @KallynGowdy's call to scope scroll-restore to browser back/forward only. The mergeability blocker, the scope issue, and findings No. 3–6 from the earlier round are all resolved as described in the author's write-up, and I didn't find further problems in `NavigationManager`'s `navigate`/`popstate` split or `BibleReadingManager`'s `applyPosition`. One real bug remains in the newly-added scroll-stamping debounce, plus a small duplication note tied to it.

### 🟡 Bug — pending scroll-stamp timer isn't cancelled when the tab closes

**File:** `packages/seed-bible/seed-bible/managers/TabsManager.tsx`, the debounce effect around line 908-925 (the one added in `3d9dfb2` to keep the current history entry's offset up to date while scrolling).

```ts
let scrollStampTimer: ReturnType<typeof setTimeout> | undefined;
effect(() => {
  const tab = selectedTab.value;
  if (!tab) {
    return; // <-- clearTimeout(scrollStampTimer) never runs on this path
  }
  const offset = tab.readingState.scrollPosition.value;
  clearTimeout(scrollStampTimer);
  scrollStampTimer = setTimeout(() => {
    ...
    navigation.stampCurrentState({ scrollPosition: offset });
  }, SCROLL_STAMP_DEBOUNCE_MS);
});
```

The early `return` when `tab` is falsy skips the `clearTimeout` a few lines below. `selectedTab` is a computed (`TabsManager.tsx` ~line 565) that only becomes `null` when `tabs.value` is empty — which happens when the *last* tab is closed: `removeTab` (`TabsManager.tsx` ~line 1171) sets `nextTabs = []` and falls through to `selectedTabId.value = nextTabs[currentTabIndex - 1]?.id ?? nextTabs[0]?.id ?? ""`, i.e. `""`, and no tab has that id. I checked `requestCloseTab` in `Tabs.tsx` (~line 722) and it calls `state.tabs.removeTab(tab.id)` unconditionally — there's no guard preventing the last remaining tab from being closed via the normal close ("x") control.

**Repro:**
1. With a single tab open, scroll the chapter (e.g. to 400px).
2. Immediately close that tab (the only tab) before `SCROLL_STAMP_DEBOUNCE_MS` elapses.
3. `selectedTab.value` goes straight to `null`; the effect re-runs, hits the early return, and the previously-scheduled timer (still closing over the stale `offset = 400`) is never cleared.
4. ~200ms later the timer fires anyway and calls `navigation.stampCurrentState({ scrollPosition: 400 })`, writing 400 onto whatever history entry is current at that point (e.g. a newly created default tab, or an unrelated page) — an entry that has nothing to do with the closed tab's scroll position. Revisiting that entry via Back/Forward later lands mid-page instead of at the top for no reason tied to its own history.

**Suggested fix:** move `clearTimeout(scrollStampTimer)` above the `if (!tab)` check (or return the cleanup from the effect itself) so a falsy `tab` still cancels any pending stamp:

```ts
effect(() => {
  clearTimeout(scrollStampTimer);
  const tab = selectedTab.value;
  if (!tab) {
    return;
  }
  ...
});
```

### 🟢 Minor — hand-rolled debounce duplicates an existing helper

Same effect: it hand-rolls `setTimeout`/`clearTimeout` for a "wait for the user to stop, then persist" pattern that `ReadingHistoryManager.tsx` already solves with `import { debounce } from "es-toolkit"` (see `saveReadingHistoryForCurrentUser`, `ReadingHistoryManager.tsx` line 1332). Reusing `debounce`'s built-in `.cancel()` here would also have sidestepped the bug above for free, since a maintained debounce cancels correctly regardless of which branch returns early. Not blocking, but worth considering since it's a very close match to an existing pattern in the same manager family (per this repo's CLAUDE.md guidance on reusing nearby helpers rather than writing a parallel version).

---

Blitzy:
Both findings addressed in `c8cb915` — and they turned out to be the same fix, which is the useful part.

### The bug: pending stamp survives the last tab closing

Confirmed exactly as described. `removeTab` (`TabsManager.tsx:1183-1190`) sets `tabs.value = []` and then `selectedTabId.value = ""`, so `selectedTab` computes to `null`; the effect re-ran, hit `if (!tab) return`, and skipped the `clearTimeout` below it. The scheduled timer still closed over the departing tab's offset and fired ~200ms later onto whatever entry was current by then.

### The minor: hand-rolled debounce duplicated an existing helper

Taking this one fixes the bug rather than just tidying it, which is why I did both together. The effect now uses `debounce` from `es-toolkit` — the same helper `ReadingHistoryManager.tsx:1332` uses for the same "wait until the user stops, then persist" shape:

```ts
const stampScrollPosition = debounce((offset: number) => {
  if (parseStaticPagePath(navigation.currentUrl.peek().pathname, navigation.basePath)) {
    return;
  }
  navigation.stampCurrentState({ scrollPosition: offset });
}, SCROLL_STAMP_DEBOUNCE_MS);

effect(() => {
  const tab = selectedTab.value;
  if (!tab) {
    // Closing the last tab leaves nothing selected. Drop any stamp still
    // waiting, or it lands the closed tab's offset on whatever entry is
    // current when it fires.
    stampScrollPosition.cancel();
    return;
  }

  stampScrollPosition(tab.readingState.scrollPosition.value);
});
```

The review's point that the shared helper "would also have sidestepped the bug above for free" is the right reason to prefer it here: cancelling is a method on the debounced function rather than a timer handle that an early return can skip past, so the same mistake is harder to make again.

### Test

`TabsManager.test.ts` gains a case that scrolls the only tab, closes it before the debounce elapses, and waits well past the window. Verified red against the pre-fix code (`expected 400 not to be 400`) and green after.

One note on how it asserts, since it looks weaker than it is: it checks that `400` specifically never reaches `history.state`, rather than that the state is unchanged. Managers built by earlier cases in that file are never disposed and go on stamping their own zero offsets, so "unchanged" is not a stable assertion there — but `400` belongs to this tab alone, and pre-fix it lands. That's a pre-existing test-hygiene wrinkle in the file, not something this PR introduces; worth a separate cleanup if it bites again.

### Verification

`6,365` tests pass, 0 type errors, 0 lint errors, Prettier clean. I also re-ran back/forward restore in a real browser after the change, since swapping the debounce mechanism is exactly the kind of edit the unit suite would not catch if it went wrong — John 3 at 1200 and John 4 at 2600 both still restore in the right direction.

---

Blitzy (approved):

PR #1826, opened 2026-09-16
Added Auto Focus on tiptap editor used it on annotaiton level both sc…

### LINKED ISSUE #1793: Auto focus annotation input when opening editor
When opening the annotation editor to create a new or edit an existing annotation, the input should auto focus so that you can start typing right away.

Kabi4:
…enarios

Fixes: #1793


Blitzy:
> [!NOTE]

## Code review — PR #1826 (Auto-focus the annotation editor)

**What this PR does:** adds an `autofocus` prop to `TipTapEditor` (typed `FocusPosition`) that, on mount, calls `editor.commands.focus(position, { scrollIntoView: false })` when set. `CreateAnnotationForm` passes `autofocus="end"`, so opening the annotation composer — for a new note or when editing an existing one — now places the caret in the editor so you can type immediately. Addresses #1793.

**Overall:** clean, well-scoped change that does what the issue asks. Typecheck passes and all touched tests pass. I verified the core test is a genuine regression test (details below). The findings below are one merge blocker and a couple of things worth a deliberate decision — no correctness bugs in the implementation.

---

### 🟠 Medium

**No. 1 — Merge conflict in `CHANGELOG.md` blocks merge.**
GitHub reports this PR as `dirty` (not mergeable). The conflict is entirely in `CHANGELOG.md`: the new autofocus entry collides with entries that have since landed on `develop`.
- *Reproduce:* `git checkout feature/1793 && git merge origin/develop` → `CONFLICT (content): Merge conflict in CHANGELOG.md`.
- *Fix:* merge (or rebase) `develop` into the branch and keep both changelog entries. Nothing else conflicts.

**No. 2 — Autofocus is unconditional, so it will pop the soft keyboard on mobile.**
The codebase already has an explicit, deliberate convention against this. In `ChatView.tsx` (around line 733):
```ts
// Don't autofocus on mobile — focusing opens the soft keyboard.
if (state.app.isMobile.value) {
  return;
}
inputRef.current?.focus();
```
The new annotation autofocus has no such guard, so on a phone, opening the annotation editor will immediately raise the keyboard and shift the layout before the user has decided to type. That may be acceptable for annotations (the issue does say "start typing right away"), but it's inconsistent with the existing chat behavior and deserves a conscious choice rather than falling out by omission.
- *Reproduce:* open the create/edit-annotation screen on a mobile viewport → the soft keyboard opens on entry.
- *Options:* (a) confirm this is the intended mobile behavior and leave it; or (b) gate it on `state.app.isMobile`. `CreateAnnotationForm` doesn't currently receive `state`/`isMobile`, so option (b) would need that plumbed through — worth weighing against the small UX benefit. Flagging for a product decision, not asserting it's wrong.

---

### 🟡 Low / nits

**No. 3 — The two `CreateAnnotationForm` autofocus tests are effectively the same test.**
Both "autofocuses when creating a new annotation" and "autofocuses when editing an existing annotation" assert the identical thing (`expect(latestAutofocus).toBe("end")`), and the component passes `autofocus="end"` unconditionally regardless of new-vs-edit. So the "editing" test's extra setup (seeded `html`) doesn't exercise any distinct code path — both pass for the same reason. They're fine as documentation of the issue's two scenarios, but the editing case adds no real coverage beyond the create case. Not a bug; just noting the redundancy.

**No. 4 — Title/description typo.** "annotaiton" in the PR title and commit message. Cosmetic.

---

### ✅ Things that are done well

- **The `TipTapEditor` tests test real behavior, not mocks.** "autofocuses the editor when autofocus is set" asserts on `editor.isFocused` and `document.activeElement === editor.view.dom` after flushing the animation frame. I confirmed it's a true regression test: reverting the focus block locally turns it red (`isFocused` false→true is exactly the behavior under test). The paired negative test ("leaves the editor unfocused when autofocus is unset") guards against accidentally making focus always-on. This matches the repo's testing guidance (assert observable behavior, cover the negative path).
- **The falsy-position guard is correct.** `if (autofocusPosition !== false && autofocusPosition != null)` deliberately uses strict `!== false` / `!= null` rather than a truthiness check, so a numeric caret position of `0` still focuses. A `!autofocusPosition` shortcut would have silently broken that case.
- **`scrollIntoView: false`** is a sensible touch — it places the caret without yanking the pane into a scroll on open.

---

PR #1827, opened 2026-09-16
Fixes Blend issue

### LINKED ISSUE #1800: Add a box shadow to the mobile search verses dialog
Currently, the box of the search verses dialog blends into the background on mobile. We should add a box shadow or some sort of gradient background so that the search box is differentiated from the scripture background.

Image

Kabi4:
Fixes: #1800


Kabi4:
image
I Tried two types of shadow sharp  fading i felt sharp shadow is what we need but upto all open changes

Blitzy:
> [!NOTE]

## Code Review — PR #1827 "Fixes Blend issue"

**Summary:** This is a small, CSS-only change that addresses issue #1800 (the mobile "Search verses" box blends into the scripture behind it). It adds a box shadow to the search input wrap inside the `max-width: 768px` media query. Overall the fix is correct, minimal, and matches exactly what the issue asked for ("add a box shadow or some sort of gradient background"). I found no blocking issues.

### Why the fix is sound
- On mobile, the panel itself is deliberately transparent (`background: transparent; box-shadow: none; border: none`) so it floats over the reader, and the input wrap keeps `background: var(--sb-reader-background)` — the *same* color as the scripture behind it. That is precisely why it blended. Outlining the box with a ring + drop shadow gives it a visible edge without changing the layout.
- Both tokens used resolve well in **both** themes, so the differentiation holds in dark mode too:
  - `--sb-divider-color`: `rgba(0,0,0,0.12)` (light) / `rgba(255,255,255,0.1)` (dark → a subtle *light* ring on the dark surface).
  - `--sb-shadow-color`: `rgba(0,0,0,0.14)` (light) / `rgba(0,0,0,0.6)` (dark).
- No new `--sb-*` variable is added and `main.css` isn't touched, so the ThemeManager/main.css "update both places" gotcha in CLAUDE.md does not apply here.
- No parent of `.sb-floating-search-input-wrap` sets `overflow: hidden` (the `overflow-y: auto` is on the sibling `.sb-floating-search-results`), so the shadow won't be clipped. There's also no competing `:focus-within` box-shadow rule that this would override.
- CHANGELOG entry is correctly placed under the unreleased "🐛 Fixed" section.

---

### Findings by severity

**High / Medium:** None.

**Low — worth a quick on-device check (not blocking)**

No. 1 — The box interior still uses `--sb-reader-background`, identical to the scripture fill, so the separation comes entirely from the edges (a 1px ring + a soft `0 2px 4px` drop shadow). In **light** mode both are faint (`0.12` ring, `0.14` shadow). This should read as "lifted," but since the issue offered a stronger alternative (a gradient/distinct background fill), it's worth confirming on a real device/light theme that the box now reads as clearly separated rather than just barely outlined. If it still feels flat, bumping the drop-shadow blur/spread or giving the wrap a slightly distinct surface color would strengthen it.

*Reproduction of the original bug (to verify the fix visually):*
1. Open the app on a mobile viewport (≤768px wide), e.g. the deploy preview linked on this PR.
2. Open the reader and tap the Search verses control so the input sheet appears over the scripture (before typing, while results are collapsed).
3. Before this change: the input box's edges are indistinguishable from the verse text/background. After: the box has a visible ring + shadow. Repeat in both light and dark themes.

---

### On tests

There are no tests in this PR, and for this change that's the right call — it's purely presentational (a `box-shadow` declaration). The rendering can't be meaningfully asserted in the jsdom/Vitest setup used here (no layout/paint computation), so a unit test would either assert the literal CSS string (a tautology) or silently pass regardless of correctness. Importantly, there is **no** always-green test being added to imply coverage that isn't there. The appropriate verification is the manual/visual check above.

---

PR #1828, opened 2026-09-16
Fix translate x for swipping issue

### LINKED ISSUE #1811: Changing from mobile layout to a larger layout causes the reader to be partially offscreen
Steps to reproduce:

1. View in mobile layout
2. Change chapters by swiping
3. Change to a larger layout (landscape orientation)
4. Notice that the reader now appears partially offscreen.

Kabi4:
Fixes: #1811


Blitzy:
> [!NOTE]

## Summary

**This PR correctly fixes #1811 and ships a genuine regression test. I recommend it for merge; the notes below are low-severity / informational only — no correctness bugs found.**

**Root cause (in plain terms):** While reading on mobile, a swipe writes an inline `transform: translateX(...)` onto the swipe track to slide chapters. When you then rotate to a larger layout, Preact re-renders the reader from the mobile branch to the desktop branch. Because both branches were plain, unkeyed fragments (`<>...</>`), Preact matched their children by position and **reused the same DOM node** — the mobile swipe track became the desktop content node, dragging the leftover `translateX` along with it. That is why the reader sat partly offscreen.

**The fix works on two fronts:**
1. `BibleReader.tsx` — the mobile and desktop branches are now `<Fragment key="mobile">` / `<Fragment key="desktop">`. Distinct keys make Preact unmount the mobile subtree and mount a fresh desktop subtree instead of reusing nodes, so no mobile inline style can leak across.
2. `TabsLayout.tsx` — a new `clearSwipeTrackInlineStyles` helper strips the inline `transition`/`transform` off the track in three spots: the touch-effect cleanup, the translation-change effect, and a render-time clear when `!isMobile`.

## Verification I performed

I checked out the branch, installed deps, and ran the suite:

- **The new test is a real regression test.** I reverted only the two source files (keeping the new test) and it failed exactly as the bug predicts:
  `expected 'translateX(-66.66666666666667%)' to be ''` on `.sb-bible-reader-main-content`. Restoring the fix turns it green. So the test genuinely reproduces the reported behavior and would not "always pass."
- **No regressions.** All 32 tests in `TabSlotReader.test.tsx` pass, and the 191 tests in `BibleReader.test.tsx` / `BibleReaderToolbar.test.tsx` pass with the Fragment-key change in place.

## Findings by severity

### Low — The two fix mechanisms are each independently sufficient
I tested them in isolation: with **only** the Fragment-key change (TabsLayout reverted) the test passes, and with **only** the TabsLayout clear (Fragment keys reverted) it also passes. So the two changes overlap for the reported scenario rather than each covering a distinct gap.

This is not a defect, but two things follow:
- The Fragment-key change is the more general fix — it stops *any* stale mobile DOM state (not just this transform) from bleeding into the desktop node, so it is the one I would keep as the primary guard.
- Because either fix alone makes the test green, the test cannot tell you if one mechanism is later removed or broken. If the team values defense-in-depth, keeping both is reasonable — a one-line comment noting they are deliberately redundant would save a future maintainer from "cleaning up" one and assuming the test still protects it.

### Low — Render-phase DOM mutation is a slight impurity
In `TabSlotReader` the body runs, before `return`:

```tsx
if (!isMobile) {
  clearSwipeTrackInlineStyles(swipeTrackRef.current);
}
```

Mutating the DOM during render (rather than in an effect) is normally something to avoid — render is meant to be side-effect free. In Preact this is safe in practice (no concurrent rendering), and the inline comment explains the *why* well (the ref still points at the old track during render, before commit nulls it). Worth being aware that it executes on every desktop render, harmlessly no-op'ing once `swipeTrackRef.current` is `null`. Given the Fragment-key change already prevents the node reuse, this block is the part that is most arguably redundant (see the finding above); if you drop one mechanism, this is the natural candidate.

### Low — Test covers LTR only
The regression test exercises a left-swipe on an LTR translation, landing the track at `translateX(-66.67%)`. The same leak is theoretically possible for an RTL translation, where the rest transform is positive. The Fragment-key fix covers both directions structurally, so this is a hardening suggestion, not a gap that hides a live bug. If you want the test to speak to RTL too, a second case with `textDirection: "rtl"` and a right-swipe would do it.

## Reproduction (for reference)
From #1811, on the pre-fix code:
1. Open the reader in a mobile (narrow) layout.
2. Swipe to change chapters.
3. Rotate to a larger / landscape layout.
4. The reader renders partially offscreen (shifted by ~⅓ of the track width).

After this PR, step 4 shows the reader correctly centered.

---

PR #1832, opened 2026-09-16
Refactor tutorial skip flow with dedicated prompt dialog

KallynGowdy:
## Summary

Restructures the tutorial skip experience to use a dedicated follow-up dialog instead of embedding skip/opt-out options on the tour popover itself. This improves UX by separating the "skip this tour" action from the "disable all tutorials" decision, and enables code reuse across the main tour and the Bible selector's embedded tour popover.

## Key Changes

- **New `TutorialSkipPrompt` component** (`TutorialSkipPrompt.tsx` + `.css`): A modal dialog that appears after the user skips a tour, asking whether to disable all tutorials or just skip the current one. Replaces the old "Don't show tutorials" button that lived on the tour popover.

- **Extracted `TutorialPopoverContent` component**: Factored the tour popover's card body (title, skip button, description, navigation buttons, progress dots) into a reusable component. Both the main `Tutorial` component and `BibleSelector`'s embedded tour popover now use this shared markup, preventing layout drift between the two.

- **Updated `TutorialManager` API**:
  - Added `skip()` method: ends the tour and raises the skip prompt (instead of immediately opting out)
  - Added `keepTutorials()` method: dismisses the skip prompt without opting out
  - Renamed `optOut()` semantics: now only records the opt-out decision (can be called directly from Settings or from the skip prompt)
  - Added `skipPromptVisible` signal to track prompt visibility

- **Reorganized tour popover layout**:
  - Title and skip button now sit on one row (flexbox header) so skip stays pinned to the dialog's upper trailing corner
  - Back/next buttons anchored to the dialog's leading/trailing bottom corners; their position never shifts as steps change
  - Progress dots flex to fill available space between back and next
  - Removed the spacer element; layout is now purely flex-based

- **Updated i18n keys**: Removed `tutorial.optOut` key; added `tutorial.skipPromptTitle`, `tutorial.skipPromptBody`, `tutorial.skipPromptKeep`, `tutorial.skipPromptDisable` for the new dialog.

- **Added test coverage**: Three new test cases in `TutorialManager.test.tsx` covering the skip flow: raising the prompt, dismissing it with `keepTutorials()`, and confirming opt-out from the prompt.

- **Integrated into main app**: `TutorialSkipPrompt` rendered in `main.tsx` alongside the existing `Tutorial` and `TutorialPrompt` components.

## Implementation Details

The skip prompt is a fixed-position overlay (z-index 10010, above the tour overlay) with a semi-transparent backdrop. Clicking outside the dialog dismisses it (calls `keepTutorials()`), matching common modal patterns. The dialog itself stops propagation to prevent accidental dismissal.

The `TutorialPopoverContent` component is a fragment that renders the popover's internal structure (header, body, actions). It receives the step, tutorial manager, and navigation state as props, making it stateless and reusable. The main `Tutorial` component wraps it in a div; the `BibleSelector` renders it directly inside its shadow-root portal.

The tour popover's flex layout now uses `margin-inline-start: auto` on the next button to pin it to the trailing edge regardless of whether back and/or progress dots are present, eliminating the need for a spacer element and making the layout more robust to content changes.

https://claude.ai/code/session_01BEyLr9gunVQGd5Cruc2yPt


Blitzy:
> [!NOTE]

## Code review — Refactor tutorial skip flow with dedicated prompt dialog

**Short version:** this is a clean, well-scoped refactor and I found no correctness bugs. The `TutorialPopoverContent` extraction genuinely de-duplicates the popover markup between the main `Tutorial` and `BibleSelector`, the `TutorialManager` API changes are consistent, and the new tests exercise real behavior. Everything below is Low severity or a nit — mostly UX and accessibility considerations, not defects.

I verified locally against the head commit (`274f438`):
- `pnpm check:ts` → 0 visible errors
- `pnpm vitest run TutorialManager.test.tsx` → 9 passed
- `eslint` on the changed files → 0 errors (only pre-existing "unused translation key" **warnings** in `en.json`, which don't fail CI)

I also confirmed the stacking is correct: the skip prompt's overlay (`z-index: 10010`) sits above the elevated tour overlay (`10001`) and the selector panel (`≤1001`), so it is never hidden behind them.

---

### Test assessment

The task asked me to look closely at whether the tests reflect real behavior and don't just always pass. They hold up:

- The three new cases assert distinct, opposite outcomes — `keepTutorials()` leaves `optedOut` false and lets a *different* contextual tour (`pane-layout`) start afterward, while `optOut()` records the opt-out and blocks that same tour. Because they're complementary, a bug in either method breaks one of them.
- To confirm they aren't vacuous, I mutated the implementation (made `keepTutorials()` also call `markOptedOut()`) and the "keepTutorials leaves future tutorials enabled" test failed as expected. So the tests are wired to the behavior, not to the mocks.

One honest caveat (not a request to change anything): these cover brand-new methods (`skip`/`keepTutorials`), so they can't double as regression tests against prior behavior — there's no pre-fix code path for them to have caught. That's fine given the feature is new.

---

### Low severity

**L1 — The "Turn off tutorials?" modal now fires after skipping *any* tour, including one-off contextual tips, and its emphasized button is the destructive one.**

Before: contextual tips showed two inline buttons — "Skip" (just end this tip) and "Don't show tutorials" (opt out). After: every tip shows a "Skip" button that ends the tip **and** raises a full-screen blocking modal asking whether to disable *all* tutorials.

Two things worth a second look:
- A single-step contextual tip (e.g. the `offline-download` glow tip, or the `search` tip) now shows both "Skip" (top corner → raises the modal) and "Done" (bottom corner → silently finishes). Both call `finish()` and mark the feature seen, so from the user's point of view "Skip" and "Done" do nearly the same thing except one throws up a disable-all-tutorials modal. That's a lot of ceremony for dismissing a one-line tip, and the two buttons are easy to confuse.
- In the modal (`TutorialSkipPrompt.tsx`), the visually primary (filled accent) button is **"Turn off tutorials"**, while the safer "Just this one" is the de-emphasized secondary button. Since this modal now appears after *every* skip, a user quickly skipping tips is nudged toward the most destructive choice. Consider swapping the emphasis (make "Just this one" primary), or only raising the prompt for the onboarding tour rather than for every contextual tip.

This is a product/UX call, so I'm flagging it rather than asserting it's wrong — worth a conscious decision.

No reproduction needed beyond: start any contextual tour (e.g. trigger the `search` tip), click "Skip", observe the disable-all modal.

**L2 — The skip modal declares `aria-modal="true"` but has no focus management or Escape handling.**

`TutorialSkipPrompt` is the first *true* blocking modal in this family — the existing `TutorialPrompt` is a non-modal corner card (`aria-modal="false"`). A dialog claiming `aria-modal="true"` sets the expectation that focus moves into it on open, is trapped while open, and returns to the trigger on close, and that Escape dismisses it. None of that is implemented here, so a keyboard or screen-reader user can tab out to the page behind the overlay, and Escape does nothing.

Reproduction: skip a tour to raise the modal, then press Tab repeatedly — focus walks onto the reader UI behind the backdrop instead of cycling within the dialog; press Escape — nothing happens.

Suggested fix: on mount, focus the dialog (or its safe/secondary button) and remember the previously focused element; add a `keydown` handler that calls `keepTutorials()` on `Escape`; restore focus on unmount. A minimal focus trap on Tab/Shift+Tab within `.sb-tutorial-skip-prompt` would complete it.

---

### Nits

**N1 — `skipPromptVisible` isn't cleared when a tour (re)starts.** `finish()`, `optOut()`, and `keepTutorials()` all reset it, but `start()` / `startContextual()` don't. In practice the modal is a blocking overlay, so a new tour can't realistically be launched underneath it — this is a latent inconsistency rather than a reachable bug — but a one-line `skipPromptVisible.value = false;` in `start()`/`startContextual()` would make the invariant explicit.

**N2 — Slightly stale comment on `optOut()`.** The comment "Record opt-out first so contextual tours can't restart during teardown" made sense when the opt-out button lived on the tour popover. Now `optOut()` is reached from the skip prompt *after* `skip()` → `finish()` has already torn the tour down (so `activeFeatureId` is already `null`), and the ordering no longer guards anything. Not a bug — the code is still correct — but the rationale reads as out of date.

---

Nice work overall — the shared-component extraction is the right call and the layout comments in `Tutorial.css` are genuinely helpful for the next reader.

---

PR #1834, opened 2026-09-17
adding  whiteLabelThemes from config

AmitPant9999:


Blitzy:
> [!NOTE]

## Code review — `adding whiteLabelThemes from config`

**What this PR does:** adds an optional `whiteLabelOverrides` field to `BrandingConfig`, reads it in `createSeedBibleState`, and layers it into `ThemeManager.currentTheme` between the base preset and the user's own custom overrides. The precedence becomes: **base preset → white‑label overrides → user custom overrides → live preview**, so a deployment can ship brand colors that a user can still customize over. That layering order is sensible.

The PR has no description and no linked issue, so I reviewed against the intent implied by the title and the code.

---

### 🔴 High — The config field type is wrong; the feature does not work as typed

`appConfig.tsx` declares the field as a full theme object:

```ts
whiteLabelOverrides?: BibleTheme;   // { id, name, variables, highlightColors }
```

but `createTheme` treats it as a flat overrides map:

```ts
whiteLabelThemeOverrides: ThemeOverrides = {}   // Partial<Record<ThemeColorKey | ThemeFontFamilyKey, string>>
```

and `applyOverrides` only spreads the override object's **top‑level** keys into `theme.variables`:

```ts
return { ...theme, variables: { ...theme.variables, ...overrides } };
```

These two types are structurally incompatible in meaning. Because `ThemeOverrides` has only optional properties and no index signature, TypeScript still accepts a `BibleTheme` value there, so this passes `pnpm check:ts` (I confirmed a minimal repro compiles clean) — the bug is silent, not caught at build time.

If a config author follows the declared type and supplies a real `BibleTheme`, the actual colors live nested under `variables`, but `applyOverrides` never looks inside `variables`. Instead it spreads the theme's top‑level keys (`id`, `name`, `variables`, `highlightColors`) straight into the CSS variable map. The result:

- The intended white‑label colors are **silently ignored** (they're one level too deep to be seen).
- `generateThemeCssVariables` then emits garbage variables like `--sb-id: brand`, `--sb-name: Brand`, and `--sb-variables: [object Object]`.

The code only actually works if the host passes a **flat** `ThemeOverrides` object (color keys at the top level) — which is exactly what the declared `BibleTheme` type steers authors away from.

**Reproduction (config JSON injected as `#app-config`):**
```jsonc
"branding": {
  "whiteLabelOverrides": {           // shaped like the declared BibleTheme type
    "id": "brand",
    "name": "Brand",
    "variables": { "backgroundColor": "#123456" },
    "highlightColors": {}
  }
}
```
Expected: page background becomes `#123456`. Actual: background unchanged; junk `--sb-*` variables (`--sb-id`, `--sb-variables: [object Object]`, …) are injected.

**Suggested fix** — make the declared type match what the code consumes (`ThemeOverrides` is already re‑exported from `../managers`):
```ts
import type { ThemeOverrides } from "../managers";
// ...
whiteLabelOverrides?: ThemeOverrides;
```
With that, a host supplies a flat map (`{ "backgroundColor": "#123456", "textColor": "#fff" }`) and it flows correctly through `applyOverrides`. If the intent really is to accept a full `BibleTheme`, then `applyOverrides` (or the call site) needs to read `whiteLabelOverrides.variables` instead — but the simpler, consistent fix is the type change above.

---

### 🟠 Medium — No tests for the new behavior

`test/unit/seed-bible/managers/ThemeManager.test.ts` already exercises `createTheme(settings)` via the "ThemeManager storage (via SettingsManager)" suite, but nothing calls the new two‑argument form. Given the type bug above slipped past the type checker, a test asserting real behavior would have caught it. Worth adding cases that assert *observable* output (the resolved `currentTheme.variables` / generated CSS), not internals:

1. White‑label override changes `currentTheme` when the user has set no custom color.
2. A user custom override for the same key **wins over** the white‑label value (precedence).
3. White‑label value still shows through for keys the user has *not* customized.
4. Empty/absent `whiteLabelOverrides` leaves the base preset untouched (the `{}` default path).

To make sure a test is genuine and not always‑green: write it against the current code first and confirm test No. 1 above **fails** as written today (a `BibleTheme`‑shaped input produces no color change / junk keys), then confirm it passes after the type fix. If it passes both before and after, it isn't testing the fix.

---

### 🟡 Low — White‑label overrides are not theme‑aware (applied to both light and dark)

`whiteLabelThemeOverrides` is applied on top of `basePresetTheme`, which switches between `LIGHT_THEME` and `DARK_THEME` with `selectedThemeId`. A single flat override set is therefore forced onto **both** presets. A host that white‑labels, say, `backgroundColor` to a light brand color will also override the dark theme's background, which can make dark mode unreadable. If per‑theme branding is ever needed, this shape can't express it. At minimum this constraint is worth a code comment or a note in whatever documents the config; ideally consider keying overrides by theme id.

Related, and by design given the `ThemeOverrides` shape: white‑label config can only override `variables` (colors + font families), not `highlightColors`. Fine if intended — just flagging that highlight branding isn't reachable through this path.

---

### ⚪ Nit — stray blank line

`ThemeManager.tsx` adds a blank line after the `withColorOverrides` assignment (line ~1264) that doesn't match the surrounding style. Trivial; drop it or let Prettier decide.

---

### Summary

The layering approach is right, but as written the feature is effectively inert (and emits invalid CSS variables) for any config author who follows the declared `BibleTheme` type — the field should be typed `ThemeOverrides`. Pairing that fix with a behavior test that fails on the current code is the key thing to land before merge. CI note: at review time `build (24.x)` was still in progress.

---

KallynGowdy:
@AmitPant9999 We should not use overrides for white label themes. Instead, the branding config should store its own themes in their entirety and those should replace the default themes

AmitPant9999:
@KallynGowdy now branding theme will replace base theme and when there is no branding theme default theme will apply

KallynGowdy (approved):

PR #1837, opened 2026-09-18
generalise adding branding logo

AmitPant9999:


Blitzy:
> [!NOTE]

## Code review — generalise adding branding logo

**Summary of what this PR does:** It renders a deployment-level branding logo in the sidebar header. When the `BrandingConfig` (from `useAppConfig()`) provides both a `logo` URL and a `websiteUrl`, an `` wrapped in an external link is shown next to the existing customization logo. A `.sb-sidebar-branding-logo` CSS rule fixes its size.

The change is small, focused, and follows the existing pattern in `icons.tsx` (`branding.icon` used directly as an ``). The `rel="noopener noreferrer"` on the new-tab link is correct. A few things are worth a look before merging.

---

### Medium

**M1 — The logo is only rendered when a `websiteUrl` is also configured.**
`Tabs.tsx:807` gates the whole block on `branding?.logo && branding?.websiteUrl`. That means a deployment that sets a `logo` but leaves `websiteUrl` empty gets **no logo at all** — the logo silently disappears rather than just being un-linked. For a change whose goal is to "generalise adding branding logo," tying the logo's visibility to an unrelated field (the website link) is a footgun. `websiteUrl` is optional in practice (it can be an empty string in `BrandingConfig`), so this is a realistic misconfiguration.

_Reproduction:_
1. In the branding config, set `logo` to a valid image URL and leave `websiteUrl` as `""` (or omit it).
2. Load the app and open the sidebar.
3. Expected: the brand logo appears. Actual: nothing renders.

_Suggested fix — render the logo whenever it exists, and only wrap it in a link when a `websiteUrl` is present:_
```tsx
{branding?.logo &&
  (branding.websiteUrl ? (
    <a
      href={branding.websiteUrl}
      target="_blank"
      rel="noopener noreferrer"
      aria-label={branding.appName || "Brand logo"}
    >
      ``&lt;img·        src={branding.logo}·        alt={branding.appName || ""}·        className="sb-sidebar-branding-logo"·      /&gt;``
    </a>
  ) : (
    ``&lt;img·      src={branding.logo}·      alt={branding.appName || ""}·      className="sb-sidebar-branding-logo"·    /&gt;``
  ))}
```
If gating on both fields is actually intended (i.e. the logo is meant to exist only as a clickable link), please say so in the PR description so it isn't mistaken for a bug later.

---

### Low

**L1 — No tests, and the render condition is exactly what a test should lock down.**
There's no coverage for the new branch logic. The conditional in M1 (logo shows / doesn't show / is or isn't wrapped in a link depending on config) is precisely the kind of behavior that quietly regresses. A small test rendering `TabsHeader` with three `branding` shapes — logo + websiteUrl, logo only, neither — and asserting on the presence/absence of the `` and its wrapping `<a href>` would guard the intended behavior (and would fail today under the M1 scenario). Per the repo's test guidance, assert on the rendered output, not internals.

**L2 — Two logos can render side by side; confirm this is intended.**
`activeLogoUrl` (the user/customization logo at `Tabs.tsx:800`) and the new `branding.logo` render in the same `.sb-sidebar-top-start` container, so a deployment with branding plus an active customization logo will show both logos next to each other. That may be the desired behavior (brand logo vs. customization logo), but it's worth a deliberate confirmation — it's easy to end up with a cramped or confusing header, especially in the collapsed rail where the two stack vertically under the hamburger.

**L3 — Collapsed-rail fit is tight.** In the collapsed sidebar the content width is roughly `4.1rem - 2×0.5rem ≈ 3.1rem`, and the logo is `3.0625rem` wide, so it sits right at the edge. It should fit, but please eyeball it in the collapsed state (and on mobile) to be sure there's no clipping or overflow.

---

### Nits

- **N1 — CSS rule placement.** `.sb-sidebar-branding-logo` (`Tabs.css:29`) was inserted between `.sb-tabs-sidebar` and `.sb-tabs-sidebar-collapsed`, splitting the sidebar-container rules. It would read better grouped next to the analogous `.sb-sidebar-logo` rule (`Tabs.css:105`).
- **N2 — Hardcoded aria-label.** `aria-label={branding.appName || "Brand logo"}` uses a raw English fallback string. This component already mixes hardcoded and `t()`-based labels, so it's not a new inconsistency, but running the fallback through i18n (`t("brand-logo", { defaultValue: "Brand logo" })`) would match the localized labels elsewhere in the file.
- **N3 — Empty PR description / no linked issue.** The PR has no description and no linked issue, so the intended behavior (e.g. the M1 question) can't be verified against a spec. A one-line description of the goal would help reviewers.

---

Nothing here is blocking on its own; **M1** is the one worth an explicit decision before merge. Happy to re-review once addressed.

---

packages/seed-bible/seed-bible/components/Tabs/Tabs.tsx
```diff
@@ -802,6 +804,20 @@ export function TabsHeader(props: TabsHeaderProps) {
...
+            rel="noopener noreferrer"
+            aria-label={branding.appName || "Brand logo"}
+          >
+            <img
+              src={branding.logo}
+              alt={branding.appName || ""}
+              className="sb-sidebar-branding-logo"
+            />
+          </a>
+        )}
```
KallynGowdy, line 820: @AmitPant9999 I think if a customization is active then this would show but the customization logo and the branding logo. Instead, it should show the customization logo first, and if that doesn't exist then it should show the branding logo

AmitPant9999: @KallynGowdy now it will show  customization logo first and if that dont exist it will show branding logo

KallynGowdy (approved):

Issue #1797, opened 2026-09-11
Remove the Playlists section from the discover pane


KallynGowdy:
Currently, the Discover pane shows three sections:

- Playlists
- Playlist History
- Notes

Instead, it should just show Playlist history and notes, like this:

- Notes
- Playlist History (if you have any history)

Notes should be above your playlist history. If you don't have any notes, then a friendly prompt should be shown to encourage you to create a note. Something like "You don't have any notes for this chapter. **Create a note**"

PR #1838, opened 2026-09-18
Feature/1797

Kabi4:
Addresses: #1797


Blitzy:
> [!NOTE]

## Code review — PR #1838 (Feature/1797)

**Short version:** This PR does exactly what issue #1797 asks for, and it does it cleanly. The Discover pane now shows **Notes first**, then **Playlist history** (only when there is history), the standalone **Playlists** list is gone, and the empty-notes state now invites you to **Create a note**. Type check, lint, and the full `DiscoverPane` test suite (53 tests) all pass locally, and the reworked tests are real regression tests (they fail on the old code). I found no correctness bugs — only two minor test-quality nits below.

### What it does, checked against the issue
- **Playlists section removed** — `PlaylistSection` and the `PlaylistRow` import are gone from `DiscoverPane.tsx`. ✅
- **Notes above Playlist history** — `AnnotationsSection` is now rendered before `PlaylistHistorySection`. ✅
- **Playlist history hidden when empty** — `PlaylistHistorySection` early-returns `null` on empty history instead of rendering an empty-state. ✅
- **Friendly "create a note" prompt** — the empty annotations state renders "You don't have any notes for this chapter. **Create a note**", with the action wired to `annotations.createNewAnnotation()`. ✅

### Verification performed
- `pnpm vitest run DiscoverPane.test.tsx` → 53 passed.
- `pnpm check:ts` → 0 visible errors.
- `pnpm lint` → 0 errors (only pre-existing baseline warnings).
- Confirmed `PlaylistRow` is still used (by `YourContentPane`), so the updated doc-comment is accurate and the component is not dead code.

---

### Issues by severity

**Critical / High:** None found.

**Medium:** None found.

**Low / Nits**

**No. 1 — Two test assertions can no longer fail (they no longer guard anything).**
`test/unit/seed-bible/components/DiscoverPane.test.tsx` now contains assertions that check for i18n strings this PR deleted, so the code path that produced them is gone and the strings can never render regardless of behavior:
- In "does not list playlists, even when the user has some":
  `expect(container.textContent).not.toContain("You haven't created any playlists yet.");`
- In "hides playlist history when there is none" / "hides playlist history when there is none" (title view):
  `expect(container.textContent).not.toContain("Play a saved playlist while signed in and it will show up here.");`

These will pass whether or not the feature works, so they're effectively dead assertions. The *primary* assertions in both tests are solid (the section title is absent, the playlist row/item is absent), so coverage is fine — these lines are just noise. Suggested fix: drop the two `not.toContain(...)` lines, or replace them with an assertion that still means something (e.g. asserting the removed section's DOM container `.sb-playlist-item` is absent, which they already partly do).

*How to see it:* revert `DiscoverPane.tsx`/`AnnotationsSection.tsx` to `develop` and re-run just those two assertions — they still pass, because those default strings are no longer emitted by any code path.

**No. 2 — Orphaned keys left in non-English locale files (informational, not actionable here).**
`discover-playlists-empty` and `discover-playlist-history-empty` were removed from `en.json` but still exist in the other 76 locale JSONs. Per this repo's convention ("only update en.json… leave those for the professionals"), that's the expected outcome and nothing to change in this PR — flagging only so it's a known follow-up for translation cleanup.

---

### Test-quality assessment (the tests do reflect real behavior)
The task for this review specifically asked whether the tests assert real behavior rather than always passing. They do:
- **"does not list playlists, even when the user has some"** seeds an actual `userPlaylist` and asserts the "Playlists" section title and the playlist title never render — this **fails on the pre-fix code**, which rendered both.
- **"shows a prompt to create a note…"** asserts the `.sb-discover-empty-action` button exists and that clicking it calls `createNewAnnotation` exactly once — **fails on the pre-fix code**, which rendered a plain `DiscoverEmpty` with no button.
- **"shows Notes above Playlist history when there is history"** asserts the ordering by section-title index — **fails on the pre-fix code**, where Notes rendered after Playlist history.
- **"hides playlist history when there is none"** asserts the "Playlist history" title is absent — **fails on the pre-fix code**, which always rendered that section (with an empty-state).

Also worth calling out as a plus: the ~354 deleted lines of DiscoverPane playlist tests (share / edit / delete / play / hero image) are **not** a coverage loss — that behavior now lives in `PlaylistRow` used by `YourContentPane`, and `YourContentPane.test.tsx` covers it ("plays a playlist from its play button", "edits a playlist from its options menu", "offers share, edit and delete, the same as the Discover list", "renders a playlist as a Discover row, with its title").

---

### Conclusion
Solid, well-scoped change that matches the issue and keeps the test suite honest. Nothing blocking from my side — addressing nit No. 1 would tidy up the tests, and nit No. 2 is just a heads-up for the translation follow-up.

---

Issue #1744, opened 2026-09-03
Add Theographic Bible Data


KallynGowdy:
Now that Theographic Bible data is supported in the Free Use Bible API, we should add it into the Seed Bible as discoverable content.

In particular, we should support:

- People
- Places
- Events

Because there is so much content, we should surface this content through the compact discover panel as custom types of discovered content. So a person is a `person_profile` type of discovered content, and a place is a `place_profile` type and an event is a `event` type. This way, we can have filters for each type.

By default, we should filter out people, places, and events until the user asks for them.

PR #1839, opened 2026-09-18
added the theological dataset to discover panel

Bored-Wizard:
task #1744


Blitzy:
> [!NOTE]

## Summary

This PR adds the Theographic Bible Metadata dataset (people, places, events) to the compact Discover panel, matching what issue #1744 asks for: three new content types (`person_profile`, `place_profile`, `event`), per-type filters, and content hidden from the "All" view by default. The implementation is well-organized and unusually well-documented, with an on-device IndexedDB cache, request de-duplication, graceful storage-failure fallback, and an expandable card that lazily fetches each entity's full record.

I ran the repo's own checks against the PR head: `pnpm check:ts` passes (0 visible errors), `pnpm eslint` on the new files is clean, and all **157** tests across the six changed/added test files pass. The test suite is genuinely strong — it asserts on observable output, covers error paths (404 vs 500, blocked/read-only storage, retry), and uses fake timers and hand-controlled promises instead of fixed sleeps.

That said, I found one real functional dead-end, a mergeability blocker, and a few smaller issues. Details below, most severe first.

---

## High

### 1. A chapter whose only discoverable content is a single Theographic type becomes unreachable

**What happens:** The filter-chip row only renders when `filters.length > 2` (`DiscoverContentPanel.tsx:204`). The "All" filter is always present, so this means "show chips only when there are at least two *other* categories." That rule was correct before this PR, because every other category is *also* shown inline under "All" — so a lone extra chip added nothing.

Theographic content breaks that assumption: it is deliberately hidden from "All". So when a chapter has exactly one Theographic type and nothing else (say a genealogy chapter with People but no Places, Events, annotations, cross-references, study notes, or reading plans), the result is:

- No chip row renders (only `all` + `people` = 2 filters, which is not `> 2`).
- Under "All", the panel shows the hint *"Choose People, Places or Events to see who and what appears in this chapter."*
- But there is no chip to choose, and the People content is never rendered anywhere. The reader is told to do something the UI gives them no way to do.

**Reproduction:** I confirmed this with a focused render test using the same fixtures as the existing panel tests — a tab whose `discoveredContent` holds a single `person_profile` result and whose annotations/cross-refs/study-notes are empty:

```
CHIPS: []   HINT: true   ADAM VISIBLE: false
```

So: hint shown, zero chips, content unreachable. To reproduce in the app, open a chapter that Theographic has people (or only places, or only events) for but that has no annotations/cross-references/study-notes/reading-plans of its own.

Note the existing panel tests don't catch this because every Theographic fixture in `DiscoverContentPanel.test.tsx` includes **two** types (People + Places), which pushes `filters.length` to 3.

**Suggested fix:** Show the chip row whenever a hidden-by-default (Theographic) filter exists, since those can't be reached any other way. For example:

```tsx
const hasHiddenOnlyFilter = hasPeople || hasPlaces || hasEvents;
// ...
{(filters.length > 2 || hasHiddenOnlyFilter) && (
  <div style={{ display: "contents" }}>
    <div className="sb-dcp-filters" role="tablist">
      {/* ... */}
```

Please also add a regression test for the single-type case (assert the chip row is present and clicking it reveals the content) — reverting the fix should turn it red.

---

## Medium

### 2. The PR does not merge cleanly into `develop`

GitHub reports the PR as `dirty`, and a local trial merge confirms conflicts in three files:

- `CHANGELOG.md`
- `packages/seed-bible/seed-bible/components/DiscoverPane/DiscoveredResultsSections.tsx`
- `test/unit/seed-bible/components/DiscoveredResultsSections.test.tsx`

`develop` has moved since this branch was cut. It needs a merge/rebase and conflict resolution before it can land. The `DiscoveredResultsSections` conflicts are worth extra care — this PR reworks `ContentSection`'s filtering there, so make sure both sides' intent survives (and re-run the tests after resolving).

---

## Low / Nits

### 3. `deathPlace` is fetched and typed but never displayed

`TheographicPersonDetail.person.deathPlace` is declared and parsed, and `PersonDetail` renders `birthPlace` (`TheographicEntityCard.tsx:258-261`) plus both `birthYear` and `deathYear` — but never `deathPlace`. Given the symmetry with birth/death year, this looks like an accidental omission rather than a deliberate choice. Consider adding a `Birthplace`/`Place of death` pair, or dropping the unused field.

### 4. Place "Type" only ever shows `featureSubType`, via an awkward expression

In `PlaceDetail` (`TheographicEntityCard.tsx:303-306`) the Type value is built as `[place.featureSubType].filter(Boolean).join("")`, which is a roundabout way of writing `place.featureSubType`. Two things worth a look: (a) the broad `featureType` (e.g. "Region"/"Water") is available on the record but never surfaced in the detail, so a place with a `featureType` but no `featureSubType` shows no Type at all; (b) the expression can be simplified. Consider `place.featureSubType ?? place.featureType`.

### 5. A large, well-tested block of name-matching code never runs in production

`REQUIRE_NAME_IN_VERSE_TEXT` is hardcoded `false` (`TheographicDiscoverProvider.tsx:530`), which means `mentionsName`, `normalizeForMatch`, `narrowToMentionedVerses`, `chapterVerseText`, and `loadVerseText` are never exercised by the running app — `discover()` always takes the `verseText === null` branch. The comment explains this is intentional (the 38%-false-drop measurement is a good call, and keeping it for a future aliases-aware API is reasonable). Flagging only because the dataset's verse links are then trusted verbatim, so entries like "Jacob (Israel)" will list verses that name the *nation* rather than the person — the PR description's "Select a verse and the lists narrow to what that verse mentions" is really "narrow to the dataset's own verse list," which is coarser. Also note that per the repo's "test real-world behavior" guidance, the `mentionsName`/`narrowToMentionedVerses` tests pass but verify a path the app doesn't take today. No change required — just make sure this disabled-by-default state is the intended shipping behavior.

### 6. `formatYear(0)` returns `"0 AD"`

`formatYear` (`TheographicEntityCard.tsx:411-416`) treats `< 0` as BC and everything else as AD, so a `0` year renders "0 AD" (there is no astronomical year 0). Vanishingly unlikely to appear in the data — noting only for completeness.

---

## On the tests specifically

The tests are real and behavior-focused, not rubber-stamps. Highlights: the client tests prove the cache actually prevents a refetch across a simulated reload and that a failure isn't cached (so retry works); the card tests drive the loading → loaded/error states with a hand-resolved promise and verify the retry path; `placeToGeoJson` locks in the deliberate `[lat, lng]` ordering with a comment explaining why the spec order is *wrong* here. The two gaps to close: a regression test for finding No. 1, and awareness that finding No. 5's tests cover a disabled path.

Nice work overall — the caching layer and the documentation around the coordinate ordering and the mention-check trade-off are especially good.

---

KallynGowdy:
@Bored-Wizard It looks like it is included in the core app? Could you make it into an optional extension for now so that it isn't included by default?

Bored-Wizard:
> @Bored-Wizard It looks like it is included in the core app? Could you make it into an optional extension for now so that it isn't included by default?

Sure, I will make it a extension that can be added to the discovery section. @amir2iqbal447 is making the new design for it so will wait till then to do it completely.

KallynGowdy:
@Bored-Wizard Designs for theographic data:

Mobile:
image

Desktop:
image

Figma: https://www.figma.com/design/vFp5AhAQwizOB6LN2MvrIB/Seed-Bible?node-id=10027-3128&t=xGwpWbM4YByShdL5-4

A key difference between mobile and desktop is that desktop can display interactive data (maps, 3d objects, etc) inline, while mobile needs a button for it.

Both mobile and desktop should display verses inline (without automatically jumping to them). If the verse is for the current chapter, then desktop can jump to that verse automatically by clicking the reference, but mobile should still display inline.

KallynGowdy:
@Bored-Wizard This is super cool!

One request I have is that events show `BC` and `AD` just like how People show it:

image

Negative dates are BC and positive ones are AD.

KallynGowdy:
Also @Bored-Wizard If there is space can we collapse the verse reference pills into one line with the title of an element? We also don't need to say "Event" or "Place" or "Person" on each item, because it already says it in the group heading:

image

Bored-Wizard:
> Also @Bored-Wizard If there is space can we collapse the verse reference pills into one line with the title of an element? We also don't need to say "Event" or "Place" or "Person" on each item, because it already says it in the group heading:
> 
> image

I can remove the person and event subtype. I will let the places subtype remain there as they have meanings. as for the reference pill, I think how they are now will look better than having them next to the title as it will look quite cluttered

PR #1840, opened 2026-09-19
Feature/1784

### LINKED ISSUE #1784: Playlist share links should always point to the first chapter referenced by the playlist
Currently when you get a link to a playlist, it will include the chapter that you're currently viewing. This can be confusing since it will load that chapter first, and then load the playlist afterwards. Instead, playlist share links should always reference the chapter of the first scripture item so that users don't have to jump into one chapter and then another.

Kabi4:
Fixes: #1784


Blitzy:
> [!NOTE]

## Code Review — PR #1840 (Feature/1784)

Reviewed against issue #1784 ("Playlist share links should always point to the first chapter referenced by the playlist"). The change makes `getPlaylistUrl` build the share URL from the first scripture item's chapter instead of the chapter the sharer happened to be reading, falling back to the current page only when the playlist has no scripture. This matches the issue's intent precisely.

### Overall
This is a clean, well-scoped fix. It reuses the existing `buildReadingUrl` / `parseReadingPath` helpers rather than hand-rolling URL construction, keeps the sharer's UI language, and clears unrelated query params. Type check passes and all 7 new tests pass.

**On test quality (the thing I looked at hardest):** the tests assert real observable behavior — the actual `url.pathname` and query keys of the generated link — not internals. I verified they are genuine regression tests, not tests that pass regardless: the pre-fix code produced the sharer's current chapter, so e.g. `expect(url.pathname).toBe("/en/AAB/genesis/1")` (from a page at `/en/AAB/john/3`) fails against the old implementation and passes against the new one. The suite also covers the meaningful branches: leading non-scripture items skipped, cross-chapter start chapter, item-level translation override, UI-language preservation, query-param stripping, and the no-scripture fallback. Good coverage.

No blocking issues. The items below are low severity / nits.

---

### 🟡 Low — First scripture item with an unresolvable `bookId` silently discards the whole feature

**File:** `packages/seed-bible/seed-bible/managers/PlaylistManager.tsx` (`firstScriptureShareRef`, ~line 165)

`firstScriptureShareRef` finds the first `bible-verse` item, then calls `getBookId(firstScripture.ref.bookId)`. If that one item's book fails to resolve, the function returns `null` and the share link falls back to the sharer's current page — even if a *later* scripture item in the playlist has a perfectly resolvable book. The issue asks for "the first chapter referenced by the playlist," so a more faithful behavior would be to use the first scripture item whose book actually resolves.

In practice this is unlikely to bite, because saved playlist items carry canonical book IDs (e.g. `"GEN"`, `"JHN"`), so resolution should essentially always succeed. That's why I've rated it Low rather than higher.

**Suggested fix** — resolve during the search instead of after picking the first candidate:

```ts
for (const item of playlist.items) {
  if (item.type !== "bible-verse") continue;
  const bookId = getBookId(item.ref.bookId);
  if (!bookId) continue;
  return { bookId, chapter: item.ref.chapter, translationId: item.translationId };
}
return null;
```

*Reproduction (theoretical):* construct a playlist whose first `bible-verse` item has `ref.bookId` set to a value `getBookId` can't resolve, followed by a valid one; the share link points at the sharer's current chapter rather than the second item's chapter. Not reproducible with normally-authored playlists.

---

### 🟢 Nit — CHANGELOG entry is missing an issue/PR reference link

**File:** `CHANGELOG.md` (line 25)

The new entry reads:

> Open a shared playlist on its first scripture chapter instead of the chapter the sharer was reading, so the reader does not load one chapter and then jump to another.

Neighboring entries in the same section link their issue/PR (e.g. `([#1825](https://github.com/HelloAOLab/seed-bible/pull/1825))`). Consider appending `([#1784](https://github.com/HelloAOLab/seed-bible/issues/1784))` for consistency.

---

### 💬 Consideration (not a defect) — UI-language vs. translation can now mismatch in the path

**File:** `packages/seed-bible/seed-bible/managers/PlaylistManager.tsx` (`getPlaylistUrl`, ~line 1669)

The translation now comes from the scripture item first (`scripture.translationId ?? parsed?.translationId ?? …`), while `buildReadingUrl` keeps the language segment from the current URL. So a user reading in Spanish (`/es/spa_onbv/...`) who shares a playlist whose first item was saved with an English translation would get a path like `/es/<english-translation>/john/3` — Spanish UI-language segment, English translation.

This is consistent with `buildReadingUrl`'s documented, existing share-link behavior ("keep whatever language the page is already being read in"), so I don't think it's a bug — just flagging it because the translation source is new here. Worth a quick confirmation that opening such a link renders sensibly (Spanish UI, English content). No test currently exercises the "item translation in a different language than the current path" combination, so it may be worth one if you want to lock the intended behavior.

---

Blitzy (approved):

PR #1841, opened 2026-09-19
Feature/1787

### LINKED ISSUE #1787: Reading plan share links should always point to the first chapter referenced by the reading plan
Currently when you get a link to a reading plan, it will include the chapter that you're currently viewing. This can be confusing since it will load that chapter first, and then load the reading plan afterwards. Instead, reading plan share links should always reference the chapter of the first scripture item so that users don't have to jump into one chapter and then another.

See #1784 for the equivalent work for playlists.

Kabi4:
Fixes: #1787


Blitzy:
> [!NOTE]

## Code review — PR No. 1841 (Fixes No. 1787)

**Short version:** The core of this change is solid and does what issue No. 1787 asks: a shared reading-plan link now carries the plan's *first scripture chapter* in its path, so opening it lands the reader on that chapter instead of loading the sharer's current chapter and then jumping. The URL-building logic is clean and the share-URL tests are genuine (they exercise book resolution, translation selection, language preservation, and query-param stripping — none of them are tautological). Below are the issues I found, ordered by severity. Nothing here is a crash-level defect; the most important item is a test-quality slip, which is directly in scope for this review.

---

### Medium

**M1. Two test cases are exact duplicates (dead coverage, and a hint of a missing test).**
`test/unit/seed-bible/managers/ReadingPlansManager.test.ts` defines the same two `loadByLocator` tests twice, verbatim:

- `"loadByLocator loads and selects the plan from a share locator"` — lines ~1568 and ~1591
- `"loadByLocator returns null for a malformed locator"` — lines ~1582 and ~1605

The second pair adds zero coverage. Vitest allows duplicate test names, so this passes silently rather than erroring — which is exactly the kind of "test that can't fail to tell you anything" worth catching. It also reads like a copy-paste that was *meant* to become a distinct test (see N2 below for the error-path case that's actually missing).

*Reproduction:* open the test file and search for `loadByLocator loads and selects` — it appears twice with identical bodies.

*Suggested fix:* delete the second copy of each (lines ~1591–1612), and consider replacing them with the error-path test described in N2.

---

### Low

**L1. A broken or private share link fails silently — the reader gets no feedback.**
In `SeedBibleStateManager.tsx`, `setupInitialReadingPlan` only logs to the console when the plan can't be loaded:

```ts
} catch (error) {
  console.error("Failed to load reading plan from URL:", error);
}
```

`loadByLocator` → `getReadingPlan` throws when `os.getData` fails (deleted plan, revoked permission, network error) or the record fails schema validation. When that happens the reader sees nothing at all — the chapter from the path loads, but no plan pane and no explanation. The playlist equivalent this PR is modeled on showed a toast in that case (`"Failed to load playlist"`, now commented out around lines 2624–2629 of the same file), and this manager already has `toast` in scope right here.

*Reproduction:* open `…?readingPlan=<recordName>.<address>` pointing at a plan you can't read (delete the plan, or use another user's private plan). The page loads a chapter and nothing else happens; only the console shows an error.

*Suggested fix:* on the catch, surface a toast, e.g. `toast(t("failed-to-load-reading-plan", { defaultValue: "Failed to load reading plan" }))` (add the key to `en.json` only, per the translation convention).

**L2. The plan pane can silently fail to open if the reading tab isn't ready.**
Still in `setupInitialReadingPlan`:

```ts
const readingState = selectedTab.peek()?.readingState;
if (!readingState) {
  return;
}
```

`selectedTab` is a `computed`, and this early-return happens after an `await` on a network round-trip, so in practice the tab is usually populated by then. But if it isn't, the plan is loaded and selected yet the pane never opens and the user is given no indication why. This is a robustness gap rather than a confirmed failure; worth either reacting to `selectedTab` becoming available, or at least logging so it's diagnosable if it ever bites.

---

### Nits

**N1. CHANGELOG entry omits the PR/issue link.** The new line under `### 🗑️ Removed`… (in `CHANGELOG.md`) reads well, but unlike the neighboring entries (e.g. the No. 1825 one right above it) it doesn't link the PR/issue. Consider appending `([#1841](https://github.com/HelloAOLab/seed-bible/pull/1841))` for consistency.

**N2. Locator error-path coverage is thin.** `parseReadingPlanLocator` also rejects a locator with *no* dot at all (`lastIndexOf(".")` returns `-1`), but no test covers that input — only `.plan-1` and `record-1.` are tested. And there's no test that `loadByLocator` propagates (rejects) when `getReadingPlan` throws, which is the behavior `setupInitialReadingPlan` relies on for its catch. If you replace one of the duplicated tests from M1, this is the natural thing to put there.

**N3. `getReadingPlanShareUrl` reads `window.location.href` directly** rather than `navigation.currentUrl`, unlike the playlist path. It's fine because the function is module-level (no navigation manager in scope) and is only called on a user click, so `window.location` is current — flagging only for awareness, not a change request.

---

### What's good
- The path-vs-query mechanism is right: the share URL puts the first chapter in the *path* (`buildReadingUrl`) and the plan in `?readingPlan=`, so the reader boots straight into the correct chapter — exactly the "don't load one chapter then jump" behavior No. 1787 wanted.
- `firstPlanScriptureShareRef` correctly skips non-scripture readings and readings whose `bookId` doesn't resolve, and falls back to keeping the current page when a plan has no usable scripture — and each of those branches has a real test.
- Translation and UI-language selection (scripture's own translation → current path's → default; language preserved from the path) is handled sensibly and covered.
- `en.json`-only translation additions follow the repo convention.

---

Blitzy (approved):

PR #1843, opened 2026-09-19
Feature/1689

Kabi4:


Blitzy:
> [!NOTE]

## Review of No. 1843 — Embedded Mode phase 1 (issue No. 1689)

**Summary:** This is a clean, well-scoped feature. It adds a `?embed=minimal` / `?embed=true` mode that renders a compact reader: the sidebar, Today screen, and bottom tab bar are dropped, and the phone reader's chrome is reused (a slimmed header of translation / open-in-new-tab / quick settings / X, plus the floating chapter-nav bar). The implementation matches the design points in the issue, the CSS refactor is careful, and the test coverage is genuine rather than always-green. I found no correctness bugs — only a few low-severity cleanups.

I verified the two things that most commonly break a feature like this, and both are handled correctly:

- **No SSR/hydration flash or mismatch.** `isMinimalEmbed` is derived from the URL, and the server seeds the state from the full request path (`standalone/entry-ssr.tsx` builds `initialHref` including the query string), so the server and the first client render agree on embed mode.
- **Embed survives navigation.** `buildReadingUrl` starts from the current URL and only strips the legacy position params, so the `embed` param is preserved when moving between chapters, through inline verse-reference links, and through the SSR language redirect (`remainingParams` keeps it). The mode does not silently switch off mid-session.

Also confirmed: only `en.json` was touched for i18n (per the repo convention), and every new translation key (`open-in-new-tab`, `close`, `settings`, `chapter-navigation`, `go-to-all-settings`) is present.

---

### Low severity

**No. 1 — `sb-bible-reader-mobile-header-embed` is applied but has no CSS rule.**
`BibleReader.tsx` adds this class to the header in embed mode:

```
}${isMinimalEmbed ? " sb-bible-reader-mobile-header-embed" : ""}`}
```

A repo-wide grep finds no matching selector in any `.css` file — the class does nothing today. Either the intended embed-header styling is missing, or this is a leftover hook.

- *Reproduction:* `grep -rn "sb-bible-reader-mobile-header-embed" packages/` returns only the JSX line, no CSS.
- *Suggested fix:* drop the class if it is not needed, or add the rule it was meant to carry.

**No. 2 — the shared `isCompactReader` signal is added but bypassed by two of its three consumers.**
`SeedBibleStateManager` now exposes `state.app.isCompactReader` (documented as "phone layout, or a compact embed"), but `BibleReader.tsx` and `TabsLayout.tsx` each re-derive the same thing inline instead of reading it:

```ts
// BibleReader.tsx
const isCompactReader = isMobile || isMinimalEmbed;
// TabsLayout.tsx
const isMobile = (state?.app.isMobile.value ?? false) || isMinimalEmbed;
```

Only `BibleReaderToolbar` consumes the shared signal (via `isSmallScreen`). This is the kind of duplication `CLAUDE.md` cautions against — three copies of one concept that can drift apart later. Since the signal already exists, consider having `BibleReader` and `TabsLayout` read `state.app.isCompactReader` directly.

---

### Informational / follow-up

**No. 3 — the X (close) button is an intentional no-op.**
Per the issue ("an X button (that does nothing for now)") this is expected, and the test asserts clicking it does nothing. Flagging only so it is not forgotten: it renders as a real, focusable button with `cursor: pointer` and `aria-label="Close"`, so a visitor may expect it to act. Worth a tracking item to wire it up (or hide it) in a later phase.

**No. 4 — one test assertion is necessarily weak while the button is a no-op.**
`BibleReader.test.tsx` → *"does nothing when the close button is clicked"* only asserts `window.open` and `openSettings` were not called. That is fine for the current spec, but it would not catch the button being wired to some *other* action. Worth strengthening once the close button gains real behavior.

---

### Tests

The tests exercise real behavior and are not tautological:

- `EmbedMode.test.ts` covers the parsing surface well, including case-insensitivity (`MINIMAL`, `True`) and the non-canonical values that must *not* trigger embed (`""`, `"1"`, `"yes"`, `"false"`, `null`).
- `BibleReaderToolbar.test.tsx` drives state from a real `?embed=` URL and asserts both directions: embed shows the floating nav and hides the bottom tabs (at phone *and* 1000px width), while `?embed=1` leaves the full toolbar intact — so the test would fail if embed did nothing.
- `TodayManager.test.ts` confirms embed suppresses Today even when `?today=open` is also set (the guard is correctly ordered before the `?today` check), and that `today.open()` is a no-op in embed.
- `MobileSettingsSheet.test.tsx` verifies "Go to all settings" is hidden in embed and present otherwise.

One structural note: the `BibleReader` embed-header tests hard-code the `isMinimalEmbed`/`isCompactReader` signals rather than driving them from a URL. That is reasonable for a component unit test, and the URL→signal wiring is covered by the `EmbedMode` and toolbar tests, so coverage overall is sound.

---

Blitzy:
Thanks for this! The embed mode itself works well. The layout, the slimmed header, and embed carrying through chapter navigation all look good.

The problems show up in the **first-run experience**. You can reproduce everything below by opening the embed in an incognito window (`https://alpha.seedbible.org/b/feature_1689/?embed=true`). That's what most visitors to a partner site will see the first time they load us. Several full-app features still appear or can still be reached, and some of them break in embed mode.

Since `embed=true` is just an alias for `minimal`, and we may add other embed types later, I think `minimal` should really be as minimal as possible.

### Needs fixing

**1. The tutorial launches in embed mode.**
- **What happens:** On a fresh load, the tutorial prompt appears and the tour runs.
- **Why it's broken:** It highlights elements in odd places, and some steps (like "select books and chapters") point at toolbar items that are hidden in minimal embed.
- **Fix:** In minimal embed, no tutorial at all: not the prompt, not the tour, not the skip prompt.

Screenshot 2026-09-24 122347

**2. The "install Seed Bible to your desktop" onboarding modal appears.**
- **What happens:** When you finish or skip the tutorial, we ask the visitor to install the app.
- **Fix:** This shouldn't appear in embed mode. A visitor on someone else's site shouldn't be asked to install our PWA from inside their iframe.

Screenshot 2026-09-24 122403

**3. The "save this translation for offline reading" prompt appears.**
- **What happens:** This prompt comes up after the install prompt.
- **Fix:** It should be gone in embed mode too.

For 1–3: all of these are rendered unconditionally in `app/main.tsx` (`OnboardingModals`, `TutorialPrompt`, `OfflineDownloadPrompt`, `Tutorial`, `TutorialSkipPrompt`). I'd suggest gating them where they get *triggered* (tutorial auto-start, onboarding, offline prompt), the same way `TodayManager` already checks for embed. That's safer than only hiding the components, because then nothing starts in the background either. Please add tests like the `TodayManager` ones that show each one stays off in embed.

**4. Remove the close (X) button.**
- **Background:** The issue asked for an X that "does nothing for now". Now that I've seen it in a real embed, I don't think it serves a purpose.
- **Why remove it:** Whether an embed can be closed, and what closing means (hide the iframe, collapse a widget, leave the page), depends on the embedding site. The embedder should add their own close button around the iframe if they want one.
- **Also:** A button that looks clickable but does nothing is confusing for visitors.
- **Fix:** Drop the button, the `close` translation key if nothing else uses it, and the "does nothing when the close button is clicked" test.

Screenshot 2026-09-24 122526

**5. Selecting a verse brings up the full mobile verse toolbar.**

Selecting a verse shows every default option: highlight, save, note, copy, Compare, and Share.

There's currently no way to sign in from the embed, so highlights, saves and notes aren't very useful there. If we keep verse selection at all, I think **Copy** is the only action that clearly belongs, and Compare is a maybe:

Screenshot 2026-09-24 122748

Two of these are also broken in embed:
- **Compare:** The pane is positioned to sit above the mobile bottom navigation bar (search, books, Today, etc.). That bar is hidden in embed, so reader text shows through at the bottom of the Compare pane. Other full-screen panes are probably affected the same way.

Screenshot 2026-09-24 122811

- **Share:** The share modal offers "Share current session", so a visitor can *start a shared session*. But every control for managing a session or seeing other participants is hidden in embed, so they end up in a session they can't see or leave.

**Open question for @KallynGowdy:** In minimal embed, should we disable verse selection entirely, or keep a reduced toolbar (Copy, maybe Compare)? Whatever we decide, Highlight, Save, Note and Share shouldn't be reachable. If Compare stays, its pane needs to be fixed so the reader text doesn't show through underneath it.

### Summary
- [ ] No tutorial (prompt, tour, or skip prompt) in minimal embed
- [ ] No install-to-desktop onboarding modal in embed
- [ ] No offline-download prompt in embed
- [ ] Remove the close (X) button and its test
- [ ] Verse selection in embed: disable it, or reduce the toolbar to Copy (and maybe Compare) pending Kallyn's answer; no Highlight, Save, Note or Share either way

KallynGowdy:
@Blitzy  I would keep a reduced toolbar with copy, share and compare. Any particular reason why you think share should be removed?

Blitzy:
@KallynGowdy I think we could keep share as long as we remove the shared session aspect of it? Currently you can start a shared session but none of the UI to actually manage the shared session.

Blitzy:
@KallynGowdy I had another thought, would we want to actually soft disable extensions in embed mode? I can see there being a number of future extensions that break minimal embed mode.

PR #1844, opened 2026-09-20
Feature/1807

### LINKED ISSUE #1807: Offer to switch to a downloaded translation when failing to load a chapter
When a chapter fails to load, we should offer to switch to a different translation that the user has downloaded. This should only show if the user has a different translation that is saved offline and is in the same language as the one requested (or the UI language if the current translation language is unknown).

Kabi4:
Fixes: #1807


Blitzy:
> [!NOTE]

## Code review — PR No. 1844 (Fixes #1807)

**Short version:** This is a solid, well-scoped implementation of the feature and it does exactly what issue #1807 asks. The core matching logic lives in a small, pure, thoroughly-tested helper (`findOfflineTranslationFallbacks`), the settings language picker was cleanly extracted into a reusable `SearchableSelect` (good deduplication), and typecheck, lint, and the full test suite for the touched files all pass locally (99 tests). I found no correctness or crash bugs. The findings below are all polish/minor.

**What I verified locally:**
- `pnpm vitest run offlineTranslationFallback.test.ts BibleReader.test.tsx` → 99 passed
- `pnpm check:ts` → 0 visible errors
- `eslint` on the four changed source files → clean

**How the behavior lines up with the issue:** "same language as requested, or the UI language when the current translation's language is unknown" is implemented faithfully — same-Bible-language matching (including aliases like `arb`/`ara`), UI-language fallback only when the requested language is empty/unknown, the failed translation itself excluded, and the hardcoded default surfaced first. The tests cover each of these branches.

---

### Findings by severity

#### Low — Plural "saved on this device" list has no spacing between names

`BibleReader.tsx` builds the plural sentence by joining translation names with a bare comma:

```tsx
names: offlineFallbackTranslations
  .map((item) => item.name)
  .join(","),
```

So with two saved English translations the reader sees:

> Accessible Ancients Bible,New International Version are saved on this device.

There is no space after the comma, and for three or more there is no "and" (`A,B,C are saved`). The changelog for this PR says the matches are "listed by name", so this is the user-facing list.

**Reproduce:** Download two English translations other than the one you are reading, open the reader on the current translation, drop the network (or navigate to an uncached chapter that fails to load). The error panel shows the run-together names above.

**Suggested fix:** at minimum `.join(", ")`. Better, since this string is localized, hand the joining to the locale rather than hardcoding the separator/conjunction, e.g. `new Intl.ListFormat(language, { style: "long", type: "conjunction" }).format(names)`, which yields "A, B and C" and adapts per language.

Note this also touches the tests — see the test note below.

#### Nit — Test asserts the un-spaced string, coupling it to the formatting above

In `BibleReader.test.tsx` the multi-download test asserts the exact run-together text:

```tsx
expect(errorPanel?.textContent).toContain(
  "Accessible Ancients Bible,New International Version are saved on this device."
);
```

That means the moment someone fixes the spacing above, this test goes red for a cosmetic reason. When you address the join, update this assertion to match the corrected output.

#### Nit — Reader reuses a Settings-scoped CSS class

`OfflineFallbackSwitch` passes `buttonClassName="sb-settings-language-select"`, a class defined in `SettingsPage.css`. It works today because that CSS is bundled globally and `SearchableSelect.css`'s own `.sb-language-picker-button` already supplies nearly identical styling — but wiring a reader component to a `sb-settings-*` class is a maintenance trap (rename/remove it in Settings later and the reader silently loses styling). Since `SearchableSelect` already styles its own button, consider dropping the `buttonClassName` here (or giving the reader its own class) so the reader doesn't depend on Settings' stylesheet.

#### Nit — Changelog entry is missing the issue link

The other entries in the "✨ Added" block link their issue (e.g. `([#1793](...))`). The new bullet for this feature doesn't. For consistency, add `([#1807](https://github.com/HelloAOLab/seed-bible/issues/1807))`.

---

### On the tests (real-world behavior check)

The tests are genuinely good and exercise observable behavior, not internals:

- `offlineTranslationFallback.test.ts` drives the pure helper across every branch (nothing downloaded, only-the-failed-one downloaded, same-language match, different-language rejection, no-fallback-to-UI-when-known, alias matching, UI-language fallback, default-first ordering, recency ordering, regional-locale normalization). These would fail on the pre-fix code and can't pass trivially.
- The component tests render the real `BibleReader`, set `error.value`, and assert on rendered DOM plus that `selectTranslationAndChapter("AAB"/"NIV", "GEN", 1)` was actually called after a click — real behavior, and they mock only at the `state.bibleData.offline.records` boundary, as the repo's testing guidance asks.

One honest caveat worth keeping in mind: the three "does not offer …" tests assert only that `.sb-reader-error-switch` is absent. On their own such negative assertions also pass against code where the feature simply never renders, so they can't distinguish "correctly withheld" from "broken/absent". They're fine here because the positive tests prove the feature renders in the intended cases — just don't rely on the negative tests alone to guard the withholding logic.

---

Blitzy:
@Kabi4 Before this merges, there's one bug and one wording change I'd like.

### Bug: we offer translations that don't contain the current book

`findOfflineTranslationFallbacks` only checks language. It never checks whether a downloaded translation actually has the book the reader is on. This comes up with the Apocrypha, or with a New Testament-only download while someone is reading the Old Testament.

**Example:** Someone is reading Tobit 3 in a translation that includes the Apocrypha. The chapter fails to load, and the only English translation they've downloaded is BSB, which has no Apocrypha.

1. **We offer BSB anyway.** The panel shows "Switch to Berean Standard Bible".
2. **Clicking Switch looks like it did nothing.** `selectTranslationAndChapter("BSB", "TOB", 3)` can't find Tobit in BSB's book list, so it fails with `Book with ID "TOB" not available for translation "BSB"`. The panel doesn't show that message. It shows the same generic "Chapter unavailable… check your internet connection" text with the same BSB offer, so the reader has no idea why the switch failed.
3. **Reload gets stuck.** `selectTranslationAndChapter` sets `lastLoadAttempt` before it runs, even if it then fails. After a failed switch, Reload retries "switch to BSB at Tobit 3" instead of the original chapter. It keeps failing even after the connection comes back, and the reader has to navigate away manually to recover.

**Suggested fix:** Only offer a translation that contains the current book. The saved download record has no book list, so look each candidate up (`offline.getTranslationBooks(id)`, or the cached book lists) and drop any without `bookId`. It's worth checking the chapter too: if the chapter is out of range, `resolveChapterInBook` quietly sends the reader to chapter 1. If nothing is left after filtering, don't show the offer.

**Test to add:** Current book is `TOB`, and the only same-language download has no `TOB`. Assert that no switch is offered. The current tests all use Genesis, which every translation has, so none of them cover this case.

### Wording: don't list translation names in the message

Right now the message lists every matching translation by name. Someone with a lot of downloads in one language would get a very long sentence. Let's keep the message generic and let the controls below it name the translations: the "Switch to {{name}}" button when there's one match, and the dropdown when there are several.

Please use i18next's `_one` / `_other` plural keys, the same way `reading-plan-cadence-length` does:

```json
"chapter-unavailable-offline-switch_one": "You have a translation saved on your device that contains {{bookName}} {{chapterNumber}}.",
"chapter-unavailable-offline-switch_other": "You have {{count}} translations saved on your device that contain {{bookName}} {{chapterNumber}}."
```

Call it once, e.g. `t("chapter-unavailable-offline-switch", { count, bookName, chapterNumber })`, and i18next picks the right one.

This replaces both `chapter-unavailable-offline-switch` and `chapter-unavailable-offline-switch-plural`. Once the names are no longer joined into a sentence, `formatConjunctionList` isn't used anywhere and can be removed. The "A and B" assertion in `BibleReader.test.tsx` and the changelog's "listed by name (for example "A and B")" wording will need updating too.

This wording only makes sense once the book check above is in place. Otherwise we'd be saying "you have a translation that contains Tobit 3" when it doesn't.

### Still open from the earlier review (minor)

- `OfflineFallbackSwitch` still passes `buttonClassName="sb-settings-language-select"`, a class from Settings. `SearchableSelect` already styles its own button, so this can probably just be removed.

